PR #48, opened 2017-03-20
get the deployment's namespace for filtering pods

exarkun:
The deployment's podspec metadata doesn't necessarily have a namespace.

This is untested because it looks like actually testing it would be tricky.


itamarst:
Thanks! Will take a look at this tomorrow.

PR #70, opened 2017-03-28
Switch to SSH forwarding

kflynn:
...and add debugging for commands being run by `entrypoint.py`.


itamarst:
This never got merged?

Issue #45, opened 2017-03-16
Better distribution mechanism


itamarst:
Right now we have file you download.

Some issues:

1. We don't notice new versions exist.
2. No update mechanism.
3. No way to use 3rd party Python libraries.

Options (non-exclusive), each of which might cover different subset of the above:

* Native packages (brew, deb, rpm)
* arx/nix-bundle/appimage/PyInstaller
* pex
* Rewrite in Go

Native packages seem best. Some thoughts on tooling:

* For Homebrew just make a brew/cask/tap whatever it's called. It's not hard. Maybe copy Flocker's (reverse patch of https://github.com/ClusterHQ/flocker/pull/2755/files includes most of it I think?).
* `fpm` makes building deb/rpm packages nice.
* https://github.com/alanfranz/fpm-within-docker wraps `fpm` in OS-specific docker images to build per-distro packages.
* https://github.com/krobertson/deb-s3 can make apt repo on s3.
* Probably something equally reasonable can be found for RPMs.
* https://kevinlaw.info/blog/distributing-flask-as-a-package.html explains how to combine Travis-CI and Bintray.

probonopd:
Providing an [AppImage](http://appimage.org/) would have, among others, these advantages:
- Works for most Linux distributions (including Ubuntu, Fedora, openSUSE, CentOS, elementaryOS, Linux Mint, and others)
- One app = one file = super simple for users: just download one AppImage file, [make it executable](http://discourse.appimage.org/t/how-to-make-an-appimage-executable/80), and run
- No unpacking or installation necessary
- No root needed
- No system libraries changed
- Works out of the box, no installation of runtimes needed
- Optional desktop integration with `appimaged`
- Optional binary delta updates, e.g., for continuous builds (only download the binary diff) using AppImageUpdate
- Can optionally GPG2-sign your AppImages (inside the file)
- Works on Live ISOs
- Can use the same AppImages when dual-booting multiple distributions

[Here is an overview](https://github.com/probonopd/AppImageKit/wiki/AppImages) of projects that are already distributing upstream-provided, official AppImages.

There are quite some Python-based AppImages, too. They bundle Python, and AppRun sets `$PYTHONHOME` to point to the Python inside the AppImage, so that the Python application is fully self-contained.

itamarst:
@probonopd thanks for the info! I recommended AppImage to someone else recently, but in the end decided it's not suitable for our particular use case.

PR #73, opened 2017-03-29
Teach Telepresence how to read K8s resource files.

### LINKED ISSUE #9: Option to parse existing Deployment object to generate the one for telepresence
Given an existing `Deployment` it should be possible to extract environment variables and the like for the Telepresence `Deployment`.

kflynn:
Fixes #9, I believe.

- `-f` understands multiple-document YAML, but you can also give it multiple files and it'll merge them all.
- Anything telepresence creates, it'll tear down on exit.

Automated tests coming shortly, but here's the PR for review.


itamarst:
Unfortunately this can't be merged unless we either stop using third party modules (might be able to do YAML->JSON with `kubectl` somehow), or fix the relevant part of #45. So going to put off looking at this for now.

itamarst:
Re optional imports, I really don't want to expose Python packaging requirements to users, so not supporting yaml by default still isn't good enough of a UX. Since Docker will soon no longer be a dependency we're at least moving closer to ability to have nicer packaging.

itamarst:
Telepresence installation now would allow us (with a bit more work, e.g. using https://pex.readthedocs.io/en/stable/) to support 3rd party libraries, so this is no longer blocked. If this matters to you immediately you can update the patch, otherwise will revisit at some point.

itamarst:
I expect to supersede this with design in #9, but might be able to reuse a bunch of the code.

itamarst:
Going to get some functionality with somewhat different code, so closing this.

PR #98, opened 2017-04-12
Context flag support

### LINKED ISSUE #3: Support specific Kubernetes context, rather than one that is currently set
Currently `telepresence` uses the currently set k8s context. It should be possible to specify it on the command line:

```console
$ telepresence --k8s-context=minikube ...
```

itamarst:
Fixes #3. Courtesy of @svend.

### DIFF
--- a/cli/telepresence
+++ b/cli/telepresence
@@ -152,6 +152,14 @@ def parse_args():
         )
     )
     parser.add_argument(
+        "--context",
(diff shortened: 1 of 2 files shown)


itamarst:
Unfortunately writing reasonable automated tests for this is... tricky. So just tested manually for now.

PR #142, opened 2017-05-05
Run as non-root user

### LINKED ISSUE #131: Support running the remote image as non-root
```shellsession
```

```shellsession
sh-4.2$ id
```

itamarst:
Fixes #131.

@elyscape will this approach work for OpenShift?


elyscape:
It should but I have a simpler version that will also work. I'll submit a PR shortly.

elyscape:
Actually, this won't quite work. The call to `runner.check_call()` will fail, and I'm not sure about sshfs. I'm testing that now.

elyscape:
Hmm, it looks like `sshfs` isn't working because it relies on `sftp`. I'm looking into a workaround.

elyscape:
Ultimately, `UsePrivilegeSeparation no` doesn't help if you need to run commands or use a subsystem, and we need to do both. The upshot is that we need to get `getpwnam()` to return the current UID. Since nss_wrapper doesn't exist on Alpine, I'm falling back to the non-ideal solution of allowing the runtime user to write to `/etc/passwd` and having a script that runs at container startup that adds a passwordless user with that UID to the file. See #143.

itamarst:
Superseded by the simpler version in #144.

PR #143, opened 2017-05-05
Run Docker container as non-root user

### LINKED ISSUE #131: Support running the remote image as non-root

elyscape:
Fixes #131.

### DIFF
--- a/cli/telepresence
+++ b/cli/telepresence
@@ -502,7 +502,7 @@ def ssh(runner, ssh_port, args):
         "-N",
         "-p",
         str(ssh_port),
-        "root@localhost",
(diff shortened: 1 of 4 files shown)


elyscape:
@itamarst There appears to be something weird going on in the Travis builds. Have you done something today that could have changed or removed the environment variables used to decrypt the service key?

itamarst:
This is expected, if you set secure env variables they won't work on PRs from forks. I'll look at this Monday - thanks for the PR!

itamarst:
Merging into non-master branch so we can do PR that will actually run tests correctly.

PR #144, opened 2017-05-08
Don't run as root

### LINKED ISSUE #131: Support running the remote image as non-root

itamarst:
Fixes #131.

@elyscape this should actually run the tests.


itamarst:
This is definitely simpler, and seems to work. Hitting same issue as my branch did, though, whereby some tests rely on running as root, so will need to tweak tests somehow.

tests/disconnect.py
```diff
@@ -17,7 +17,7 @@ def disconnect():
         "kubectl", "exec",
         "--container=" + os.environ["TELEPRESENCE_CONTAINER"],
         os.environ["TELEPRESENCE_POD"], "--", "/bin/sh", "-c",
-        r"kill $(ps xa | grep 'sshd: root' | " +
+        r"kill $(ps xa | grep 'sshd: ' | " +
```
elyscape: This can be simplified into `pkill -f 'sshd: '`.

PR #173, opened 2017-06-12
Alpine Linux apk packages

drslump:
Builds client .apk packages for Alpine Linux 3.5 and 3.6. This distro is very popular for containers given it's small footprint.

The publishing/uploading hasn't been tested but hopefully it works straight away 😳 

Note that for building on Alpine it first generates a local image with the FPM requirement.

### DIFF
--- a/packaging/Dockerfile.alpine
+++ b/packaging/Dockerfile.alpine
@@ -1,0 +1,4 @@
+ARG DISTRO
+FROM alpine:$DISTRO
+RUN apk --no-cache add ruby ruby-bundler ruby-dev ruby-rdoc ruby-irb libffi-dev build-base tar git bash
+RUN gem install fpm
--- a/packaging/build-package.sh
+++ b/packaging/build-package.sh
@@ -4,13 +4,15 @@
 #
 # Inputs:
 # $PACKAGE_VERSION is the package version to use.
-# $PACKAGE_TYPE is rpm or deb.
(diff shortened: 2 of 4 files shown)


itamarst:
Thanks for the patch!

Unfortunately at the moment we're using https://packagecloud.io for package hosting, and they don't support APK. So to make this similar to Deb and RPM support we'd need to:

1. Create code to manage an Alpine repository.
2. Change upload code to use that code.
3. Create infrastructure to manage Alpine packages.
4. Write appropriate docs.

I'm wondering if perhaps it's better for me to focus first on first-class Docker support for Telepresence. That would probably cover most use cases for running Telepresence on Alpine: I doubt many people run their main computer environment on Alpine. Make sense?

drslump:
Ooops I just assumed packagecloud.io would support arbitrary binary uploads. As you say final users won't be running Alpine, the idea behind this was to have ready `telepresence` for running from inside a container for *first-class Docker support* track (`wget` + `apk add file.apk`).

What do you think if I modify the publishing script so it sends the Alpine binaries to GitHub releases?

itamarst:
If we're just publishing a Docker image (`docker run datawire/telepresence-local` say) then it doesn't even need to be an APK. So I suggest not bothering with further work for now. I will however import this branch into the main repository so it's not lost, since it might be useful. Sound OK?

drslump:
sounds good! 👍

PR #176, opened 2017-06-13
Ignore local ssh config

drslump:
Solves https://github.com/datawire/telepresence/issues/174

Additionally there are some improvements for troubleshooting by enabling verbose logging on ssh and kubectl with the `--verbose` flag.

### DIFF
--- a/cli/telepresence
+++ b/cli/telepresence
@@ -14,7 +14,7 @@ import signal
 import socket
 import ssl
 import sys
-from functools import wraps
+from functools import wraps, partial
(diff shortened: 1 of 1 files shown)


itamarst:
This looks great, thank you!

I need to merge this into a branch in original repo in order to run tests, so I can make any necessary changes. The things I will do are:

1. Not using `global`. It's really a last resort feature, it's mutate-y enough that I prefer not to use it.
2. Add a changelog entry.

PR #181, opened 2017-06-19
Docker support

### LINKED ISSUE #175: First class support for Docker containers
> Related to the discussion on https://gitter.im/datawire/telepresence?at=593eaed0c59bd9c4640b6f8d

## Preface

Given the current maturity of local Docker solutions, for many projects, developing against a local container offers more benefits than drawbacks. Not only can a developer benefit from having an execution environment that closely matches the production one, it also helps aligning it with the environment on CI which is a common source of pain specially when tuning a build procedure.

Moreover, with the current trend of having many *small projects* to map the *micro services* architectures running on production, it becomes very important to improve the on-boarding experience for new hires or existing company personal, allowing to isolate arbitrarily complex build schemes from the peculiarities of personal computers.

## Audience

Developers that work on different projects, probably based on different technologies, and that see containerization as a means to simplify their development setup effort but won't give up on having a quick feedback loop while developing.

## Problem

Currently `telepresence` cannot directly proxy a Docker container which forces the developer down an uncomfortable path of modifying their containers so they run `telepresence` from inside Docker. Even if doable the UX is not great.

Subjectively it's hard to grasp why a tool specifically designed to simplify the development with *container orchestrators* doesn't offer great support for their atomic units. It certainly came as a surprise to me since like many others I've meet, we started working with containers locally before stepping into  tools like Kubernetes.

## Hypothesis

@itamarst proposed a clever approach to solve the issue at the lower level, yet without a great UX but something easy enough to mechanise if technically viable.

Building on top of Docker's support for sharing a network namespace among different containers (similar to Kubenertes' Pod concept), it becomes possible to isolate the telepresence VPN setup from the developer's container. This is a great solution since it not only hides the magic via UX, it actually offers guaranties that the proxying setup won't interfere with your own program.

## Proof of concept

After putting together a quick test against a kubernetes cluster on AWS and minikube, the approach seems to work reliably and even solves some of the caveats with running the proxying directly on the host machine:

 - `.local` DNS resolution works properly since we can control the *resolv* configuration on the proxying container
 - `minikube` is also fully supported and shows no problems with DNS resolution
 - Cloud resources seem to be properly routed without requiring `--also-proxy` (haven't tested `externalName` yet but I don't see why it wouldn't work)
 - Multiple proxying containers can be launched on a single host
 - Proxying container requires some extra capabilities but not full `--privileged` mode
 - And obviously, the proxying doesn't interfere with other software running on the host :)

The solution is really simple and elegant, just requires a proper UX to make it useable at the same level as the other approaches.

## Proposal

Some random thoughts about how this could be implemented:

 - Define a new `--method` named `container` since the caveats of `vpn-tcp` do not apply, so it'll be less confusing in the docs. Also it can come handy in the future if we can find a better vpn software (i.e. UDP support).
 - `--run-docker [docker flags] <image> [program flags]` seems a sensible UI being an alias for `docker run *args`. It would complain if `--method` is defined but it is not `container`. Also paves the way for a `--run-rkt` command.
 - Proxying container is launched (random name suffix?), cluster is updated, then the developer's container with the shared network and kube's config mount.
 - publish an optimized image with all the requirements for the proxying container (`telepresence`, `kubectl`, `oc`?)
 - telepresence will need to mount the `$KUBECONFIG`, taking into account that it can point to external files (i.e certs). Maybe we can use `kubectl` on the *host* to generate a materialized dump of a specific context?
 - workaround for core tools (i.e. ping) should not be used. They seem to work properly.
 - How to sync the life cycle of the proxying container and the developer's container? it's important to clean up properly.
 - This is more general but setting up and tearing down the VPN and volumes takes a bit of time. It'll be great if there is some trick to speed it up for repeated executions.

Note that while the proposal mechanizes the proof of concept approach, perhaps a more sensible solution would be to do all the `kubectl` interaction from the host, splitting the vpn setup code so it can be run on its own inside the container.

itamarst:
Fixes #175.


itamarst:
Mac error seems unimportant, will investigate tomorrow.

PR #182, opened 2017-06-20
Update telepresence-deployment.yaml

danisla:
Version bump


itamarst:
Thanks! BTW, I would recommend using `--swap-deployment` in lieu of manually editing the YAML (it was added after the GCP tutorial was written). At some point I should go back and update the GCP tutorial to do that.

PR #199, opened 2017-06-26
Add back support for ports smaller than 1024

### LINKED ISSUE #194: Telepresence can't expose ports < 1024
Telepresence no longer runs as root, so can't expose ports <1024.

Solution:

1. If `--expose P` has P < 1024, run as root.
2. If possible, detect clusters that don't support this and complain in advance, rather than failing after long timeout.

itamarst:
Fixes #194.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,6 @@
 language: python
+python:
+  - "3.5"
 sudo: required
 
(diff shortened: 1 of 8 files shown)


itamarst:
Included in #200, which has some more fixes to make Travis happy.

PR #219, opened 2017-07-05
Document limitations on startup.

### LINKED ISSUE #217: VPN breaks telepresence
See https://github.com/datawire/telepresence/issues/210#issuecomment-312868407

A priori I would expect starting a VPN *after* telepresence is started in `vpn-tcp` mode to break DNS.

itamarst:
Fixes #217.


itamarst:
I guess this doesn't really fix #217, just documents it.

PR #238, opened 2017-07-22
Allow to override service net when heuristics fail.

farcaller:
This allows to override the service net via env's `KUBERNETES_OVERRIDE_SERVICE_NET` in case it's not /16.


itamarst:
Thanks for the patch! I think I'll approach this is a different way (see #243) but will get this PR open until I've fixed it or decide to take up this approach.

PR #271, opened 2017-08-17
Add sharing buttons to documentation pages and add css to email signup

sipplified:


sipplified:
This pr adds some css to make the email subscription look better as well as adds sharing buttons to the documentation pages. You can preview it here: https://deploy-preview-271--telepresence.netlify.com. To complete it someone needs to go into the hubspot form and add a placeholder for the email input, otherwise there is no label for the field.

richarddli:
Looks good. A couple comments:

1. Can we have the link sharing buttons only show up on tutorial pages?
2. Can you add a text blurb that says (before the buttons) "If this tutorial was useful to you, we'd appreciate it if you'd share it online." (and then add the buttons)

kelseyevans:
Added placeholder in HubSpot form

PR #281, opened 2017-08-30
Update scout reporting for new_install and future expansion

ark3:


plombardi89:
I've implemented a similar `new_install` marker in scout library. I need to make a implementation decision however wrt to data types for the metadata map. The backend store only handles Map<String,String> for the metadata map (for simplicity reasons). I think I can expand it to support richer types but I had some issues doing that in the past. If I can't reasonably support a more diverse set of types using lowercase "true" and "false" would be preferable to sending Boolean True|False over the wire which then get converted to uppercase "True" and "False" by the backend before writing to database. I'd prefer to avoid normalizing metadata to all lowercase in case it's significant on the backend side. 

All food for thought. I need to think through these issues before we merge this.

ark3:
Probably best to "update" to Scout 0.2 once that's ready.

ark3:
May I land this (updated) PR for this release @plombardi89? I'll upgrade to Scout 0.2 in a separate PR, possibly after figuring out updated build/packaging stuff.

PR #283, opened 2017-08-31
Use built-in venv instead of virtualenv to avoid dependency

ark3:


ark3:
This approach isn't fun. I'm going to try to address the associated issues by fixing packaging/delivery instead. Dropping this PR.

PR #313, opened 2017-10-07
Add support for deployments using `automountServiceAccountToken: false`

dhensen:
We are using non default deployments. Meaning that we prevent the default service account to be mounted. This makes `/var/run/secrets/kubernetes.io/serviceaccount/namespace` unavailable to be used by forwarder.py.


dhensen:
Oh build failed. I was hoping I could use your CI, but ofcourse encrypted env vars are not available because this PR comes from an external repo... Can we somehow work around this?

itamarst:
Someone with commit access will have to merge it into branch and open another PR. So just wait until they're back at work next week (Monday or Tuesday) and they can do that.

ark3:
Merging #314 instead. Thank you for this PR!

PR #314, opened 2017-10-07
Add support for deployments using `automountServiceAccountToken: false`

ark3:
This is the main repo version of #313.

### DIFF
--- a/cli/telepresence
+++ b/cli/telepresence
@@ -1295,7 +1295,6 @@ def swap_deployment(runner: Runner,
         TELEPRESENCE_REMOTE_IMAGE,
         args.method == "vpn-tcp" and args.in_local_vm,
         args.needs_root,
-        args.namespace,
     )
     apply_json(new_deployment_json)
     return deployment_name, run_id, orig_container_json
@@ -1308,7 +1307,6 @@ def new_swapped_deployment(
     telepresence_image: str,
     add_custom_nameserver: bool,
     as_root: bool,
-    namespace: str,
 ) -> Tuple[Dict, Dict]:
     """
     Create a new Deployment that uses telepresence-k8s image.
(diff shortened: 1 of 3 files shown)


dhensen:
Thanks, I'm coming back with fixed tests soon.

dhensen:
I fixed a test, and ran tests locally. Can this branch be synced with mine? Some tests were already failing before my PR.  Thanks!

dhensen:
Nice, at least the linux build had success. I'm hoping to get this merged soon because we are doing a telepresence integration and are blocked on this now. Thanks for mirroring my PR :+1:

ark3:
The macOS breakage seems like a configuration issue unrelated to your change. I may end up ignoring that here and fixing it elsewhere. As far as the timeline goes, I'd like to land this and a few other small things, then release in the next day or so, definitely by the middle of the week. 

I forgot to say this before: Thanks very much for your PR! Can you tell us about your use case? Telepresence development is driven by users like you.

dhensen:
Thanks a lot, looking forward to this being merged. You're welcome, it's awesome that telepresence exists and I'm happy to contribute.

Our use case is that we are developing [riddles.io](https://www.riddles.io/) on top of kubernetes. We have built our own tooling to support our workflow, packaging, building images, publishing them etc.

One of the things the tool can do is run a given microservice locally from source or run the container locally. Both options make the microservice run against a k8s cluster environment. Basically it means that we connect to backing directly, such as mysql, rabbitmq, redis, etc. Because we are then connected to rabbitmq, our microservice runs alongside other similar microservices transparently. 

This works fine for us, but it requires us to keep local application configuration in sync with that of the cluster environment. Also we have to fake some folder mounts and because in one case we are not running inside a container, we have a different startup script than for each microservice not running locally inside a container. I'm one of two guys who knows this bigger picture at our company, but having to manage all this is taxing on developers. That's why we built our own tool.

Now with telepresence, I can basically cut away all the environment variable injecting and just swap deployment :+1:

ark3:
@dhensen Setting aside MacOS Travis issues, I've landed this PR. Thanks for your help! If you send me (ark3@datawire.io) an email with your t-shirt size and shipping address, Datawire would be happy to send you a Telepresence t-shirt.

PR #324, opened 2017-10-12
Add support for deployments containing Services of type ExternalName

nikovanmeurs:
We have some services in our cluster of type ExternalName. These are essentially service definitions pointing to external services. As such, the spec does not contain a `clusterIP` field, but an `externalName` field instead. Telepresence currently crashes on these service definitions because it expects a service spec to always have a `clusterIP` field.

### DIFF
--- a/cli/telepresence
+++ b/cli/telepresence
@@ -1816,7 +1816,7 @@ def get_proxy_cidrs(
         )["items"]
         return [
             svc["spec"]["clusterIP"] for svc in services
-            if svc["spec"]["clusterIP"] != "None"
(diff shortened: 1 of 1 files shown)


nikovanmeurs:
It seems CI fails because this PR comes from an external repo.

ark3:
Thanks for the fix! I'll get this into 0.69. 

Separately, we're working to make external pull requests perform useful CI without Travis secrets.

nikovanmeurs:
Awesome! Is there an eta for 0.69?

ark3:
I'm targetting next week.

ark3:
Merging #329 instead. Thank you for the PR!

Issue #193, opened 2017-06-22
--run-shell doesn't need to suppress shell startup script in vpn-tcp mode


itamarst:
`inject-tcp` needs to suppress shell startup script on OS X since DYLD_INSERT_LIBRARIES breaks Go programs like `kubectl`. `vpn-tcp` does not have that problem.

donnyyung:
I believe this is no longer an issue in Telepresence 2. Here are the docs on how to install Telepresence (https://www.telepresence.io/docs/latest/install/), please re-open if you still see this issue in our latest version!

PR #338, opened 2017-11-03
PS1 environment functions

azatoth:
I've __git_ps1 in my PS1, and `. /etc/bash_completion` is run from `~/.bashrc`, and because telepresence is running bash with --no-rc this file is not executed and thus I'm getting "bash: __git_ps1: command not found"

Would it be possible to have a `~/.telepresencerc` instead that could include relevant stuff?


ark3:
Thanks for the PR. We would like to improve the run-shell features such that running bash with `--no-rc` is only done when necessary (#193). In the meantime, you can achieve a similar result by using `telepresence --run bash` in place of `telepresence --run-shell` and detecting Telepresence in your dotfiles by looking for the `TELEPRESENCE_ROOT` environment variable.

I'm closing this PR. Again, thank you.

PR #405, opened 2018-01-12
Move publish args from docker run into the network container (fixes #387)

### LINKED ISSUE #387: (enhancement) expose a container port locally while telepresenc-ing
It would be useful to have the option to specify port(s) to expose *locally* in combination with telepresence's `--docker-run` option (all other traffic being proxied as usual).

I believe this is currently unsupported (I tried using `--docker-run` in combination with the `--expose` or `-p` options for `docker run`):

```
docker: Error response from daemon: conflicting options: port publishing and the container type network mode.
```

The use case I have in mind is to connect a development tool to a process running in the local telepresenced container.

Specifically: I have a Clojure web service running in a container, if I use telepresence with `--docker-run` then I am unable to connect a local `nREPL` client to it. If I use `--method=vpn-tcp` and run the Clojure process directly on the host, then this is possible, but this is not ideal.

ark3:


telepresence/container.py
```diff
@@ -78,6 +78,13 @@ def run_docker_command(
     remote_env["TELEPRESENCE_ROOT"] = mount_dir
     remote_env["TELEPRESENCE_METHOD"] = "container"  # mostly just for tests :(
 
+    # Extract --publish flags and add them to the sshuttle container, which is
+    # responsible for defining the network entirely.
+    parser = argparse.ArgumentParser(allow_abbrev=False)
+    parser.add_argument("--publish", "-p", action="append", default=[])
+    publish_ns, docker_args = parser.parse_known_args(args.docker_run)
+    publish_args = ["-p={}".format(pub) for pub in publish_ns.publish]
```
exarkun: `run_docker_command` is probably already too long.  Maybe this new logic could go into a separate  function that returns `publish_args` and `docker_args`?  Not the nicest signature but it would keep the size of `run_docker_command` down and make it clear that `publish_ns` isn't used anywhere except in this logic.

Issue #363, opened 2017-12-12
Add copy-deployment and swap-service as a way to use Telepresence


ark3:
See also #257 and PR #407

ofpiyush:
Our use case/experience report:

We already have openvpn added to our cluster. Simplifying that would be sweet but isn't required as such. https://github.com/pieterlange/kube-openvpn

Having a separate openvpn instance allows us to add other kind of urls to be forwarded without needing `--also-proxy`.
eg: `c.<project_id>.internal` urls with GCP.

Separating out the VPN part makes it straight forward to have multiple different services running locally. It also makes the network very very stable. I run into cases where urls sometimes do not get proxied with telepresence. Those class of issues can be completely avoided by allowing a separate VPN setup.

If you baked that configurability into telepresence, that would be cool.

I envision separate `telepresence config(s)` tied to `kubectl config current-context`.

Different kind of needs we regularly run into:

1. Add a local process as a pod parallel to the ones running, some traffic comes to my pod. This has added benefit of being useful in really bad cases of production debugging as well.

2. Let the cluster pods deal with partial background functionality, redirect all active service traffic to my local.

We have rabbitmq listeners etc in some of our services, they can continue to operate in the cloud while we debug some API level stuff/dev locally, but we want requests to the service to come to the local instance.

3. Stop everything and let us completely take over. This is the current behaviour of `--swap-deployment`

We would not be opposed to having an always on, telepresence deployment with it's own URL and such if that made things easier, esp around networking and managing everyone's connection / multiple swappable shells.

This central deployment would also help you to keep 1 source of truth and return the cluster to it's original state if things went sideways on someone's local.

donnyyung:
This seems to be pretty much how Telepresence 2 works. Here are the docs on how to install Telepresence (https://www.telepresence.io/docs/latest/install/), please re-open if you still see this issue in our latest version!

PR #407, opened 2018-01-16
[WIP] Copy Deployment

ofpiyush:
WRT: #363

Waiting for @ark3 to move things around.

<details><summary>Original comment</summary><p>

### Planned steps
- [x] Make POC for `--copy-deployment` flag on k8s.
- [x] Open PR and ask about it's direction.
- [ ] Write tests for k8s
- [ ] Figure out what needs to be done for openshift.
- [ ] Add docs.
- [ ] Release 🎉 

Feel free to add steps that I have missed.

PS: this will completely break for open shift clusters right now as I do not know how to work with them yet. 
If it looks like the correct direction for k8s, I will start working on that and open shift

PPS: there is some need for refactor here, but I'd hold it off till we have the direction etc look solid.
</p></details>

### DIFF
--- a/telepresence/cli.py
+++ b/telepresence/cli.py
@@ -231,9 +231,24 @@ def parse_args(args=None) -> argparse.Namespace:
             "Swap out an existing deployment with the Telepresence proxy, "
             "swap back on exit. If there are multiple containers in the pod "
             "then add the optional container name to indicate which container"
+            " to use.\n"
+            "This automatically sets --forward-traffic to true"
+        )
+    )
+
+    group_deployment.add_argument(
+        "--copy-deployment",
+        "-c",
+        dest="copy_deployment",
+        metavar="DEPLOYMENT_NAME[:CONTAINER]",
+        help=(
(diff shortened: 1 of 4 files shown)


ofpiyush:
The more I think about it, the more it looks like swap deployment can be implemented with `--copy-deployment` + 0 replicas on the original deployment.

It has an added benefit of leaving the original deployment there in case we died and can't replace it.

One just needs to set replicas back to their original number.

ofpiyush:
I've made some changes to make `swap-deployment` behaviour = `copy-deployment` + `forward-traffic` + resize original deployment to 0 replicas.

If that doesn't sound good, I can switch back to how things were and add `copy-deployment` as a completely separate command.

ark3:
I like where you're headed with this. We can help with Openshift stuff, so please keep going. I should have more time to help you tomorrow (Friday).

ofpiyush:
Want all the tests to run, closing and resending a PR

ark3:
I'm closing this stale PR for now, but let's keep #363 open as a reference/link to this work. Thanks for your contribution!

PR #409, opened 2018-01-16
Make the build script perform linting like the old ci/lint.sh script

ark3:


build
```diff
@@ -207,19 +214,18 @@ def _push(registry, version):
                 image,
             )
 
+
 def _parser():
-    parser = ArgumentParser(
-        description="Test Telepresence",
-    )
+    parser = ArgumentParser(description="Test Telepresence", )
```
exarkun: I really liked the trailing comma on line 172.  Not so much a fan of this one. :)  Seems like a funny thing for yapf to do... Let's get rid of it, though?  yapf seems happy enough for every other function call to not have such a trailing comma...

Issue #378, opened 2017-12-15
Add `--no-fs` or similar option to skip the sshfs step


ark3:
Skipping the sshfs step may speed up startup.

Issue #411, opened 2018-01-17
Project based configuration


tristanpemble:
I would like to define and version control a configuration file that Telepresence can use, similar to Forge's `service.yaml`.

Ideally if I am fully bought into the Datawire ecosystem, my developers can do something as straight forward as this to begin development with Kubernetes:

```
git clone git@github.com:my-org/my-project.git
cd my-project
minikube start
forge deploy
telepresence run web
```

and begin editing code locally to see changes as they work. being able to define different scenarios would be helpful -- basically all the same usecase for Telepresence:

- running a daemon from your local project folder
- entering an interactive shell for debugging
- etc.

it might also be worth considering letting Telepresence integrate with Forge, somehow. that way we could reuse the environment configuration/Jinja2 variables to configure Telepresence.

this type of use case also might be better suited for another tool. I was considering writing a simple wrapper if it does not seem like an appropriate addition to Datawire's toolset.

this could easily be done with a Makefile. I created this issue for a discussion generally on how Datawire sees bridging a gap between these tools

ofpiyush:
I would also love some variant of this. We've not tried forge out yet, but the ability to set 1 global and 1 local config file is high in our list of needs as well.

See #363

richarddli:
The use case makes perfect sense. We've been exploring ways to integrate Forge and Telepresence together. Our users seem to be pretty happy with the Forge/Ambassador integration which uses Jinja2 etc like you propose.

If you're following the commits right now, we're mostly working on speeding up the development cycle for Telepresence, which requires disentangling us from Travis and revamping the test suite. In the meantime, we can use this issue to discuss requirements & design in more detail.

donnyyung:
I think forge has been mostly discontinued? But otherwise, Telepresence 2 does have a notion of a config file: https://www.getambassador.io/docs/telepresence/latest/reference/config . Here are the docs on how to install Telepresence (https://www.telepresence.io/docs/latest/install/), please re-open if you still see this issue in our latest version!

PR #412, opened 2018-01-18
adds `--no-fs`  flag

ofpiyush:
Adds #378

### DIFF
--- a/telepresence/cli.py
+++ b/telepresence/cli.py
@@ -246,6 +246,14 @@ def parse_args() -> argparse.Namespace:
             "the run subprocess will be proxied."
         )
     )
+    parser.add_argument(
+        "--no-fs",
+        dest="no_fs",
+        action="store_true",
+        help=(
+            "If you do not wish to mount volumes, use this flag."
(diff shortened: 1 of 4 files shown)


ark3:
Looks good! Any thoughts on how to test?

ofpiyush:
@ark3 It's slightly difficult to test as the temp folder has a random name.

One solution, that would also work as a larger separation of concerns/ refactor, is to have  a `Config` object and make the configuration step completely separate from running.

This configuration step would do things like handle precedence of config files and flags #411, and in context of temp files, set a random name for root temp folder for our process in a format similar to, say `/tmp/telepresence-<run-id>/`

Once configuration step is over, the `Config` object must be treated as strictly immutable.

Then mount path etc can have a fixed relative path from the random tmp root path.

In case of tests, we can set our config object and know for sure where the files will be, what the <run-id> of remote deployment will be and so on...

ark3:
We finally landed a variant of this in the form of #610 `--mount`. Thanks for your help.

Issue #289, opened 2017-09-12
Allow other shells to join existing telepresence session.


james-atwill-hs:
With `vpn-tcp`, my whole machine is teleported into the pod. It would speed up debugging if I could open another shell and have it "join" the existing session such that the command I run has `PRESENCE_ROOT` and the various other Kubernetes environment variables without having to do the heavy lifting of VPN again.

I imagine a usecase where I do `telepresence --swap-deployment myservice --run-shell` in a shell, but in my IDE I run my application as `telepresence --join ./myapp`.

ark3:
See also #257.

donnyyung:
I believe this is no longer an issue in Telepresence 2. Here are the docs on how to install Telepresence (https://www.telepresence.io/docs/latest/install/), please re-open if you still see this issue in our latest version!

PR #413, opened 2018-01-18
Allow running telepresence without any network

ofpiyush:
WRT: #289, comment on #363 

Waiting for @ark3 to move things around.
<details><summary>Original comment</summary><p>

I am not entirely sure if this is the correct way to achieve that goal.

Please advise

### What this PR does

* This PR adds a method called "none"
* When selected, no attempt of forwarding connections/DNS is made.

### Why?
* If I have the a VPN setup with my cluster, I have to switch it off for telepresence.
* When I had to work on a feature across multiple dependent go programs, I had no way of proxying all of them without risking network issues.

### The fix
If I already have a VPN, I can run all telepresence instances with this method.

If I don't, I  can run 1 telepresence instance locally with `vpn-tcp` and other instances with `none` to allow multiple shells for multiple deployments.

</p></details>

### DIFF
--- a/telepresence/cli.py
+++ b/telepresence/cli.py
@@ -196,7 +196,7 @@ def parse_args(args=None) -> argparse.Namespace:
     parser.add_argument(
         "--method",
         "-m",
-        choices=["inject-tcp", "vpn-tcp", "container"],
+        choices=["inject-tcp", "vpn-tcp", "container", "none"],
         help=(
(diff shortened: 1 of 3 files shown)


ark3:
I'm closing this stale PR for now, but let's keep #289 open as a reference/link to this work. Thanks for your contribution!

PR #415, opened 2018-01-18
Switch OSX CI from TravisCI to CircleCI

### LINKED ISSUE #153: Switch Travis CI to use build stages
Right now Linux and OS X builds run in parallel. Since the OS X build depends on the Linux build, this only works because OS X builds are so slow and so always get scheduled after the Linux one.

We should instead be doing build stages, now that Travis supports them: OS X after Linux. Need to figure out how to this works re releasing packages.

### LINKED ISSUE #397: Enable CircleCI builds
Travis builds, particularly on OS X, can take an unreasonable amount of time (particularly in the afternoon when builds queue up and it might be dozens of minutes before one even starts).

CircleCI also offers OS X support.  It seems worth investigating further, including gaining some real-world experience with that system, to determine whether it might provide a better experience than the current travis-based experience.

exarkun:
Fixes #397 
Fixes #153 

Stacked on #414 - review that first.

This also includes groundwork for moving Linux to CircleCI but the state of OSX on TravisCI seems to warrant getting that part of the change landed as quickly as possible.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,6 +10,11 @@ jobs:
       - "checkout"
 
       - run:
+          name: "dump git tags"
+          command: |
+            git tag
+
+      - run:
           name: "decrypt gcloud keys"
           command: |
             # To update:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,7 +14,7 @@ jobs:
   build-image:
     machine:
       enabled: true
-      image: "circleci/python:3"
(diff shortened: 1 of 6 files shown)


exarkun:
Hm.  Out of repo branch means TravisCI doesn't trigger.  Probably would be nice to be able to see those results...

exarkun:
Ugh.  GitHub helpfully gave me the wrong base branch.

PR #416, opened 2018-01-18
Switch OSX CI from TravisCI to CircleCI

### LINKED ISSUE #153: Switch Travis CI to use build stages

### LINKED ISSUE #397: Enable CircleCI builds

exarkun:
Fixes #397
Fixes #153

Stacked on #414 - review that first.

This also includes groundwork for moving Linux to CircleCI but the state of OSX on TravisCI seems to warrant getting that part of the change landed as quickly as possible.


exarkun:
Until Datawire CircleCI is configured for OSX, OSX results can be seen https://circleci.com/gh/exarkun/workflows/telepresence/tree/397.circleci-investigation

environment-setup.sh
```diff
@@ -4,7 +4,7 @@
 # suite.  These steps should typically only be required once to prepare the
 # environment.
 
-if [ "$#" -ne 3 ]; then
+if [ "$#" -ne 4 ]; then
     echo "Usage: $0 <gcloud project name> <gcloud cluster name> <gcloud compute zone>"
```
ark3: Need to add osx|linux to usage

exarkun:
A couple additional changes were required:

  * The gcloud key was encrypted for my personal CircleCI account.  I had to re-encrypt it with keys available to the Datawire CircleCI account.  This was a trivial change.
  * Apparently the CircleCI builds on my account were only really succeeding because a particular version of kubectl was present in one of the caches for the build.  The cache must have been populated by an old version of the branch.  The most recent code in the branch installed a different version of kubectl but this codepath was never followed because of the build cache.  When I did the first build on the Datawire account, the cache did not exist (to that account) so the new install code ran - and got a too-new version of kubectl, so the build failed.  I changed the code to install the same version of kubectl on OSX as on Linux and the build is now happy.

So, merging (as soon as the TravisCI Linux build finishes - exciting to see that the CircleCI OSX build actually completed first!).

PR #418, opened 2018-01-18
CircleCI for Linux

### LINKED ISSUE #417: Use CircleCI for Linux builds
Follow-up for #397 which switched OSX builds from TravisCI to CircleCI.  There were some issues with the Linux build on CircleCI without immediately obvious causes.  Fix those issues and enable CircleCI Linux builds.

Don't disable TravisCI Linux builds yet because they're related to the release process in some way.

exarkun:
Fixes #417 

Stacked on #416 - review that first.


ark3:
Maybe this? https://stackoverflow.com/a/43847627/8060653

> This message digital envelope routines: EVP_DecryptFInal_ex: bad decrypt can also occur when you encrypt and decrypt with an incompatible versions of openssl.

exarkun:
Ah right, those changes.  That was how I tracked down the kubectl / kube-apiserver incompatibility that was initially causing the Linux builds to fail.  Without the extra verbosity, the only report was a nonsensical "no such resource" error from kubectl.  The extra verbosity reveals that the resource being requested is something to do with swagger which apparently 1.6.x doesn't provide.

So ... the change isn't strictly necessary but something like it might come in handy to someone sometime if they need to debug a similar failure.  I'd say it makes sense to drop it from here but track the problem it was addressing in a new issue.  (It's not clear that version mismatch between k8s client and server is our problem but reporting enough information to diagnose such a case could be.)

Issue #391, opened 2017-12-28
Telepresence should send all traffic to the cluster (container method)


ark3:
It seems #320 is also an issue in container mode. We should consider having all container traffic go to the cluster, much like inject-tcp mode works.

ark3:
FWIW, @plombardi89 worked around this issue (and #320) in Kubernaut by putting an IP address in the generated Kubernetes configuration for Kubernaut clusters. This avoids the DNS lookup entirely. Regardless, we should still consider sending all container traffic to the cluster. Among other things, that would avoid some cluster analysis on startup, thereby speeding things up.

ark3:
Plan:
- [ ] Use `kubectl exec` to grab the hostname and resolv.conf of the proxy pod. See `telepresence.remote_env._get_remote_env(...)` for an example. Perhaps even do this in the same place; these are components of the remote environment in some sense.
- [ ] Use the captured info to modify the appropriate `docker run` command with `--hostname` and `--dns*` flags as described in the options link above. You will need to figure out whether these flags need to be used for the network container or the user container (or both).
- [ ] Replace the sshuttle invocation in the network container with one that proxies `0/0` instead of a subset of the IP space. You should not need `--dns` if the above DNS configuration step worked correctly for the user container.
- [ ] Actively disallow (by erroring out) Telepresence options that are intended for the vpn-tcp method.

ptemmer:
How can I direct traffic to a specific IP address to NOT go to the cluster, and instead go to my local machine (which in the end gets routed through a VPN)?

PR #450, opened 2018-02-07
Improve DNS suffix probe robustness

### LINKED ISSUE #393: Travis Mac tests fail intermittently due to environmental issues
DNS requests from the Travis Mac environment have domain names that vary during a single Telepresence run. For example, one request will be for hellotelepresence10.**jobs**.macstadium.blah and the next will be for hellotelepresence11.macstadium.blah. Once Telepresence captures the first request's domain for automatic stripping, the second request yields an NXDOMAIN, which causes either the Telepresence connection or the test to fail.

exarkun:
Fixes #393

### DIFF
--- a/build
+++ b/build
@@ -177,6 +177,7 @@ def _test(registry, version, method, test_env, test_args):
         *test_args,
         "tests/test_run.py",
         "k8s-proxy/test_socks.py",
+        "k8s-proxy/test_forwarder.py",
         test_env=test_env,
     )
 
--- a/docs/reference/changelog.md
+++ b/docs/reference/changelog.md
@@ -2,6 +2,11 @@
 
 #### 0.76 (Unreleased)
 
+Bug fixes:
+
+* Telepresence now makes a greater effort to account for local DNS search domain configuration when bridging DNS to Kubernetes.
+  ([#393])(https://github.com/datawire/telepresence/issues/393))
+
 Misc:
 
 * A new end-to-end test suite setup will help us reduce the cycle time associated with testing Telepresence as we port over existing tests.
@@ -54,7 +59,7 @@ Misc:
   ([#269](https://github.com/datawire/telepresence/issues/269))
 * An install script is included for installing Telepresence from source.
   ([#347](https://github.com/datawire/telepresence/issues/347))
-  
+
 
 #### 0.71 (November 1, 2017)
 
--- a/docs/reference/changelog.md
+++ b/docs/reference/changelog.md
@@ -4,7 +4,7 @@
 
 Bug fixes:
 
-* Telepresence now makes a greater effort to account for local DNS search domain configuration when bridging DNS to Kubernetes.
+* Telepresence now makes a greater effort to account for local DNS search domain configuration when bridging DNS to Kubernetes when using `--method=vpn-tcp`.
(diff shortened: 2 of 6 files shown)


exarkun:
> I see that the DNS server code collects multiple suffixes. Shouldn't the client side try to make multiple successful requests to increase the likelihood that the server has seen all the relevant suffixes?

I think the client side does issue multiple requests (ignoring "successful"), inadvertently.  In just about every Telepresence log I've looked at, there's been `hellotelepresenceN` requests with N of 10 or higher.  I think this is because the DNS probing starts very soon after `sshuttle` is launched and `sshuttle` takes a little while to start up and settle into a working state.  As far as "successful" goes, the DNS proxy doesn't need the request to succeed in order to detect a new suffix.  Any request that starts with "hellotelepresence" will be examined to detect a new prefix, whether it ultimately succeeds or fails.  So, effectively, I think we have multiple probes - but again, only inadvertently due to the time it takes `sshuttle` to start working.

So, I'll add an explicit probe count and require that at least so many probes are issued before we get out of the `sshuttle` is-it-working loop, even if probes start to succeed earlier than this.

> Should the DNS server try to strip off the longest known suffix first?

Aaahhh...  Let's see.

If the search domains are "a" and "b.a" and someone queries for "c" then maybe the resulting query will be for "c.a" or "c.b.a".  For the case of "c.a" we will only try stripping "a" and this yields the "correct" name of "c".  For the case of "c.b.a" we might try stripping "a" or "b.a".  If we strip "a" then we end up with a name of "c.b" which is not what the application searched for (shortest -> wrong).  If we strip "b.a" then we end up with a name of "c" which is the "correct" name (longest -> right).  So far so good.
 
If the search domains are "a" and "b.a" and someone queries for "c.b" then maybe the resulting query will be for "c.b.a" or "c.b.b.a" (I don't _think_ any systems will give you "c.b.a" in this case).  For the case of "c.b.a" we might try stripping "a" and this yields the "correct" name of "c.b" (shortest -> right).  We might also try stripping "b.a" which is not what the application searched for (longest -> wrong).  For the case of "c.b.b.a" we might try stripping "a" or "b.a".  If we strip "a" then we end up with a name of "c.b.b" which is not what the application searched for (shortest -> wrong).  If we strip "b.a" then we end up with a name of "c.b" which is the "correct" name (longest -> right).

So it seems mixed.  In some cases, stripping the longest suffix gives the highest fidelity but in other cases stripping a shorter suffix produces a better result.

What do you think?

> Suffix detection stuff only happens for the vpn-tcp method (and container method for now, because it uses the same code -- see also #391, which posits that all container traffic should go to the cluster, as it does with inject-tcp). We should clarify that in the change log entry.

Good call.  I've made a change that I think clarifies but #391 confuses me a little so let me know if you were thinking something different.

ark3:
In your list of suffix stripping examples, it seems that longest -> right occurs more often, and longest -> wrong only occurs in a case that might not happen in practice. Have I misunderstood? I think the current behavior, choosing at random, is strictly inferior because it's inconsistent. If we have to pick a behavior, longest first seems correct intuitively, even if that isn't strictly true.

rhs:
It feels to me like in practice this won't come up that often regardless of which choice is made, and I do agree that random is not good, so I'm +1 on longest first.

Issue #481, opened 2018-02-28
Make --init optional when using --docker-run


CMajeri:
Hi,

I'm running full system containers that run their own init, and I noticed that my init gets switched in favor of docker's `--init` option.

Could it be possible to make passing that flag optional? Possibly by having the user have to specify `--init` themselves?
If that is considered a big change for existing users, maybe consider a `--no-init` parameter that will just remove that flag?

There doesn't seem to be a need for that flag for telepresence itself, since it can seemingly run on older versions of docker that do not have that flag available.

I can participate to the pull request, depending on the option you guys prefer!

Cheers!

ark3:
I would love to drop the automatic `--init`, but that would most certainly be a breaking change. On the other hand, I am not excited about adding yet another option (`--no-init`) to the Telepresence command line given how complicated things already are (see #455). What do other folks think?

CMajeri:
Does kubernetes even support passing the `--init` flag to containers? In my very biased opinion, since the `--docker-run` option is meant more as a dev tool to connect a container to a cluster, I'd expect the container to run about the same as they would in the cluster, just so I could catch bugs more easily.

CMajeri:
For now, the workaround I'm using is passing `--init=false` to the command, which works but depends on `docker run --init --init=false` parsing the arguments correctly, which it does, at least on my version of docker, but that's not the most reliable thing.

A way I see that doesn't affect current users and lets people turn that feature off would be to look for the `--init` flag in the user specified arguments. If there's a `--init`, maybe drop the automatic `--init`, and just pass the args as is. Such a check wouldn't be too hard to add, and can still be kept somewhat clean, as long as it is clearly documented and we limit the number of such cases. What do you guys think? 
This would also prevent stuff like `docker run --init --init` from happening (which again, depends on docker doing the right thing, although that's not too far fetched for this case).

ark3:
You are correct: adding `--init` automatically makes the container behave differently from how it would act in the cluster. Telepresence does this to handle signals better (Ctrl-C, a signal from your IDE, a `kill` command, etc.), which improves its responsiveness when restarting a session repeatedly. This scenario cannot occur in the Kubernetes cluster. We felt it was okay to make things behave a bit differently in exchange for having a more comfortable developer experience.

[...]

PR #483, opened 2018-03-01
Lets us specify init flags

CMajeri:
As discussed in issue #481, this PR makes the use of `--init` flags when using `--docker-run` more reliable by detecting the use.
---
Before landing
- Update `docs/reference/changelog.md`
- Link to applicable GH issues in the changelog


ark3:
Your implementation (in container.py) is more complicated than I expected. Why modify `parse_docker_args(...)` at all? It seems all we need is an additional test in the `if` statement on (original) line 163. If `docker_args` does not already have an `--init`, then add one.

Sadly there is no straightforward approach to testing if you make the change that way... But that's a symptom of the poor organization of the code as a whole.

CMajeri:
I thought about doing that until you mentioned the `--publish`. I figured it would make more sense to treat all "specfic" parameters in one place (and when/if that becomes too big, move them and change the way they're treated).

I'm okay with moving the if inside the run command, but that makes it untestable as you've said.
And I find it makes the code slightly muddier.

CMajeri:
I refactored it as you asked, it's much simpler and smaller, but unfortunately it can't be tested that way.
I tried it out by `print`ing the `docker_command`, and it seems to do what is expected.

PR #501, opened 2018-03-12
Port swapdeployment tests to new testing framework

exarkun:
Advances #400

### DIFF
--- a/tests/test_endtoend.py
+++ b/tests/test_endtoend.py
@@ -3,11 +3,8 @@ End-to-end tests which launch Telepresence and verify user-facing
 behaviors.
 """
 
-from os import (
-    urandom,
-)
-from base64 import (
-    b64encode,
+from time import (
+    time,
 )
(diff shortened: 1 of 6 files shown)


tests/test_endtoend.py
```diff
@@ -369,3 +338,121 @@ def probe_also_proxy(probe_result, hostname):
...
+
+
+def _get_swap_result(probe):
+    if probe.operation.name != "swap":
+        pytest.skip("Test only applies to --swap-deployment usage.")
+
+    # Telepresence won't try to swap anything back until it believes its job
+    # is done.  So make sure its job is done before we make any assertions
+    # about whether things were swapped back.
+    return probe.ensure_dead()
```
ark3: The name of that method is confusing, even here when being used. The name should express that this is like `probe.get_result()` with an additional guarantee. How about `probe.get_result_after_dead()`?

PR #513, opened 2018-03-19
Fixing Ubuntu install instructions

ianwalter:
Adding `sudo apt update` otherwise you get `E: Unable to locate package telepresence`

(Assuming small doc fixes are not added to changelog. Let me know if they are.)

### DIFF
--- a/docs/macros.html
+++ b/docs/macros.html
@@ -32,7 +32,7 @@ brew install socat datawire/blackbird/telepresence
 
 <h4>Ubuntu 16.04 or later</h4>
 <p>Run the following to install Telepresence:</p>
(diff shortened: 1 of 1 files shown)


ark3:
The PackageCloud script performs the `apt update` for me. What version of Ubuntu are you running?

ianwalter:
Sorry should have checked there first and also my comment is misleading since I'm on Pop_OS! a derivative of Ubuntu (16.04 I think). Not sure why it didn't actually update then. I'll close since it could be specific to my system.

ark3:
I wonder whether you ran into a variant of the core problem I describe in https://github.com/datawire/telepresence/issues/486#issuecomment-369968226. However, running an additional `apt update` doesn't resolve that issue, so I'm not sure what's going on here. If you have time to investigate, we would love to improve the story around installation further.

PR #519, opened 2018-03-21
Caching of cluster info

rhs:
This PR introduces some basic caching of cluster info that is queried whenever telepresence starts. This caching improves startup time for all invocations of telepresence. My non-scientific measurements show the speedup to be roughly in the realm of 5 seconds.

---
Before landing
- Update `docs/reference/changelog.md`
- Link to applicable GH issues in the changelog


rhs:
Are you supposed to assume that `~/.cache` exists or create it if it is not present?

rhs:
I've implemented an arbitrary 12 hour cache invalidation, so I think that makes this PR functionally complete. I suppose hypothetically I could add a command line argument for overriding the timeout, but I'm inclined to wait for someone to ask for it.

The one thing missing (that I'm aware of) at this point is testing of the cache invalidation. Please chime in if you have any brilliant ideas about how to do this.

rhs:
oops, I should also stop ignoring the build failures now that circleci has recovered

PR #520, opened 2018-03-21
Resolve crash on address lookup

### LINKED ISSUE #195: Telepresence segfaults (sometimes) in inject-tcp mode
On my home network (but not at the office), the following will segfault somewhere inside torsocks (both v2.1 and v2.2) in `inject-tcp` mode:

```console
$ python3 -c "import socket; socket.getfqdn('')"
```

exarkun:
Fixes #195 

This also includes an upgrade from alpine:3.6 to alpine:3.7 in order to have Python 3.6.3 instead of Python 3.6.1 in the Telepresence proxy pod.  Python 3.6.1 has a string formatting bug which caused misbehavior in the SOCKSv5 proxy used to support Telepresence's networking features.

A consequence of this change is that the Kubernetes volume must now be found at `$TELEPRESENCE_ROOT/run` rather than `$TELEPRESENCE_ROOT/var/run`.


ark3:
I got this to pass by banging on reload, but this needs some discussion. I'm comfortable landing the Telepresence subset, but we should figure out what we want to do with the torsocks patch beyond just enabling it for Linux test runs.

PR #527, opened 2018-03-23
Answer RESOLVE-PTR with a protocol-compliant error

exarkun:
Relates to #195 

Does not exactly fix the problem but eliminates the most offensive symptom.

See #520 for the complete fix which also depends on getting a new Torsocks out to folks.


ark3:
I'm still seeing a segfault on my machine that has valid reverse DNS. Debug next week?

PR #529, opened 2018-03-23
Use the Linux OS package manager to get torsocks

### LINKED ISSUE #528: Remove the custom-built Torsocks 2.1.0 support with use of distro-packaged Torsocks
The distros currently supported (Fedora 26 & 27; Ubuntu Xenial & Artful) package Torsocks 2.1.0 or newer.  Prefer to configure CI to use the distro package.

exarkun:
Fixes #528


exarkun:
Turns out the OS-packaged version of Torsocks is not acceptable.  We need a newer version.  So this will have to wait.

PR #543, opened 2018-03-28
[Do not merge] Switch CI back to using gcr.io

ark3:
This PR represents the set of changes required to get CI back onto gcr.io, just in case we need it.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -38,7 +38,7 @@ jobs:
           command: |
             ./build --no-tests --lint
 
-  build-image-gcloud:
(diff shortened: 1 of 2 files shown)


LukeShu:
Wouldn't the `REGISTRY = os.environ.get("TELEPRESENCE_REGISTRY", "datawire")` line in `telepresence/__init__.py` also need changed?

ark3:
No, this is just about images for CI.

LukeShu:
CI isn't responsible for pushing the release Docker images?

PR #548, opened 2018-03-30
Replace suffix stripping with fast NXDOMAIN.

### LINKED ISSUE #192: confusing dns behavior
Using telepresence to proxy my local shell into a cluster was awesome, but it took me a few minutes to figure out that it was actually working. This was because when I typed 'host <servicename>', the output is confusing since it reports the correct ip, but the fully qualified hostname reflects the laptop's local configuration and hence references something that doesn't actually exist:

```
[rhs@venture hello-kubernetes]$ telepresence -m vpn-tcp -n foo
Starting proxy...
[sudo] password for rhs: 
@gke|bash-4.3$ host users
users.wework.com has address 10.91.248.50
users.wework.com has address 10.91.248.50
@gke|bash-4.3$ 
```

exarkun:
Fixes #192 

master behavior:

> $ telepresence --method vpn-tcp        
> Starting proxy with method 'vpn-tcp', ...                                                                 
> Volumes are rooted at ...
> 
> No traffic is being forwarded ...
> 
> [sudo] password for exarkun:                   
> @gke_datawireio_us-central1-a_telepresence-testing|(virtualenv) $ host kubernetes
> kubernetes.local has address 10.3.240.1
> kubernetes.local has address 10.3.240.1
> Host kubernetes.local not found: 3(NXDOMAIN)
> @gke_datawireio_us-central1-a_telepresence-testing|(virtualenv) $ 
> 

branch behavior:

> $ TELEPRESENCE_REGISTRY=exarkun TELEPRESENCE_VERSION=0.78-4-g3409ce0-LNX-1522412298 telepresence --method vpn-tcp                                  
> Starting proxy with method 'vpn-tcp', ...
> Volumes are rooted at ...
> 
> No traffic is being forwarded ...
> 
> [sudo] password for exarkun:                   
> Guessing that Services IP range is 10.3.240.0/20. Services started after this point will be inaccessible if are outside this range; restart telepresence if you can't access a new Service.   
> 
> @gke_datawireio_us-central1-a_telepresence-testing|bash-4.4$ host kubernetes                   
> kubernetes has address 10.3.240.1              
> kubernetes has address 10.3.240.1              
> Host kubernetes not found: 3(NXDOMAIN)         
> @gke_datawireio_us-central1-a_telepresence-testing|bash-4.4$              
> 

The branch does not have a changelog entry yet because it will immediately conflict #520.

### DIFF
--- a/k8s-proxy/resolver.py
+++ b/k8s-proxy/resolver.py
@@ -2,6 +2,7 @@ import socket
 from copy import deepcopy
 from typing import Callable, List, Tuple, Optional, Union
 
+from twisted.python.failure import Failure
 from twisted.internet import defer
 from twisted.names import client, dns, error
 from twisted.internet.threads import deferToThread
@@ -167,26 +168,13 @@ class LocalResolver(object):
     def _handle_search_suffix(self, query, parts, timeout):
         stem = self._strip_search_suffix(parts)
         if stem != parts:
-            new_query = deepcopy(query)
-            new_query.name.name = b".".join(stem)
-            print(
-                "Updated query of type {} from {} to {}".
-                format(query.type, query.name.name, new_query.name.name)
-            )
-
-            def failed(f):
-                print(
-                    "Failed to lookup {} due to {}, falling back to {}".
-                    format(new_query.name.name, f, query.name.name)
-                )
-                return self.fallback.query(query, timeout=timeout)
-
-            return defer.maybeDeferred(
-                self.query,
-                new_query,
-                timeout=(1, 1),
-                real_name=query.name.name,
-            ).addErrback(failed)
+            # This is a query including a search domain from the client's
+            # network context.  The search domain doesn't make sense in the
+            # cluster's network context.  Fail this query fast.
+            print("Fast-failing query with search suffix: {}".format(
+                query.name,
+            ))
+            return Failure(error.DomainError())
 
     def query(
         self,


ark3:
What's the difference between "Fast-failing" and "Name does not resolve" in this trace? The former is from the suffix we're stripping, the latter is from a suffix we should be stripping, but aren't.

```
  26.1 13 | 2018-03-30T14:54:16+0000 [stdout#info] Fast-failing query with search suffix: qotm.c.datawireio.internal
  26.1 22 | c : DNS request from ('10.142.0.2', 57468) to None: 38 bytes
  26.1 13 | 2018-03-30T14:54:16+0000 [stdout#info] A query: b'qotm.google.internal'
  26.1 13 | 2018-03-30T14:54:16+0000 [stdout#info] getaddrinfo error: [Errno -2] Name does not resolve
  26.1 22 | c : DNS request from ('10.142.0.2', 52369) to None: 22 bytes
  26.2 13 | 2018-03-30T14:54:16+0000 [stdout#info] A query: b'qotm'
  26.2 13 | 2018-03-30T14:54:16+0000 [stdout#info] Result for b'qotm' is ['10.3.250.232']
```

Why should we be stripping the latter? Because it will be searched:

```
$ cat /etc/resolv.conf 
# Dynamic resolv.conf(5) file for glibc resolver(3) generated by resolvconf(8)
#     DO NOT EDIT THIS FILE BY HAND -- YOUR CHANGES WILL BE OVERWRITTEN
nameserver 169.254.169.254
search c.datawireio.internal google.internal
```

We don't strip it because we don't see it, and that's because we always return success for hellotelepresence queries.

exarkun:
It sounds like #393 may not have had the intended consequences (at least not in all environments), then?

ark3:
Yeah, sure, but my real question is: why fast fail is a special case? Why do we have to figure out the suffixes in the first place?

exarkun:
I think it's because the local search domains are meaningless in the context of the cluster but that's not a guarantee that looking them up will fail.  If the goal is to make lookup of those names succeed with cluster-supplied results then, yes, all of the suffix logic can just go away.  I don't think that would make a whole lot of sense but maybe that's an arbitrary or subjective conclusion.

ark3:
That's how a typical split-tunnel VPN works. This is what I was trying to get at in [my comment in #192](https://github.com/datawire/telepresence/issues/192#issuecomment-373037982). If we can dump all the suffix logic, I think that would be a huge win, _unless_ there are common or important use cases that would become more annoying.

PR #678, opened 2018-06-07
Make Telepresence a single-file executable and build packages in one container

### LINKED ISSUE #642: Make Telepresence a single-file executable
The goal is to simplify
- OS packaging: no Python stuff is required to build a package
- initial distribution, once Telepresence becomes better about reporting missing dependencies and explaining how to install them
- distribution of one-off development builds and such if you have the dependencies already

Steps
- [x] Write a script to build an executable Python zip file for Telepresence with an appropriate `__main__.py` file. Use `setup.py build` or `setup.py sdist` to extract source code. 
- [x] Use the existing `packaging/build-sshuttle.py` script to construct a single-file executable for sshuttle using Pex. The zip file approach doesn't work because sshuttle grabs its own source code to send over the ssh connection.
- [x] Update `install.sh` to use the above scripts. Update the install-from-source docs if necessary.
- [x] Update deployment stuff to use the above to build both executables in a container and add them to the artifact tarball.
- [x] If #641 has been completed, update CI to push the executables to S3 with the names `telepresence-$version` and `sshuttle-telepresence-$version`. This will work for non-released master commits too.

In the future, solve #503 by putting these executables at the head of the PATH. Use CircleCI's workspace feature to carry the executables forward to subsequent stages

### LINKED ISSUE #643: It's hard to add Linux package support
... because we rely on [fpm-within-docker](https://github.com/alanfranz/fpm-within-docker) images for each OS and version we support.

Plan
- [x] Implement #641 and #642
- [x] Build a container image that has `fpm`, `rpm`, and `dpkg`
- [x] In this container, build a package for every combination of supported package format (rpm and deb) and Python version (3.5 and 3.6) using the executables from #642 with package versions based on #641
- [x] Push these four packages to every OS we support (2 Fedoras and 3 Ubuntus)
- [ ] Push these packages for other rpm/deb operating systems that offer the dependencies with the same package names (e.g., Mint, Debian)
- [ ] Use the new machinery to build packages for other operating systems where dependencies have different package names (e.g., CentOS)

ark3:
Closes #642 
Also closes #643

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -78,7 +78,6 @@ jobs:
     steps:
       # https://circleci.com/docs/2.0/building-docker-images/
       - setup_remote_docker:
-          version: 17.11.0-ce
           docker_layer_caching: true
 
       - "checkout"
@@ -98,6 +97,7 @@ jobs:
     steps:
       # https://circleci.com/docs/2.0/building-docker-images/
       - setup_remote_docker:
+          version: 17.11.0-ce
           docker_layer_caching: true
 
       - "checkout"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -89,23 +89,21 @@ jobs:
             ./build --registry datawire --version-suffix="" --build-and-push --no-tests
 
   deploy:
-    machine:
-      enabled: true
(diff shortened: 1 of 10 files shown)


ark3:
Blah. CircleCI's Docker is too old.

Issue #673, opened 2018-06-04
On lost or failed connection, Telepresence should report proxy pod status/info to simplify troubleshooting


yuzumikan15:
### What were you trying to do?
I tried to swap a remote k8s deployment with my local docker image.
```
$ docker pull udacity/example-hello:1.0.0
$ telepresence --swap-deployment hello --docker-run \
  --rm -it udacity/example-hello:1.0.0
```

- The source code that I used for my k8s cluster is here: https://github.com/udacity/ud615/tree/master/kubernetes .
- I created the cluster on AWS EC2.
- I could access to my cluster on EC2 via telepresence: `telepresence --docker-run -i -t alpine /bin/sh` and curl command.

### What did you expect to happen?
I wanted to swap only `hello` deployment ([docker image](https://hub.docker.com/r/udacity/example-hello/)) just like the tutorial [here](https://www.telepresence.io/tutorials/docker).

### What happened instead?

`RuntimeError: SSH isn't starting.`
- It look like a pod was recreated on the remote cluster, but it said ssh is not starting.

### Automatically included information

Command line: `['/usr/local/bin/telepresence', '--swap-deployment', 'hello', '--docker-run', '--rm', '-it', 'udacity/example-hello:1.0.0']`
Version: `0.88`
Python version: `3.6.5 (default, Apr 25 2018, 14:23:58) 
[GCC 4.2.1 Compatible Apple LLVM 9.1.0 (clang-902.0.39.1)]`
kubectl version: `Client Version: v1.9.6 // Server Version: v1.9.6`
oc version: `(error: [Errno 2] No such file or directory: 'oc': 'oc')`
OS: `Darwin yuzuco.local 17.3.0 Darwin Kernel Version 17.3.0: Thu Nov  9 18:09:22 PST 2017; root:xnu-4570.31.3~1/RELEASE_X86_64 x86_64`
Traceback:

```
Traceback (most recent call last):
  File "/usr/local/Cellar/telepresence/0.88/libexec/lib/python3.6/site-packages/telepresence/cli.py", line 87, in call_f
    return f(*args, **kwargs)
  File "/usr/local/Cellar/telepresence/0.88/libexec/lib/python3.6/site-packages/telepresence/main.py", line 79, in main
    subprocesses, socks_port, ssh = connect(runner, remote_info, args)
  File "/usr/local/Cellar/telepresence/0.88/libexec/lib/python3.6/site-packages/telepresence/proxy.py", line 122, in connect
[...]

PR #760, opened 2018-08-24
Show pod info when ssh fails to start.

rhs:
---


telepresence/connect/ssh.py
```diff
@@ -69,7 +69,7 @@ def bg_command(self, additional_args: List[str]) -> List[str]:
             ]
         )
 
-    def wait(self) -> None:
+    def wait(self) -> bool:
         """Return when SSH server can be reached."""
```
ark3: Now it returns whether the server can be reached (t/f)

ark3:
This passes CI now. But I'm not convinced the UX is genuinely better than before. I think we need to iterate a bit around what `remote_info.probe(...)` does, what Tel displays at the console, and what gets recorded in the logfile. Thoughts?

ark3:
Closing this, as the spirit of the work is contained in #1128. But I'm leaving the original issue (#673) open, as there is more work to be done.

Issue #756, opened 2018-08-24
Telepresence crashes if --new-deployment names an existing deployment


ark3:
See #748

PR #771, opened 2018-08-31
Fix: crash if --new-deployment names an existing deployment

rohan47:
This is a fix for #756 

Previously when running the following command, and the deployment `qotm` being present in the cluster, Telepresence would crash
```console
telepresence --new-deployment qotm
```

So, after this fix, Telepresence will give an error message stating that deployment already exists in the cluster.

Signed-off-by: rohan47 <rohanrgupta1996@gmail.com>

---

### DIFF
--- a/telepresence/proxy/deployment.py
+++ b/telepresence/proxy/deployment.py
@@ -14,7 +14,7 @@
 
 import json
 from copy import deepcopy
-from subprocess import STDOUT
+from subprocess import STDOUT, CalledProcessError
 from typing import Tuple, Dict, Optional
 
 from telepresence.cli import PortMapping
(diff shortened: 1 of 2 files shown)


telepresence/proxy/deployment.py
```diff
@@ -78,9 +78,20 @@ def remove_existing_deployment():
...
+    except CalledProcessError:
+        command = [
+            "get",
+            "deployment",
+            deployment_arg,
+        ]
+        deployment_already_exists = runner.get_output(runner.kubectl(command))
+        if deployment_already_exists:
+            raise runner.fail("Deployment already exixts")
+        return deployment_arg, run_id
```
ark3: What will this `return` statement do? I think you need to re-raise the original failure here, not pretend there was success.

rohan47: Okay, Got it

ark3:
Thanks for the PR. We ended up addressing this in #871.

PR #814, opened 2018-10-29
Disable tracking by default and make it opt in

avinassh:
I had no idea the TP was collecting data. How come it is not mentioned in the main `README`? I had look into `docs/reference/usage_reporting.md`. I am really disappointed that such default/opt-out behaviour is not made obvious.

Second, it should be opt in, not opt out.

### DIFF
--- a/docs/reference/usage_reporting.md
+++ b/docs/reference/usage_reporting.md
@@ -29,8 +29,8 @@ The reporting code can be found in [`telepresence/usage_tracking.py`][1].
 
 Telepresence collects and reports usage every time a session is launched.
 
-## Can it be disabled?
+## Can it be enabled?
 
 Yes. Set the environment variable `SCOUT_DISABLE`.
 
-    export SCOUT_DISABLE=1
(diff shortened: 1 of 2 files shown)


richarddli:
Thanks for your concern. Telepresence is extremely sensitive to your environment. The data we collect has been crucial in designing some of the new features that we are introducing to improve robustness and stability. We also use it to improve our test suite coverage. Finally, we also have support (not currently used) to give updated version information to end users for security errata and such.

We believe, on balance, that an opt-out vs opt-in approach is appropriate for where we are at this stage of development. I understand that you may have a different point of view. However, we have carefully considered the pros/cons of opt-out vs opt-in in this specific situation.

We may revisit this in the future, but for now, our commitment is to providing robust & reliable software, and this anonymized telemetry data is invaluable.

That said, happy to take a PR that does a better job explaining this all.

ark3:
Closing, as this was largely handled by #815; more work can happen based on #816.

Issue #817, opened 2018-11-02
Broken link at the bottom of telepresence.io/discussion/overview


vemv:
This one:

![image](https://user-images.githubusercontent.com/1162994/47909802-42f86780-de91-11e8-90e4-be648042e233.png)

Thanks - Victor

vemv:
Fix:

```diff
- <a href="tutorials/kubernetes-rapid">quick start that uses the VPN method</a>
+ <a href="/tutorials/kubernetes-rapid">quick start that uses the VPN method</a>
```

ark3:
Fixed in #819

PR #819, opened 2018-11-02
Fixed the broken links on overview page

chenrui333:
Looks like the three links on the overview page are broken, here is the fix.

### DIFF
--- a/docs/discussion/overview.md
+++ b/docs/discussion/overview.md
@@ -32,7 +32,7 @@ Typical alternatives to Telepresence include:
 
 ## Getting started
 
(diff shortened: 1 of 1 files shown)


ark3:
Thanks for the PR! The type of fix proposed by @vemv in #817, i.e. adding a leading slash, seems more correct to me.

chenrui333:
interesting, I thought about the relative link references. but i found the other locations on the same page are actually doing the direct link references. I will make another update to see if it fits.

chenrui333:
@ark3 yeah, indeed that is much better, I also correct the other locations. Please review again.

chenrui333:
@ark3 can you give another look? Thanks!

PR #840, opened 2018-11-21
Merge `./build` in to `Makefile`

LukeShu:
(re-creating #839 from a branch rather than a fork, for CI reasons)

When I'm reading code, I tend to modify it, as part of my learning process.  So I usually end up with a stack of commits that I throw away before I make my first purposeful commit.

In this case, I think they're changes that do make for a more positive new-developer experience.  So maybe I won't throw them away this time :)

This is what my testbench commit is on top of, but I'm sure it wouldn't be too much trouble to rewrite it on top of master instead of on top of this, if you don't like this change.

---

Begone with the `./build` script, move its functionality in to `Makefile`.  This PR does create a new executable `./tools/license-check` for the `_license_check()` function, instead of trying to rewrite that as a Makefile snippet.  Both `developing.md` and the CircleCI config have been updated to use `make` instead of `./build`

This:
 - Reduces the number of tools/scripts a developer has to learn coming on to the project
 - The targets to `make` are easier to understand than the flags to `./build`; both from the `./build --help`/`make help` documentation, and from reading the source.


LukeShu:
Re: `ci/circleci: build-linux` and `build-macos` failing: Am I to understand that CircleCI is still using GNU Make 3.81?

ark3:
That's an interesting question. Basically all Macs are running that version:

```shell
$ make --version
GNU Make 3.81
Copyright (C) 2006  Free Software Foundation, Inc.
This is free software; see the source for copying conditions.
There is NO warranty; not even for MERCHANTABILITY or FITNESS FOR A
PARTICULAR PURPOSE.

This program built for i386-apple-darwin11.3.0
```

LukeShu:
Huh, I knew that Mac avoided recent versions, because GPLv3, which I'd thought was just 4.x.  But apparently the GPLv2→3 switch happened in 3.82 (the last 3.x release).

LukeShu:
`private` is also new to 3.82.

ark3:
Netlify is _special_. The initial deployment runs `jekyll build` from some really old configuration that hasn't been valid for a long time and is nowhere to be found on their settings UI. Then somehow it decides to re-run with the correct configuration, though it often does not have all the latest tags, so it builds the home page with the wrong version number. When I do a real release, I have to clear the Netlify cache and rebuild to get the right version number on the home page. I would like to move the website build into Circle and push the generated HTML to Netlify, but I haven't gotten there yet.

richarddli:
Why not just do the doc build in Netlify itself?

ark3:
Hmm? That's what we do now. It's just that Netlify gets it wrong every release, and I have to log in and hit the "clear cache and rebuild" button every time.

PR #841, opened 2018-11-21
add lukeshu as maintainer

richarddli:


ark3:
We need to get Luke on the CNCF lists et al too.

PR #883, opened 2018-12-19
Avoid teleproxy startup race

LukeShu:
This modifies `telepresence outbound` to wait for teleproxy to send READY=1 over the sd_notify(3) interface during start-up.  It assumes that https://github.com/datawire/teleproxy/pull/33 will be present in teleproxy 0.3.6, and that 0.3.6 will be released by the time the next telepresence release is cut.

This has been tested on Parabola, and on macOS 10.13.6.

### DIFF
--- a/packaging/retrieve-teleproxy.py
+++ b/packaging/retrieve-teleproxy.py
@@ -8,7 +8,7 @@ from urllib.request import urlopen
 import sys
 
 TELEPROXY_BASE = "https://s3.amazonaws.com/datawire-static-files/teleproxy/"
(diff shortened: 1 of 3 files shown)


LukeShu:
"deploy" is failing because teleproxy 0.3.6 isn't yet released.  I assume other jobs will have similar failures.

LukeShu:
Now that there's a new teleproxy, I think this is mergable.

LukeShu:
At @ark3's request, I've edited Runner.launch() to use `self.temp` instead of creating a new temporary directory.

LukeShu:
Did Circle die?

Issue #299, opened 2017-09-21
Crash trying to --swap-deployment (openshift)


ghaskins:
### What were you trying to do?

Wanted to swap an existing deployment with telepresence so that I would have access to its Secrets/ConfigMaps and could debug it.

### What did you expect to happen?

I expected a telepresence shell in the context of the swapped application

### What happened instead?

telepresence crashes immediately.

### Automatically included information

Command line: `['/usr/local/bin/telepresence', '--swap-deployment', 'xxx']`
Version: `0.67`
Python version: `3.6.2 (default, Jul 17 2017, 16:44:45) 
[GCC 4.2.1 Compatible Apple LLVM 8.1.0 (clang-802.0.42)]`
kubectl version: `Client Version: v1.7.6`
oc version: `oc v3.6.0+c4dd4cf
kubernetes v1.6.1+5115d708d7
features: Basic-Auth

Server https://127.0.0.1:8443
openshift v3.6.0+c4dd4cf
kubernetes v1.6.1+5115d708d7`
OS: `Darwin MacBook-Pro.haskins.net 16.7.0 Darwin Kernel Version 16.7.0: Thu Jun 15 17:36:27 PDT 2017; root:xnu-3789.70.16~2/RELEASE_X86_64 x86_64`
Traceback:

```
Traceback (most recent call last):
  File "/usr/local/bin/telepresence", line 257, in call_f
    return f(*args, **kwargs)
  File "/usr/local/bin/telepresence", line 2350, in go
    runner, args
  File "/usr/local/bin/telepresence", line 1486, in start_proxy
    swap_deployment_openshift(runner, args)
  File "/usr/local/bin/telepresence", line 1402, in swap_deployment_openshift
    )[0].split("/", 1)[1]
IndexError: list index out of range

```

Logs:

```
   0.0 TL | Telepresence launched at Thu Sep 21 19:18:21 2017
   0.0 TL |   ['/usr/local/bin/telepresence', '--swap-deployment', 'xxx']
   0.0 TL | Scout info: {'latest_version': '0.63', 'application': 'telepresence', 'notices': []}
   0.0 TL | Context: myproject/127-0-0-1:8443/developer, namespace: myproject, kubectl_command: oc
   0.0 TL | [1] Capturing: (['oc', '--context', 'myproject/127-0-0-1:8443/developer', 'status'],)...
   0.3 TL | [1] captured.
   0.3 TL | [2] Capturing: (['ssh', '-V'],)...
[...]

Issue #825, opened 2018-11-08
Telepresence crashed in OpenShift


snorthov:
### What were you trying to do?

Running the standard OpenShift example: https://docs.openshift.com/online/getting_started/basic_walkthrough.html

telepresence --swap-deployment nodejs-mongo-persistent --run-shell --expose 8080 --expose 9229

### What did you expect to happen?

I'd like a telepresence shell.

### What happened instead?

Telepresence crashed.

### Automatically included information

Command line: `['/usr/local/bin/telepresence', '--swap-deployment', 'nodejs-mongo-persistent', '--run-shell', '--expose', '8080', '--expose', '9229']`
Version: `0.93`
Python version: `3.7.0 (default, Oct  2 2018, 09:20:07) 
[Clang 10.0.0 (clang-1000.11.45.2)]`
kubectl version: `Client Version: v1.12.2 // Server Version: v1.11.0+d4cacc0`
oc version: `oc v3.11.16 // kubernetes v1.11.0+d4cacc0 // features: Basic-Auth //  // Server https://api.starter-us-west-2.openshift.com:443 // openshift v3.11.16 // kubernetes v1.11.0+d4cacc0`
OS: `Darwin steves-mbp-3.ottriv.can.ibm.com 17.7.0 Darwin Kernel Version 17.7.0: Thu Jun 21 22:53:14 PDT 2018; root:xnu-4570.71.2~1/RELEASE_X86_64 x86_64`
Traceback:

```
Traceback (most recent call last):
  File "/usr/local/bin/telepresence/telepresence/cli.py", line 129, in crash_reporting
    yield
  File "/usr/local/bin/telepresence/telepresence/main.py", line 61, in main
    remote_info = start_proxy(runner)
  File "/usr/local/bin/telepresence/telepresence/proxy/__init__.py", line 90, in start_proxy
    run_id=run_id,
  File "/usr/local/bin/telepresence/telepresence/proxy/remote.py", line 209, in get_remote_info
    format(deployment_name)
RuntimeError: Telepresence pod not found for Deployment 'nodejs-mongo-persistent'.

```

Logs:

```
.8 TEL | [84] Capturing: oc --context snorthov-example/api-starter-us-west-2-openshift-com:443/stevenorthover --namespace snorthov-example get pod -o json --export --selector=telepresence=5b5e9813e3a54833b6a92b2c8a18ecc7
 129.5 TEL | [85] Running: sudo -n echo -n
[...]

Issue #793, opened 2018-10-03
Crash when swapping Deployments on OpenShift


zeonin:
### What were you trying to do?

Swap a Kubernetes Deployment in an OpenShift (Minishift) cluster

### What did you expect to happen?

The Deployment to be swapped out like they are on k8s or like how DeploymentConfigs are swapped on OpenShift.

### What happened instead?

Telepresence panicked trying to find the deployment. It appears that telepresence assumes that only DeploymentConfigs are used on OpenShift.

### Automatically included information

Command line: `['/nix/store/bkf8lj2znv5lgc02x5adfhkdga425f41-python3.6-telepresence-0.90/bin/telepresence', '--method', 'inject-tcp', '--swap-deployment', 'a', '--expose', '8000', '--run', 'python3', '-m', 'http.server', '8000']`
Version: `0.90`
Python version: `3.6.6 (default, Jun 27 2018, 05:47:41) 
[GCC 7.3.0]`
kubectl version: `Client Version: v1.9.7 // Server Version: v1.9.1+a0ce1bc657`
oc version: `oc v3.9.0+191fece // kubernetes v1.9.1+a0ce1bc657 // features: Basic-Auth GSSAPI Kerberos SPNEGO //  // Server https://192.168.42.217:8443 // openshift v3.9.0+71543b2-33 // kubernetes v1.9.1+a0ce1bc657`
OS: `Linux charon 4.14.71 #1-NixOS SMP Wed Sep 19 20:43:49 UTC 2018 x86_64 GNU/Linux`
Traceback:

```
Traceback (most recent call last):
  File "/nix/store/bkf8lj2znv5lgc02x5adfhkdga425f41-python3.6-telepresence-0.90/lib/python3.6/site-packages/telepresence/cli.py", line 87, in call_f
    return f(*args, **kwargs)
  File "/nix/store/bkf8lj2znv5lgc02x5adfhkdga425f41-python3.6-telepresence-0.90/lib/python3.6/site-packages/telepresence/main.py", line 76, in main
    remote_info = start_proxy(runner, args)
  File "/nix/store/bkf8lj2znv5lgc02x5adfhkdga425f41-python3.6-telepresence-0.90/lib/python3.6/site-packages/telepresence/proxy.py", line 195, in start_proxy
    swap_deployment_openshift(runner, args)
[...]

Issue #666, opened 2018-05-28
Running Hello World example in OpenShift(MiniShift)


lordofthejars:
Currently, I am following https://blog.openshift.com/telepresence-local-development/ with telepresence 0.88, minishift 1.13.1, Openshift 3.9 and oc client and kubectl installed.

The problem that I am facing is that when I run _telepresence --swap-deployment hello-world --expose 8000_ then it seems that everything works as expected:

_Forwarding remote port 8000 to local port 8000.

Password:
Guessing that Services IP range is 172.30.0.0/16. Services started after this point will be inaccessible if are outside this range; restart telepresence if you can't access a new Service.

@myproject/192-168-64-38:8443/developer|bash-3.2_ 

But after 10 seconds I get:

_Proxy to Kubernetes exited. This is typically due to a lost connection.
Error in atexit._run_exitfuncs:
Traceback (most recent call last):
  File "/usr/local/Cellar/telepresence/0.88/libexec/lib/python3.6/site-packages/telepresence/mount.py", line 85, in cleanup
    runner.get_output(sudo_prefix + ["umount", "-f", mount_dir])
  File "/usr/local/Cellar/telepresence/0.88/libexec/lib/python3.6/site-packages/telepresence/runner.py", line 193, in get_output
    raise CalledProcessError(cpe_exc.returncode, cpe_exc.cmd, output)
subprocess.CalledProcessError: Command '['umount', '-f', '/tmp/tmpvq8wkd5r']' returned non-zero exit status 1._

I don't know if someone else has faced the same problem, or if the problem is that Minishift and Telepresence do not work pretty well together.

Thank you very much.

lordofthejars:
@ mhausenblas this is the problem I found when running telepresence with OpenShift.

PR #895, opened 2019-01-11
fix: OpenShift Deployment/DeploymentConfig support

bartoszmajsak:
This PR improves usage of Telepresence with OpenShift. 

Potentially fixes: #299 #581 #582 #663 #666 #793 #825 

While using telepresence with Openshift (v3.11.0) I came across several problems.

#### Checks if cluster is openshift by inspecting api groups

Current version calls `server + '/version/openshift'` which returns 404 thus the code concludes it needs to use `kubectl` instead of `oc`. 

See d1817c1.

#### Checks if openshift is using `DeploymentConfig` or k8s deployment

As OpenShift support both `DeploymentConfig` and `Deployment` resources we have to check which is used in the current namespace for `--swap-deployment` mode before proceeding.  

See 1b12fc5.

#### Swaps deployment config instead of replication controller

Current implementation attempts to swap `ReplicationController`, but if deployment config has `ImageStream` trigger this will result with discarding telepresence deployment and roll back to original one.

See b48ff8a.

### DIFF
--- a/telepresence/proxy/deployment.py
+++ b/telepresence/proxy/deployment.py
@@ -291,19 +291,14 @@ def swap_deployment_openshift(
     add_custom_nameserver: bool
 ) -> Tuple[str, str]:
     """
-    Swap out an existing DeploymentConfig.
+    Swap out an existing DeploymentConfig and also clears any triggers
+    which were registered, otherwise replaced telepresence pod would
+    be immediately swapped back to the original one because of
+    image change trigger.
 
     Returns (Deployment name, unique K8s label, JSON of original container that
     was swapped out.)
 
-    In practice OpenShift doesn't seem to do the right thing when a
-    DeploymentConfig is updated. In particular, we need to disable the image
-    trigger so that we can use the new image, but the replicationcontroller
-    then continues to deploy the existing image.
-
-    So instead we use a different approach than for Kubernetes, replacing the
-    current ReplicationController with one that uses the Telepresence image,
-    then restores it. We delete the pods to force the RC to do its thing.
     """
 
     run_id = runner.session_id
@@ -336,14 +331,14 @@ def swap_deployment_openshift(
             runner.kubectl("replace", "-f", "-"),
             input=json.dumps(json_config).encode("utf-8")
         )
-        # Now that we've updated the replication controller, delete pods to
-        # make sure changes get applied:
+        # Now that we've updated the deployment config,
+        # let's rollout latest version to apply the changes
         runner.check_call(
             runner.kubectl("rollout", "latest", "dc/{}".format(deployment))
         )
 
     runner.add_cleanup(
-        "Restore original replication controller", apply_json,
+        "Restore original deployment config", apply_json,
         dc_json_with_triggers
     )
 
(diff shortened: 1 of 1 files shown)


bartoszmajsak:
A bit sloppy demo https://asciinema.org/a/yvc4dDGsptJOLpFC7zpKvgUFS :)

bartoszmajsak:
@ark3 no worries, lemme know if you need any help.

ark3:
Thank you for submitting a well-structured PR. I landed the first few commits on master. Regarding the last commit:

- There's a comment in the doc string about why Telepresence messes with RCs instead of DCs. I assume that stuff is no longer true, since your changes presumably work for you on modern OpenShift. Can you please update the comment?
- You changed `oc apply` to `oc replace`. Why? The rest of the code (with the notable exception of the new deployment operation...) uses declarative-style commands to modify the cluster. I don't feel strongly about this, but I'm curious about the motivation for the change.
- Please use `.format`-style string formatting, not `%`-style.

Because I landed some commits on master with my own commit following, I've made a minor mess of the commit graph. Sorry about that. If you want to rebase and force-push or create another PR, that's totally fine. I don't feel strongly about that either, and am happy to land this PR (with the above addressed) and handle the merge at my end.

bartoszmajsak:
Thanks for taking time incorporating some of my changes to master and reviewing the code. Much appreciated!

> There's a comment in the doc string about why Telepresence messes with RCs instead of DCs. I assume that stuff is no longer true, since your changes presumably work for you on modern OpenShift. Can you please update the comment?

That's correct - it also needs to remove triggers which are created when you use `oc new-app` like in the hello world example on the official blog https://blog.openshift.com/telepresence-local-development/

I tested against v3.11. Is there a way we can set up automated testing for OpenShift? Happy to help with that.

Updated the comment in 507092f. Somewhat overlooked - my bad. Probably I was so excited that it finally works I forgot to skim through doc string :)

> You changed `oc apply` to `oc replace`. Why? The rest of the code (with the notable exception of the new deployment operation...) uses declarative-style commands to modify the cluster. I don't feel strongly about this, but I'm curious about the motivation for the change.

My line of thinking is that we actually do `replace` those resources, so it's more self-explanatory of what is going on. In addition when using `apply` you will see following warning in the `telepresence.log`

`Warning: oc apply should be used on resource created by either oc create --save-config or oc apply  `

which might be a bit confusing. WDYT?

> Please use .format-style string formatting, not %-style.

Sure, done in f3984d1. That was the only one I quickly learned and missed other style in the code base. Lesson learned :)

> Because I landed some commits on master with my own commit following, I've made a minor mess of the commit graph. Sorry about that. If you want to rebase and force-push or create another PR, that's totally fine. I don't feel strongly about that either, and am happy to land this PR (with the above addressed) and handle the merge at my end.

No worries, you can also "Squash and merge" this PR - that would probably be the easiest. Let me know what works best for you.

ark3:
I would love to have automated testing against OpenShift. Two problems:

1. We don't know how to get access to an OpenShift cluster from CircleCI.
2. We don't use OpenShift, so if we break something, it's often not clear how to fix it.

What do you think? We have had #584 outstanding for a long time. Maybe continuing the discussion there would make sense.

Anyhow, this is good stuff and I'm landing it. Thank you!

PR #904, opened 2019-01-24
Minor fix to doc for local debugging

rohansingh:
"background" -> "foreground" (`fg`, brings the process
to the foreground).


rohansingh:
Foregoing a changelog update for this.

PR #911, opened 2019-01-25
Openshift docs

bartoszmajsak:
* mentions adding security context constraints for the user when using OpenShift
* creates doc-related targets in Makefile
* minor polishing in dev guide with mention how to run docker registry locally (useful when building)


ark3:
Thank you!

Issue #917, opened 2019-02-02
Feature Request: Automatically generate list of remote volumes


turettn:
I've recently discovered telepresence, and quickly got hooked.  I have a whole bunch of micro-services running on k8s, each with a different list of volume mounts.  Instead of having to manually create sym-links or use `proot` to create fake bindings, what I really want is to be able to get a list of mounts from telepresence, which I can then use to automatically handle the mounts.

To solve these problems, I propose adding a `TELEPRESENCE_MOUNTS` environment variable to the local container that holds a colon-separated list of remote mount points, ignoring those that docker automatically creates (`/sys/*`, `/proc/*`, `/dev/*`, `resolv.conf`, `hostname`, `hosts`).

With this in hand, I can do something like this in the entry-point of my container:

```python
def telepresence_remote_mounts():
    mounts = os.environ.get('TELEPRESENCE_MOUNTS')
    if not mounts:
        return

    tele_root = os.environ.get('TELEPRESENCE_ROOT')

    for mount in mounts.split(':'):
        dir_name, link_name = os.path.split(mount)
        os.makedirs(dir_name, exist_ok=True)
        
        link_src = os.path.join(tele_root, mount[1:])
        os.symlink(link_src, mount)
```

If I'm running telepresence, I get all the volumes where they would normally be, and everything works as expected.  If I'm in production, the environment variable is not present and nothing changes.

This saves me from having to manually maintain a list of volume mounts for each micro-service, and standardizes the telepresence workflow across all of them.

I will submit a pull request for consideration shortly.

PR #918, opened 2019-02-02
Add env var for remote volume mounts as colon-seperated list

turettn:
Proposed solution to #917.

I've recently discovered telepresence, and quickly got hooked.  I have a whole bunch of micro-services running on k8s, each with a different list of volume mounts.  Instead of having to manually create sym-links or use `proot` to create fake bindings, what I really want is to be able to get a list of mounts from telepresence, which I can then use to automatically handle the mounts.

This pull request adds a `TELEPRESENCE_MOUNTS` environment variable to the local container that holds a colon-separated list of remote mount points, ignoring those that docker automatically creates (`/sys/*`, `/proc/*`, `/dev/*`, `resolv.conf`, `hostname`, `hosts`).

With this in hand, a manually-updated list of locations that must be sym-link'd or bind-mount'd from under `$TELEPRESENCE_ROOT` to their usual locations is no longer required, and can be replaced with something like:

```python
def telepresence_remote_mounts():
    mounts = os.environ.get('TELEPRESENCE_MOUNTS')
    if not mounts:
        return

    tele_root = os.environ.get('TELEPRESENCE_ROOT')

    for mount in mounts.split(':'):
        dir_name, link_name = os.path.split(mount)
        os.makedirs(dir_name, exist_ok=True)
        
        link_src = os.path.join(tele_root, mount[1:])
        os.symlink(link_src, mount)
```

If the container is running locally via telepresence, all the volumes end up where they would normally be, and everything works as expected.  In production, the environment variable is not present and no action occurs.

### DIFF
--- a/docs/howto/volumes.md
+++ b/docs/howto/volumes.md
@@ -4,6 +4,10 @@ Volume support requires a small amount of work on your part.
 The root directory where all the volumes can be found will be set to the `TELEPRESENCE_ROOT` environment variable in the shell run by `telepresence`.
 You will then need to use that env variable as the root for volume paths you are opening.
 
+Telepresence will attempt to gather a list of all mount points that exist and put them into
+the `TELEPRESENCE_MOUNTS` variable, seperated by `:` characters.  This allows automated discovery
+of mount-points.  (See below for an example).
+
 For example, all Kubernetes containers have a volume mounted at `/var/run/secrets` with the service account details.
 Those files are accessible from Telepresence:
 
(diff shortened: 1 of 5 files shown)


ark3:
Thanks for the PR! I will try to get this landed later this week.

LukeShu:
@turettn To avoid leaving you hanging: @ark3 is offline for a while because of an injury. I'm not sure what all he wanted to do before landing this.

turettn:
@LukeShu Thanks for the update!  Hope @ark3 heals up soon.

PR #933, opened 2019-02-22
Fix outbound network access with `kind`

### LINKED ISSUE #932: DNS loop when using kind (kubernetes-in-docker)
Using [kind](https://sigs.k8s.io/kind) for a local Kubernetes cluster results in a DNS loop similar to #736, since the cluster name is not recognized as "local".

rohansingh:
`kind` has the same issue as minikube/minishift, so we need to detect
and handle it the same way.

Fixes #932.


LukeShu:
Can you amend the PR to add the "(kube-in-docker)" parenthetical to the news fragment, as you did in the in-code comment?

rohansingh:
Sure, will do!

rohansingh:
News fragment is fixed in the latest commit.

ark3:
Thanks for the PR. We need to do this the right way by detecting a local cluster directly. Until we get that worked out, I appreciate your help getting the various cases covered.

PR #940, opened 2019-02-26
Don't hardcode the kubernetes.default FQDN in local-docker

lminaudier:
The default search domain for clusters can be changed and
kubernetes.default.svc.cluster.local may not respond leading to
telepresence timing out when starting the local container.

So I propose to not use a FQDN.

---

### DIFF
--- a/local-docker/entrypoint.py
+++ b/local-docker/entrypoint.py
@@ -115,7 +115,7 @@ def wait():
     start = time()
     while time() - start < 30:
         try:
-            gethostbyname("kubernetes.default.svc.cluster.local")
+            gethostbyname("kubernetes.default")
             sleep(1)  # just in case there's more to startup
             sys.exit(100)
         except gaierror:


LukeShu:
Hey, just to avoid leaving you hanging: The lead Telepresence developer, @ark3, is offline for a while because of an injury, and this is something I think we should get his opinion on before merging.

I suspect the more "correct" thing to do would be to construct the FQDN based on the zone, rather than hard-coding a fixed string.  I also suspect that this may cause funny things if/when the DNS support switches over to teleproxy-based DNS, which does some really funny things with the DNS search path.

lminaudier:
Oh I am so sorry to hear that. Feel better @ark3. 
Thanks for the headsup. No worries. Totally understand.

ark3:
Thank you for the PR. I think your solution is correct. As `parse_resolv_conf` in [container.py](https://github.com/telepresenceio/telepresence/blob/0.97/telepresence/outbound/container.py#L65) already sets up the network container's DNS based directly on the proxy pod's resolv.conf, doing a lookup for `kubernetes.default` should be correct.

And let's worry about Teleproxy integration later. 😉

PR #968, opened 2019-03-28
version split allows for private registries

arroo:
---


ark3:
Thank you for the PR.

Issue #1003, opened 2019-04-21
Telepresence exits when attaching to process in Goland


the-gigi:
I run a local minikube cluster and want to debug a Go microservice locally. I managed to get everything working just fine. telepresence created the proxy. I was able to hit the cluster and see my modified service getting called. so far so good.
```
$ telepresence --swap-deployment link-manager --run go run main.go
T: How Telepresence uses sudo: https://www.telepresence.io/reference/install#dependencies
T: Invoking sudo. Please enter your sudo password.
Password:
T: Starting proxy with method 'vpn-tcp', which has the following limitations: All processes are affected, only one telepresence can run per machine, and you can't use other VPNs. You may need to add cloud hosts and headless services with --also-proxy.
T: For a full list of method limitations see https://telepresence.io/reference/methods.html
T: Volumes are rooted at $TELEPRESENCE_ROOT. See https://telepresence.io/howto/volumes.html for details.
T: Starting network proxy to cluster by swapping out Deployment link-manager with a proxy
T: Forwarding remote port 8080 to local port 8080.

T: Guessing that Services IP range is 10.96.0.0/12. Services started after this point will be inaccessible if are outside this range; restart telepresence if you can't access a new Service.
T: Setup complete. Launching your command.
2019/04/20 01:17:06 DB host: 10.100.193.162 DB port: 5432
2019/04/20 01:17:06 Listening on port 8080...
```
but, then I tried to attach to the running process with Goland and that seems to kill telepresence.
```
T: Your process has exited.
T: Exit cleanup in progress
T: Swapping Deployment link-manager back to its original state
```

Here is the relevant output of `ps aux` before trying to attach to the process:

```
$ ps aux | rg -v rg | rg "go run main.go"
the-gigi      91302   0.0  0.1  4402496  17992 s004  S+    2:07AM   0:01.52 go run main.go
the-gigi      90795   0.0  0.2  4391716  34836 s004  S+    2:04AM   0:03.07 /usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/Resources/Python.app/Contents/MacOS/Python /usr/local/bin/telepresence --swap-deployment link-manager --expose 8080 --run go run main.go
```

After trying to attach to the local service (process 91302) the telepresence process (90795) is gone, but the local service is still running.

```
$ ps aux | rg -v rg | rg "go run main.go"
[...]

PR #1005, opened 2019-04-23
Add support for attaching a debugger to processes under telepresence

the-gigi:
This fixes [bug 1003](https://github.com/telepresenceio/telepresence/issues/1003). 

I added a [detailed comment](https://github.com/telepresenceio/telepresence/issues/1003#issuecomment-485592803) that explains the issue + sample code that demonstrates it.

This change was tested successfully by on macOS by attaching the Goland (delve) debugger to a Go service and the Pycharm debugger to a Python service. Before this change attaching a debugger would cause telepresence to incorrectly assume the process exited and shutdown.

### DIFF
--- a/newsfragments/1003.bugfix
+++ b/newsfragments/1003.bugfix
@@ -1,0 +1,1 @@
+Add support for attaching a debugger to Go services under telepresence
--- a/newsfragments/1003.bugfix
+++ b/newsfragments/1003.bugfix
@@ -1,1 +1,1 @@
-Add support for attaching a debugger to Go services under telepresence
+Add support for attaching a debugger to services under telepresence (tested on Go and Python)
--- a/newsfragments/1003.bugfix
+++ b/newsfragments/1003.bugfix
@@ -1,1 +1,1 @@
-Add support for attaching a debugger to services under telepresence (tested on Go and Python)
+Add support for attaching a debugger to processes under telepresence (tested on Go and Python)
--- a/telepresence/runner/runner.py
+++ b/telepresence/runner/runner.py
@@ -679,7 +679,7 @@ class Runner(object):
         self.write("Everything launched. Waiting to exit...")
         main_code = None
         span = self.span()
-        Thread(target=wait_for_process, args=(main_process,))
+        Thread(target=wait_for_process, args=(main_process, ))
         while not self.quitting:
             sleep(0.1)
         span.end()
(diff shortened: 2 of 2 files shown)


LukeShu:
Re: the news fragment: This isn't just for Go, right?  Isn't the problem also present for C programs being debugged with GDB?

the-gigi:
I don't know. I observed it for Go programs. The Go debugger (Goland and VSCode at least) is delve. I'm not a debugger expert and can't tell if it's something that gdb does too for C/C++ programs.

newsfragments/1003.bugfix
```diff
@@ -0,0 +1 @@
+Add support for attaching a debugger to Go services under telepresence
```
ark3: Could you please update this since we now know it's not specific to go? 
 
> the-gigi [9:27 PM]
> Also, tested on attaching PyCharm to a swapped Python service. Current telepresence exits immediately when debugger attaches, telepresence+PR works like a [py]charm

😆

telepresence/runner/runner.py
```diff
@@ -664,12 +664,23 @@ def wait_for_exit(self, main_process: Popen) -> None:
...
+            """
+            nonlocal main_code
+            main_code = p.wait()
+            self.quitting = True
+
         self.write("Everything launched. Waiting to exit...")
         main_code = None
         span = self.span()
-        while not self.quitting and main_code is None:
+        Thread(target=wait_for_process, args=(main_process,))
```
ark3: Can you clarify in a comment (or maybe the doc string above) why this is necessary? The intent of this change is clearly described in #1003, of course. But it would be even better to have a bit of context in the code itself. Perhaps something as simple as `# See https://github.com/telepresenceio/telepresence/issues/1003`.

ark3:
Also, please make sure you've run `make lint` on the code. 

Soon we will fix the issue with CI not running correctly PRs...

newsfragments/1003.bugfix
the-gigi: done

telepresence/runner/runner.py
the-gigi: sure thing. added mention of 1003 in the docstring

the-gigi:
@ark3 I addressed all the comments and ran `make lint` successfully.

the-gigi:
@ark3 no problem. I restored the order of the imports. PyCharm changed it automatically

the-gigi:
@ark3 any ETA on merging this? let me know if there's anything else I can do

PR #1058, opened 2019-06-19
Log some stuff that may help diagnose vpn-tcp method failures

### LINKED ISSUE #1049: vpn-tcp should include more information in the logs if it fails to connect
- `/etc/resolv.conf`
- `networksetup ...` stuff on the Mac
- `/etc/resolver/...` on the Mac (see also `man 5 resolver`)

ark3:
Fixes #1049

### DIFF
--- a/telepresence/outbound/vpn.py
+++ b/telepresence/outbound/vpn.py
@@ -308,7 +308,26 @@ def connect_sshuttle(
         dns_lookup(runner, many_dotted_name, 1)
 
     if countdown != 0:
+        runner.add_cleanup("Diagnose vpn-tcp", log_info_vpn_crash, runner)
         raise RuntimeError("vpn-tcp tunnel did not connect")
 
     subspan.end()
     span.end()
+
+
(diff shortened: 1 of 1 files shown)


cauli:
@ark3 This is some information that may also be relevant to Linux:
`/etc/nsswitch.conf`

```
For Ubuntu 18.04 vanilla install (or any install that uses `systemd-resolve`) and telepresence to properly connect to K8s, you need to change the following line in `/etc/nsswitch.conf`:
from: `hosts:          files mdns4_minimal [NOTFOUND=return] resolve [!UNAVAIL=return] dns myhostname`
to:     `hosts:          files mdns4_minimal [NOTFOUND=return] resolve dns myhostname`
That is: remove `[!UNAVAIL=return]`  for `resolve` module. The problem is with systemd-resolve that will query only local interface over LLMNR for single named hostnames and not use the actual DNS query. Removing that condition will allow fallback to dns module which still uses systemd-resolve stub but will end up as a DNS query because the local interface search will be bypassed.
```

Issue #875, opened 2018-12-15
swap deployment fails when securityContext contains unprivileged user


kunickiaj:
### What were you trying to do?

trying to use the swap-deployment feature with one of my deployments.

### What did you expect to happen?

expected to expose two ports for a local process and have traffic directed to them

### What happened instead?

telepresence died with the attached traceback
full log in gist: https://gist.github.com/kunickiaj/080328802f437cdc1fbb6722856de4ee

It seems that the root cause is the securityContext in the container I wished to swap.
Other (more privileged) containers do not have this issue. Was able to confirm that removing the following securityContext from the affected container allowed me to work around the issue:

```yaml
securityContext:
  runAsNonRoot: true
  runAsUser: 500
```

Probably related to #617 #737 and #723 
A possible fix might be to have telepresence replace the relevant parts of the security context if it does in fact need root (e.g. removing the runAsNonRoot). Would also suggest alerting the user to those kind of modifications.

### Automatically included information

Command line: `['/usr/local/bin/telepresence', '--swap-deployment', 'sch-control-hub-pipelinestore:pipelinestore', '--expose', '18631', '--expose', '18632']`
Version: `0.96`
Python version: `3.6.6 (default, Oct  4 2018, 20:50:27) 
[GCC 4.2.1 Compatible Apple LLVM 10.0.0 (clang-1000.11.45.2)]`
kubectl version: `Client Version: v1.13.0 // Server Version: v1.10.0`
oc version: `oc v3.11.0+0cbc58b // kubernetes v1.11.0+d4cacc0 // features: Basic-Auth //  // Server https://192.168.37.162:8443 // kubernetes v1.10.0`
OS: `Darwin streamsam381331.nerdworld.xyz 18.2.0 Darwin Kernel Version 18.2.0: Mon Nov 12 20:24:46 PST 2018; root:xnu-4903.231.4~2/RELEASE_X86_64 x86_64`

```
Traceback (most recent call last):
  File "/usr/local/bin/telepresence/telepresence/cli.py", line 131, in crash_reporting
    yield
  File "/usr/local/bin/telepresence/telepresence/main.py", line 70, in main
    socks_port, ssh = do_connect(runner, remote_info)
  File "/usr/local/bin/telepresence/telepresence/connect/connect.py", line 99, in do_connect
    return connect(runner_, remote_info, is_container_mode, args.expose)
  File "/usr/local/bin/telepresence/telepresence/connect/connect.py", line 57, in connect
[...]

PR #1067, opened 2019-07-01
Add --serviceaccount option to set anyuid scc on it separately

dbazhal:
https://github.com/telepresenceio/telepresence/issues/875

https://github.com/telepresenceio/telepresence/issues/875#issuecomment-501257612

We've been running telepresence on local okd kubernetes cluster, so there are security restrictions, such as anyuid, that are breaking telepresence proxy container. So we must to allow serviceaccount, which telepresence uses, anyuid security context. This PR proposes to run telepresence with separate serviceaccount, so privileged mode could be granted not to all apps running in selected namespace.

### DIFF
--- a/telepresence/cli.py
+++ b/telepresence/cli.py
@@ -270,6 +270,12 @@ def parse_args(args=None) -> argparse.Namespace:
         )
     )
     parser.add_argument(
+        "--serviceaccount",
+        dest="custom_serviceaccount",
+        default=None,
+        help="Service account to set in the proxy pod spec."
+    )
+    parser.add_argument(
         "--expose",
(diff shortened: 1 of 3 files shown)


dbazhal:
@ark3 could you please take a look at my smaaall pr? :)

ReSearchITEng:
Another option is running a docker image that does not need elevated perms. The Dockerfile resides in this PR, should you want to use it instead of the original one: #1114. 
OCP/OKD bonus: it's centos/rhel based

ark3:
I did this work in #1126, which is very similar to your PR and #1094. Thanks for your help!

PR #1070, opened 2019-07-08
Fix/docker deployment

netag:
- Adding ssl certs to k8s dockerfile
- Swapped deployment: explicitly add docker command to run

### DIFF
--- a/k8s-proxy/Dockerfile
+++ b/k8s-proxy/Dockerfile
@@ -20,7 +20,7 @@ COPY requirements.txt /usr/src/app
 
 # For some reason pip doesn't install incremental (a Twisted dependency) so do
 # so manually. When done, remove unneeded packages for a smaller image.
-RUN apk add --no-cache python3 python3-dev openssh gcc libc-dev && \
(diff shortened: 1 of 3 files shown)


ark3:
Thank you for the PR. Can you please explain why these changes are needed? You didn't refer to an issue in the tracker, and I don't immediately see what you're accomplishing, so some context would be helpful. Thanks!

netag:
@ark3 thanks for the prompt response. 

Since in our Kubernetes cluster we use an annotation which injects a wrapper for the command to allow access to secrets (like vault). The tool that does the injection needs the command to be part of the deployment (this way it knows how to "replace" the original command with a command that is wrapped by the tool).

In telepresense we get another wrapper, but since this wrapper command is not explicitly part of the deployment, our wrapper misses it and runs the command from the telepresense deployment, which is an empty string.

This pr helps us support our use case with full backward compatibility for the current version.

vladshub:
Any movement on this?

netag:
hi @ark3, do you have any update around this?

vladshub:
Hi, any updates on this?

PR #1079, opened 2019-07-18
Support for vieux/sshfs docker volume driver when using container mode

skwasiborski:
The change adds support for using [vieux/sshfs](https://github.com/vieux/docker-volume-sshfs) docker volume driver instead of using sshfs directly. This should help at least partially resolve [#115](https://github.com/telepresenceio/telepresence/issues/115). From my testing it works when telepresence is run inside a container. It does not work when run from WSL.

### Implementation details
new `--docker-mount` cli option is added. It is mutually exclusive with `--mount`. It needs to be set to absolute path and can be used only in conjunction with `--method container`.

When `--docker-mount` is specified instead of using `sshfs` a randomly named docker volume will be created using [vieux/sshfs](https://github.com/vieux/docker-volume-sshfs) volume driver. The volume will be mounted in the user container at the mount point specified by `--docker-mount`. As a side benefit `sudo` is no longer required for container method. 

One downside is that it is not possible to mount volume sub-directory to a container. This means that using `TELEPRESENCE_ROOT` is obligatory.


skwasiborski:
Hi, any comments to the PR?

ark3:
Sorry to take so long to get back to you.

1. You ensure that `--mount` and `--docker-mount` cannot both be active at the same time. This is good because it offers an opportunity to simplify the implementation. I think mount.py could be left essentially unchanged, and a separate docker_mount.py could do setup and teardown work for your new code path. That would reduce the size of the changeset and simplify testing a lot.

2. There is no need to carry a randomly-generated volume name all the way through the code. Instead, `mount_dir` can stay as it is (as the target of the mount). The volume name should just be `runner.session_id`. That is already available everywhere, nearly guaranteed to be unique, etc. This would reduce the size of the changeset too.

3. Can you add some documentation in docs/howto/volumes.md to the PR? Just adding what you wrote in the PR description above would be great.

4. Do you have any thoughts on how to test this in an automated fashion? I have some ideas but I'd like to know what you think.

All in all, this is really good. Most likely I will work on it some more next week, probably pursuing the ideas above. If you can get to it before then, great!

Thank you very much.

skwasiborski:
I made changes as suggested in 2 and 3. I did modify mount.py to provide cleaner separation of new and old code. I have no Python experience whatsoever so I was not sure how to effectively split the code into to files.

As far as testing goes I do not have any ideas at hand. That said we used the modified version quite a bit and it seems to function quite well.

ark3:
Excellent work. The new changeset is much smaller and more focused. Thank you!

I'm going to think about a testing story around this for next week. I am excited to land this functionality as soon as I can get a test worked out.

skwasiborski:
Hi @ark3 thanks for finalizing the work and merging the PR. Hope it makes into a release soon.

PR #1087, opened 2019-08-02
Add check for docker-desktop (MacOS) in startup

willaustin:
Docker Desktop for Mac has changed the context name it uses. I added a check for this new context name next to docker-for-desktop in startup.py. I wonder if there is a better way to check if telepresence is running locally. Thanks!


willaustin:
Do I need to open an issue to be tied to this PR?

PR #1094, opened 2019-08-12
Allows Telepresence to set the service account when a new deployment is created

### LINKED ISSUE #1093: Allow Telepresence to set the serviceAccount or serviceAcountName that is used by a deployment.
**Feature Request:**
Allow the service account to be set on a new deployment.

**Rational:**
We purposely run cluster that doesn't have enough permissions on the default service account to launch a new deployment or service.  This would allow Telepresence to start pods in our cluster.

billmcmath:
Allows Telepresence to set the service account when a new deployment is created

Fixes #1093

### DIFF
--- a/1093.feature
+++ b/1093.feature
@@ -1,0 +1,1 @@
+Telepresence can now set the service account that is used in a new deployment.  This is done using the `--service-account <foo>` flag.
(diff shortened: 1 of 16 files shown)


ark3:
I did this work in #1126, which is very similar to your PR and #1067. Thanks for your help!

PR #1114, opened 2019-09-10
new proxy image which solves the UID runAs issues with sshd

ReSearchITEng:
the generated image will no longer fail due to missing privileges in secured/restrictive ocp or k8s environments
Also, the image is Centos/Rhel based, the usual base images for ocp (which also makes it bigger unfortunately: ~445Mb)

It's for the security focused envs, willing to trade the bigger size for solving issues like #875

PS: more explanations inside the file.


ReSearchITEng:
For anyone interested to test it:
docker.io/researchiteng/telepresence:0.101

PR #1121, opened 2019-09-20
remove cpu resource limit

zhuopeng:
Thanks for this wonderful tool! enjoying it so far :), but found a problem

Having CPU limit on telepresence pod causes noticeable latency increasing when there is a lot of simultaneously connections being proxied by telepresense. I observed almost 4X faster with my case( proxying a lot of connections) by disabling CPU limit. 

For latency sensitive applications, turning off CPU limit is sort of what's being suggested give the current kubernetes status. Thinking telepresense is one of the type.

### DIFF
--- a/telepresence/proxy/deployment.py
+++ b/telepresence/proxy/deployment.py
@@ -118,7 +118,7 @@ def create_new_deployment(
     command = [
         "run",  # This will result in using Deployment:
         "--restart=Always",
-        "--limits=cpu=100m,memory=256Mi",
+        "--limits=cpu=1000m,memory=256Mi",
         "--requests=cpu=25m,memory=64Mi",
         deployment_arg,
         "--image=" + get_image_name(expose),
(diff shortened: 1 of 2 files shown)


ark3:
You're right about the issue, but unfortunately some clusters _require_ limits to schedule pods, and will reject pods with no limit set:

    No nodes are available that match all of the following predicates:: Insufficient cpu (3).

So I'm hesitant to land your PR, as it may break Telepresence mysteriously for other users. Does raising the CPU limit help? The discussion in the linked issues implies that it doesn't, but can you try it and see for your particular case? If raising the limit helps, that would be a safer fix.

zhuopeng:
setting CPU limit to 1 seems yielding same performance with removing it as far as I tested. Looks like at least not throttled by CPU time when limit is less than 1.

telepresence/proxy/deployment.py
```diff
@@ -118,7 +118,7 @@ def remove_existing_deployment(quiet=False):
     command = [
         "run",  # This will result in using Deployment:
         "--restart=Always",
-        "--limits=cpu=100m,memory=256Mi",
+        "--limits=1000m,memory=256Mi",
```
ark3: I'm not sure how this is interpreted. Can you try `--limits=cpu=1000m,memory=256Mi` (i.e. add back the `cpu=` bit) on your setup?

zhuopeng: oops :P. will re do my tests with correct setting

zhuopeng: Tested both ```cpu=1000m``` and without cpu limit, almost getting same performance. 

Also updated PR

PR #1132, opened 2019-09-27
Add --container-to-host argument

ark3:


jiezhang0917:
Hi @ark3 , is this option removed on version 2? Is there a solution to achieve the same thing with version 2 then?

Issue #320, opened 2017-10-10
kubectl under Telepresence vpn-tcp & Kubernaut doesn't work


ark3:
When using Telepresence with a [Kubernaut][k] cluster, the DNS lookup that the `kubectl` command performs to find the Kubernetes master gets the EC2-internal IP address because all DNS lookups are done in the cluster. Said `kubectl` command then fails because it cannot reach the master at that IP address from the local network.

This bug might apply to all Kops/EC2 Kubernetes clusters. We should find out.

[k]: https://github.com/datawire/kubernaut#kubernaut

richarddli:
I don't see this?
```
lost-cause:qotm richard$ kubectl get svc
NAME         CLUSTER-IP      EXTERNAL-IP   PORT(S)   AGE
kubernetes   10.96.0.1       <none>        443/TCP   1h
qotm         10.110.20.176   <none>        80/TCP    4s
lost-cause:qotm richard$ telepresence
Starting proxy with method 'vpn-tcp', which has the following limitations: All processes are affected, only one telepresence can run per machine, and you can't use other VPNs. You may need to add cloud hosts with --also-proxy. For a full list of method limitations see http://www.telepresence.io/reference/methods.html.
Volumes are rooted at $TELEPRESENCE_ROOT. See http://www.telepresence.io/howto/volumes.html for details.

No traffic is being forwarded from the remote Deployment to your local machine. You can use the --expose option to specify which ports you want to forward.

Guessing that Services IP range is 10.96.0.0/12. Services started after this point will be inaccessible if are outside this range; restart telepresence if you can't access a new Service.

@admin@kubernetes|bash-4.4$ curl qotm
{
  "hostname": "qotm-108851151-jmzpr",
  "ok": true,
  "quote": "The light at the end of the tunnel is interdependent on the relatedness of motivation, subcultures, and management.",
  "time": "2017-10-10T20:09:07.552793",
  "version": "1.1"
}
@admin@kubernetes|bash-4.4$ kubectl cluster-info
Kubernetes master is running at https://i-03c5184c782a8983c.kubernaut.io:6443
```

richarddli:
I should note that I am running 0.64, if that makes a difference.

ark3:
@richarddli If I follow your script, I can get to where you got, but my `kubectl` commands hang on master access. So, for example, I'll see that the Kubernetes master is running at blah.kubernaut.io but then my `kubectl` hangs there for a while, then times out.

```shell
[...]

PR #1138, opened 2019-10-03
Update OpenShift tutorial

marun:
- Replace DeploymentConfig with Deployment
   - Otherwise the use of `--swap-deployment=<name>` will fail with error:
      
```
T: Failed to find deployment <name>:
```

- Document workaround for [API endpoint DNS issue](https://github.com/telepresenceio/telepresence/issues/320#issuecomment-537757455)


bartoszmajsak:
@marun can you have a look at #1139? Is that related to the problems you are facing?

PR #1142, opened 2019-10-06
Flush DNS cache after connecting on MacOS clients

### LINKED ISSUE #1118: Consider option for flushing DNS cache after telepresence is active
Every time after running `telepresence`, I get a few more host entries available for access. E.g.: `http://remote-server:5050`.
 
But in macOS, due to DNS caching, I have to manually run `sudo killall -HUP mDNSResponder` every time after `telepresence` is loaded for it to actually start seeing the new entries.

My proposal is a new flag for `telepresence` that detects the OS and runs a dns-cache-killing command automatically.

Telepresence already has `sudo`, which probably makes this more acceptable.

KarolisL:
Fixes #1118

If there is a better way to implement this, please point me to it, I would be more than happy to implement it.

### DIFF
--- a/telepresence/outbound/local.py
+++ b/telepresence/outbound/local.py
@@ -121,6 +121,7 @@ def launch_vpn(
     Launch sshuttle and the user's command
     """
     connect_sshuttle(runner, remote_info, also_proxy, ssh)
+    _flush_dns_cache(runner)
     env = get_local_env(runner, env_overrides, False)
     runner.show("Setup complete. Launching your command.")
     try:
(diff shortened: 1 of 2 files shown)


KarolisL:
Thank you for review
! 
As I understand, I don't need to copy ENV vars to my `sudo pkill` process (as an arg to `check_call`), do I?

ark3:
You are correct, you should not need environment variables for that.

telepresence/outbound/local.py
```diff
@@ -137,5 +137,8 @@ def launch_vpn(
 def _flush_dns_cache(runner: Runner):
     if runner.platform == "darwin":
         runner.show("Connected. Flushing DNS cache.")
-        pkill_cmd = ["sudo", "-n", "/usr/bin/pkill", "-HUP", "mDNSResponder"]
-        runner.check_call(pkill_cmd)
+        pkill_cmd = ["sudo", "-n", "/usr/bin/pkill", "-HUP", "mDNSResponder_Z"]
```
ark3: Why the `_Z` at the end?

KarolisL: Sorry about that, I've used it to test error condition. I've removed it.

ark3:
Thank you again for the PR.

KarolisL:
Thank you for great tool! :-)

PR #1143, opened 2019-10-07
fix: calls /users endpoint to determine if oc should be used

### LINKED ISSUE #1139: Unable to use `oc` when api-resources are secured
Currently, in order to determine if the target cluster is vanilla k8s or openshift, telepresence calls `${SERVER_URL}/apis`. 

https://github.com/telepresenceio/telepresence/blob/1f95e10da1b803c306f709f96aecd63f80cad451/telepresence/startup.py#L28-L50

If this endpoint is protected, OpenShift will return `403` and Telepresence will assume it's dealing with vanilla k8s. Therefore it will use `Deployment` instead of `DeploymentConfig`. 

This leads to errors while using `telepresence` in case of the latter, e.g.:

```
T: Starting network proxy to cluster using the existing proxy Deployment ratings-v1-v1-circleci-jufar

T: Failed to find deployment ratings-v1-v1-circleci-jufar:
T: Error from server (NotFound): deployments.extensions "ratings-v1-v1-circleci-jufar" not found

T: Exit cleanup in progress
```

Expected to use `DeploymentConfig` instead.

<details>
  <summary>telepresence.log</summary>

```
   0.0 TEL | Telepresence 0.101 launched at Wed Oct  2 15:37:34 2019
   0.0 TEL |   /usr/bin/telepresence --namespace ike-tests-epyskbtjlxzngrwx --deployment ratings-v1-v1-bartek-dmaxu --expose 9080 --method inject-tcp --run ike watch --dir '"."' --exclude '"*.log,.git/"' --interval 500 --run 'ruby ratings.rb 9080'
   0.0 TEL | Platform: linux
   0.0 TEL | Python 3.7.4 (default, Jul  9 2019, 16:32:37)
   0.0 TEL | [GCC 9.1.1 20190503 (Red Hat 9.1.1-1)]
   0.0 TEL | [1] Running: uname -a
   0.0   1 | Linux hydrobook 5.2.16-200.fc30.x86_64 #1 SMP Thu Sep 19 16:14:04 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
   0.0 TEL | [1] ran in 0.00 secs.
   0.0 TEL | BEGIN SPAN main.py:40(main)
   0.0 TEL | BEGIN SPAN startup.py:74(__init__)
   0.0 TEL | Found kubectl -> /home/bartek/.asdf/shims/kubectl
   0.0 TEL | Found oc -> /home/bartek/.asdf/shims/oc
   0.0 TEL | [2] Capturing: kubectl version --short
   0.8 TEL | [2] captured in 0.74 secs.
   0.8 TEL | [3] Capturing: kubectl config current-context
   1.0 TEL | [3] captured in 0.20 secs.
   1.0 TEL | [4] Capturing: kubectl config view -o json
   1.1 TEL | [4] captured in 0.19 secs.
   1.1 TEL | [5] Capturing: kubectl --context *** get ns ike-tests-epyskbtjlxzngrwx
   1.9 TEL | [5] captured in 0.77 secs.
   2.5 TEL | Command: kubectl 1.11.9
   2.5 TEL | Context: -------, namespace: ike-tests-epyskbtjlxzngrwx, version: 1.14.6+73b5d76
   2.5 >>> | Warning: kubectl 1.11.9 may not work correctly with cluster version 1.14.6+73b5d76 due to the version discrepancy. See https://kubernetes.io/docs/setup/version-skew-policy/ for more information.
   2.5 >>> | 
   2.5 TEL | END SPAN startup.py:74(__init__)    2.5s
   2.5 TEL | Found ssh -> /usr/bin/ssh
   2.5 TEL | [6] Capturing: ssh -V
   2.5 TEL | [6] captured in 0.02 secs.
   2.5 TEL | Found ike -> /home/bartek/code/golang/src/github.com/maistra/istio-workspace/dist/ike
   2.5 TEL | Found torsocks -> /usr/bin/torsocks
   2.5 TEL | Found sshfs -> /usr/bin/sshfs
   2.5 TEL | Found fusermount -> /usr/bin/fusermount
   2.5 TEL | [7] Running: kubectl --context *** --namespace ike-tests-epyskbtjlxzngrwx get pods telepresence-connectivity-check --ignore-not-found
   3.5 TEL | [7] ran in 0.96 secs.
   4.2 TEL | Scout info: {'latest_version': '0.101', 'application': 'telepresence', 'notices': []}
   4.2 >>> | Starting network proxy to cluster using the existing proxy Deployment ratings-v1-v1-bartek-dmaxu
   4.2 TEL | [8] Running: kubectl --context *** --namespace ike-tests-epyskbtjlxzngrwx get deployment ratings-v1-v1-bartek-dmaxu
   5.2   8 | No resources found.
   5.2   8 | Error from server (NotFound): deployments.extensions "ratings-v1-v1-bartek-dmaxu" not found
   5.2 TEL | [8] exit 1 in 1.03 secs.
   5.2 >>> | 
   5.2 >>> | Failed to find deployment ratings-v1-v1-bartek-dmaxu:
   5.2 >>> | No resources found.
   5.2 >>> | Error from server (NotFound): deployments.extensions "ratings-v1-v1-bartek-dmaxu" not found
   5.2 >>> | 
   5.2 TEL | EXITING with status code 255
   5.2 >>> | Exit cleanup in progress
   5.2 TEL | (Cleanup) Stop time tracking
   5.2 TEL | END SPAN main.py:40(main)    5.2s
   5.2 TEL | (Cleanup) Remove temporary directory
   5.2 TEL | (Cleanup) Save caches
```
</details>

bartoszmajsak:
Previously called `/apis` endpoint which lists all resources in the cluster can be secured making it impossible to determine if the actual cluster is Openshift. In such a case Telepresence rolls back to `kubectl` and assumes `Deployment` is used rather than `DeploymentConfig` which is commonly used in Openshift.

The proposed solution calls Openshift-specific endpoint and based on the
response decides which cli tool to use.

Fixes #1139

### DIFF
--- a/telepresence/startup.py
+++ b/telepresence/startup.py
@@ -24,6 +24,10 @@ from urllib.request import urlopen
 
 from telepresence.runner import Runner
 
+KUBECTL = "kubectl"
+
+OC = "oc"
+
 
 def kubectl_or_oc(server: str) -> str:
     """
@@ -31,23 +35,31 @@ def kubectl_or_oc(server: str) -> str:
 
     :param server: The URL of the cluster API server.
     """
-    if which("oc") is None:
-        return "kubectl"
+    if which(OC) is None:
+        return KUBECTL
     # We've got oc, and possibly kubectl as well. We only want oc for OpenShift
-    # servers, so check for an OpenShift API endpoint:
+    # servers, so check for an OpenShift API endpoint exposing users
+    # (it's also used by oc whoami command):
     ctx = ssl.create_default_context()
     ctx.check_hostname = False
     ctx.verify_mode = ssl.CERT_NONE
     try:
-        with urlopen(server + "/apis", context=ctx) as response:
+        with urlopen(
+            server + "/apis/user.openshift.io/v1/users/", context=ctx
+        ) as response:
             api_group_list = str(response.read())
-    except (URLError, HTTPError):
-        return "kubectl"
+    except HTTPError as err:
+        if err.code == 403:
+            return OC
+        else:
+            return KUBECTL
+    except URLError:
+        return KUBECTL
 
     if "openshift" in api_group_list:
-        return "oc"
+        return OC
     else:
-        return "kubectl"
+        return KUBECTL
 
 
 def _parse_version_component(comp: str) -> int:
--- a/telepresence/startup.py
+++ b/telepresence/startup.py
@@ -24,10 +24,6 @@ from urllib.request import urlopen
 
 from telepresence.runner import Runner
 
-KUBECTL = "kubectl"
-
-OC = "oc"
-
 
 def kubectl_or_oc(server: str) -> str:
     """
@@ -35,8 +31,11 @@ def kubectl_or_oc(server: str) -> str:
 
     :param server: The URL of the cluster API server.
     """
-    if which(OC) is None:
-        return KUBECTL
+    kubectl = "kubectl"
+    oc = "oc"
+
+    if which(oc) is None:
+        return kubectl
     # We've got oc, and possibly kubectl as well. We only want oc for OpenShift
     # servers, so check for an OpenShift API endpoint exposing users
     # (it's also used by oc whoami command):
@@ -50,16 +49,16 @@ def kubectl_or_oc(server: str) -> str:
             api_group_list = str(response.read())
     except HTTPError as err:
         if err.code == 403:
-            return OC
+            return oc
         else:
-            return KUBECTL
+            return kubectl
     except URLError:
-        return KUBECTL
+        return kubectl
 
     if "openshift" in api_group_list:
-        return OC
+        return oc
     else:
-        return KUBECTL
+        return kubectl
 
 
 def _parse_version_component(comp: str) -> int:
--- a/newsfragments/1139.bugfix
+++ b/newsfragments/1139.bugfix
@@ -1,0 +1,1 @@
+Changing strategy to determine if `oc` (Openshift) cli tool should be used for a given telepresence session. Previously called `/apis` endpoint can be secured making `telepresence` defaulting to `kubectl` where the actual cluster is Openshift and uses `DeploymentConfig`s.


telepresence/startup.py
```diff
@@ -24,30 +24,42 @@
 
 from telepresence.runner import Runner
 
+KUBECTL = "kubectl"
+
+OC = "oc"
+
```
ark3: Could you move these "constants" into the function? They're not used anywhere else, and that's probably for the best.

Thank you for the fix!

bartoszmajsak: Fixed in 2dc56cd

telepresence/startup.py
```diff
@@ -31,23 +31,34 @@ def kubectl_or_oc(server: str) -> str:
...
-        with urlopen(server + "/apis", context=ctx) as response:
+        with urlopen(
+            server + "/apis/user.openshift.io/v1/users/", context=ctx
+        ) as response:
             api_group_list = str(response.read())
-    except (URLError, HTTPError):
-        return "kubectl"
+    except HTTPError as err:
+        if err.code == 403:
+            return oc
```
ark3: Why is this special-cased? I'm tripping on this now with my old-ish GKE cluster.

```console
$ oc whoami
error: No Auth Provider found for name "gcp"
```

bartoszmajsak: Sorry @ark3, you caught me during my holidays with this one. Can you provide me some more details about the GKE cluster which is causing the problem?

bartoszmajsak: The general idea is/was - if we have both `oc` and `kubectl` on the `$PATH` then the assumption is we most likely want to use OpenShift tooling and we connect to OpenShift cluster. If we call for api groups and getting `unauthorized` for the Openshift-specific endpoint, then we stick with OpenShift anyway.

Maybe we could also provide an optional flag for the cli (e.g. `--cluster-type`) where user can specifically say what is the actual vendor of the target cluster. If you think that would make sense I can open an issue and I'm happy to work on it.

ark3: GKE appears to return `unauthorized` for _all_ endpoints if it needs to grab a new access token or whatever using the auth provider helper. The code ends up thinking the cluster is OpenShift when really it's just a general failure to auth to the cluster. The `oc` error above is actually not the point, though it implies that `oc` doesn't know how to use Google Cloud's auth helper. But yeah, sorry to misdirect you with that.

Anyhow, I replaced your code (which is still there, but not called) with a different approach. Now Tel calls `kubectl api-versions` (or `oc api-versions` if `kubectl` is not available) and searches the output for "openshift.io". The implementation details are somewhat hackish (bad error handling in some cases) but the general approach seems sound. What do you think?

I ended up releasing this in 0.103. Can you please check it out? I hope I didn't break what you were fixing with your PR.

Issue #973, opened 2019-04-01
e2e tests rely on external httpbin.org


ark3:
... and that service, which we do not control, changes. 

1. We used three different IPs for httpbin.org to test three different `--also-proxy` invocations, relying on the existence of those three non-overlapping IPs to avoid masking a failure in one case with a success in another. But httpbin.org no longer has three different IPs (as of April 1, 2019). So https://github.com/telepresenceio/telepresence/commit/1de52bcf0ac29a9510d09640f7e03c6fd238a145 doubles the _two_ IPs we get, so that tests pass. This means masking of a failure is possible. Also, if two IPs turns into one, tests will start failing again.

    This is not so bad. We can test that the different `--also-proxy` invocations yield the correct underlying data structure with unit tests. Then we can use just one `--also-proxy` in end-to-end tests to reduce the scope of this problem.

2. The format of the answer returned by http://httpbin.org/ip has changed. It may change again. https://github.com/telepresenceio/telepresence/commit/2ba42e54729e8335b5a0c256f58f5fa7e2dbdcc4 handles the current (as of April 1, 2019) result as well as the prior one. But this sort of thing may break tests again in the future.

ark3:
Let's address this like so: https://github.com/telepresenceio/telepresence/pull/1167#issuecomment-549055271

donnyyung:
I believe this is no longer an issue in Telepresence 2. Here are the docs on how to install Telepresence (https://www.telepresence.io/docs/latest/install/), please re-open if you still see this issue in our latest version!

Issue #379, opened 2017-12-15
"--also-proxy hostname" finds one address for the hostname but misses any more


exarkun:
`--also-proxy` is not much documented but https://www.telepresence.io/reference/methods suggests using it with a domain name to have traffic destined for that domain proxied via telepresence.

The implementation resolves names given in this way on the cluster, turning them into IPs which can be given to sshuttle which will do the developer-local part of the proxying.  But the implementation only looks up one IP per hostname:

```
import socket, sys, json
result = []
for host in sys.argv[1:]:
    result.append(socket.gethostbyname(host))
sys.stdout.write(json.dumps(result))
sys.stdout.flush()
```

because `socket.gethostbyname` only returns one result.  If there are multiple addresses associated with the domain name then all but one are missed and will not be proxied.

`socket.getaddrinfo` can be used to get all the addresses (at least at the time of its call; if DNS is reconfigured afterwards, telepresence won't notice but the user's application might; separate issue):

```
>>> pprint(set(x[-1] for x in socket.getaddrinfo('duckduckgo.com', 0) if x[:2] == (socket.AF_INET, socket.SOCK_STREAM)))
set([('107.21.1.61', 0),
     ('184.72.106.52', 0),
     ('184.72.115.86', 0),
     ('23.21.193.184', 0)])
>>> 
```

ark3:
Fixed by #1167

PR #1167, opened 2019-10-24
resolve and proxy all ips for a given hostname when using --also-proxy

Flydiverny:
An attempt at solving #379 :)

disclaimer: I'm not a python dev :D

### DIFF
--- a/telepresence/outbound/vpn.py
+++ b/telepresence/outbound/vpn.py
@@ -49,7 +49,11 @@ import socket, sys, json
 
 result = []
 for host in sys.argv[1:]:
-    result.append(socket.gethostbyname(host))
+    host_ips = []
+    for x in socket.getaddrinfo(host, None):
+         if x[:2] == (socket.AF_INET, socket.SOCK_STREAM):
+            host_ips.append(x[4][0])
+    result.append(host_ips)
 sys.stdout.write(json.dumps(result))
 sys.stdout.flush()
 """
@@ -110,14 +114,17 @@ def k8s_resolve(
             continue
 
         if proxy_target in ipcache:
-            ip_ranges.append(ipcache[proxy_target])
+            if isinstance(ipcache[proxy_target], list):
+                ip_ranges += ipcache[proxy_target]
+            else:
+                ip_ranges.append(ipcache[proxy_target])
             continue
 
         hostnames.append(proxy_target)
 
     if hostnames:
         try:
-            resolved_ips = json.loads(
+            hostname_ips = json.loads(
                 runner.get_output(
                     runner.kubectl(
                         "exec", "--container=" + remote_info.container_name,
@@ -136,10 +143,12 @@ def k8s_resolve(
                 )
             )
     else:
-        resolved_ips = []
+        hostname_ips = []
 
-    for host, ip in zip(hostnames, resolved_ips):
-        ipcache[host] = ip
+    resolved_ips: List[str] = []
+    for host, ips in zip(hostnames, hostname_ips):
+        ipcache[host] = ips
+        resolved_ips += ips
 
     return resolved_ips + ip_ranges
 
--- a/telepresence/outbound/vpn.py
+++ b/telepresence/outbound/vpn.py
@@ -142,7 +142,7 @@ def k8s_resolve(
     else:
         hostname_ips = []
 
-    resolved_ips: List[str] = []
+    resolved_ips = []  # type: List[str]
     for host, ips in zip(hostnames, hostname_ips):
         ipcache[host] = ips
         resolved_ips += ips
--- a/telepresence/outbound/vpn.py
+++ b/telepresence/outbound/vpn.py
@@ -102,7 +102,7 @@ def k8s_resolve(
     hostnames = []
     ip_ranges = []
 
-    ipcache = runner.cache.child(runner.kubectl.context).child("ips")
+    ipcache = runner.cache.child(runner.kubectl.context).child("ip-list")
 
     for proxy_target in hosts_or_ips:
         try:
@@ -114,10 +114,7 @@ def k8s_resolve(
             continue
 
         if proxy_target in ipcache:
-            if isinstance(ipcache[proxy_target], list):
-                ip_ranges += ipcache[proxy_target]
-            else:
-                ip_ranges.append(ipcache[proxy_target])
+            ip_ranges += ipcache[proxy_target]
             continue
 
         hostnames.append(proxy_target)
--- a/newsfragments/379.bugfix
+++ b/newsfragments/379.bugfix
@@ -1,0 +1,1 @@
+Using --also-proxy will now proxy all resolved ips for a hostname.
--- a/tests/cluster/parameterize_utils.py
+++ b/tests/cluster/parameterize_utils.py
@@ -362,12 +362,17 @@ def create_deployment(deployment_ident, image, args, environ, ports, replicas):
 
     deployment = dumps({
         "kind": "Deployment",
-        "apiVersion": "extensions/v1beta1",
+        "apiVersion": "apps/v1",
         "metadata": {
             "name": deployment_ident.name,
             "namespace": deployment_ident.namespace,
         },
         "spec": {
+            "selector": {
+                "matchLabels": {
+                    "name": deployment_ident.name
+                }
+            },
             "replicas": replicas,
             "template": {
                 "metadata": {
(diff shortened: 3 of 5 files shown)


telepresence/outbound/vpn.py
```diff
@@ -110,14 +114,17 @@ def k8s_resolve(
             continue
 
         if proxy_target in ipcache:
-            ip_ranges.append(ipcache[proxy_target])
+            if isinstance(ipcache[proxy_target], list):
+                ip_ranges += ipcache[proxy_target]
+            else:
+                ip_ranges.append(ipcache[proxy_target])
```
Flydiverny: This really only needs to be `ip_ranges += ipcache[proxy_target]`, but it will give exciting results if user has an old cache. Is the cache cleared if you upgrade telepresence or so?

ark3: The cache is cleared when it is 24 hours old. That said, perhaps a better approach is to use a different cache key. So it'd be 

    cache[context]["ips"]["hostname"] -> one IP string

and

    cache[context]["ip-list"]["hostname"] -> list of IP strings

telepresence/outbound/vpn.py
```diff
@@ -136,10 +143,12 @@ def k8s_resolve(
                 )
             )
     else:
-        resolved_ips = []
+        hostname_ips = []
 
-    for host, ip in zip(hostnames, resolved_ips):
-        ipcache[host] = ip
+    resolved_ips: List[str] = []
```
ark3: This syntax does not work for Python 3.5, which Telepresence must support.

ark3: Use something like this instead:

    resolved_ips = []  # type: List[str]

ark3: (I'm annoyed that CI did not catch that, but I will fix that elsewhere...)

Flydiverny: Aight changed :D just had the linter complain on me as it wanted a type hint

telepresence/outbound/vpn.py
Flydiverny: Yes, was considering changing the key but wasn't sure if that would break something else. Couldn't really find anything related but figured I'd get feedback :) 
Changed it now!

ark3:
As discussed in Slack, we're trying to figure out how to adjust the automated tests to cover this.

Flydiverny:
Would it be possible to setup two subdomains or so on telepresence.io for use in the tests? Ie like `
multi-ip.test.telepresence.io` `single-ip.test.telepresence.io` and just have them point to some IPs and verify that these are proxied in the tests? Instead of relying on httpbin?

ark3:
Setting up domains is easy enough. But how do we verify they are proxied?

Flydiverny:
> Setting up domains is easy enough. But how do we verify they are proxied?

Yeah, got a bit stuck here. Is there any way we can validate the command used to run sshuttle? Ie. would it be enough to verify that the expected IPs are passed to sshuttle. Assuming other tests cover that sshuttle proxying works as expected? Not sure how to go about that tho :)

ark3:
Okay, I'll try to reorganize the code so that `--also-proxy` can be tested independently of a Telepresence invocation. That won't be an end-to-end test, but it'll resolve #973 and make it easier to test on local clusters (where the current `--also-proxy` tests fail).

I'll try my best to get this done soon, but I have a bunch of other stuff to do before KubeCon that'll take priority. That said, it may make sense to land your PR, assuming we can test it manually somehow, and keep testing all of this as part of #973.

Flydiverny:
One way to test this pr manually:

1. Get httpbin.org ip addresses:
```
dig httpbin.org
```

2. Curl each ip for your ip to see the non proxied result:
```sh
curl http://52.200.159.44/ip
curl http://3.225.168.125/ip
```

3. Start telepresence, from this pr
```sh
telepresence --also-proxy httpbin.org
```

4. Curl each ip for your ip to see the cluster IP:
```sh
curl http://52.200.159.44/ip
curl http://3.225.168.125/ip
```


should yield something like
```
~/Code/github/telepresence (⎈ |my-cluster:default)
virtualenv λ dig httpbin.org

; <<>> DiG 9.10.6 <<>> httpbin.org
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 37384
;; flags: qr rd ra; QUERY: 1, ANSWER: 2, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 1452
;; QUESTION SECTION:
;httpbin.org.			IN	A

;; ANSWER SECTION:
httpbin.org.		60	IN	A	52.200.159.44
httpbin.org.		60	IN	A	3.225.168.125

;; Query time: 32 msec
;; SERVER: 192.168.50.3#53(192.168.50.3)
;; WHEN: Sat Nov 02 19:19:13 CET 2019
;; MSG SIZE  rcvd: 72


~/Code/github/telepresence (⎈ |my-cluster:default)
virtualenv λ curl http://52.200.159.44/ip
{
  "origin": "<my-private-ip>, <my-private-ip>"
}

~/Code/github/telepresence (⎈ |my-cluster:default)
virtualenv λ curl http://3.225.168.125/ip
{
  "origin": "<my-private-ip>, <my-private-ip>"
}

~/Code/github/telepresence (⎈ |my-cluster:default)
virtualenv λ telepresence --also-proxy httpbin.org
T: Starting proxy with method 'vpn-tcp', which has the following limitations: All processes are affected, only one telepresence can run per machine, and you can't use other VPNs. You may need to add cloud
T: hosts and headless services with --also-proxy. For a full list of method limitations see https://telepresence.io/reference/methods.html
T: Volumes are rooted at $TELEPRESENCE_ROOT. See https://telepresence.io/howto/volumes.html for details.
T: Starting network proxy to cluster using new Deployment telepresence-1572718767-9866261-50483

T: No traffic is being forwarded from the remote Deployment to your local machine. You can use the --expose option to specify which ports you want to forward.

T: Connected. Flushing DNS cache.
T: Setup complete. Launching your command.

The default interactive shell is now zsh.
To update your account to use zsh, please run `chsh -s /bin/zsh`.
For more details, please visit https://support.apple.com/kb/HT208050.
@my-cluster|bash-3.2$ curl http://3.225.168.125/ip
{
  "origin": "<my-cluster-ip>, <my-cluster-ip>"
}
@my-cluster|bash-3.2$ curl http://52.200.159.44/ip
{
  "origin": "<my-cluster-ip>, <my-cluster-ip>"
}
```

----------




With out this PR one of the two curls in step 4 would yield the same IP as in step 2, as it was not proxied.

example with telepresence 0.102:

```sh
~/Code/github/telepresence master* (⎈ |my-cluster:default)
λ curl http://3.225.168.125/ip
{
  "origin": "<my-private-ip>, <my-private-ip>"
}

~/Code/github/telepresence master* (⎈ |my-cluster:default)
λ curl http://52.200.159.44/ip
{
  "origin": "<my-private-ip>, <my-private-ip>"
}

~/Code/github/telepresence master* (⎈ |my-cluster:default)
λ telepresence --also-proxy httpbin.org
T: How Telepresence uses sudo: https://www.telepresence.io/reference/install#dependencies
T: Invoking sudo. Please enter your sudo password.
Password:
T: Starting proxy with method 'vpn-tcp', which has the following limitations: All processes are affected, only one telepresence can run per machine, and you can't use other VPNs. You may need to add cloud
T: hosts and headless services with --also-proxy. For a full list of method limitations see https://telepresence.io/reference/methods.html
T: Volumes are rooted at $TELEPRESENCE_ROOT. See https://telepresence.io/howto/volumes.html for details.
T: Starting network proxy to cluster using new Deployment telepresence-1572719163-4271932-51725

T: No traffic is being forwarded from the remote Deployment to your local machine. You can use the --expose option to specify which ports you want to forward.

T: Setup complete. Launching your command.

The default interactive shell is now zsh.
To update your account to use zsh, please run `chsh -s /bin/zsh`.
For more details, please visit https://support.apple.com/kb/HT208050.
@my-cluster|bash-3.2$ curl http://3.225.168.125/ip
{
  "origin": "<my-cluster-ip>, <my-cluster-ip>"
}
@my-cluster|bash-3.2$ curl http://52.200.159.44/ip
{
  "origin": "<my-private-ip>, <my-private-ip>"
}
```

Flydiverny:
> Okay, I'll try to reorganize the code so that `--also-proxy` can be tested independently of a Telepresence invocation. That won't be an end-to-end test, but it'll resolve #973 and make it easier to test on local clusters (where the current `--also-proxy` tests fail).
> 
> I'll try my best to get this done soon, but I have a bunch of other stuff to do before KubeCon that'll take priority. That said, it may make sense to land your PR, assuming we can test it manually somehow, and keep testing all of this automatically as part of #973.

While I'd love to land this PR and remove some internal tooling, take your time! :) Wish I could be of more help 😅

PR #1227, opened 2020-01-13
Added the ability to set a node selector for a new deployment

### LINKED ISSUE #1096: Allow Telepresence to specify nodeSelector on proxy deployment when using --swap-deployment
We are considering using Telepresence in EKS with most of the nodes running on spot instances. Due to the nature of these type of instances, pods are often moved from one node to another by Kubernetes. 

The problem is that if you have Telepresence proxy deployment moved that disrupts the session created with `--swap-deployment` or `--new-deployment`.

A way to solve it could be by adding a flag to Telepresence specifying `nodeSelector` for proxy deployments (https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#nodeselector). This would allow to run proxy deployments on a long lived non spot EC2 instance, and all other deployments would go onto more ephemeral spot nodes.

billmcmath:
Added the ability to set a node selector for a new deployment

Fixes #1096

### DIFF
--- a/newsfragments/1096.feature
+++ b/newsfragments/1096.feature
@@ -1,0 +1,1 @@
+Telepresence can now set the node selector that is used in new deployments.  This is done using the `--node-selectors '{"key1": "value1", "key2": "value2"}'` flag.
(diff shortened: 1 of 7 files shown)


JorgenRingen:
Any updates on this? This would be be extremely useful.

vaidik:
+1

This change will be super helpful

Issue #1097, opened 2019-08-14
hostAlias entries are not working in a local docker container injected with --swap-deployment


MFAshby:
Steps to reproduce:
Have a kubernetes deployment with some hostAliases configured https://kubernetes.io/docs/concepts/services-networking/add-entries-to-pod-etc-hosts-with-host-aliases/

When you run telepresence with --swap-deployment, the hostAliases are not preserved, so the application running locally cannot use them.

MFAshby:
hmm, looks like they should be, see https://github.com/telepresenceio/telepresence/blob/0d9f33e77e09191c23c4880272c8e8a7d9365554/telepresence/proxy/deployment.py#L281

Not sure why this isn't working

MFAshby:
seems it is, but for some reason the application can't use the host alias.

MFAshby:
Possibly the proxy container that's deployed in the cluster doesn't check it's /etc/hosts file. 

https://github.com/telepresenceio/telepresence/blob/master/k8s-proxy/resolver.py

MFAshby:
I think it's specific to using --docker-run. If I run a local shell using --swap-deployment, I can do `nslookup` and results are returned from the hostAlias entry. If I run a local docker container using --docker-run, and execute nslookup from within that, then the results are not returned.

charliecatney:
I have the same issue, is there a fix or workaround.

Thanks

MFAshby:
Workaround: instead of using a hostAlias entry, I deployed a proxy instance (haproxy) and exposed it to my cluster using a service.

ark3:
To fix this for the container method we could either

1. Grab the proxy pod's `/etc/hosts` file like we do for `/etc/resolv.conf`
2. Process the hostAlias entries in the Pod spec

and then emit the appropriate `--add-host` arguments to `docker run` the way we do for the DNS stuff.

charliecatney:
Pull request submitted.

PR #1241, opened 2020-01-30
hostAlias injected into local docker container during --swap-deployment (Issue 1097)

charliecatney:
Adds remote pod host entries form k8s cluster to local running telepresence container.


charliecatney:
Fix for issue:
https://github.com/telepresenceio/telepresence/issues/1097

charliecatney:
@ark3, where is the test (before swapping) pod configured and can you give advice on how I could pass the host Alias?

It appears to be created using kubernaut in setup-for-ci.sh?
Thanks

ark3:
Take a look at `create_deployment(...)` in `tests/cluster/parameterize_utils.py`. It's essentially a normal Deployment manifest written out in Python. That Deployment is used for both the existing and swap test cases.

tests/cluster/parameterize_utils.py
```diff
@@ -406,6 +406,10 @@ def create_deployment(deployment_ident, image, args, environ, ports, replicas):
                     "containers": [container, sidecar_container],
                 },
             },
+            "hostAliases": [{
+                "ip": "127.0.0.1",
+                "hostnames": ["foo.local"]
+            }]
```
ark3: The `hostAliases` field needs to be in the Pod spec, not the Deployment spec. This Deployment does not validate.

tests/cluster/test_endtoend.py
```diff
@@ -460,6 +460,17 @@ def test_resolve_names(probe):
...
+def test_resolve_host_alias(probe):
+    """
+    Name resolution is performed in the context of the Kubernetes cluster.
+    """
+    result = probe.result()
+    result.write("gethostbyname foo.local")
+    success, reply = loads(result.read())
+    assert success and IPv4Address(reply) and "127.0.0.1" in reply[0], reply
+
+
```
ark3: This is a good test! How do I know? Because it fails! I modified your test Deployment so that the `hostAliases` field is in the correct spot. With that change, the cluster tests actually run, and this test fails.

charliecatney: Hi Ark, I've updated how the hosts file is parsed and manually validated that the local docker container is started with the hosts entries including a cat of the /etc/hosts.

I am however struggling to get a green run of the end to end tests locally. Not isolated to the new test.

Would you mind taking a look if possible, to see if my latest changes resolve the issues. As I say I have manually validated the change.

Thanks

tests/cluster/parameterize_utils.py
charliecatney: moved

ark3:
Thank you again for the PR.

Issue #1245, opened 2020-02-05
Users want to use a Kubernetes client in a swapped container


ark3:
and the best example of how to do that is hidden away in the Local Java Development docs.

https://www.telepresence.io/tutorials/java#kubernetes-client

We should bubble that up to a general item so that more users are likely to find it.

PR #1247, opened 2020-02-06
1245 users want to use a kubernetes client in a swapped container

gurayyildirim:
Hi!

I created a separate page for client libraries and moved Java example to there. Also prepared a Python example with officially supported Kubernetes Python library and documented the process.

The image I have used as an example is based on a [repo](https://github.com/gurayyildirim/python-kubernetes-get-info/) I created for this example.

Related to #1245

### DIFF
--- a/docs/SUMMARY.md
+++ b/docs/SUMMARY.md
@@ -18,6 +18,7 @@
 * [Rapid development with Kubernetes](tutorials/kubernetes-rapid.md)
 * [Get started with OpenShift](tutorials/openshift.md)
 * [Minikube VPN access](tutorials/minikube-vpn.md)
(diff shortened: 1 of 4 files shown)


docs/SUMMARY.md
```diff
@@ -18,6 +18,7 @@
 * [Rapid development with Kubernetes](tutorials/kubernetes-rapid.md)
 * [Get started with OpenShift](tutorials/openshift.md)
 * [Minikube VPN access](tutorials/minikube-vpn.md)
+* [Using Telepresence with Golang](tutorials/kubernetes-client-libs.md)
```
ark3: This link could say something like "Using a Kubernetes Client Library"

gurayyildirim: I forgot it, now fixed. Thank you!

PR #1259, opened 2020-02-21
add/patch securityContext runAsUser

AtzeDeVries:
The ssh server inside the `datawire/telepresence-k8s` must be run as user `0` because the ssh keys are only readable as root

```shell
➜  ~ docker run --rm docker.io/datawire/telepresence-k8s:0.104 ls -lha /etc/ssh
total 592
drwxr-xr-x    1 root     root        4.0K Jan 23 21:12 .
drwxr-xr-x    1 root     root        4.0K Feb 24 07:48 ..
-rw-r--r--    1 root     root      540.2K Mar  4  2019 moduli
-rw-r--r--    1 root     root        1.6K Mar  4  2019 ssh_config
-rw-r-----    1 root     root         672 Jan 23 21:12 ssh_host_dsa_key
-rw-r--r--    1 root     root         607 Jan 23 21:12 ssh_host_dsa_key.pub
-rw-r-----    1 root     root         227 Jan 23 21:12 ssh_host_ecdsa_key
-rw-r--r--    1 root     root         179 Jan 23 21:12 ssh_host_ecdsa_key.pub
-rw-r-----    1 root     root         411 Jan 23 21:12 ssh_host_ed25519_key
-rw-r--r--    1 root     root          99 Jan 23 21:12 ssh_host_ed25519_key.pub
-rw-r-----    1 root     root        1.6K Jan 23 21:12 ssh_host_rsa_key
-rw-r--r--    1 root     root         399 Jan 23 21:12 ssh_host_rsa_key.pub
-rw-r--r--    1 root     root        3.5K Jan 23 21:12 sshd_config
➜  ~ 
```

If you deployment is configured with a `runAsUser` which is not `0` and you have a `PodSecurityPolicy` with something like

```yaml
  runAsUser:
    ranges:
    - max: 65535
      min: 1
    rule: MustRunAs
```

then the telepresence image will not start. This PR will override or add the rusAsUser to be 0. If you use PSP, then you should supply a PSP which allows `runAsUser: 0`. You can combine this with the telepresence `--serviceaccount` switch and bind this policy to a specific SA

### DIFF
--- a/newsfragments/1239.misc
+++ b/newsfragments/1239.misc
@@ -1,2 +1,1 @@
-For the telepresence deployment, the runAsUser will now be set to 0. This is a usefull setting when you
-have a different runAsUser for your deployment or if you have Pod Security Policies defined.
+For the telepresence deployment, the runAsUser will now be set to 0. This is a usefull setting when you have a different runAsUser for your deployment or if you have Pod Security Policies defined.
--- a/newsfragments/1239.misc
+++ b/newsfragments/1239.misc
@@ -1,0 +1,1 @@
+For the telepresence deployment, the runAsUser will now be set to 0.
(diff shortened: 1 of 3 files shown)


AtzeDeVries:
i think error is because securityContext key does not exist in test deployment. Will fix way key is added

Issue #560, opened 2018-04-04
Kubernetes DNS Resolution Failing Spuriously on Mac


esorey:
I'm opening a telepresence session to our K8s cluster using method `vpn-tcp` without swapping out any deployments just to access K8s resources. Roughly half of the time I do this, it works perfectly. The other times, I get errors complaining that DNS resolution of the K8s addresses failed. However, when I run `dig <K8s-IP>`, it reports status NOERROR, so I know that there is no real issue with the addresses themselves. The only workaround I've found thus far is to restart my machine entirely. I've also tried the same setup on Linux and have not seen the issue after many runs. Specifically, this is on macOS Sierra 10.12.6. This seems like a particularly hairy issue that may be solved by the plans to run DNS locally, but I still wanted to document it.

Thank you!

ark3:
Your Mac ends up in some sort of broken state such that DNS resolution of cluster resources with vpn-tcp fails, and you have to restart the machine to fix things. Is that right? The next time this happens, could you please link to a gist of telepresence.log? Please try curl, dig, etc., so the DNS lookups are visible in the log. 

If you have to redact the logfile contents, it may be easier to extract the `sshuttle` and  pod logs. See the trace below for an example.

Thank you for your help!

```shell
$ telepresence
Starting proxy with method 'vpn-tcp', which has the following limitations: All processes are affected, only one telepresence can run per machine, and you can't use other VPNs. You may need to add cloud hosts with --also-proxy. For a full list of method limitations see https://telepresence.io/reference/methods.html
Volumes are rooted at $TELEPRESENCE_ROOT. See https://telepresence.io/howto/volumes.html for details.

No traffic is being forwarded from the remote Deployment to your local machine. You can use the --expose option to specify which ports you want to forward.

Guessing that Services IP range is 10.3.240.0/20. Services started after this point will be inaccessible if are outside this range; restart telepresence if you can't access a new Service.

@gke_datawireio_us-central1-a_telepresence-testing|bash-4.4$ curl -sk https://kubernetes/api/
{
  "kind": "APIVersions",
  "versions": [
    "v1"
  ],
  "serverAddressByClientCIDRs": [
    {
      "clientCIDR": "0.0.0.0/0",
      "serverAddress": "35.184.xx.xx"
    }
  ]
}@gke_datawireio_us-central1-a_telepresence-testing|bash-4.4$ 
@gke_datawireio_us-central1-a_telepresence-testing|bash-4.4$ dig kubernetes

[...]

PR #1268, opened 2020-03-13
Handle more kube DNS resolution formats

agrenott:
Add support of xxx.yyy.zzz.svc kube addresses DNS resolution, on top of
already supported "service" and "service.namespace".

This is required to handle Strimzi kafka auto-generated addresses for
example (see https://github.com/strimzi/strimzi-kafka-operator/issues/2656).


agrenott:
Changes performed.
Attached this PR to #560, as it should fix the latest issue reported in https://github.com/telepresenceio/telepresence/issues/560#issuecomment-484476698

ark3:
Thank you!

Issue #1297, opened 2020-04-08
`kubectl run` creates a pod instead of a deployment as of kube 1.18


marun:
Kubernetes 1.18 removes support from `kubectl run` for creating anything but a pod:

https://kubernetes.io/docs/setup/release/notes/#kubectl

This would appear to break many of the documented examples for telepresence that rely on `kubectl run` to spin up a deployment that can be targeted for replacement.

jtyers:
I haven't looked into the code to see whether this is a big fix or not, but there is a workaround I've just found, which is to create the deployment manually.

Create a telepresence deployment:

```
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  name: my-telepresence-deployment
spec:
  template:
    metadata:
      labels:
        app: telepresence
    spec:
      containers:
      - name: my-telepresence-deployment
        image: datawire/telepresence-k8s:0.104
        env:
        - name: MY_POD_IP
          valueFrom:
            fieldRef:
              fieldPath: status.podIP
```

Then run telepresence, instructing it to use the existing deployment:

```
telepresence -d my-telepresence-deployment [your other args here]
```

This seems to fix it for me; Telepresence starts fine and appears to proxy through to pods normally as before.


I think this isn't the first time changes in new versions of Kubernetes have broken Telepresence. As another workaround, you might want to consider allowing users to specify what `kubectl` binary to use. That way, I can download `kubectl` 1.17.x and point telepresence to use that instead of the latest version, which is installed by my distro's packaging system.

```
# if only I could do this...
telepresence --kubectl=/usr/local/bin/kubectl-1.17.4 [...etc]
```

Working with different clusters, I find I need multiple versions of kubectl installed anyway, and it is a more durable solution to when upstream changes in `kubectl` cause errors, until the underlying issue is fixed.

marun:
@jtyers In the absence of direct support for specifying the binary:

```bash
mkdir ~/tp-bin
ln -s /usr/local/bin/kubectl-1.17.4 ~/tp-bin/kubectl
PATH=~/tp-bin:$PATH telepresence
```

marun:
Beyond breaking documented examples that expect that `kubectl run` has been executed to create a deployment prior to executing `telepresence`, a bare invocation that depends on `kubectl run` creating a deployment:

https://github.com/telepresenceio/telepresence/blob/master/telepresence/proxy/deployment.py#L129

[...]

PR #1302, opened 2020-04-15
Fix doc compatibility with kubectl >= 1.18

marun:
As of 1.18, `kubectl run` only creates pods. Rather than document the different behavior between versions, update the docs to always create deployments with `kubectl create deployment` which is compatible with all versions of `kubectl`.

Intended to address #1297 from a docs perspective 

TODO
 - [ ] - Figure out how best to update the [transcript](https://github.com/telepresenceio/telepresence/blob/master/docs/internal/transcript.json) to reflect the changes in `demo-script.txt`

### DIFF
--- a/docs/howto/golang.md
+++ b/docs/howto/golang.md
@@ -5,10 +5,11 @@ Because of the way Go is implemented you will need to use `--method vpn-tcp` wit
 
 ### A Go program talking to Kubernetes
 
-First, start a web server inside Kubernetes:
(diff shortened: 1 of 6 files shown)


marun:
Given that I only changed docs in this PR, I'm assuming master is broken wrt CI?

ark3:
Thank you for the PR.

> Given that I only changed docs in this PR, I'm assuming master is broken wrt CI?

Yes, for the reason you noted and fixed in #1313.

Issue #462, opened 2018-02-21
telepresence --docker-run fails from an ec2 instance in a VPC


blak3mill3r:
I'm trying without success to run `telepresence --docker-run` ... from an ec2 instance inside our VPC

I've spent a lot of time trial-and-error narrowing down what's wrong

`telepresence --docker-run -it debian:latest /bin/bash` <-- pretty simple failing case

in the `telepresence.log` I can see that it is trying to resolve a strange DNS name:

```
9.4 42 | ssh: Could not resolve hostname ip-172-17-0-1.e: Name does not resolve^M
```

I have a feeling this might have to do with EC2's DNS server behavior, which involves certain names resolving differently from inside the AWS network ... DNS names that are public and resolve from outside as a public IP will resolve as a private IP if the DNS query is from inside EC2.

I wonder if anyone else is trying to do this? `telepresence --docker-run` from an ec2 instance?

Interestingly, `telepresence --method=vpn-tcp` works fine from this ec2 instance.

Sometimes when I run the above, `telepresence` hangs forever, and in the log I can see it waiting for `20.7 TL | [54] Running: (['docker', 'run', '--network=container:telepresence-1519197605-8019295-17852', '--rm', 'datawire/telepresence-local:0.75', 'wait'],)...`, which never happens because no container `telepresence-1519197605-8019295-17852` is running.

The ec2 instance is Ubuntu 16.04 LTS from an official AMI from https://cloud-images.ubuntu.com/locator/ec2/ which I launched earlier today (specifically: `ami-0b383171`), I installed `docker-ce` from the [docker apt repo](https://docs.docker.com/install/linux/docker-ce/ubuntu/) and installed `telepresence` `0.75` via apt, all today.

It does launch the `telepresence` deployment on the remote end, and the logs from that pod look normal.

the `Would you like to file an issue in our issue tracker?` bit does not work from the ec2 instance (presumably because it cannot launch a web browser?)

here's the part that is relevant (pretty certain)

```
   8.9 41 | Starting sshuttle proxy.
   9.1 41 | firewall manager: Starting firewall with Python version 3.6.1
   9.1 41 | firewall manager: ready method name nat.
   9.1 41 | IPv6 enabled: False
   9.1 41 | UDP enabled: False
   9.1 41 | DNS enabled: True
   9.1 41 | TCP redirector listening on ('127.0.0.1', 12300).
   9.1 41 | DNS listening on ('127.0.0.1', 12300).
[...]

PR #1303, opened 2020-04-15
Delete statements about limitation on EC2 in documents

shibataka000:
As far as I see https://github.com/telepresenceio/telepresence/issues/462#issuecomment-376244255 , limitation on EC2 described at #462 was solved >=0.76 . So I delete statements about that limitations in documents.


ark3:
Thank you for the PR.

PR #1304, opened 2020-04-15
Fix pod CIDR detection when falling back to using pod IPs

marun:
If the pod CIDR can not be determined by looking at node configuration, an attempt will be made to collect pod ips and find a CIDR that covers them. Previously only pods in the current namespace were collected which may not result in an accurate CIDR. This change ensures that the IPs of pods from all namespaces are considered.


marun:
Rebased to pick up ci fix.

ark3:
Thank you! Landing this via #1322 shortly.

Issue #1305, opened 2020-04-15
ocp image not configured for password-less sudo


marun:
Attempting to invoke telepresence against an ocp cluster results in the output that follows, indicating that password-less sudo is not configured correctly. Using the default remote image works fine against the same cluster.

```
telepresence --namespace=openshift-authentication-operator --swap-deployment authentication-operator --run run-local-operator.sh
T: Failed to find OpenShift deploymentconfig authentication-operator:
T:   Error from server (NotFound): deploymentconfigs.apps.openshift.io "authentication-operator" not found
T: Will try regular Kubernetes Deployment.
T: Starting proxy with method 'vpn-tcp', which has the following limitations: All processes are affected, only one telepresence can run per machine, and you can't use other VPNs. You may need to add cloud hosts and headless services with --also-proxy. For a full list of
T: method limitations see https://telepresence.io/reference/methods.html
T: Volumes are rooted at $TELEPRESENCE_ROOT. See https://telepresence.io/howto/volumes.html for details.
T: Starting network proxy to cluster by swapping out Deployment authentication-operator with a proxy
telepresence@127.0.0.1's password:
telepresence@127.0.0.1's password:
telepresence@127.0.0.1's password:
telepresence@127.0.0.1's password:
telepresence@127.0.0.1's password:
telepresence@127.0.0.1's password:

Looks like there's a bug in our code. Sorry about that!

Traceback (most recent call last):
  File "/home/maru/src/telepresence/telepresence/cli.py", line 135, in crash_reporting
    yield
  File "/home/maru/src/telepresence/telepresence/main.py", line 68, in main
    socks_port, ssh = do_connect(runner, remote_info)
  File "/home/maru/src/telepresence/telepresence/connect/connect.py", line 119, in do_connect
    args.from_pod
  File "/home/maru/src/telepresence/telepresence/connect/connect.py", line 70, in connect
    raise RuntimeError("SSH to the cluster failed to start. See logfile.")
RuntimeError: SSH to the cluster failed to start. See logfile.
```

thallgren:
I believe this is no longer an issue in Telepresence 2. Here are the docs on how to install Telepresence (https://www.telepresence.io/docs/latest/install/), please re-open if you still see this issue in our latest version!

PR #1306, opened 2020-04-15
WIP Allow the default remote image to be used with ocp via env var

marun:
The default remote image is lightweight compared to the centos-based ocp image and compatible with ocp when invoked with sufficient privilege. This change allows the use of the default remote image on ocp when `TELEPRESENCE_OCP_USE_DEFAULT_IMAGE` is set to a non-empty value.

Marking WIP because I have no attachment to the proposed implementation. I just want to allow use of the default image on ocp. For me the centos image [doesn't work](https://github.com/telepresenceio/telepresence/issues/1305) due to password-less sudo not being configured correctly. While the centos image can and should be fixed, but I don't think everyone using ocp wants or needs the less-supported option that the centos image represents.

### DIFF
--- a/telepresence/__init__.py
+++ b/telepresence/__init__.py
@@ -46,9 +46,15 @@ TELEPRESENCE_REMOTE_IMAGE = "{}/telepresence-k8s:{}".format(
 TELEPRESENCE_REMOTE_IMAGE_PRIV = "{}/telepresence-k8s-priv:{}".format(
     REGISTRY, image_version
 )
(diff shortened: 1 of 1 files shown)


marun:
Rebased to pick up ci fix.

ark3:
Thanks for the PR. I went with a different implementation in #1323 that will let you achieve your goal by setting `TELEPRESENCE_USE_OCP_IMAGE=NO`.

PR #1312, opened 2020-04-19
Create deployments with `kubectl create` for compatibility with kubectl 1.18

### LINKED ISSUE #1297: `kubectl run` creates a pod instead of a deployment as of kube 1.18

marun:
As of kube 1.18, `kubectl run` is no longer capable of creating a deployment. This commit switches to supplying deployment yaml to `kubectl create -f` to ensure compatibility with all supported versions of kubectl. If `--expose` is specified, `kubectl expose` will be used to create a service.

Note that the yaml specifies `apiVersion: apps/v1`, which is compatible with kube >= 1.9.

Fixes #1297

### DIFF
--- a/telepresence/proxy/__init__.py
+++ b/telepresence/proxy/__init__.py
@@ -74,11 +74,7 @@ def setup(runner: Runner, args):
     if args.new_deployment is not None:
         # This implies --new-deployment
         deployment_arg = args.new_deployment
-        if runner.kubectl.cluster_is_openshift:
-            # FIXME: This will not clean up the new dc
-            deployment_type = "deploymentconfig"
-        else:
-            deployment_type = "deployment"
+        deployment_type = "deployment"
         operation = create_new_deployment
         args.operation = "new_deployment"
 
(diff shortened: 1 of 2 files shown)


marun:
Rebased to pick up ci fix.

marun:
> Thank you for the PR.
> 
> 1. Telepresence supports Python 3.5, so the `f"..."` strings won't work. Can you please change that to `"...".format(...)` or whatever is appropriate?
> 2. Telepresence manages all temporary files using a single temporary directory. See `runner.make_temp(...)`.
> 3. That said, the `kubectl apply` case doesn't need a temporary file. The subprocess management helpers (`runner.check_call(...)` et al) support passing an input string as stdin via the `input` keyword argument, so you should be able to `kubectl apply -f -` and pass the manifest in directly.

Updated, PTAL.

ark3:
Looks good! Landing shortly via #1318

PR #1313, opened 2020-04-19
Fix mypy lint errors

marun:
I've submitted a few PRs this week and all of them failed CI due to these lint errors. I finally took the time to figure out how to fix them. Not sure how the errors crept in to master though. Is the version of mypy floating such that a newer and stricter release got pulled in automatically?


ark3:
Thank you for the PR!

>  Is the version of mypy floating such that a newer and stricter release got pulled in automatically?

Yes, exactly.

Issue #1307, opened 2020-04-16
Failed to connect to proxy in remote cluster


ZolotarenkoM:
### What were you trying to do?

`telepresence --mount=/tmp/known --swap-deployment app_deployment --namespace provisioners --docker-run --rm -it -v=/tmp/known/var/run/secrets:/var/run/secrets app_new_image`

### What did you expect to happen?

expected to get an interactive shell in the context of the running pod

### What happened instead?

It crashed with issue `Failed to connect to proxy in remote cluster.`

### Automatically included information

Command line: `['/usr/local/bin/telepresence', '--mount=/tmp/known', '--swap-deployment', 'app_deployment', '--namespace', 'provisioners', '--docker-run', '--rm', '-it', '-v=/tmp/known/var/run/secrets:/var/run/secrets', 'app_new_image']`
Version: `0.104`
Python version: `3.7.7 (default, Mar 10 2020, 15:43:03) 
[Clang 11.0.0 (clang-1100.0.33.17)]`
kubectl version: `Client Version: v1.14.3 // Server Version: v1.15.9`
oc version: `(error: [Errno 2] No such file or directory: 'oc': 'oc')`
OS: `Darwin mykhailozolotarenko-r 18.7.0 Darwin Kernel Version 18.7.0: Thu Jan 23 06:52:12 PST 2020; root:xnu-4903.278.25~1/RELEASE_X86_64 x86_64`

```
Traceback (most recent call last):
  File "/usr/local/bin/telepresence/telepresence/cli.py", line 135, in crash_reporting
    yield
  File "/usr/local/bin/telepresence/telepresence/main.py", line 82, in main
    runner, remote_info, env, socks_port, ssh, mount_dir, pod_info
  File "/usr/local/bin/telepresence/telepresence/outbound/setup.py", line 169, in launch
    pod_info,
  File "/usr/local/bin/telepresence/telepresence/outbound/container.py", line 165, in run_docker_command
    TELEPRESENCE_LOCAL_IMAGE, "wait"
  File "/usr/local/bin/telepresence/telepresence/runner/runner.py", line 462, in check_call
    env,
  File "/usr/local/bin/telepresence/telepresence/runner/runner.py", line 437, in _run_command_sync
    None if stderr_to_stdout else err_logger.get_captured(),
[...]

PR #1330, opened 2020-04-28
Detect  cluster domain and build FQDN for hostnames involved in connectivity tests

ipleten:
My attempt to address #1307.

### DIFF
--- a/local-docker/entrypoint.py
+++ b/local-docker/entrypoint.py
@@ -65,7 +65,7 @@ def main() -> None:
     if command == "proxy":
         proxy(loads(sys.argv[2]))
     elif command == "wait":
-        wait()
+        wait(sys.argv[2])
 
 
 def proxy(config: typing.Dict[str, typing.Any]) -> None:
@@ -113,13 +113,13 @@ def proxy(config: typing.Dict[str, typing.Any]) -> None:
     runner.wait_for_exit(main_process)
 
 
-def wait() -> None:
+def wait(hostname) -> None:
     """Wait for proxying to be live."""
     start = time()
     while time() - start < 30:
         try:
-            gethostbyname("kubernetes.default")
             sleep(1)  # just in case there's more to startup
+            gethostbyname(hostname)
             sys.exit(100)
         except gaierror:
             sleep(0.1)
--- a/newsfragments/1307.bugfix
+++ b/newsfragments/1307.bugfix
@@ -1,0 +1,1 @@
+Getting cluster domain from pod's /etc/resolv.conf to construct FQDN and use it for connectivity test (--method inject-tcp).
--- a/telepresence/outbound/container.py
+++ b/telepresence/outbound/container.py
@@ -25,6 +25,7 @@ from telepresence.connect import SSH
 from telepresence.proxy import RemoteInfo
 from telepresence.runner import Runner
 from telepresence.utilities import find_free_port, random_name
+from telepresence.utilities import get_cluster_domain
 
 
 def make_docker_kill(runner: Runner, name: str) -> Callable[[], None]:
@@ -156,6 +157,7 @@ def run_docker_command(
     if "resolv" in pod_info:
         dns_args.extend(parse_resolv_conf(pod_info["resolv"]))
 
+    cluster_domain = get_cluster_domain(pod_info['resolv'])
     # Image already has tini init so doesn't need --init option:
     span = runner.span()
     runner.launch(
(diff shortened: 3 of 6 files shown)


ark3:
Can you help me understand the goal for this change?

You've set ndot=1 in your Pod, presumably for a good reason. Otherwise you could fix your problem by not modifying ndots in your Pod. Is that fair?

So I think the correct solution would be for Telepresence not to misbehave when ndots is set, presumably by doing fully-qualified lookups only. Does that make sense?

Or do you _really_ want networking to behave differently in the cluster and locally?

ipleten:
We changed settings to `ndots:1` as default value `5` dramatically increases latency for our services. (https://pracucci.com/kubernetes-dns-resolution-ndots-options-and-why-it-may-affect-application-performances.html)
We can't change it to default without impacting performance. We plan to use Telepresence for TEST cluster and probably for PROD for deeper debugging.  While TEST cluster might work fine with "ndots:5" that requires us to use a different configurations for TEST/PROD clusters.
Currently it's impossible due to issue #1307.
So for now workflow looks like: 
1) Edit deployment and set ndots
2) Wait till pod got recreated :(
3) Swap deployment and do some debugging developing
4) Revert everything back.

ipleten:
I tried to figure out how to get an  k8s 'cluster-domain' value and seems there is no reliable (via API) way to do that.

ipleten:
The funny thing, while python works and is able to resolve, problem seems come from `torsocks`:

python:
> root@python-test-ndots-ipleten-5bdbb59c74-zjwdv:/# python3 -c "import socket; socket.socket().connect(('kubernetes.default', 443))"

>root@python-test-ndots-ipleten-5bdbb59c74-zjwdv:/# torsocks python3 -c "import socket; socket.socket().connect(('kubernetes.default', 443))"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
socket.gaierror: [Errno -4] Non-recoverable failure in name resolution

curl:
>root@python-test-ndots-ipleten-5bdbb59c74-zjwdv:/# curl google.com -s >/dev/null
root@python-test-ndots-ipleten-5bdbb59c74-zjwdv:/# torsocks curl google.com
curl: (6) Could not resolve host: google.com

As well as I could confirm that proxy is working .
>root@python-test-ndots-ipleten-5bdbb59c74-zjwdv:/# torsocks curl http://192.168.0.1 -s > /dev/null

I have ndots:1 on this pod and create ssh SOCKS5 proxy to another machine on localhost:9050.


I tried to call socket.getaddinfo and getfqdn() but none of them works via `torsocks`.

ark3:
I'm missing some context here (or I have forgotten... sorry!). The ssh-based socks proxy doesn't support the TOR extensions for DNS, does it? And why is torsocks relevant here, anyhow?

You can construct FQDNs by reading `/etc/resolv.conf` in the proxy pod. We already pass that file back to the CLI application via podinfo.py, and parse that file in container.py, but we don't use it for this purpose. Yes, this is not a reliable or supported mechanism, but that is the unfortunate truth of working on Telepresence.

ipleten:
Yeah `torsocks` doesn't work with ssh proxy :(
I tried to figure out the whole chain and reproduce it without telepresence, my bad. 
Sorry for noise. I'll try to concentrate on FQDN.

ark3:
Ah, okay. I'm sorry you had to figure out/reverse engineer the chain. We need to have far better documentation!

Thank you very much for your help on this.

ark3:
Not stale. I'd like to poke at fixing the FQDN stuff everywhere in Tel. And maybe @ipleten can get us most of the way there! :smile:

ipleten:
Sorry, I'll try to find time to fix that.

ipleten:
Should I fix formating?

ipleten:
hm... it also hardcoded for 'container' mode

ipleten:
Not sure that I need to do with failed test.
I tested it a bit manually and it works on my cluster now.

ipleten:
Hello, @ark3 
Could you please review, if you have a moment, and let me know if it looks more or less fine?

ark3:
Sorry I haven't had a chance to work on this. At a glance, this looks excellent! But I need to try it out on a few machines and clusters before I land it, and that will have to wait for next week. 

MacOS CI failing is its own problem (#1455)...

Thank you for this great stuff!

ipleten:
not stale.

ark3:
I'm concerned about `get_cluster_domain(...)` itself. It has no error checking and it can potentially return a bogus answer. There's no fallback path to the current behavior. Aside from that function, the rest looks good. Thoughts?

ipleten:
> I'm concerned about `get_cluster_domain(...)` itself. It has no error checking and it can potentially return a bogus answer. There's no fallback path to the current behavior. Aside from that function, the rest looks good. Thoughts?

Well, while it's possible to fallback to current behavior, I'm not sure how to detect that cluster domain was detected properly. Looks like yet another `try`  https://github.com/telepresenceio/telepresence/blob/b9661e48c05a41dd457273dce5014d73ef315ec7/telepresence/outbound/local.py#L66  is needed, hence if it fails for new logic we could try current 'kubernetes.default' hostname?

Issue #1226, opened 2020-01-13
[privacy] Support to turn off verbose k8s-proxy logging


KarolisL:
At least in VPN mode, k8s-proxy logs EVERY query executed from telepresence host machine. 
I see this as a **privacy breach**, since any cluster or logging system user can see which websites were accessed by a telepresence user while they were connected to telepresence.

Example of log messages:
```
2020-01-13T10:44:45+0000 [stdout#info] A query: b'github.com'
2020-01-13T10:44:45+0000 [stdout#info] Result for b'github.com' is ['140.82.118.3']
```

Ideally, such logs messages would be disabled by default, but it would be alright for us to have it as a CLI option (since we're using wrapper shellscript anyway).

machine424:
@KarolisL, if you run Telepresence manually: (You have control on the `docker` image to be used in the `Telepresence`'s pod); As a simple workaround you can prevent `Twisted` (the framework that runs, catches `Telepresences`'s stdout and displays it as logs) from displaying its logs, you can do that by sending logs into `/dev/null` by replacing https://github.com/telepresenceio/telepresence/blob/master/k8s-proxy/run-priv.sh#L4 with `exec env PYTHONPATH=/usr/src/app twistd -l /dev/null--pidfile= -n -y ./forwarder.py` and rebuilding the docker image (`Dockerfile` is in the same folder), make sure that the docker image contains `/telepresence-` or `/telepresence-k8s` otherwise `Telepresence` will raise an error here: https://github.com/telepresenceio/telepresence/blob/master/telepresence/proxy/remote.py#L46.
To prevent `/dev/null` from being rotated if it receives SIGUSR1 you can use `exec env PYTHONPATH=/usr/src/app twistd -l=- --pidfile= -n -y ./forwarder.py 2>&1 >/dev/null` instead.
@ark3, Does `Telepresence`'s CLI need to control `Telepresence`'s output (in this case we need to do sth about `prints`'s, especially in https://github.com/telepresenceio/telepresence/blob/master/k8s-proxy/resolver.py or should we make `Twisted`'s logging customisable ?

donnyyung:
I believe this is no longer an issue in Telepresence 2, since our log levels are customizable https://www.telepresence.io/docs/latest/reference/config/#log-levels. Here are the docs on how to install Telepresence (https://www.telepresence.io/docs/latest/install/), please re-open if you still see this issue in our latest version!

Issue #985, opened 2019-04-05
Optional configuration


ark3:
Users want to configure many aspects of Telepresence. However, the Telepresence command line experience is already too complicated; adding even more options would be overwhelming. It might be useful to offer **completely optional** configuration from another source (not the cli) for Telepresence. 

- configuration of the proxy Pod
  - resources (#1422)
  - node selector (#1096)
  - others?
- hosts/URLs for connectivity checks #389
- various timeouts #777 and #939 (cli `--timeout` option PR)
- executable to use in place of kubectl https://github.com/telepresenceio/telepresence/issues/914#issuecomment-479951009
  - use a specific executable (e.g., `kubectl1.10` for an old cluster)
  - use command line options Tel doesn't support (e.g., `kubectl --server foo --token $BLAH`)
  - force `kubectl` vs `oc` vs ...
- mount options
  - mount to a known directory
  - no volume mounting at all
- how `--verbose` modifies the helpers called by Telepresence
  - `ssh -vv` #1015
  - `kubectl --v=4` https://github.com/telepresenceio/telepresence/issues/807#issuecomment-430334967
- proxy to host polling interval
- default logfile location
- cache location, expiration times
- default method?
- other logging configuration (#1226)

Tel already uses environment variables for optional configuration (TELEPRESENCE_VERSION and TELEPRESENCE_REGISTRY). It could also look for a configuration file. 

See also #411 and #455.

davidmichaelkarr:
I'm trying to consider the POV of whether adding additional options could be "overwhelming".  Certainly when a list of options is considered from the POV of the entire list of options, the more options you have, the more difficult it is to find the ones you need.

The simple solution implemented in numerous tools is simply to present the options list in groups.  I'm sure tp is already doing this.  kubectl already has many options. I don't find it "overwhelming". I look through the docs for what I need and ignore the rest.

Compare this with the steps that people might have to take to work around the lack of certain options. In my case, putting an "alternate" kubectl in the PATH really won't be helpful. That will be quite a bit messier than simply accepting the fact that I'll have to move to using contexts, which isn't as convenient in my situation.

However, if as in the description of this issue, you were able to implement some unusual method of supplying ANOTHER command-line argument that points tp to some sort of alternate "command-line processor" so I can still have "--server" and "--token" on the tp command line and not having those options added to tp itself, then I guess it's entirely possible that could work, but I don't see how this could be possible.

tobiasfriden:
[...]

PR #1336, opened 2020-05-02
Make it possible to hide the Telepresence proxy output

machine424:
- Adds the flag `--suppress-proxy-output` (may reconsider its name, maybe an env var is sufficient) to hide the output of the proxy Pod.
- An attempt to address: https://github.com/telepresenceio/telepresence/issues/1226
- I have some questions here: https://github.com/telepresenceio/telepresence/issues/1226#issuecomment-576449777
- I think we can add an e2e test that checks the `kubectl logs` of the proxy pod, can we set the flag for `_ExistingDeploymentOperation` e.g., so we don't have to add another dimension to the tests parameters (probe) ?

### DIFF
--- a/k8s-proxy/run-ocp.sh
+++ b/k8s-proxy/run-ocp.sh
@@ -3,4 +3,8 @@ set -e
 echo -e ",s/1000/`id -u`/g\\012 w" | ed -s /etc/passwd
 ssh-keygen -A
 /usr/sbin/sshd -e
-exec env PYTHONPATH=/usr/src/app twistd --pidfile= -n -y ./forwarder.py
+if [ "$TELEPRESENCE_SUPPRESS_PROXY_OUTPUT" = "1" ]; then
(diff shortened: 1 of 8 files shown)


ark3:
Thank you for the PR! We can't accept this as-is for two reasons:

1. Your changeset will effectively turn off all logging in the Pod. Experience has shown that this causes the `kubectl log` command to disconnect from inactivity, which causes the session to end. I know it's not _supposed to_ work that way, but that is what our users have seen.
2. You've added a command line flag for a configuration option that most users will not need. We want to have all new configuration of that nature in the optional config file (#985, which is not implemented yet...). 

Can I ask you to make a more targeted change? All the logging that has privacy ramifications is in `resolver.py`. Can you modify just that file to look for the environment variable and log less as is appropriate? Then we can address how to set that variable in a different PR.

Thank you for your contribution.

ark3:
@machine424 Just checking in. Will you be able to make a more targeted change in resolver.py?

Issue #1220, opened 2019-12-16
HTTP/1.1 301 Moved Permanently connection redirect within --docker-run times out


josephmate:
The problem is reproducible every time by asking for an HTTP GET, and the connection gives back an HTTP/1.1 301 Moved Permanently

```
telepresence --docker-run --rm centos:centos7   curl -i -v http://ec2.amazonaws.com
T: Volumes are rooted at $TELEPRESENCE_ROOT. See https://telepresence.io/howto/volumes.html for details.
T: Starting network proxy to cluster using new Deployment telepresence-1576532273-094672-1474

T: No traffic is being forwarded from the remote Deployment to your local machine. You can use the --expose option to specify which ports you want to forward.

T: Setup complete. Launching your container.
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0* About to connect() to ec2.amazonaws.com port 80 (#0)
*   Trying 54.239.29.8...
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0* Connected to ec2.amazonaws.com (54.239.29.8) port 80 (#0)
> GET / HTTP/1.1
> User-Agent: curl/7.29.0
> Host: ec2.amazonaws.com
> Accept: */*
>
  0     0    0     0    0     0      0      0 --:--:--  0:07:30 --:--:--     0* Empty reply from server
  0     0    0     0    0     0      0      0 --:--:--  0:07:30 --:--:--     0
* Connection #0 to host ec2.amazonaws.com left intact
curl: (52) Empty reply from server
T: Your process exited with return code 52.
T: Exit cleanup in progress
T: Cleaning up Deployment telepresence-1576532273-094672-1474
```

Summary:

telepresence --swap-deployment my-component --docker-run my-image results in my java application throwing `javax.net.ssl.SSLException: Read timed out` - I think the connection stalls, and never receives the HTTP 301 Moved Permanently because:
- `kubectl exec -t <podId> -- curl -i https://ec2.amazonaws.com` works on the original remote pod before running telepresence [1]
- `telepresence --docker-run --rm centos:centos7   curl -i https://ec2.amazonaws.com` times out [2]
- `telepresence --docker-run --rm centos:centos7   curl -i http://ec2.amazonaws.com` times out [6]
[...]

PR #1371, opened 2020-05-31
Share CIDR routing to container method

isra17:
This workaround an issue in sshuttle on some host (See: https://github.com/telepresenceio/telepresence/issues/1220).

- A bit of refactoring was done to share the CIDR code between VPN and Container outbounds. 
- I also added a faster, more reliable way to get the serviceCIDR from the kube-apiserver command line parameters.
- I also added a filter on guessed pods CIDR to avoid collapsing into public network. While testing my PR, I had a cluster with pods in both `172.x.x.x` and `192.x.x.x` which end up collapsing in `128.0.0.0/1`


ark3:
Not stale! I’ll get to this next week.

ark3:
Your CIDR refactor is great! I wish you had posted that as a separate PR, or at least a separate commit. I would cherry-pick that onto master immediately; it's a pure win.

The core of this PR basically reverts #858 and makes the container method uses more likely to need things like `--also-proxy`. I'm flabbergasted that this solved #1220 for you, but I admit I _really_ don't understand that issue yet, and I'm frustrated that I have no way to replicate it myself.

isra17:
My theory is that #1220 is caused by the sshuttle component and a mix of the cloud and host environment... Luckily I was able to reproduce on my macbook, but not on my Desktop. I wasn't able to find the root cause, but it's probably one of those bugs that goes down to the network packet. that said, I can understand why the current behavior would be wanted. What do you think of (yet another) parameters like `--proxy-cidr` that would allow overriding `0.0.0.0`?

Otherwise I can work on splitting the PR along with #1373. I'm a bit busy right now, so maybe this will come in the next weeks.

ptemmer:
As to not revert #858. What about a parameter like `--never-proxy` or what have you, to exclude certain ips or domains to go through sshuttle, opposite to `--also-proxy`

isra17:
My issue with `--never-proxy` is that it isn't really the case, you still want to proxy the cluster cidrs to the remote cluster.

ptemmer:
Yes, that seems valid to me, it's just a different approach to the problem. While `--never-proxy amazonaws.con` would proxy everything, except the domain name specified, `--no-global-proxy` would only proxy the service CIDR and require you to explicitly specify any others. In my case, where I only have problems with amazonaws.com, it seems easier to just exclude that domain from being proxied, and leave all the rest as it is.

ark3:
This PR is stale now. The CIDR logic refactor (#1420) has landed and there is a separate PR (#1456) that addresses the original goal here. Closing this in favor of the other PR. Thank you!

aviau:
There are still lines of code in this PR that are needed in master! I'll open a new PR.

PR #1373, opened 2020-06-02
Split collapse

isra17:
Additional work on top of #1371 

This allow `covering_cidr` to return a list of `cidr` that doesn't cross global boundaries, sush that `["192.168.0.1", "192.168.1.1", "172.0.0.1"]` get collapsed into `["192.168.0.0/16", "172.0.0.0/24"]` instead of a single CIDR `128.0.0.0/1` (Actual case that happened to me in minikube).


ark3:
Shall we close this in favor of #1420? Thank you!

Issue #1041, opened 2019-06-07
Rationalize proxy pod management


ark3:
With #1404 landed, we are most of the way there. For release 0.106 we will keep a fallback path to the original deployment-based code (via the `TELEPRESENCE_USE_DEPLOYMENT` environment variable). Assuming we don't discover fundamental or unfixable issues with the new approach, we should remove the old code path for the next release and make these associated changes:

- [ ] Remove the legacy code path. Warn if the legacy environment variable is set.
- [ ] Migrate existing unit tests to test the new code path. This will require some refactoring of the new code to enable and improve testability. Note that the end-to-end tests already test the new code path.
- [ ] Update the documentation to say Pod instead of Deployment. This shouldn't make a practical difference for users but it's good to be accurate in our docs.

---

Original issue description follows

---

The control flow to set up the proxy pod is very confusing and makes too many `kubectl` calls. Tel supports roughly six methods using only three names:

1. Create a new Deployment
2. Create a new DeploymentConfig (not implemented)
3. Use an existing Deployment
4. Use an existing DeploymentConfig
5. Swap a Deployment
6. Swap a DeploymentConfig

The intent phase tries to figure out what the user really wants by making some read-only `kubectl` calls. It makes decisions about user intent and then throws away the results of those calls. Then the action phase rolls around and Tel has to figure out what's going on in the cluster by querying it again.

We should refactor this code into three steps.

1. Intent: Figure out what workflow the user wants (new, existing, or swap) and capture the deployment argument if specified. (simplified version of what's there now)
2. Intent stage 2: (new)
    - Existing: Query the cluster for the existing DeploymentConfig's manifest, falling back to querying for the existing Deployment's manifest. Check that the proxy version matches the CLI version. Report errors and fail if needed. Otherwise, pass along the queried manifest. 
    - Swap: Query the cluster for the existing DeploymentConfig's manifest, falling back to querying for the existing Deployment's manifest. Report errors and fail if needed. Otherwise, pass along the queried manifest to the appropriate swap code (deploy vs dc).
    - New: Compute the new Deployment's manifest. Use a DeploymentConfig if `oc` is in use. Pass along the generated manifest. Perform a (hopefully quick) check for cluster connectivity (`kubectl get pods telepresence-connectivity-check --ignore-not-found`), so it's possible to report errors and fail if needed. Maybe it would be okay to skip this and leave cluster connectivity checking to the action stage for this workflow.
 3. Action: (modified)
[...]

PR #1404, opened 2020-07-18
Proxy as a Pod

ark3:
Implements https://github.com/telepresenceio/telepresence/issues/1041#issuecomment-635537160

- [x] Manually check the error cases mentioned in the issue
- [x] More comments in the new code
- [x] Changelog entry

I'm going to put off these tasks for a future issue/PR
- [ ] Tests (and some refactoring for testability)
- [ ] Documentation updates

### DIFF
--- a/telepresence/proxy/operation.py
+++ b/telepresence/proxy/operation.py
@@ -1,0 +1,74 @@
+# Copyright 2020 Datawire. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Callable, Dict, NamedTuple, Optional, Tuple
+
+from telepresence.cli import PortMapping
+from telepresence.runner import Runner
+
+from .remote import RemoteInfo, get_remote_info
+
+ProxyIntent = NamedTuple(
+    "ProxyIntent", [
+        ("name", str),
+        ("container", str),
+        ("expose", PortMapping),
+        ("env", Dict[str, str]),
+        ("service_account", str),
+    ]
+)
+
+
+class ProxyOperation:
+    def __init__(self, intent: ProxyIntent) -> None:
+        self.intent = intent
+        self.remote_info = None  # type: Optional[RemoteInfo]
+
+    def prepare(self, runner: Runner) -> None:
+        pass
+
+    def act(self, _: Runner) -> RemoteInfo:
+        raise NotImplementedError()
+
+
+LegacyOperation = Callable[[Runner, str, PortMapping, Dict[str, str], str],
+                           Tuple[str, Optional[str]]]
+
+
+class Legacy(ProxyOperation):
+    def __init__(self, intent: ProxyIntent, legacy_op: LegacyOperation):
+        super().__init__(intent)
+        self.op = legacy_op
+
+    def act(self, runner: Runner) -> RemoteInfo:
+        deployment_arg = self.intent.name
+        if self.intent.container:
+            deployment_arg += ":" + self.intent.container
+
+        tel_deployment, run_id = self.op(
+            runner,
+            deployment_arg,
+            self.intent.expose,
+            self.intent.env,
+            self.intent.service_account,
+        )
+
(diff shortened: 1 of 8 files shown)


telepresence/proxy/operation.py
```diff
@@ -0,0 +1,312 @@
...
+
+    def act(self, _: Runner) -> RemoteInfo:
+        raise NotImplementedError()
+
+
+LegacyOperation = Callable[[Runner, str, PortMapping, Dict[str, str], str],
+                           Tuple[str, Optional[str]]]
+
+
+class Legacy(ProxyOperation):
```
marun, line 61: I'm a bit confused by the convention of generic class names like `Legacy` or `New` vs something more descriptive. Is this a py3 convention or precedent you could point me to?

telepresence/proxy/operation.py
```diff
@@ -0,0 +1,312 @@
...
+            self.intent.service_account,
+            self.intent.env,
+        )
+        self.manifests.append(pod)
+
+        # Construct a Service manifest as needed
+        if self.intent.expose.remote():
+            svc = make_svc_manifest(
+                self.intent.name,
+                dict(telepresence=runner.session_id),
```
marun, line 137: There doesn't appear to be a need to vary labels from selectors, maybe accept a map of labels and reuse it?

telepresence/proxy/operation.py
```diff
@@ -0,0 +1,312 @@
...
+                dict(telepresence=runner.session_id),
+                {p: p
```
marun, line 139: If there isn't a need to vary port from target port, maybe just pass the list?

telepresence/proxy/remote.py
```diff
@@ -28,27 +30,30 @@ class RemoteInfo(object):
...
     :ivar container_name str: The name of the container.
     """
-    def __init__(
-        self,
-        runner: Runner,
-        deployment_name: str,
-        pod_name: str,
-        deployment_config: dict,
-    ) -> None:
+    def __init__(self, pod_name: str, pod_spec: Manifest) -> None:
```
marun, line 33: Given what this function does, how likely is it that `pod_spec` is ever written again? And will the pod spec ever have more than the telepresence container? If not, would it make sense to capture tel_container as an instance variable and skip retaining `_container_index`?

Also, maybe just accept the pod rather than requiring the caller to pull out the name and spec themselves?

telepresence/proxy/operation.py
```diff
@@ -0,0 +1,312 @@
...
+from .manifest import (
+    Manifest, make_k8s_list, make_new_proxy_pod_manifest, make_pod_manifest,
+    make_svc_manifest
+)
+from .remote import (
+    RemoteInfo, get_deployment, get_pod_for_deployment, get_remote_info,
+    make_remote_info_from_pod, wait_for_pod
+)
+
+ProxyIntent = NamedTuple(
```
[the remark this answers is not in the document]
ark3: This is mostly only documented in few comments in `main` but basically all of the code is (supposed to be) organized as _intent_ (figure out what the user wants us to do, fast) and _action_ (do what we've decided we're going to do). This particular object records the user's intent for the proxy stage.

telepresence/proxy/operation.py
```diff
@@ -0,0 +1,312 @@
...
+) -> None:
+    """
+    Merge container ports into the expose list
+    """
+    containers = pod["spec"]["containers"]  # type: Iterable[Manifest]
+    for container in containers:
+        if not container_name or container["name"] == container_name:
+            expose.merge_automatic_ports([
+                port["containerPort"] for port in container.get("ports", [])
+                if port["protocol"] == "TCP"
```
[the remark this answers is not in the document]
ark3: Telepresence only supports TCP...? I copy-pasted this from the original code (in deployment.py) more or less.

telepresence/proxy/operation.py
```diff
@@ -0,0 +1,312 @@
...
+        )
+
+        # Perform the relevant swap changes to the pod spec
+        pod_spec = deployment["spec"]["template"]["spec"]
+        pod_spec["restartPolicy"] = "Never"
+        if self.intent.service_account:
+            pod_spec["serviceAccount"] = self.intent.service_account
+
+        # Find the relevant container
+        if self.intent.container:
```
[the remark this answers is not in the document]
ark3: Nope! Good catch. The legacy code just crashes, but at least with a useful error message. This would have crashed with a NameError. Anyhow, now we fail with a similar useful error message rather than crashing, at least in the new code path.

telepresence/proxy/operation.py
ark3: I used generic names because these classes are not used outside of this package. I don't see there being a chance of confusion, since I don't expose them by importing them in this package's `__init__.py`. That said, perhaps this is a bit of Golang-style naming leaking into my Python code...

marun:
LGTM

PR #1414, opened 2020-08-13
(docs) Preventing indexing of future deploy previews

hdoro:
As it stands, search engines are currently crawling our deploy previews from GitHub and indexing them, making the published telepresence.io website less valuable in their eyes due to the content duplication.

This PR is an attempt to block all future branch deploys and deploy previews from being indexed. For past deploys, I'm waiting on Netlify's support to see what we can do. I imagine **we might have to create a new site on Netlify for the docs, do the migration and delete the current one**. We'd lose our history of deploys but at least Telepresence.io would be reachable by more people on Google, Bing and others.


hdoro:
Okay, so the `robots.txt` file didn't work (see https://deploy-preview-1414--telepresence.netlify.app/robots.txt). This is probably due to the fact that the Netlify site is configured to have the root of the repository as the base for the site, and so it doesn't identify the `docs/netlify.toml` file I created (which tells Netlify to build previews differently).

I'll try to move it into the root of the repository, but I understand this is suboptimal for onboarding for new contributors. Let me know what you think.

richarddli:
Thanks! Netlify does also let us add a simple auth to previews -- maybe that would work better?

hdoro:
Hm... that's actually a great idea if possible, @richarddli ! I imagine it'd work backwards for all previews, solving the previously indexed content problem. Would this be OK for the coordination with contributors?

Also, I do think it'd still be a good idea to create the `robots.txt` file for next deploys, but currently I'm a bit lost about how the build command works. I just saw in Netlify's preview log that it usually uses `jekyll build` to create the docs, as opposed to `npm run build`, which is what I was envisioning. Do you think I could get access to the Netlify site to take a look at how the build is configured?

LukeShu:
Closing in favor of https://github.com/telepresenceio/telepresence/pull/1707

PR #1420, opened 2020-08-23
Refactor CIDR logic

isra17:
@ark3 

I did cut out the refactor as asked in https://github.com/telepresenceio/telepresence/pull/1371#issuecomment-658364838

I also included https://github.com/telepresenceio/telepresence/pull/1373 as part of the refactor.


telepresence/outbound/cidr.py
```diff
@@ -0,0 +1,292 @@
...
+from telepresence.runner import Runner
+from telepresence.utilities import random_name
+
+
+def covering_cidr(ips: List[str]) -> List[str]:
+    """
+    Given list of IPs, return CIDR that covers them all.
+
+    Presumes it's at least a /24.
+    """
```
ark3, line 31: Please update the function name and doc string.

telepresence/outbound/cidr.py
```diff
@@ -0,0 +1,292 @@
...
+
+    # Run script to convert --also-proxy hostnames to IPs, doing name
+    # resolution inside Kubernetes, so we get cloud-local IP addresses for
+    # cloud resources:
+    result = set(k8s_resolve(runner, remote_info, hosts_or_ips))
+    context_cache = runner.cache.child(runner.kubectl.context)
+    result.update(context_cache.lookup("podCIDRs", lambda: podCIDRs(runner)))
+    result.update(
+        context_cache.lookup("serviceCIDR", lambda: serviceCIDR(runner))
+    )
```
ark3, line 108: Please update the lookup key to "serviceCIDRs" with an "s" at the end.

If you don't do this, then every existing user will get weird crashes from `sshuttle` until the cache entry is cleared. The old cache stored a string, the new one stores a list, and the code doesn't check that it got back what it expected.

telepresence/outbound/cidr.py
isra17: Done

telepresence/outbound/cidr.py
isra17: Right! Done

isra17:
>Oh, could you also add newsfragments/1373.bugfix to this PR?

Done!

PR #1425, opened 2020-08-27
Fix broken also-proxy parameter handling

lightoze:
There's a straightforward bug with preexisting variable `name` being overwritten by the loop.


ark3:
Thank you!

PR #1456, opened 2020-10-29
Add no-global-proxy flag to stop proxying all traffic

jrdi:
Workaround added after reading #1371 

* Add no-global-proxy flag to avoid proxying all traffic. Otherwise, everything works as expected.


ark3:
Sorry we let this get stale. This is a valid solution to #1220 and friends.

aviau:
Can we merge this? This is a good feature to have in addition to better cidr guessing.

PR #1462, opened 2020-11-04
Update Docker tutorial, add hello-world manifest

### LINKED ISSUE #1409: Missing --expose in tutorial command
When following: 
https://www.telepresence.io/tutorials/docker

We come to the part where we do a `swap-deployment`:
```
$ telepresence --swap-deployment hello-world --docker-run --rm -it -v $(pwd):/usr/src/app hello-dev
T: Volumes are rooted at $TELEPRESENCE_ROOT. See https://telepresence.io/howto/volumes.html for details.
T: Starting network proxy to cluster by swapping out Deployment hello-world with a proxy
T: Forwarding remote port 8000 to local port 8000.

T: Setup complete. Launching your container.
 * Serving Flask app "server" (lazy loading)
[...]
```
The output of the command shows the port forward, however the command is missing the `--expose` to make that happen.

The command should be:
```
$ telepresence --swap-deployment hello-world --expose 8000 --docker-run --rm -it -v $(pwd):/usr/src/app hello-dev
```

mattmcclure-dw:
resolves #1409

- Replaced `kubectl create` and `kubectl expose` commands with a manifest that creates the deployment and service, with port info included in the deployment, the lack of which broke the tutorial.

- Made a few minor revisions across the doc

- After this is merged I'll go back and review the other tutorials, assuming they all need to be updated to use this manifest.

### DIFF
--- a/docs/tutorials/docker.md
+++ b/docs/tutorials/docker.md
@@ -1,8 +1,8 @@
 # Fast development workflow with Docker and Kubernetes
 
-Keeping development environments in sync is a constant pain. Containerizing your development environment enables your service to run in the exact same environment everywhere: from your laptop to production (for more details on the benefits of a container native development workflow, see [this post by Matt Butcher](https://open.microsoft.com/2018/04/23/5-reasons-you-should-be-doing-container-native-development/).)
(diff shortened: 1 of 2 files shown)


LukeShu (approved):
lgtm

LukeShu (changes requested):
Actually, you need to fix the DCO (I'm sending you more info on Slack).

LukeShu (approved):
Thanks!

PR #1473, opened 2020-11-18
fix(docs): cleans up Telepresence 2 announcement

bartoszmajsak:
This commit removes unneeded bullet points, as they are rendered double (because of `<li>`).

image

I'm super excited about this announcement! Is the code already open? I would love to contribute!


ark3:
Thank you! Stand by for more information soon.

PR #1476, opened 2020-11-20
Prevent iptables --list from hanging on reverse lookup

peterjanes:
[`iptables --list` will occasionally hang](https://serverfault.com/questions/286422/iptables-hangs-when-listing-rules) while performing a reverse lookup for IP addresses. As it's just a sanity check, the lookup isn't needed.


ark3:
Thank you for the PR!

PR #1498, opened 2021-01-06
Add Fedora 33 to packaging

tommi-gangverk:
The builds for Fedora 33, which has been out for a couple of months, are missing. Hopefully this pull request will nudge things forward. :smile:


ark3:
Thank you for the PR! It looks like this happened in #1494.

PR #1502, opened 2021-01-08
Only add docker args without env-file

hudelgado:
Hello and thanks for the great work on this tool.
I've created a workaround that could help starting telepresence in a cluster with several services, mainly it leverage the docker cli to use the environment variables from a file instead of from parameters.

It would be good if it could incorporated into telepresence to ease the start in clusters with lot of services.
Is it something that worth to be incorporated?

---

## Helpful information

    ```shell
    make lint check-local
    ```

    ```shell
    ```

### DIFF
--- a/docs/reference/proxying.md
+++ b/docs/reference/proxying.md
@@ -144,3 +144,14 @@ Currently unsupported:
 
 * Fully qualified Kubernetes DNS names that end with `.local`, e.g. `redis-master.default.svc.cluster.local`, won't work on Linux (see [the relevant ticket](https://github.com/datawire/telepresence/issues/161) for details.)
 * UDP messages in any direction.
+
+### Docker and Kubernetes in clusters with severall services
+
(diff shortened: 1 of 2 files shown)


ark3:
This is a good idea! My only concern is the lack of documentation and the associated surprise factor.

1. We need a bit of text explaining this exception in `docs/reference/proxying.md`
2. Then something like 
   ```python
        if runner.chatty:
            runner.show(
                "Not proxying cluster environment variables to your container "
                "because you are passing '--env-file' to 'docker run'. See "
                "https://www.telepresence.io/reference/proxying#environment-variables "
                "for more information"
            )
   ```
   would avoid the surprise when this case fires.

Thank you for the PR, and my apologies for taking so long to get back to you.

hudelgado:
Hi, thanks for the feedback.
I've applied your suggestions, happy to refactor things as you feel best

PR #1530, opened 2021-02-11
fixed type hinting for setup method's return type to be more specific

enkeyz:
Fixed type hinting for setup method's return type, from generic Callable, to be more specific.


LukeShu:
My apologies that one got to this before stalebot hid it from us.

Woo! Thanks!

enkeyz:
No problem!

PR #1551, opened 2021-03-08
[v2] Rename the Go module github.com/datawire/telepresence2/v2 => github.com/telepresenceio/telepresence/v2

LukeShu:


donnyyung (changes requested):
I think you need to run make generate (unless you intentionally didn't for some reason, which then my follow up learning question would be wondering why) since there are some conflicts in the generated files (mainly around import order since we went from `github.com/datawire...` to `github.com/telepresence...`

donnyyung (approved):

PR #1560, opened 2021-03-10
Handle port-less containers better

donnyyung:
## Description
When we are selecting which container the agent for an intercept should be installed on, if we see a container that has no containerPorts, we automatically select that container.  While there are use cases where we do indeed want to select that container, we should make sure there aren't any better fits in the container list by seeing if there's a container where the containerPort matches the targetPort. 

See the test yaml I added for an example of this problem. 

## Checklist


thallgren (approved):
LGTM

PR #1566, opened 2021-03-11
Don't auto-skip mapping of kube-* and ambassador namespaces

thallgren:


donnyyung (approved):
LGTM!

PR #1564, opened 2021-03-11
[v2] Miscellaneous dev quality of life improvements

LukeShu:
1. Drop a now-pointless `replace golang.org/x/sys` line from `go.mod`.  The commit message explains why it's not needed anymore.  Removing it lets me sleep easier.
2. Add a stub docs-build to shut up the failures from Netlify
3. Fix Thomas' complaint that he can't cross-compile by running `GOOS=linux make build` anymore.
4. Add a `./build/file-pr.go` script that will use the correct target branch, and the PR template from telepresence2.git.  The template part requires a counterpart PR to `master` to work (https://github.com/telepresenceio/telepresence/pull/1565).

PR #1567, opened 2021-03-11
.github: Add a separate PR template for telepresence v2

LukeShu:
This adds a separate PR template for the `release/v2` branch, which is selected by the `file-pr.go` script added in https://github.com/telepresenceio/telepresence/pull/1564, which is necessary because GitHub's ability to have multiple PR templates without extra tooling is surprisingly bad.  Why can't it use the template from the branch you're on, or the branch you're PR'ing in to, instead of always using the one from the default (`master`) branch?  Why can't it prompt you about the type of PR, like it does for issues when you have multiple issue templates?  Why can't it let you set different defaults for different branches, like it does for branch protection?


donnyyung (commented):
LGTM

PR #1571, opened 2021-03-12
Handle services that have ports declared without a TargetPort field

thallgren:
## Description

When a TargetPort field is missing, the target port defaults to the
value of the Port field. This commit ensures that this case is handled
correctly.

## Checklist

### DIFF
--- a/pkg/client/connector/install.go
+++ b/pkg/client/connector/install.go
@@ -643,10 +643,21 @@ func addAgentToDeployment(
 		Number   uint16
 		Protocol corev1.Protocol
 	}
+
 	// Start by filling from the servicePort; if these are the zero values, that's OK.
-	containerPort.Name = servicePort.TargetPort.StrVal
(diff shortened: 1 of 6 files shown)


donnyyung (approved):
LGTM and I especially like the simplified test yaml you used for your tests here!

pkg/client/connector/install.go
```diff
@@ -297,6 +297,10 @@ func findMatchingPort(dep *kates.Deployment, portName string, svcs []*kates.Serv
 	for _, svc := range svcs {
 		// For now, we only support intercepting one port on a given service.
 		ports := svcPortByName(svc, portName)
+		if len(ports) == 0 {
+			// this may happen when portName is specified but non of the ports match
```
LukeShu, line 301: typo: none of

LukeShu (approved):
Hit it!

PR #1574, opened 2021-03-12
[v2] Improve install_actions.go

LukeShu:
## Description

@thallgren Remember back in December when I had a PR that had 1 commit in it that made a whole bunch of hard-to-follow cleanup changes to `install_actions.go`, and we agreed I'd remove it from that PR?  This PR is all of those cleanup changes, but split out in to separate commits and hopefully justified in comments and commit messages.

I do suggest a commit-by-commit review.

## Checklist

 - [x] My change is adequately tested. - CI should cover that

### DIFF
--- a/pkg/client/connector/install_actions.go
+++ b/pkg/client/connector/install_actions.go
@@ -88,25 +88,17 @@ func (ma multiAction) explain(
 	out io.Writer,
 	ef func(partialAction partialAction, obj kates.Object, out io.Writer),
 ) {
-	last := len(ma) - 1
-	if last < 0 {
-		return
-	}
-
-	switch last {
-	case 0:
-	case 1:
-		ef(ma[0], obj, out)
-		fmt.Fprint(out, " and ")
-	default:
-		for _, partialAction := range ma[:last] {
-			ef(partialAction, obj, out)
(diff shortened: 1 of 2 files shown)


pkg/client/connector/install_actions.go
```diff
@@ -88,25 +88,17 @@ func (ma multiAction) explain(
...
-	default:
-		for _, partialAction := range ma[:last] {
-			ef(partialAction, obj, out)
-			fmt.Fprint(out, ", ")
+	for i, action := range ma {
+		switch i {
+		case 0:
+			// nothing
+		case len(ma) - 1:
+			_, _ = io.WriteString(out, ", and ")
```
thallgren, line 96: This removes my special rule of not inserting a comma when `len(ma) == 2`. Not sure what I had in mind there. The clauses are independent so the comma should definitely always be present.

thallgren (approved):
I like this a lot better than the last clean-up attempt. The only thing that I don't agree fully with is the capitalized method names. IMO, such names indicate possible package-external use, which is very different from implementing a package-internal interface. When I look at code, I think it's very helpful to know what's package internal and what isn't, and whether or not a function is implementing an interface is made very apparent by my IDE.

PR #1579, opened 2021-03-15
[v2] Adding service uid metadata to InterceptSpec

donnyyung:
## Checklist


pkg/client/connector/intercept.go
```diff
@@ -184,6 +184,18 @@ func (tm *trafficManager) addIntercept(c context.Context, ir *rpc.CreateIntercep
 		return result, nil
 	}
 
+	// Once we know the agent has been added, we get the service from the
+	// deployment's annotation and add its UID to the InterceptSpec.
+	svc, err := tm.getSvcFromDepAnnotation(c, spec.Namespace, spec.Agent)
```
thallgren, line 189: This means that we do two new queries over the wire, one to get the deployment and then one to get the service which in fact the same service that we've just used in `ensuredAgent()` (called from `addAgent()` just above). An alternative could be to have those two methods return that service (or the UUID of that service) in addition to the current `result`.

thallgren (changes requested):

pkg/client/connector/intercept.go
donnyyung: Ah yeah that was how I initially did it. But since ensureAgent just returned an error (if there was one), I thought it might be clunky to start returning the UUID as well, but you're right that it increases the # of calls and is getting the same dep + svc that it just got, so I'll revert back to doing it in `ensureAgent` and then adding the UUID to the InterceptResult in `addAgent`

thallgren (approved):
LGTM!

donnyyung:
(force push was bc I rebased to fix merge conflict with the changelog)

PR #1583, opened 2021-03-17
Ensure that connector only acts on mapped namespaces.

thallgren:
## Description

This changes the behavior when encountering an unmapped namespace for
the following commands:
- `list` returns an empty list
- `intercept` returns a NO_ACCEPTABLE_DEPLOYMENT error.
- `uninstall --agent` returns an "unable to find a deployment" error.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### 2.1.2 (TBD)
 - Bugfix: Uninstalling agents now only happens once per deployment instead of once per agent.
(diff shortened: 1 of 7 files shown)


donnyyung (commented):
Woops I got to this late, but all in all this LGTM. My only potential concern is that sometimes it is helpful to see all the errors at once instead of getting one and then failing fast, but I think you chose good places to use the require and we can always tweak it if there are places where it doesn't make sense to use require.  I've definitely done some test runs locally where it's clear that everything was going to fail because the port-forward didn't work for instance, so I think this is a robust change!

PR #1584, opened 2021-03-17
[v2] Fix bug where IPv6 routes broke all pfctl rule writes

donnyyung:
So I think this issue only affects us when people use externalNames where the lookupIP includes an IPv6 address.  This results in routes that have an IP that is IPv6 but our newRouteMessage only supported an IPv4 RTAX_GATEWAY and similarly our pfctl rules only worked for IPv4.  This adds support for both of those.

example of what some pfctl rules look like after they are written:
```
rdr pass on lo0 inet proto tcp from any to 10.43.223.167 port = 80 -> 127.0.0.1 port 62772
rdr pass on lo0 inet proto tcp from any to 10.43.223.167 port = 443 -> 127.0.0.1 port 62772
rdr pass on lo0 inet proto tcp from any to 10.43.233.56 port = 443 -> 127.0.0.1 port 62772
rdr pass on lo0 inet proto tcp from any to 10.43.244.24 port = 8080 -> 127.0.0.1 port 62772
rdr pass on lo0 inet proto tcp from any to 10.43.249.54 port = 2746 -> 127.0.0.1 port 62772
rdr pass on lo0 inet proto tcp from any to 172.217.11.46 -> 127.0.0.1 port 62772
rdr pass on lo0 inet6 proto tcp from any to 2607:f8b0:4006:803::200e -> ::1 port 62772
``` 

## Checklist

 - [x] My change is adequately tested. - I tested it locally but since the firewall is going away and likely the need for this patch very soon, I didn't try to write more exhaustive tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### 2.1.2 (TBD)
 - Bugfix: Uninstalling agents now only happens once per deployment instead of once per agent.
 - Bugfix: The list command no longer shows agents from namespaces that aren't mapped.
+- Bugfix: IPv6 routes now work + don't prevent other pfctl rules being written in MacOS
 - Change: Service UID was added to InterceptSpec to better link intercepts and services.
 
 ### 2.1.1 (March 12, 2021)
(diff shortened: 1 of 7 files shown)


thallgren (approved):
This is great. We should have an automated test that verifies that intercepts work when using ipv6.
We need to get a release out nowish!

PR #1587, opened 2021-03-18
[v2] Add ability to configure timeouts (and future other things)

thallgren:
## Description

Adds a generic configuration ability which reads a `config.yml` file
stored under the users config directory. Files under the system config
directories will also be considered on platforms that supports this.

The config contains only timeouts at present but it's expected to grow.

## Checklist

### DIFF
--- a/pkg/client/connector/install.go
+++ b/pkg/client/connector/install.go
@@ -532,7 +532,8 @@ func (ki *installer) ensureAgent(c context.Context, namespace, name, portName, a
 }
 
 func (ki *installer) waitForApply(c context.Context, namespace, name string, dep *kates.Deployment) error {
-	c, cancel := context.WithTimeout(c, 1*time.Minute)
+	tos := &client.GetConfig(c).Timeouts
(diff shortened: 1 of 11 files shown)


pkg/client/connector/install.go
```diff
@@ -532,7 +532,7 @@ func (ki *installer) ensureAgent(c context.Context, namespace, name, portName, a
 }
 
 func (ki *installer) waitForApply(c context.Context, namespace, name string, dep *kates.Deployment) error {
-	c, cancel := context.WithTimeout(c, 1*time.Minute)
+	c, cancel := context.WithTimeout(c, client.GetConfig(c).Timeouts.Apply)
```
donnyyung, line 535: Small suggestion, could we include the config timeout ussed in the timeout error message so if people see failures, they know which timeout to tweak?

donnyyung (approved):
LGTM and even on such short notice. Thanks for taking my feedback about alerting users to which timeout was the one that timed out :) 

I think this will be a great place for us to start for our config values.

PR #1589, opened 2021-03-19
[v2] Improve config.yml parser and timeout handling.

thallgren:
## Description

This PR contains some improvements to the config parser and how timeouts are handled.

## Checklist

### DIFF
--- a/pkg/client/config.go
+++ b/pkg/client/config.go
@@ -178,6 +178,9 @@ func loadConfig(c context.Context) (*Config, error) {
 	cfg := defaultConfig // start with a by value copy of the default
 
 	readMerge := func(dir string) error {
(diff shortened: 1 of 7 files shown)


pkg/client/config.go
```diff
@@ -79,45 +119,57 @@ func CheckTimeout(c context.Context, which *time.Duration, err error) error {
 }
```
donnyyung, line 119: I can't comment on it for some reason, but the error message above should be "can be configured as timeouts.%s" right?

thallgren: Fixed!

donnyyung (approved):
LGTM and tested setting the timeouts on my machine which now works 🎉

PR #1592, opened 2021-03-21
[v2] Defer startup of k8s watchers

thallgren:
## Description

Starting the k8s watchers in parallel with the kubectl port-forward
creates intermittent problems on some platforms (most notably
Minikube with Hyperkit). Deferring until the traffic-manager is
fully connected lessens the concurrency at connect time and that
makes the intermittent problem go away.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+### 2.1.3 (TBD)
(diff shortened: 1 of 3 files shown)


donnyyung (approved):

donnyyung:
Good catch + thanks for digging into this, I know it was a weird one.

PR #1598, opened 2021-03-23
[v2] Properly retry on hiccups on the SOCKS server

LukeShu:
## Description

@esmet reports that he sees AES CI flakes where a hiccup in the connection to the cluster causes the whole connector to quit.  It should have retries!  Well, it turns out that while most of the things had retries, the `/connector/server-socks` goroutine didn't retry, and so it'd error out, and cause all the rest of it to shut down.

As usual, I suggest a commit-by-commit review:

- The first 3 commits are doing prep-work for the fix, with an eye for auditing uses of `tm.sshPort` to make sure that all users of it properly retry.
- The 4th commit is the actual fix.
- The final 2 commits are generic cleanup tasks that are now-trivial and now-obvious because of the changes made above.

## Checklist

 - [ ] My change is adequately tested. - no


thallgren (approved):
LGTM!

PR #1599, opened 2021-03-24
[v2] Upgrade dlib, consistently use dlib/dexec

LukeShu:
## Description

We got a report of the https://github.com/datawire/dlib/pull/14 crash happening in a real telepresence install on CentOS 7, so pull in the fix from dlib.

And while we're at it, now that the bug is fixed, switch (most) everything over to dexec instead of os/exec.  I left `run.go` alone though, because it's pretty hairy.

## Checklist

 - [x] My change is adequately tested. - CI should do it

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -27,8 +27,10 @@ issues:
 linters-settings:
   depguard:
     list-type: blacklist
+    include-go-root: true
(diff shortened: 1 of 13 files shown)


thallgren (approved):
Yes!

PR #1600, opened 2021-03-25
[v2] Adding support for ReplicaSets

donnyyung:
Here's my quite big PR about support ReplicaSets.  This got a little hairy bc we assumed Deployments in a lot of places and I didn't want to add copies of all those functions for each new Workload (ReplicaSet, StatefulSets), that we add in the future.  So this ended up being a substantial refactor.

Notes from the author:
I encourage a commit-by-commit review, I will leave comments on a few commits where I urge you to actually just look at the final version bc I either: 
A. Changed things substantially 
B. cleaned it up a lot and I'd rather you comment on the clean version instead of commenting on stuff I may have fixed in later commits.
C. The rebase on top of release/v2 really made that commit gross

If this makes it more confusing to review, let me know and I'll squash everything into one commit and then try to break that commit into logical commits afterwards.
This will be followed up by a PR adding StatefulSets soon, so I'll be taking another fresh look + catching any mentions of "deployment" I find where it doesn't make sense, but let me know if you find any while looking at this PR. 

Lastly, there are a lot of name changes to be inclusive of things aside from Deployments.. I tried to name things what made sense according to k8s documents, but I'm not married to any of the names I chose, so if you hate "workloads", feel free to recommend something else.   

## Checklist

 - [x] I made sure to either submit a docs PR, or tell Matt about the necessary documentation changes. - PR is up
 - [x] My change is adequately tested. - I have added some tests, I'd like to add more (and will either on top of this PR or when we add stateful sets) 


pkg/client/connector/install.go
```diff
@@ -273,6 +273,31 @@ func (ki *installer) findMatchingServices(portName string, dep *kates.Deployment
 	return matching
 }
 
+func (ki *installer) findMatchingServicesByLabels(portName string, labels map[string]string) []*kates.Service {
```
donnyyung: This goes away in future commits, decided to just meld it into findMatchingServices, so you can ignore it for now

pkg/client/connector/traffic_manager.go
```diff
@@ -85,6 +85,18 @@ func newTrafficManager(_ context.Context, env client.Env, cluster *k8sCluster, i
 	return tm, nil
 }
 
+type uniqueWorkload struct {
```
donnyyung: This struct + associated function go away in future commits

pkg/client/connector/traffic_manager.go
```diff
@@ -85,6 +85,18 @@ func newTrafficManager(_ context.Context, env client.Env, cluster *k8sCluster, i
 	return tm, nil
 }
```
donnyyung: I'd just look at the final version of this file, since I did some cleaning in the final commit

pkg/client/connector/install.go
```diff
@@ -147,14 +147,33 @@ func (ki *installer) removeManagerAndAgents(c context.Context, agentsOnly bool,
 		ai := ai // pin it
```
donnyyung, line 147: similar to the traffic_manager, I'd look through this but then look at the final, cleaned up version before leaving comments, since I likely addressed some of the more obvious comments you had in the cleanup.

pkg/client/connector/install.go
```diff
@@ -372,202 +403,290 @@ func findMatchingPort(dep *kates.Deployment, portName string, svcs []*kates.Serv
...
-		if err != nil {
-			return "", err
-		}
-		fallthrough
 	case agentContainer == nil:
 		dlog.Infof(c, "no agent found for deployment %s.%s", name, namespace)
 		dlog.Infof(c, "Using port name %q", portName)
-		matchingSvcs := ki.findMatchingServices(portName, dep)
+		matchingSvcs := ki.findMatchingServices(portName, svcName, podTemplate.Labels)
 		if len(matchingSvcs) == 0 {
```
donnyyung, line 522: Does it make sense here to, similar to when a service has more than one port we tell the user they have to specify a port, to tell users if they have more than one service matching their workload, they also need to specify the service?  I'm not sure why we ever wanted to proceed when we match more than one service to begin with.

donnyyung:
sigh apparently an optimization I just added broke the tests, fixing asap but marking as wip until they are fixed.

pkg/client/cli/telepresence_test.go
```diff
@@ -370,15 +370,15 @@ func (cs *connectedSuite) TestH_ListOnlyMapped() {
 
 	stdout, stderr = telepresence(cs.T(), "list", "--namespace", cs.ns())
 	require.Empty(stderr)
-	require.Contains(stdout, "No deployments")
+	require.Contains(stdout, "No Workloads (Deployments or ReplicaSets")
```
thallgren, line 393:
```suggestion
	require.Contains(stdout, "No Workloads (Deployments or ReplicaSets)")
```

pkg/client/cli/telepresence_test.go
```diff
@@ -370,15 +370,15 @@ func (cs *connectedSuite) TestH_ListOnlyMapped() {
...
 
 	stdout, stderr = telepresence(cs.T(), "connect", "--mapped-namespaces", "all")
 	require.Empty(stderr)
 	require.Empty(stdout)
 
 	stdout, stderr = telepresence(cs.T(), "list", "--namespace", cs.ns())
 	require.Empty(stderr)
-	require.NotContains(stdout, "No deployments")
+	require.NotContains(stdout, "No Workloads (Deployments or ReplicaSets")
```
thallgren, line 401:
```suggestion
	require.NotContains(stdout, "No Workloads (Deployments or ReplicaSets)")
```

pkg/client/connector/traffic_manager.go
```diff
@@ -197,6 +209,89 @@ func (tm *trafficManager) session() *manager.SessionInfo {
 	return tm.sessionInfo
 }
 
+func (tm *trafficManager) getInfosForWorkflow(ctx context.Context, names []string, objectKind, namespace string, iMap map[string]*manager.InterceptInfo, aMap map[string]*manager.AgentInfo, filter rpc.ListRequest_Filter) []*rpc.WorkloadInfo {
```
thallgren: Documentation of this method and it's parameters would make the code easier to read. Also, please split this very long line into several lines.

pkg/client/connector/install.go
```diff
@@ -455,30 +455,30 @@ var agentNotFound = errors.New("no such agent")
 // is installed alongside the proper deployment.  In doing that, it also ensures that
```
thallgren: This comment needs an update. It's no longer "deployment" and now the function also returns the object's kind.

pkg/client/connector/install.go
```diff
@@ -524,21 +524,21 @@ func (ki *installer) ensureAgent(c context.Context, namespace, name, portName, a
...
+			return "", "", errors.New(errMsg)
 		}
 		var err error
 		obj, svc, err = addAgentToDeployment(c, portName, agentImageName, obj, matchingSvcs)
 		if err != nil {
-			return "", err
+			return "", "", err
 		}
 	case agentContainer.Image != agentImageName:
 		var actions deploymentActions
```
thallgren, line 534: deploymentActions => workloadActions ?

pkg/client/connector/install.go
```diff
@@ -489,36 +498,16 @@ func (ki *installer) ensureAgent(c context.Context, namespace, name, portName, a
 		}
 	}
 
-	svcPortChanged, err := portChanged(c, obj, portName)
-	if err != nil {
-		return "", "", err
+	if err := checkSvcSame(c, obj, svcName, portName); err != nil {
+		return "", "", errors.Wrap(err, "Workload already being used for intercept. To intercept this with your new configuration, please use telepresence uninstall --agent <nameOfWorkload> first. This will cancel any intercepts that already exist")
```
thallgren: Please split into concatenated strings to make the line shorter.

pkg/client/connector/install.go
thallgren: Yes, I think it sounds reasonable to report ambiguities. Especially if the user can remove them by providing more info.

pkg/client/connector/install.go
```diff
@@ -372,202 +402,290 @@ func findMatchingPort(dep *kates.Deployment, portName string, svcs []*kates.Serv
...
-		}
-		// Since the agent has been removed, we find the updated version of
-		// the deployment and fallthrough to the no agent case
-		dep, err = ki.findDeployment(c, namespace, name)
-		if err != nil {
-			return "", err
-		}
-		fallthrough
 	case agentContainer == nil:
 		dlog.Infof(c, "no agent found for deployment %s.%s", name, namespace)
```
thallgren, line 518:
```suggestion
		dlog.Infof(c, "no agent found for %s %s.%s", kind, name, namespace)
```

thallgren (approved):
Looks like a very useful addition. I had some minor nit-picks that you can fix if you want.

PR #1606, opened 2021-03-27
[v2] Don't hardcode port 1080 for dynamic ssh port forwards

thallgren:
## Description

Ensure that the port used for dynamic port-forwards is dynamically
allocated by the connector when the port-forward is set up, and then
passed on to the daemon proxy.

## Checklist


pkg/client/connector/service.go
```diff
@@ -504,9 +505,16 @@ func run(c context.Context) error {
 		if !ok {
 			return nil
 		}
-		tm.sshPortForward(c,
-			"-D", "localhost:1080",
-		)
+		ln, err := net.Listen("tcp4", "127.0.0.1:0")
```
donnyyung, line 508: I know it's in the function, but maybe a brief comment saying we are using this to randomly select an available port, just so people know why in the future.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ### 2.1.4 (TBD)
 
 - Bugfix: Fix race condition that occurred when intercepting a ReplicaSet while another pod was terminating in the same namespace (this fixes a transient test failure)
+- Bugfix: Telepresence no longer require that port 1080 must be available.
```
donnyyung, line 6: I would say this is a change, since requiring 1080 wasn't really a _bug_ per say

LukeShu: +1 to what Donny said

pkg/client/connector/service.go
```diff
@@ -504,9 +505,16 @@ func run(c context.Context) error {
...
+		if err != nil {
+			return err
+		}
+		dynamicAddr := ln.Addr().String()
+		_ = ln.Close()
+		_, portStr, _ := net.SplitHostPort(dynamicAddr)
+		dynamicPort, _ := strconv.Atoi(portStr)
+		tm.dynamicSshPort = int32(dynamicPort)
+		tm.sshPortForward(c, "-D", dynamicAddr)
```
LukeShu, line 517: Any reason not to use `-D localhost:0` and parse the output like we do with the `kubectl port-forward`s?

pkg/client/connector/service.go
LukeShu, line 508: Is there a specific reason why you used `tcp4`/`127.0.0.1` instead of maybe-IPv6 `tcp`/`localhost`?

pkg/client/daemon/proxy/proxy.go
```diff
@@ -19,9 +20,10 @@ import (
 
 // A Proxy listens to a port and forwards incoming connections to a router
 type Proxy struct {
-	listener    net.Listener
-	connHandler func(*Proxy, context.Context, *net.TCPConn)
-	router      func(*net.TCPConn) (string, error)
+	dynamicSshPort int32
```
LukeShu, line 23: Likewise about naming it `socksPort`

rpc/daemon/iptables.proto
```diff
@@ -13,4 +13,5 @@ message Route {
 message Table {
   string name = 1;
   repeated Route routes = 2;
+  int32 dynamic_ssh_port = 3;
```
LukeShu, line 16: likewise again

LukeShu (changes requested):
Changes requested:
- CHANGELOG entry
- Naming

Questions:
- Why IPv4?

pkg/client/connector/service.go
thallgren: I think using `localhost` consumes more CPU-cycles since that name must be resolved somehow. Do you see that doing so would bring any advantages?

pkg/client/connector/service.go
thallgren: That behaviour is only for `-R`. Trying it with `-D` results in `"Bad dynamic forwarding specification 'localhost:0'"` (which is consistent with the [man page](https://man.openbsd.org/ssh).

LukeShu (approved):
Thanks!

PR #1611, opened 2021-03-29
[v2] Fix ReplicaSet intercept race condition

donnyyung:
When refreshing the ReplicaSet pods, it was looking at all pods in the
namespace so if we had a pod that was terminating, the ki.findPod call
would error when it got to that pod. This has been fixed by only calling
findPod on pods that have the ReplicaSet name in their name.  This was
causing
TestTelepresence/TestB_Connected/TestG_SuccessfullyInterceptsReplicaSet
to be a little flakey.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+### 2.1.4 (TBD)
+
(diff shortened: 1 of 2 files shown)


thallgren (approved):
The fix makes sense. Perhaps we should also consider making `kindNames()` exclude terminating resources? When would we even want them?

donnyyung:
Good suggestion Thomas, I'll see if that's possible in my StatefulSet work!

PR #1629, opened 2021-04-01
[v2] Pt 3: Move login to happen in the connector, use refresh tokens

LukeShu:
## Description

- The first 8 commits are moving login things to happen in the connector instead of an ephemeral server in the CLI.  This requires implementing a way of streaming output from the connector to the CLI, which is something we've talked about before, and will allow us to address several other UX concerns.
- The next 11 commits are implementing refresh tokens.  This work requires the "move login things to happen in the connector" because we really do want a background keepalive process.  This work is important for API keys because the change in how we auth to Ambassador Cloud changes some of the failure conditions, so having better knowledge the the access token lets us make those go away.

I'm saddened that this is adding more than 1000 lines of code, but
1. broadcastqueue can be merged with the connector's watchable and go away... once Go has generics.  Slated for Go 1.18 in February 2022.  That's so far away, but will get here :/
2. A decent chunk of functionality is duplicated in `pkg/client/cli` and `pkg/client/cli/cliutil`.  The stuff in `cli` can go away once everything migrates to `cliutil`.
  
## Checklist

 - [x] My change is adequately tested. - yes


pkg/client/cli/cliutil/systema.go
```diff
@@ -3,33 +3,57 @@ package cliutil
...
 func Logout(ctx context.Context) error {
-	return auth.Logout(ctx)
+	err := WithConnector(ctx, func(ctx context.Context, connectorClient connector.ConnectorClient) error {
+		_, err := connectorClient.Logout(ctx, &empty.Empty{})
+		return err
+	})
+	if grpcStatus.Code(err) == grpcCodes.NotFound {
+		err = errors.New(grpcStatus.Convert(err).Message())
+	}
+	if err != nil {
```
donnyyung, line 37: it would seem we are effectively returning `err` in either case bc we return `err` if err != nil and `nil` if err == nil.  Is this done to make it more legible and clear that we are explicitly returning an err? Or is it something we can optimize?

pkg/client/cli/cliutil/systema.go
```diff
@@ -3,33 +3,57 @@ package cliutil
...
-	if err == auth.ErrNotLoggedIn {
+	err := WithConnector(ctx, func(ctx context.Context, connectorClient connector.ConnectorClient) error {
+		_, err := connectorClient.Logout(ctx, &empty.Empty{})
+		return err
+	})
+	if grpcStatus.Code(err) == grpcCodes.NotFound {
 		err = nil
 	}
-	return err
+	if err != nil {
```
donnyyung, line 53: Same Q as above.

pkg/client/connector/auth/login.go
```diff
@@ -84,10 +89,12 @@ func NewLoginExecutor(
 		stdout:           stdout,
 		scout:            scout,
 
-		callbacks: make(chan oauth2Callback),
+		callbacks:    make(chan oauth2Callback),
+		refreshTimer: time.NewTimer(1 * time.Minute),
 	}
 	ret.oauth2ConfigMu.Lock()
 	ret.loginMu.Lock()
+	ret.resetRefreshTimer(0)
```
donnyyung, line 97: Does this do anything? In the function we only do something if it's > 0 so this seems like a no-op whenever we call this with 0?

pkg/client/cli/cliutil/systema.go
LukeShu: Yes, it's done for style reasons.  Part of that is legibility.  Part of that is that there are two cases, so having it this way lets coverage testing tell us if we're testing both cases (not that we're doing coverage testing on telepresence right now).  Part of it is paranoia about "nil isn't always nil".

pkg/client/connector/auth/login.go
LukeShu: It stops the timer when <= 0.  AFAICT there's not a way to create a timer in a stopped state, so I create it started with 1 minute to go, and then immediately stop it.

I should add comments about this.

donnyyung (approved):
I think this LGTM. There's obviously a lot here and most of my comments were questions (that you clarified so thank you!).  In general, the code is well-commented which made it easier to follow.  In an ideal world, it'd be nice to have Thomas take a look since he's going to take over some work related to this, but I think it's more important to have this in the next release since it unblocks a lot of work.

PR #1633, opened 2021-04-02
[v2] Add support for StatefulSets 

donnyyung:
This will enable users to intercept StatefulSet services, and largely builds on top of the ReplicaSet PR I merged last week, I've also cleaned up some functions + names that I introduced in that PR that weren't the best.  Would recommend a commit-by-commit review. 

## Checklist

 - [x] I made sure to either submit a docs PR, or tell Matt about the necessary documentation changes. -- will do this in the morning, should just be rbac changes in the docs

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -147,8 +147,8 @@ github.com/cpuguy83/go-md2man/v2 v2.0.0-20190314233015-f79a8a8ca69d/go.mod h1:ma
 github.com/cpuguy83/go-md2man/v2 v2.0.0/go.mod h1:maD7wRr/U5Z6m/iR4s+kqSMx2CaBsrgA7czyZG/E6dU=
 github.com/creack/pty v1.1.7/go.mod h1:lj5s0c3V2DBrqTV7llrYr5NG6My20zk30Fl46Y7DoTY=
 github.com/cyphar/filepath-securejoin v0.2.2/go.mod h1:FpkQEhXnPnOthhzymB7CGsFk2G9VLXONKD9G7QGMM+4=
-github.com/datawire/ambassador v1.12.2-0.20210325164646-9d653a4f2cfa h1:sd5jtI+QMQQRghLh2zsEMTnp+iqnSc3jcbjkkh6hZ0U=
-github.com/datawire/ambassador v1.12.2-0.20210325164646-9d653a4f2cfa/go.mod h1:2grBLdYgILzrgTpenDMB5OeyhObIUaT+KwkLkZI1KDE=
+github.com/datawire/ambassador v1.12.2-0.20210329195933-976e68296977 h1:w5AavQPFbXbLSIBn2OLkbI65aVi44PJsVHRvUWfDxa0=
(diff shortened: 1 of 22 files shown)


pkg/client/connector/install_test.go
```diff
@@ -509,21 +509,150 @@ func TestAddAgentToReplicaSet(t *testing.T) {
...
+			sanitizeWorkload(actualDep)
 
 			assert.Equal(t, expectedDep, actualDep)
 			assert.Equal(t, expectedSvc, actualSvc)
 		})
 	}
 }
 
-func sanitizeDeployment(obj kates.Object) {
+// I (Donny) would like to unify this w/ the "TestAddAgentToWorkload
```
donnyyung, line 526: I know this is gross, the thing tripping me up on this is the DeepCopy doesn't exist for kates.Objects, so I can't replace kates.StatefulSet/ReplicaSet/Deployment with kates.Object.  I do want to fix these three tests to be less copy paste and remove redundancies, but I didn't think it should hold up getting the feature out.

LukeShu: `kates.Object` implements `proto.Message`, right? You should be able to use `proto.Clone(obj)` on it for a slightly-less-efficient `DeepCopy`.

go.sum
```diff
@@ -147,8 +147,8 @@ github.com/cpuguy83/go-md2man/v2 v2.0.0-20190314233015-f79a8a8ca69d/go.mod h1:ma
 github.com/cpuguy83/go-md2man/v2 v2.0.0/go.mod h1:maD7wRr/U5Z6m/iR4s+kqSMx2CaBsrgA7czyZG/E6dU=
 github.com/creack/pty v1.1.7/go.mod h1:lj5s0c3V2DBrqTV7llrYr5NG6My20zk30Fl46Y7DoTY=
 github.com/cyphar/filepath-securejoin v0.2.2/go.mod h1:FpkQEhXnPnOthhzymB7CGsFk2G9VLXONKD9G7QGMM+4=
-github.com/datawire/ambassador v1.12.2-0.20210325164646-9d653a4f2cfa h1:sd5jtI+QMQQRghLh2zsEMTnp+iqnSc3jcbjkkh6hZ0U=
-github.com/datawire/ambassador v1.12.2-0.20210325164646-9d653a4f2cfa/go.mod h1:2grBLdYgILzrgTpenDMB5OeyhObIUaT+KwkLkZI1KDE=
+github.com/datawire/ambassador v1.12.3-0.20210401195424-3d91930ec3fd h1:yt9n2dnenlvOeTyOHuMKQRrFjFJNlMAURVrCiTUJjCM=
+github.com/datawire/ambassador v1.12.3-0.20210401195424-3d91930ec3fd/go.mod h1:2grBLdYgILzrgTpenDMB5OeyhObIUaT+KwkLkZI1KDE=
```
LukeShu, line 151: Why? I'm not objecting, but I don't see anything else in the PR that looks like it would have prompted this.

LukeShu (approved):
<3 tests

Yeah, there's a lot of code duplication, but we can clean that up whenever. (Looking at this code, I'm wishing Go had C#'s thing where you can attach methods to an existing type.)

go.sum
donnyyung: Ah I had to add the StatefulSet alias: https://github.com/datawire/ambassador/pull/3323

Issue #1625, opened 2021-03-31
[v2] 2.1.3 found multiple services with a selector matching labels


scottillogical:
After upgrading from 2.12 to 2.1.3 I get this error a lot on intercept 
```
telepresence --context CONTEXT -n sandbox-scott intercept dev-tools --port 8081:http
 Found multiple services with a selector matching labels map[app:dev-tools cheetah:true]: dev-tools,dev-tools,dev-tools and a port named http
```

khussey:
This is fixed in the 2.1.4 release, which is now available.

PR #1634, opened 2021-04-02
[v2] Make FindMatchingServices namespace-bound

donnyyung:
This fixes the bug mentioned in https://github.com/telepresenceio/telepresence/issues/1625 . Pretty self-explanatory fix, but findMatchingServices should have _only_ been looking at the namespace we were trying to intercept in instead of looking at all namespaces (which caused collisions with services that shouldn't have been collisions). 

## Checklist


LukeShu (approved):
LGTM!

PR #1640, opened 2021-04-05
[v2] OpenRotatingLogFile: Fix the os.MkdirAll call

LukeShu:
## Description

It was calling os.MkdirAll with the filepath.Dir of `fileName`, which is
*already* the basename, so the filepath.Dir of it was *always* just ".".

So rename the variables to hopefully make this clearer:

- fullPath → logfilePath
- dirName  → logfileDir
- fileName → logfileBase

## Checklist

 - [ ] I made sure to update `./CHANGELOG.md`.
 - [ ] I made sure to either submit a docs PR, or tell Matt about the necessary documentation changes.
 - [ ] My change is adequately tested.
 - [ ] I updated `DEVELOPING.md` with any any special dev tricks I had to use to work on this code efficiently.

PR #1641, opened 2021-04-05
DEVELOPING.md: Add a note about debugging early setup

LukeShu:
## Description

Add some notes to DEVELOPING.md about the tricks I used to debug #1640.

## Checklist

 - [x] I updated `DEVELOPING.md` with any any special dev tricks I had to use to work on this code efficiently. - that's all this is


donnyyung (approved):
Thanks for this, I've seen mentions of it in the code, but always nice to have it explicit in a doc!

PR #1644, opened 2021-04-06
[v2] Support intercepting by port name and number

donnyyung:
We already had support for doing `telepresence intercept <svc> --port <localPort>:<portName>` and this change now broadens that command so you can do something like `telepresence intercept <svc> --port <localPort>:<svcPort>` which is something that people could do in tel1.

## Checklist


rpc/manager/manager.proto
```diff
@@ -88,8 +88,8 @@ message InterceptSpec {
   string target_host = 6;
   int32 target_port = 7;
 
-  // Name of ther service port that is being intercepted
-  string service_port_name = 10;
+  // Identifier for the service port: either the name or port number
+  string service_port_identifier = 10;
```
donnyyung, line 92: I thought about making this servicePortNameOrNumber to match how it's referenced later in the connector code, but thought that might be a really long rpc name, but if we don't think that's a concern, I can change this so it's the same throughout.

pkg/client/connector/install.go
```diff
@@ -262,19 +262,31 @@ func (ki *installer) updateDeployment(c context.Context, env client.Env, current
...
 		port := &ports[i]
-		if name == "" || name == port.Name {
+		matchFound := false
+		// If no nameOrNumber has been specified, we include it
+		if nameOrNumber == "" {
+			matchFound = true
+		}
+		// If a nameOrNumber has been specified, we check if it matches the Name
+		// or Port number, and add it if either case is true
+		if nameOrNumber == port.Name || nameOrNumber == strconv.Itoa(int(port.Port)) {
```
LukeShu, line 279: Part of me feels weird about using `strconv.Itoa` instead of going the other direction.  What if the user wants to pass the number on the CLI with a leading 0?

And the other part of me thinks that I'm being ridiculous.

You could use `k8s.io/apimachinery/pkg/util/validation.IsValidPortName()` and `strconv.Atoi()` to see which one the string is before you even look at the Service.

LukeShu (approved):
Woo!

PR #1653, opened 2021-04-09
[v2] Replace firewall-routers with a TUN-device

thallgren:
## Description

This PR contains a replacement of the current firewall base routers in the form of a TUN device. The current shape of the PR is intended for review and testing, not production. The following needs to happen before it is production ready

1. The implementation must be thoroughly tested, especially with regards to TCP since I've written a complete workflow for negotiation.
2. Some aspects of MTU aren't covered just yet. It can be made more efficient.
3. IPv6 is not fully supported.
4. There's an abundance of logging going on. That must be minimized because it's only useful when debugging protocol issues.

## Checklist


pkg/subnet/byteset.go
```diff
@@ -0,0 +1,101 @@
+package subnet
```
donnyyung, line 1: Perhaps a nit but could you change the name of `*byteset*.go` files to be `*byte_set*.go` just so it's easier to read?

pkg/subnet/byteset.go
```diff
@@ -0,0 +1,101 @@
...
+// Equals returns true if this ByteSet equals the argument
+func (b *ByteSet) Equals(other *ByteSet) bool {
+	if other == nil {
+		return false
+	}
+	return *b == *other
+}
+
+// ToSlice returns the number of bytes in this ByteSet
+func (b *ByteSet) Len() (l int) {
```
donnyyung, line 32: I had no idea you could have named return variables that are automatically created and returned in functions in Go, thinking of numerous places I could have done this 🤦

pkg/subnet/byte_set.go
```diff
@@ -0,0 +1,101 @@
+package subnet
+
+import "fmt"
+
+// ByteSet represents 0 - 255 unique bytes.
+type ByteSet [4]uint64
```
LukeShu, line 6: This comment and that type did not give me much of any idea what this is/does.

Perhaps:

```go
// Bitfield256 is a 256-bit bitfield.
type Bitfield256 [4]int64
```

pkg/subnet/byte_set.go
```diff
@@ -0,0 +1,101 @@
...
+
+// Add adds a byte to this ByteSet
+func (b *ByteSet) Add(bv byte) {
```
LukeShu, line 9:
```go
// SetBit sets the 1<<bv bit of the bitfield to 1.
//
// Compare: math/big.Int.SetBit
func (b *Bitfield256) SetBit(bv byte) {
```

pkg/subnet/byte_set.go
```diff
@@ -0,0 +1,101 @@
...
+	b[bv>>6] |= uint64(1) << uint64(bv&0x3f)
+}
+
+// Remove removes a byte from this ByteSet
+func (b *ByteSet) Remove(bv byte) {
```
LukeShu, line 14:
```go
// ClearBit clears the 1<<bv bit of the bitfield to 0.
//
// Compare: math/big.Int.SetBit
func (b *Bitfield256) ClearBit(bv byte) {
```

pkg/subnet/byte_set.go
```diff
@@ -0,0 +1,101 @@
...
+	b[bv>>6] &^= uint64(1) << uint64(bv&0x3f)
+}
+
+// Contains returns true if this ByteSet contains the given byte
+func (b *ByteSet) Contains(bv byte) bool {
```
LukeShu, line 19:
```go
// GetBit returns the value of the 1<<bv bit of the bitfield (0 is false, 1 is true).
//
// Compare: math/big.Int.Bit
func (b *Bitfield256) GetBit(bv byte) bool {
```

pkg/subnet/byte_set.go
```diff
@@ -0,0 +1,101 @@
...
+// Equals returns true if this ByteSet equals the argument
+func (b *ByteSet) Equals(other *ByteSet) bool {
+	if other == nil {
+		return false
+	}
+	return *b == *other
+}
+
+// ToSlice returns the number of bytes in this ByteSet
+func (b *ByteSet) Len() (l int) {
```
LukeShu, line 32: 1. I don't think you meant to write `ToSlice`?
2.
```go
// OnesCount returns the number of 1 bits in the bitfield.
//
// Compare: math/bits.OnesCount
func (b *Bitfield256) OnesCount() int {
```

pkg/subnet/byte_set.go
```diff
@@ -0,0 +1,101 @@
...
+	for _, g := range b {
+		if g != 0 {
+			for bit := 0; bit < 64; bit++ {
+				if g&(uint64(1)<<bit) != 0 {
+					l++
+				}
+			}
+		}
+	}
```
LukeShu, line 41: The inner loop could surely be improved by using `math/bits.OnesCount64()`.

pkg/subnet/subnet.go
```diff
@@ -15,6 +17,84 @@ type ipAndNetwork struct {
 // stubbable version
 var interfaceAddrs = net.InterfaceAddrs
 
+// IPv4 subnet key. Identifies a class B subnet
+type ipv4SubnetKey [2]byte
+
+// IPv6 subnet key. This is the 48-bit route and 16-bit subnet identifier. Identifies a 64 bit subnet.
+type ipv6SubnetKey [8]byte
```
LukeShu, line 24: Style suggestion:

Because these don't have any methods, and are only used inside of `AnalyzeIPs()`, go ahead and define the types inside of the `AnalyzeIPs()` body.

pkg/subnet/subnet.go
```diff
@@ -15,6 +17,84 @@ type ipAndNetwork struct {
...
+
+// AnalyzeIPs returns the ip networks needed to cover the given IPs with as
+// big mask as possible for each subnet. The analyze starts by finding all
+// subnets using a 16-bit mask for IPv4 and a 64 bit mask for IPv6 addresses.
+// Once the subnets are established, the mask for each one will be increased
+// to the maximum value that still masks all IPs that it was created for.
+func AnalyzeIPs(ips []net.IP) []*net.IPNet {
```
LukeShu, line 31: 1. I think `AnalyzeIPs` is a poor name, it doesn't communicate what it's analyzing for.  In the Telepresence 1 code the equivalent function is named `covering_cidrs`.
2. It's probably worth adding a comment about how this logic relates to the `covering_cidrs` logic from Tel 1.  If the answer is "This was written, but wasn't compared to the code in Tel 1", that's fine, but we should write it down.  FWIW, I'm pretty sure an external contributor has an open PR to improve this logic in Tel 1.

thallgren: The whole `subnet` package is completely removed later on when the need to figure out IP's and merge them into subnets is gone.

pkg/tun/route_darwin.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+	_, err = unix.Write(routeSocket, wb)
+	if err == unix.EEXIST {
+		// addrMessage exists, that's OK
+		err = nil
+	}
+	return err
+}
+
+// nolint:unused
```
donnyyung: I think we can remove this since it's used in `tun_darwin`, right?

pkg/tun/tun_darwin.go
```diff
@@ -0,0 +1,246 @@
...
+	name, err := unix.GetsockoptString(fd, sysProtoControl, uTunOptIfName)
+	if err != nil {
+		return nil, err
+	}
+	return &Device{File: os.NewFile(uintptr(fd), ""), name: name}, nil
+}
+
+// AddSubnet adds a subnet to this TUN device and creates a route for that subnet which
+// is associated with the device (removing the device will automatically remove the route).
+func (t *Device) AddSubnet(_ context.Context, subnet *net.IPNet) error {
```
LukeShu: A common idiom (at least in the stdlib) is to declare `t.Method()` in a common file, and have it just call `t.method()` which is defined in in the platform-specific file.  That way the godoc only needs to be written once.  But that probably matters more in libraries, and not so much in applications.

pkg/tun/tun_darwin.go
```diff
@@ -0,0 +1,246 @@
...
+func OpenTun() (*Device, error) {
+	fd, err := unix.Socket(unix.AF_SYSTEM, unix.SOCK_DGRAM, sysProtoControl)
+	if err != nil {
+		return nil, fmt.Errorf("failed to open DGRAM socket: %v", err)
+	}
+	defer func() {
+		if err != nil {
+			_ = unix.Close(fd)
+		}
+	}()
```
LukeShu: 1. Probably need to close it even if `err != nil`?
2. Does the daemon shell out for anything?  If it might fork, then when dealing with raw file descriptors we need to set up close-on-exec and also mess with `syscall.ForkClock`

```go
	syscall.ForkLock.RLock()
	fd, err := unix.Socket(unix.AF_SYSTEM, unix.SOCK_DGRAM, sysProtoControl)
	if err == nil {
		syscall.CloseOnExec(fd)
	}
	syscall.ForkLock.RUnlock()
```

Or, on platforms with `unix.SOCK_CLOEXEC` you can do it al atomically (IDK if macOS has it; GOOS=linux has it, but depends on Linux 2.6.27 (2008-10-09), which is slightly newer than Go itself depends on, 2.6.23 (2007-10-09), so maybe need to check for EINVAL and try again without it?  Or just declare that we need 2.6.27?):

```go
	fd, err := unix.Socket(unix.AF_SYSTEM, unix.SOCK_DGRAM|unix.SOCK_CLOEXEC, 0)
```

See also:
- https://git.lukeshu.com/go/libsystemd/tree/sd_daemon/notify_linux.go#n93
- https://git.lukeshu.com/go/libsystemd/tree/sd_daemon/notify_nonlinux.go#n93
- https://github.com/golang/go/blob/go1.15.12/src/net/sock_cloexec.go#L18-L50

pkg/tun/tun_darwin.go
```diff
@@ -0,0 +1,246 @@
...
+
+	if err = unix.SetNonblock(fd, true); err != nil {
+		return nil, err
+	}
+
+	name, err := unix.GetsockoptString(fd, sysProtoControl, uTunOptIfName)
+	if err != nil {
+		return nil, err
+	}
+	return &Device{File: os.NewFile(uintptr(fd), ""), name: name}, nil
```
LukeShu: IMO, it helps readability to write struct literals with 1 field per line.  As a 1-liner is fine if there's just one field, but once I have a second field, I do it as multiple lines.

pkg/tun/tun_linux.go
```diff
@@ -0,0 +1,146 @@
+// tun_linux.go: Open a Tunnel (L3 virtual interface) using the Universal TUN/TAP device driver.
+package tun
```
LukeShu: Without a blank line between the comment and the `package` line, this will show up as package-level documentation.  Which is mostly-fine, except then it probably shouldn't say `tun_linux.go: `.

pkg/tun/tun_linux.go
```diff
@@ -0,0 +1,146 @@
...
+// Index returns the index of this device
+func (t *Device) Index() uint32 {
+	return t.index
+}
+
+func (t *Device) SetMTU(mtu int) error {
+	return withSocket(unix.AF_INET, func(fd int) error {
+		var mtuRequest struct {
+			name [unix.IFNAMSIZ]byte
+			mtu  uint32
```
LukeShu: FWIW, both `<linux/if.h>` and GNU libc `<net/if.h>` define `ifru_mtu` to be `int`, not `uint`.

pkg/tun/tun_linux.go
```diff
@@ -0,0 +1,146 @@
...
+		}
+		copy(mtuRequest.name[:], t.name)
+		mtuRequest.mtu = uint32(mtu)
+		err := ioctl(fd, unix.SIOCSIFMTU, unsafe.Pointer(&mtuRequest))
```
LukeShu: This is a very short race condition.  Once the pointer is an `unsafe.Pointer`, the garbage collector doesn't know that `mtuRequest` is still reachable, and might garbage collect it after taking the `unsafe.Pointer` but before the actual syscall is made.

You should call `runtime.KeepAlive` after calling ioctl.  (See how `unix.IoctlSetIfreqMTU` does it on GOOS=darwin)

pkg/tun/tun_darwin.go
```diff
@@ -0,0 +1,246 @@
...
+//
+// See https://www.unix.com/man-page/osx/4/netintro/
+type addrIfReq6 struct {
+	name           [unix.IFNAMSIZ]byte
+	addr           unix.RawSockaddrInet6
+	dest           unix.RawSockaddrInet6
+	mask           unix.RawSockaddrInet6
+	flags          int32 // nolint:structcheck
+	expire         int64 // nolint:structcheck
+	preferred      int64 // nolint:structcheck
```
LukeShu: Assuming that `time_t` is 32 bits makes me hella nervous, especially when there's `unix.Time_t` to make this easy to get right.

LukeShu: ... and for some reason it's not defined on GOOS=darwin

pkg/tun/tun_linux.go
```diff
@@ -0,0 +1,146 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to open TUN device %s: %v", devicePath, err)
+	}
+
+	var flagsRequest struct {
+		name  [unix.IFNAMSIZ]byte
+		flags int16
+	}
+	copy(flagsRequest.name[:], "tel%d")
+	flagsRequest.flags = unix.IFF_TUN | unix.IFF_NO_PI
```
LukeShu: Without `unix.IFF_NO_PI`, it would have 4-byte prefix indicating some flags, and indicating the L3 proto.  We clearly don't need that, but having the 4 byte prefix might be nice, since it would mean that we could use the same `buffer` code on GOOS=darwin and GOOS=linux.  Sure, the 4 byte-prefix doesn't get parsed the same between the two platforms, but having the offset be the same might be nice.

thallgren: Parsing is half the problem. How would we set those four bytes when we write stuff back to the TUN-device? I assume they need to contain correct flags, etc.

pkg/tun/tun_darwin.go
thallgren: The close must only happen if `err != nil` because `fd` must remain open on success. I use this construct instead of repeatedly do a `unix.Close(fd)` on all error returns that follow.

LukeShu: Gotcha, I was thinking that `os.NewFile` `dup(2)`'d the fd, but I was mixing it up with `net.FileConn`.

pkg/tun/tun_linux.go
LukeShu: One thing I like to do to help manage things like this is to have a separate `syscall.go` file for things I wish were in `x/sys/unix`/`syscall` but aren't; to keep that type of thing self-contained.

pkg/tun/tun_linux.go
```diff
@@ -0,0 +1,153 @@
...
+
+type Device struct {
+	*os.File
+	name  string
+	index uint32
+}
+
+// OpenTun creates a new TUN device and ensures that it is up and running.
+func OpenTun() (*Device, error) {
+	// https://www.kernel.org/doc/Documentation/networking/tuntap.txt
```
LukeShu: That link is outdated; the kernel docs got restructured a couple of years ago.  The current link is https://www.kernel.org/doc/html/latest/networking/tuntap.html

LukeShu: or https://www.kernel.org/doc/html/latest/_sources/networking/tuntap.rst.txt if you like plaintext

pkg/tun/tun_darwin.go
```diff
@@ -0,0 +1,247 @@
...
+		return withSocket(unix.AF_INET, func(fd int) error {
+			ifreq := &addrIfReq{
+				addr: unix.RawSockaddrInet4{Len: 16, Family: unix.AF_INET},
+				dest: unix.RawSockaddrInet4{Len: 16, Family: unix.AF_INET},
+				mask: unix.RawSockaddrInet4{Len: 16, Family: unix.AF_INET},
+			}
+			copy(ifreq.name[:], t.name)
+			copy(ifreq.addr.Addr[:], sub4.IP)
+			copy(ifreq.mask.Addr[:], sub4.Mask)
+			copy(ifreq.dest.Addr[:], to4)
```
LukeShu: IMO, it would be a big help to readability to define some `Cast{IfName,IPv{4,6}{,Mask}` helper functions to just define these as struct literals.

pkg/tun/tun_linux.go
```diff
@@ -0,0 +1,153 @@
...
+	"github.com/datawire/dlib/dexec"
+	"github.com/telepresenceio/telepresence/v2/pkg/tun/buffer"
+)
+
+const devicePath = "/dev/net/tun"
+
+type Device struct {
+	*os.File
+	name  string
+	index uint32
```
LukeShu: Is there somewhere else that defines it as a `uint`? Because `<linux/if.h>` defines `ifru_index` as an `int`.

LukeShu: `<linux/ipv6_route.h>` also has `rtmsg_ifindex` as an `int`.

pkg/tun/tun_linux.go
```diff
@@ -0,0 +1,153 @@
...
+	// Passing a network ordered short here is required.
+	provisioningSocket, err := unix.Socket(unix.AF_PACKET, unix.SOCK_RAW, int(htons(unix.ETH_P_ALL)))
+	if err != nil {
+		return nil, err
+	}
+	defer unix.Close(provisioningSocket)
+
+	if err = unix.BindToDevice(provisioningSocket, name); err != nil {
+		return nil, err
+	}
```
LukeShu: Where's this come from?

If I look at net-tools `ifconfig` (the usual `ifconfig` on Linux, unless you're using BusyBox `ifconfig`), it uses
`socket(AF_INET, SOCK_DGRAM, IPPROTO_IP)` and then doesn't bother binding it to a device.

Can we get a comment with explanation or references or something?

pkg/tun/tun_linux.go
```diff
@@ -0,0 +1,153 @@
...
+	flagsRequest.flags |= unix.IFF_UP | unix.IFF_RUNNING
+	if err = ioctl(provisioningSocket, unix.SIOCSIFFLAGS, unsafe.Pointer(&flagsRequest)); err != nil {
+		return nil, err
+	}
```
LukeShu: How confident are we that it starts with `flags=0`? And/or that these are the only flags we want?  The normal thing to do is to get the flags first, and then or that with the flags you want to set; what you're doing will clear existing flags.

pkg/tun/tun_linux.go
```diff
@@ -0,0 +1,153 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Set non-blocking so that Read() doesn't hang for several seconds when the
+	// fd is Closed. Read() will still wait for data to arrive.
+	//
+	// See: https://github.com/golang/go/issues/30426#issuecomment-470044803
+	_ = unix.SetNonblock(fd, true)
+	return &Device{File: os.NewFile(uintptr(fd), devicePath), name: name, index: index}, nil
```
LukeShu: `devicePath` (`/dev/net/tun`) here seems a little funny to me.  I would think that `name` (`telXX`) would make more sense.

pkg/tun/tun_darwin.go
```diff
@@ -0,0 +1,247 @@
...
+		return nil, fmt.Errorf("failed to getBuffer IOCTL info for %s: %w", uTunControlName, err)
+	}
+
+	if err = unix.Connect(fd, &unix.SockaddrCtl{ID: info.Id, Unit: 0}); err != nil {
+		return nil, err
+	}
+
+	if err = unix.SetNonblock(fd, true); err != nil {
+		return nil, err
+	}
```
LukeShu: It's weird that this is so far down, after the `ip link up`-stuff.

pkg/tun/tun_darwin.go
```diff
@@ -0,0 +1,247 @@
...
+		if err != nil {
+			err = fmt.Errorf("set MTU on %s failed: %w", t.name, err)
+		}
+		return err
+	})
+}
+
+// Read reads as many bytes as possible into the given buffer.Data and returns the
+// number of bytes actually read
+func (t *Device) Read(into *buffer.Data) (int, error) {
```
LukeShu: Having `Read` and `Write` methods that are different than `io.Reader` and `io.Writer` is confusing.  Perhaps name them `ReadPacket` and `WritePacket`?

pkg/tun/tcp/header.go
```diff
@@ -0,0 +1,283 @@
...
+	"golang.org/x/sys/unix"
+
+	"github.com/telepresenceio/telepresence/v2/pkg/tun/ip"
+)
+
+// HeaderLen is the length of a TCP header that doesn't have any options
+const HeaderLen = 20
+const HeaderMaxLen = 60
+
+// Header represents a TCP header. The header is obtained by simply casting the IP headers payload.
```
LukeShu: header's

pkg/tun/tun_linux.go
thallgren: I don't think that would work. `/dev/net/tun` is the path of a pre-existing character device. TP doesn't create it. The actual device name is determined later by passing `"tel%d" to the `TUNSETIFF` ioctl.

pkg/tun/tun_linux.go
thallgren: I should have made a comment because now I cannot find the link. I think I got it from the ifconfig source but perhaps it was the wrong one. In any case, using the `socket()` parameters that you suggest and removing the `BindToDevice()` call works just fine, so I'll do that.

pkg/tun/tun_linux.go
thallgren: `int` is is, or rather `int32` since it's the go correspondance.

thallgren:
I've addressed most comments and rebased on top of release/v2. I didn't change anything in the subnet package because it is completely removed in a future PR.

New commits start at "Close TUN file descriptor on exec".

pkg/tun/tun_darwin.go
thallgren: It's not clear to me what you mean here. Can you please give an example?

LukeShu:
My mode of thinking when thinking about the dispatcher and all of the `pkg/tun/*/` code that goes with it is different enough from my mode of thinking about the stuff in `pkg/tun/` from before that commit that I'm thinking it should be a different package?  `pkg/netstack/`?

LukeShu:
Am I going crazy, or is this all single-threaded-enough that the buffer pool thing is unnecessary, and we could just have each goroutine have its own buffer on its stack?

LukeShu:
I'm curious if you looked in to using readv/writev to avoid the buffer thing?

pkg/subnet/subnet.go
```diff
@@ -45,6 +125,38 @@ func FindAvailableClassC() (*net.IPNet, error) {
 	return nil, errors.New("no available CIDR")
 }
 
+func Available(subnet *net.IPNet) (bool, error) {
+	addrs, err := interfaceAddrs()
+	if err != nil {
+		return false, fmt.Errorf("failed to obtain interface addresses: %v", err)
```
LukeShu: Use `%w` instead of `%v` when wrapping errors (I think there are a couple of other places too).

thallgren:
I haven't looked at readv/writev for this. Perhaps something to look at as a future optimization but I'm not sure we'd gain much. It might also be hard to find a good Go implementation on Windows.

I don't think we can have go-routines that keep buffers on the stack. Especially not for the TCP handler. It keeps track of packages that have been ack-ed and performs resends of those that don't. Another complicating factor is fragmentation and that the fragments not necessarily arrives in correct order.

Regarding package names, let's revisit that when you've reviewed all the PRs. I'm all for better naming and I know that some of the names I chose may be confusing. Perhaps some of the code needs to move too, to make things clearer.

LukeShu:
Do TUN devices give us IP packets in wire format, or in the platform-specific not-quite-wire-format of golang.org/x/net/ipv4?

Between GOOS=linux and GOOS=darwin, it would make a difference with the `TotalLen` field.

Edit: Wait, maybe?  It looks like they're different in `Parse`, but are the same in `Marshal`?

LukeShu:
Not that it actually matters, but a handful of the commits are from @tada.se not from @datawire.io

pkg/tun/ip/header.go
```diff
@@ -0,0 +1,132 @@
...
+	PayloadLen() int
+
+	// L4Protocol is the protocol of the layer-4 header.
+	L4Protocol() int
+
+	// PacketBase returns the full packet that this header is backed by.
+	Packet() []byte
+
+	// Payload returns the payload of the package that this header is backed by.
+	Payload() []byte
```
LukeShu, line 42: Since this includes the payload and not just the header, shouldn't this be "ip.Packet", not "ip.Header"?

thallgren: It doesn't really "include" anything. It's just a cast of a `[]byte` that exposes all header fields. I use the term `Packet` for things that holds on to the buffer and have methods to retrieve the IPHeader. That terminology is used throughout so there is also `tcp.Header` and `tcp.Packet` for instance.

thallgren:
What we receive and send to the TUN device is always in wire format and it's retained that way because it's never parsed or serialized. Just wrapped. Hence all the `binary.BigEndian` usage.

pkg/tun/ip/v4header.go
```diff
@@ -0,0 +1,273 @@
...
+		s += int(h[i])<<8 | int(h[i+1])
+	}
+	for s > 0xffff {
+		s = (s >> 16) + (s & 0xffff)
+	}
+	c := ^uint16(s)
+	if c == 0 {
+		// From RFC 768: If the computed checksum is zero, it is transmitted as all ones.
+		c = 0xffff
+	}
```
LukeShu: Was this mis-copy-pasted?  This is the IPv4 (RFC 791) checksum, not the UDP (RFC 768) checksum.  I see the language you refer to in RFC 768, but I don't see any such language in RFC 791.

LukeShu:
I asked about wire-format vs raw IP socket format because the old Linux TUN code used x/net/ipv4 which did the raw IP socket format, so my thinking was along the lines that if that was correct before, how is it correct to change it?  And I think the answer is "on Linux the raw wire-format and the raw IP socket format happen to be the same".

pkg/tun/ip/v4header.go
thallgren: Good catch. The IP checksum is not optional so if it indeed is zero, it should not be presented as all ones.

pkg/tun/ip/v4header.go
```diff
@@ -0,0 +1,273 @@
...
+		// last fragment hasn't arrived yet.
+		return nil
+	}
+
+	// Ensure that there are no holes in the fragment chain
+	lastPayload := 0
+	expectedOffset := 0
+	for _, data := range fragments {
+		eh := V4Header(data.Buf())
+		if eh.FragmentOffset()*8 != expectedOffset {
```
LukeShu: Fragments are allowed to overlap.  This should be checking `>` not `!=`.

pkg/tun/ip/v4header.go
```diff
@@ -0,0 +1,269 @@
...
+	t := h.HeaderLen()
+	for i := 0; i < t; i += 2 {
+		s += int(h[i])<<8 | int(h[i+1])
+	}
+	for s > 0xffff {
+		s = (s >> 16) + (s & 0xffff)
+	}
+	c := ^uint16(s)
+	binary.BigEndian.PutUint16(h[10:], c)
+}
```
LukeShu, line 200: I think this will panic on an odd-length header?

thallgren: It would, but there will never be an odd-length IPv4-header. Without extension headers, the IPv4 header is 20 bytes (or 5 * groups of 4 bytes). It can be extended by more groups of 4, but never with odd bytes.

LukeShu: You're right. I was thrown off by RFC 1071 specifically handling odd-length inputs, but that clearly doesn't apply to IPv4 headers.

LukeShu:
This commit set adds a lot of code just to delete it a few commits later.  That makes review tricky.

thallgren:
I know, and I apologize for that. It's hard to create a consistent series of commits that follows one line of thought after having had tens (or hundreds) of ideas that you've tried out. In my defense, I have spent a fair amount of time squashing stuff to get closer to that line but I know the result is far from perfect.

pkg/tun/tcp/packet.go
```diff
@@ -0,0 +1,105 @@
...
+func (p *packet) Header() Header {
+	return p.IPHeader().Payload()
+}
+
+func (p *packet) PayloadLen() int {
+	return p.IPHeader().PayloadLen() - p.Header().DataOffset()*4
+}
+
+func (p *packet) String() string {
+	b := bytes.Buffer{}
```
LukeShu, line 69: `strings.Builder` would be better for this case, since it would avoid an alloc+copy when you call `.String()`.

LukeShu:
Now, the `L4Checksum` _does_ have the "panics for odd payload lengths" problem, right?

Edit: Nevermind, I was looking for the odd-byte to be handle after the loop; you handle it before.

LukeShu:
I don't see the "a zero checksum is transmitted as all 1s" for TCP (RFC 793) either.  I think it's just a UDP thing?

thallgren:
Yes, the zero check is only for UDP. It is covered by today's commit 095a72a16193d76f9c62aaa692bde0eef9453452

pkg/tun/tcp/packet.go
thallgren: The problem with `strings.Builder` is that it doesn't allow `Truncate()` which complicates the `AppendFlags()` function. Also, performance it not an issue for this function as it is only used for debugging.

pkg/client/connector/outbound.go
```diff
@@ -153,7 +153,7 @@ func (kc *k8sCluster) refreshNamespaces(c context.Context, accWait chan<- struct
 }
 
 func (kc *k8sCluster) shouldBeWatched(namespace string) bool {
-	if len(kc.mappedNamespaces) == 0 {
+	if len(kc.mappedNamespaces) == 0 || namespace == "kube-system" {
```
LukeShu, line 158: Probably add a comment saying what the commit message says?

pkg/client/connector/outbound.go
```diff
@@ -300,7 +305,7 @@ func (kc *k8sCluster) updateDaemonNamespaces(c context.Context) {
 	for _, ns := range namespaces {
 		paths = append(paths, ns+clusterServerSuffix+".")
 	}
-	dlog.Debugf(c, "posting search paths to %s", strings.Join(paths, " "))
+	dlog.Debugf(c, "posting search paths [%s]", strings.Join(paths, " "))
```
LukeShu: FWIW, I'm pretty sure that's equivalent to
```go
dlog.Debugf(c, "posting search paths %v", paths)
```

pkg/tun/icmp/packet.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+	icmpHdr := Header(iph.Payload())
+	icmpHdr.SetMessageType(msgType)
+	icmpHdr.SetCode(int(code))
+	roh := icmpHdr.RestOfHeader()
+	binary.BigEndian.PutUint16(roh[2:], mtu)
+	copy(icmpHdr.Payload(), origHdr.Packet()[:origHdr.HeaderLen()+8])
+	icmpHdr.SetChecksum(iph)
+	return pkt
+}
```
LukeShu: OK, my last comment on this PR:

1. What's the deal with putting the MTU in the roh?  Are you looking at the same section of ICMP as I am?  I see the entire `roh` as being unused/zero for type="Destination Unreachable" packets. https://datatracker.ietf.org/doc/html/rfc792#page-4 https://datatracker.ietf.org/doc/html/rfc4443#section-3.1

2. That `origHdr.HeaderLen()+8` is _probably_ good enough for both, but is ICMPv4 specific; ICMPv6 says "just shove in as much of `origHdr.Packet()` as you can without blowing the MTU".  So for v6 it should be more like `min(mtu-ip.V6HeaderLen-HeaderLen, len(origHdr.Packet()))`

pkg/tun/ip/header.go
LukeShu: So I guess I was saying that it's surprising to me that `header.Payload()` is possible; it's surprising that `len(header) != header.HeaderLen()`.

LukeShu (approved):
Approving since (1) receivers should ignore the unused bytes in ICMP, so that's fine, and (2) The v4 ICMP payload len should be good enough for v6, even if v6 can do better.

pkg/tun/icmp/packet.go
thallgren: Doh! That ROH shouldn't be there. It's a left over from an experiment with MTU. I'll remove it.

PR #1655, opened 2021-04-09
[v2] Fix hcp error log

donnyyung:
hcp.PortName was correct when we were adding the agent but it is
hcp.HiddenName when we do an uninstall, so in the error log we should use the
input to be correct in both cases.

---

## Checklist

### DIFF
--- a/pkg/client/connector/install_actions.go
+++ b/pkg/client/connector/install_actions.go
@@ -571,7 +571,7 @@ func (hcp *hideContainerPortAction) getPort(obj kates.Object, name string) (*kat
 			}
 		}
 	}
-	return nil, nil, fmt.Errorf("unable to locate port %s in container %s in %s %s.%s", hcp.PortName, hcp.ContainerName, objKind, obj.GetName(), obj.GetNamespace())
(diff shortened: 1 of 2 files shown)


pkg/client/connector/install_actions.go
```diff
@@ -571,7 +571,7 @@ func (hcp *hideContainerPortAction) getPort(obj kates.Object, name string) (*kat
 			}
 		}
 	}
-	return nil, nil, fmt.Errorf("unable to locate port %s in container %s in %s %s.%s", hcp.PortName, hcp.ContainerName, objKind, obj.GetName(), obj.GetNamespace())
+	return nil, nil, fmt.Errorf("unable to locate port %s in container %s in %s %s.%s", name, hcp.ContainerName, objKind, obj.GetName(), obj.GetNamespace())
```
donnyyung, line 574: Use the input for the log bc that's what we are checking in the logic and the provided input is different in 598 (hcp.PortName) than 618 + 627 (hcp.HiddenName)

thallgren (approved):
LGTM

PR #1661, opened 2021-04-12
Fix checkSvcSame when portNameOrNumber empty

donnyyung:
When introducing portNameOrNumber instead of just portName, I made an
error in the logic and it was checking if the annotations matched the
value of portNameOrNumber, even when that value is empty.  This fixes
the logic so that we only validate portNameOrNumber with the annotations
when it is non-empty

---
## Checklist

 - [x] I made sure to update `./CHANGELOG.md`. - bugfix for new feature that hasn't been released yet so omitted from changelog


thallgren (approved):
Hit it!

PR #1663, opened 2021-04-12
Ensure that we don't create port names longer than 15 characters.

thallgren:

## Description

Don't generate port names longer than 15 characters.

## Checklist

### DIFF
--- a/pkg/client/connector/install_actions.go
+++ b/pkg/client/connector/install_actions.go
@@ -579,6 +579,11 @@ type hideContainerPortAction struct {
 	PortName      string `json:"port_name"`
 	// HiddenName is the name that we swapped it to; this is set by Do(), and read by Undo().
 	HiddenName string `json:"hidden_name"`
+
+	// ordinal is only used for avoiding ambiguities when generating the HiddenName. It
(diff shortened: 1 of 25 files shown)


pkg/client/connector/install_actions.go
```diff
@@ -579,6 +579,11 @@ type hideContainerPortAction struct {
 	PortName      string `json:"port_name"`
 	// HiddenName is the name that we swapped it to; this is set by Do(), and read by Undo().
 	HiddenName string `json:"hidden_name"`
+
+	// ordinal is only used for avoiding ambiguities when generating the HiddenName. It
+	// is the zero based order of all hideContainerPortAction instances for a workload.
+	// Right now we only use one port so this is always zero.
+	ordinal int
```
donnyyung, line 586: I think we might need a re-factor of a lot of things once we have more than one port intercepted at a time, but I think it's good to add this here just so don't forget this issue when it comes time to do that!

donnyyung (approved):
LGTM, thanks for doing this and in a forward-thinking manner!

LukeShu (approved):

PR #1668, opened 2021-04-15
[v2] Support license for air-gapped environment

donnyyung:
This is the first set of work to enable people to use selective intercepts with telepresence in air-gapped environments. It add a new command that creates a license secret that users can than apply to their kubernetes cluster, and then the traffic-manager can send that information to the traffic-agent, which can then validate the license.

---
## Checklist

 - [ ] I made sure to either submit a docs PR, or tell Matt about the necessary documentation changes. - todo
 - [ ] My change is adequately tested. - would have liked to write some additional tests, might still do that if I have time / if this is deemed sufficient
 - [x] I updated `DEVELOPING.md` with any any special dev tricks I had to use to work on this code efficiently. -- no tricks


pkg/client/cli/cliutil/systema.go
```diff
@@ -93,3 +96,51 @@ func GetCloudAPIKey(ctx context.Context, description string, autoLogin bool) (st
 	}
 	return keyData.GetApiKey(), nil
 }
+
+// GetCloudLicense communicates with system a to get the jwt version of the
+// license, puts it in a kubernetes secret, and then writes that secret to the
+// output file for the user to apply to their cluster
+func GetCloudLicense(ctx context.Context, outputFile, id string) (string, error) {
```
LukeShu, line 103: The "writing to a file" part seems like something that should be part of the `cli/license.go` subcommand implementation, not something in `cliutil`; it's a mixing of concerns.

pkg/client/cli/license.go
```diff
@@ -0,0 +1,40 @@
...
+
+type licenseInfo struct {
+	outputFile string
+}
+
+func LicenseCommand() *cobra.Command {
+	li := &licenseInfo{}
+	cmd := &cobra.Command{
+		Use:  "license",
+		Args: cobra.MinimumNArgs(1),
```
LukeShu, line 20: If you're going to accept positional arguments, the `Use` string should tell the user what they are.

For example, `intercept_state.go`'s
```go
Use:  "intercept [flags] <intercept_base_name> [-- <command with arguments...>]",
Args: cobra.MinimumNArgs(1),
```

pkg/client/cli/license.go
```diff
@@ -0,0 +1,40 @@
...
+
+		Short: "Get License from Ambassador Cloud",
+		Long:  "Get License from Ambasssador Cloud",
```
LukeShu, line 23: typo: `Ambasssador` has too many `s`

pkg/client/cli/license.go
LukeShu: And it looks like it expects exactly 1 arg, not ≥1 args, so this should be `cobra.ExactArgs(1)`.

pkg/client/cli/license.go
```diff
@@ -0,0 +1,40 @@
...
+		RunE:  li.getCloudLicense,
+	}
+	flags := cmd.Flags()
+
+	flags.StringVarP(&li.outputFile, "output-file", "f", "/tmp/license", "The file where you want the license secret to be output to")
```
LukeShu, line 28: Use `filepath.Join(os.TempDir(), "license")` instead of hard-coding `/tmp`.

pkg/client/cli/license.go
```diff
@@ -0,0 +1,40 @@
...
+	return cmd
+}
+
+func (li *licenseInfo) getCloudLicense(cmd *cobra.Command, args []string) error {
```
LukeShu, line 32: FWIW, it's my preference to have the separate `RunE` implementation  take arguments like it would if it were a regular Go function, and translate cobra to that in-line; something like:

```go
func LicenseCommand() *cobra.Command {
	var flags struct {
		outputFile string
	}
	cmd := &cobra.Command{
		Use:  "license [flags] <id>",
		Args: cobra.imumNArgs(1),

		Short: "Get License from Ambassador Cloud",
		Long:  "Get License from Ambassador Cloud",
		RunE: func(cmd *cobra.Command, args []string) error {
			return getCloudLicense(cmd.Context(), cmd.OutOrStdout(),
				args[0], flags.outputFile)
		},
	}
	cmd.Flags().StringVarP(&flags.outputFile, "output-file", "f", "/tmp/license", "The file where you want the license secret to be output to")
	return cmd
}
```

pkg/client/connector/service.go
```diff
@@ -299,6 +299,29 @@ func (s *service) GetCloudAPIKey(ctx context.Context, req *rpc.KeyRequest) (*rpc
...
 
+func (s *service) GetCloudLicense(ctx context.Context, req *rpc.LicenseRequest) (*rpc.LicenseData, error) {
+	ctx = s.callCtx(ctx, "GetCloudLicense")
+	license, err := s.getCloudLicense(ctx, req.GetId())
+	if err != nil {
+		return nil, err
+	}
+	return &rpc.LicenseData{License: license}, nil
+}
+func (s *service) getCloudLicense(ctx context.Context, id string) (string, error) {
```
LukeShu, line 310: Is there a reason to have `.getCloudLicense` separate from `.GetCloudLicense`?  The reason I have a separate `.getCloudAPIKey` is because `traffic_manager.go` wants to use it (and the separate `.getCloudAccessToken` is vestigial; a few different things _used to_ want to use it); I don't see anywhere else that wants to use `.getCloudLicense`.

pkg/client/cli/license.go
LukeShu: Mostly because of `curl` using `-o` for output and `kubectl apply` using `-f` for input; I'd expect the short flag here to be `-o` and not `-f`.

pkg/client/cli/license.go
LukeShu: Dumb question: How do I know what to pass to Telepresence as the license key ID?  I assume you're writing docs to explain that... but the `Long` help should at least point the user in the right direction to find that info.

pkg/client/cli/license.go
```diff
@@ -0,0 +1,40 @@
...
+	outputFile := strings.TrimSpace(li.outputFile)
+	id := strings.TrimSpace(args[0])
```
LukeShu, line 34: Why `TrimSpace`? I wouldn't expect there to be extra spaces here.

pkg/client/connector/auth/license.go
```diff
@@ -0,0 +1,85 @@
...
+		// includes it for us.
+		return "", err
+	}
+	// Output is different depending on the status code so we return custom errors depending
+	// on the status of the request
+	switch resp.StatusCode {
+	case 404:
+		return "", fmt.Errorf("license JWT not found")
+	case 500:
+		return "", fmt.Errorf("Error getting license jwt for %s: %s", licenseID, string(bodyBytes))
```
LukeShu, line 62: 1) Error messages usually start with a lowercase letter

2) I'd probably change that to `server error` instead of just `error`.  Maybe even include the `resp.Status` message string?

3) I'd use `%q` instead of `%s` for the `licenseID`.  I like to use `%q` pretty much any time we're echoing the user's input back at them; avoid any confusion if they accidentally copied an unprintable character or something like that.

pkg/client/connector/auth/license.go
```diff
@@ -0,0 +1,85 @@
...
+	defer l.loginMu.Unlock()
+	if l.tokenSource == nil {
+		return "", fmt.Errorf("GetLicense: %w", ErrNotLoggedIn)
+	} else if tokenInfo, err := l.tokenSource.Token(); err != nil {
+		return "", err
+	} else if license, err := getLicenseJWT(ctx, l.env, tokenInfo.AccessToken, id); err != nil {
+		return "", err
+	} else {
+		if license == "" {
+			return "", fmt.Errorf("No licenses found for %s", id)
```
LukeShu, line 81: likewise; lowercase letter and use `%q`

pkg/client/connector/install.go
```diff
@@ -1029,6 +1035,30 @@ func (ki *installer) managerDeployment(env client.Env) *kates.Deployment {
...
+			},
+		}
+		licenseVolumeMount = []corev1.VolumeMount{
+			{
+				Name:      "license",
+				ReadOnly:  true,
+				MountPath: "/home/telepresence/",
+			},
+		}
+	}
```
LukeShu, line 1061: Now wait a minute, if we're going to get it as a volume from a secret, wouldn't it be way simpler to have it get mounted directly in the agent workloads? Instead of mounting it in the manager Pod and passing it to the agents over gRPC?

pkg/client/connector/install.go
```diff
@@ -1029,6 +1035,30 @@ func (ki *installer) managerDeployment(env client.Env) *kates.Deployment {
...
+						SecretName: managerLicenseName,
+					},
+				},
+			},
+		}
+		licenseVolumeMount = []corev1.VolumeMount{
+			{
+				Name:      "license",
+				ReadOnly:  true,
+				MountPath: "/home/telepresence/",
```
LukeShu, line 1058: That's an... interesting place to put it.

rpc/manager/manager.proto
```diff
@@ -255,10 +255,16 @@ message VersionInfo2 {
...
+message License {
+  string license = 1;
+  string host = 2;
+}
+
 service Manager {
   // Version returns the version information of the Manager.
   rpc Version(google.protobuf.Empty) returns (VersionInfo2);
 
+  rpc GetLicense(google.protobuf.Empty) returns (License);
```
LukeShu, line 267: No doc comment? :)

rpc/manager/manager.proto
```diff
@@ -255,10 +255,16 @@ message VersionInfo2 {
   string version = 2;
 }
 
+message License {
+  string license = 1;
+  string host = 2;
```
LukeShu, line 260: Probably add a doc comment clarifying that `host` is the server that issued the license (i.e. `SYSTEMA_HOST`).

LukeShu: Except is that really even true?  It's set to the manager's `SYSTEMA_HOST`, which isn't necessarily how it was set on the laptop when the license was created.

pkg/client/cli/intercept_state.go
```diff
@@ -110,6 +112,22 @@ func leaveCommand() *cobra.Command {
...
+		timeout := 2 * time.Second
+		conn, err := net.DialTimeout("tcp", fmt.Sprintf("%s:%s", env.SystemAHost, env.SystemAPort), timeout)
+		if err != nil {
+			isAirGapped = true
+			return
+		}
+		conn.Close()
+	}
+	return
+}
```
LukeShu, line 129: Everyone keeps saying "air-gapped", but I think the scenario is actually just the cluster having a locked-down firewall.  I think the laptop is still able to talk to the Internet; so I'm not sure this is the right place for this check.

pkg/client/cli/license.go
LukeShu: Or maybe have the default be to dump it to stdout instead of using `os.TempDir()`.

rpc/manager/manager.proto
LukeShu: So probably have `telepresence license` shove the host in as part of the secret value? And have the manager read that?

pkg/client/connector/install.go
LukeShu: Oh, probably because RBAC, right?

pkg/client/cli/intercept_state.go
```diff
@@ -189,7 +212,7 @@ func interceptMessage(r *connector.InterceptResult) string {
 	case connector.InterceptError_NO_PREVIEW_HOST:
 		msg = `Your cluster is not configured for Preview URLs.
 (Could not find a Host resource that enables Path-type Preview URLs.)
-Please specify one or more header matches using --match.`
+Please specify one or more header matches using --http-match.`
```
LukeShu, line 215: Wow, I did a really bad job of searching for all of the `--match` references when I mad that change, didn't I?

LukeShu (changes requested):
On the whole, pretty solid!

- Lots of I have lots of nit-picks I'm asking you to change
- I have one question (about why not have the agent mount the license directly; RBAC, right?)
- I have one substantive change complaint: The `isAirGapped` logic.

I'd change the extension thing to say `RequiresAPIKeyORLicense` instead of `RequiresAPIKey` and if that's why it triggers a `login`, have have it print a message about "if your cluster doesn't have access to the public Internet, you should hit Ctrl-C and use `telepresence license` instead".

pkg/client/cli/license.go
donnyyung: Yup, I'm just waiting for the UI page to be set in staging so I A. know it works B. can point users to how to use it.

Right now I get my license via the API but I'm not expecting users to do that... and there will be a more in-depth tutorial in the docs.

pkg/client/cli/license.go
donnyyung: Was just being overly cautious, but agreed it is unnecessary

pkg/client/connector/install.go
donnyyung: Yeah since I was using secrets, it'd be an RBAC escalation and normal users would need the ability to read secrets which I think some organizations are rightly opposed to allowing their developers to do.  

Additionally secrets are namespace bound so then we'd have to put that secret in every namespace intercepted workloads were in, which I thought might be a headache.  

All that being said, secrets might not be the best™️ way to do this forever, so I will revisit that in the next iteration of license work.

pkg/client/connector/install.go
donnyyung: Do you have a strong opinion on where to put it? I mostly just picked it bc that dir already exists and didn't have a strong opinion / thought maybe y'all would.

pkg/client/cli/intercept_state.go
```diff
@@ -148,7 +148,7 @@ func (ii *interceptInfo) intercept(cmd *cobra.Command, args []string) error {
 	// We don't particularly care about errors here because if we do
 	// error then we'll assume they aren't in an air-gapped environment
 	// and if there are issues with the connector, it will fail later
-	_ = ii.withConnector(true, func(cs *connectorState) error {
+	_ = ii.withConnector(false, func(cs *connectorState) error {
```
thallgren, line 151: This is very expensive. Especially for the case when an intercept is done in a disconnected state. Unless a connect has been done in advance, this will start the connector and daemon, do the call, and then kill them both. To avoid that, I suggest the following:

1. Always perform the check if the user is required to log in but isn't (it's cheap) but don't return the error. Instead, assign it to an specific error variable.
2. Defer the call to `CanConnectAmbassadorCloud()` to a point where the connector is enabled (on line 178 and 185) and don't call at all unless the error is set.

thallgren (changes requested):

pkg/client/cli/intercept_state.go
donnyyung: Ah thank you Thomas! This is what was bugging me last night bc it was expensive but I was over-tired and figured I should sleep on it and return to it in the morning.

pkg/client/connector/auth/license.go
```diff
@@ -76,9 +76,10 @@ func (l *loginExecutor) GetLicense(ctx context.Context, id string) (string, erro
 		return "", err
 	} else if license, err := getLicenseJWT(ctx, l.env, tokenInfo.AccessToken, id); err != nil {
 		return "", err
-	} else if license == "" {
-		return "", fmt.Errorf("No licenses found")
 	} else {
+		if license == "" {
```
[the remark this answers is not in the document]
donnyyung: That was what I had but the linter didn't like it  and wanted to replace it with a switch which I thought would be even more ugly, let me see if I can find a way to make it work.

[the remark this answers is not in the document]
thallgren: The linter is also wrong. It cannot be replaced with a switch because some of the if's contain assignments. Switch cases cannot have assignments. So, never mind.

LukeShu (approved):
- Question: Why have the agent ask the manager to try connecting to Ambassador Cloud, instead of just doing that themselves?
- I'd still like to see a better `Long` help message, but I understand you're waiting on the UI side.

LGTM!

pkg/client/connector/auth/license.go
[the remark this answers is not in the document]
LukeShu: That's weird that that's what the linter said--the linter is usually smart enough to not suggest a switch when there are assignments.

donnyyung:
Yeah I'll get a better `Long` message once I have the UI in place.

PR #1670, opened 2021-04-15
[v2] Add --docker-run option to telepresence intercept.

thallgren:
## Description

Adds a --docker-run option to the `intercept` subcommand which will
start a `docker run` with the intercepted environment and remote volume
mount. An additional `--docker-mount` option is also provided for the
case where defaulting to `--mount` is undesirable.

## Checklist


thallgren:
The docs PR is here: https://github.com/datawire/getambassador.io/pull/694

pkg/client/cli/intercept_state.go
```diff
@@ -275,30 +301,61 @@ func (is *interceptState) createRequest() (*connector.CreateInterceptRequest, er
...
+	is.localPort = port
+	spec.TargetPort = int32(port)
+
+	switch len(portMapping) {
+	case 1:
 	case 2:
-		port, err := strconv.ParseUint(portMapping[0], 10, 16)
-		if err != nil {
-			return nil, errors.Errorf("Port numbers must be a valid, positive int, you gave: %q", portMapping[0])
+		if port, err = parsePort(portMapping[1]); err == nil && is.dockerRun {
```
donnyyung, line 332: This might just be the case of my first time seeing this, but will we hit the else if `err != nil`? If so, I think we may want to split this out to error check correctly if err != nil && is.dockerRun is false. Or is there some special behavior when checking err inline that the else only looks at the 2nd condition

donnyyung (approved):
This LGTM. I had one comment (that you may have seen) but I deleted it bc I had just misunderstood something in the code 🤦  thanks for the quick work on this!

LukeShu:
The kinda loosey goosey parsing of Docker's flags makes me nervous, but it's good enough for v1.  See also: https://github.com/datawire/docker-build-wrapper/

LukeShu (approved):

PR #1679, opened 2021-04-20
Better detect presence of ambassador namespace

donnyyung:
There was a race condition in determining presence of the ambassador
namespace that was dependent on trying to create the namespace and
ignoring the error if the namespace existed.  This is an unneccesary
rbac escalation, so I have fixed that here.

## Checklist

### DIFF
--- a/pkg/client/connector/install.go
+++ b/pkg/client/connector/install.go
@@ -94,14 +94,18 @@ func (ki *installer) createManagerSvc(c context.Context) (*kates.Service, error)
 	}
 
 	// Ensure that the managerNamespace exists
-	if !ki.namespaceExists(managerNamespace) {
+	_, err := ki.findNamespace(c, managerNamespace)
+	if err != nil {
 		ns := &kates.Namespace{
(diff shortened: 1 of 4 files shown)


pkg/client/connector/install.go
```diff
@@ -94,14 +94,18 @@ func (ki *installer) createManagerSvc(c context.Context) (*kates.Service, error)
 	}
 
 	// Ensure that the managerNamespace exists
-	if !ki.namespaceExists(managerNamespace) {
+	_, err := ki.findNamespace(c, managerNamespace)
+	if err != nil {
 		ns := &kates.Namespace{
```
thallgren, line 99: Should this proceed for other errors than `errors2.IsNotFound(err)` ?

thallgren (approved):
LGTM, but I'm always a bit anxious to catch all errors.

pkg/client/connector/install.go
donnyyung: Ah good point, fixed!

PR #1686, opened 2021-04-25
Get rid of sshd in the traffic-agent

thallgren:
## Description

This PR contains what is needed to get rid of sshd in the traffic-agent and all use of ssh in Telepresence. It sits on top of #1685.

## Checklist

 - [x] I made sure to update `./CHANGELOG.md`.
 - [x] My change is adequately tested.

PR #1687, opened 2021-04-25
Move all DNS resolution to the cluster.

thallgren:
## Description

This PR contains what's needed to move the DNS resolution to the cluster. Intercepted agents get a first say and if they fail, the traffic-manager is used. The PR sits on top of #1686.

## Checklist


pkg/client/connector/traffic_manager.go
```diff
@@ -499,3 +511,94 @@ func (tm *trafficManager) uninstall(c context.Context, ur *rpc.UninstallRequest)
...
+		// from the kubeDNS IP. This is cheating but a cluster may only have one service subnet and the mask is
+		// unlikely to cover less than half the bits.
+		dlog.Debugf(c, "Deriving serviceSubnet from %s (the IP of kube-dns.kube-system)", kubeDNS)
+		bits := len(kubeDNS) * 8
+		ones := bits / 2
+		mask := net.CIDRMask(ones, bits) // will yield a 16 bit mask on IPv4 and 64 bit mask on IPv6.
+		serviceSubnet = &daemon.IPNet{Ip: kubeDNS.Mask(mask), Mask: int32(ones)}
+	}
+
+	if err = tm.client.List(c, kates.Query{Kind: "Node"}, &nodes); err != nil {
```
donnyyung: I haven't thought about this much, but is there an alternative way for us to get this information without the nodes.. _or_ could we calculate this when the operator is launching telepresence and then get that value via gRPC call to the manger?  

I only say this bc requiring users to list nodes is a cluster permission and, at least at my last job, that wasn't given to developers and was rarely even given to my team as DevOps.  So it's kind a substantial permission escalation (imo). But if this is the only way we can get this info, then I think it's a worthy trade-off and maybe we come up w/ some workaround or other way to handle people who _can't_ list nodes.

donnyyung (changes requested):
Just left one comment about something that I think we should change if possible, but I'm fully open to discussion so there might not _actually_ be changes and if you were online now I'd just ping ya to ask, but feel free to slack me in the morning if you want to talk about it!

donnyyung:
Here's a list of the things I tested on this branch that worked as expected:

ExternalName services
headless services
local intercept
smoke tests up to ones requiring prop agent (Expected behavior)
gRPC intercept
login + logout
intercept multiple ports (name + number)
connect w/ mapped-namespaces

Potential bug:

This is a bit contrived, but I noticed (After I was away from my computer for a bit) the User daemon had stopped running. So I ran a telepresence connect and services in the cluster weren't resolving. When I tried to resolve a service, the errors looked like this in the daemon log:
```
2021/05/10 16:52:31 daemon/server-dns/Server LookupHost hello.default
2021/05/10 16:52:31 daemon/server-dns/Server rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing dial tcp 127.0.0.1:53980: connect: connection refused"
2021/05/10 16:52:31 daemon/server-dns/Server QTYPE[1] hello.default. -> NOT FOUND
```

So then I quit and connected again, this time the daemon logs seemed stuck here:

```
(⎈ |donnyyung-1:default)➜  ~/go/telepresenceio git:(thallgren/cluster-dns) ✗ cat ~/Library/Logs/telepresence/daemon.log
2021/05/10 16:29:34 daemon ---
2021/05/10 16:29:34 daemon Telepresence daemon v2.2.2-59-g5fefbc82-1620657280 (api v3) starting...
2021/05/10 16:29:34 daemon PID is 33581
2021/05/10 16:29:34 daemon
2021/05/10 16:29:34 daemon/server-grpc Server starting
2021/05/10 16:29:34 daemon/server-router/MGR stream Waiting until manager gRPC is configured
2021/05/10 16:29:34 daemon/server-router/TUN reader Waiting until manager gRPC is configured
```

Then I did one more quit and connect and it worked again.

I was able to reproduce this by killing the connector-foreground. (in our latest released version of telepresence, when I do the first telepresence connect after the user daemon has died, everything works).

Additional notes: If after killing the connector I do a `telepresence quit` before a `telepresence connect` everything works.

I can look more into what the root cause is (I think it is likely in one of the earlier PRs, not this one) but since I did all my testing on this branch, I thought I'd mention it here. Once we have that resolved, I will approve all the other PRs leading up to this one.

pkg/client/connector/traffic_manager.go
```diff
@@ -499,3 +511,102 @@ func (tm *trafficManager) uninstall(c context.Context, ur *rpc.UninstallRequest)
...
+		// from the kubeDNS IP. This is cheating but a cluster may only have one service subnet and the mask is
+		// unlikely to cover less than half the bits.
+		dlog.Debugf(c, "Deriving serviceSubnet from %s (the IP of kube-dns.kube-system)", kubeDNS)
+		bits := len(kubeDNS) * 8
+		ones := bits / 2
+		mask := net.CIDRMask(ones, bits) // will yield a 16 bit mask on IPv4 and 64 bit mask on IPv6.
+		serviceSubnet = &daemon.IPNet{Ip: kubeDNS.Mask(mask), Mask: int32(ones)}
+	}
+
+	if err := tm.client.List(c, kates.Query{Kind: "Node"}, &nodes); err != nil {
```
donnyyung, line 589: Edit: my 2nd suggestion might not even apply since we have an open issue to provide manager yaml that people can use to deploy the traffic-manager, so perhaps there isn't a way around it.. I'll sleep on it.

----

I haven't thought about this much, but is there an alternative way for us to get this information without the nodes.. _or_ could we calculate this when the operator is launching the traffic-manager  and then get that value via gRPC call to the manger?  

I only say this bc requiring users to list nodes is a cluster permission and, at least at my last job, that wasn't given to developers and was rarely even given to my team as DevOps.  So it's kind a substantial permission escalation (imo). But if this is the only way we can get this info, then I think it's a worthy trade-off and maybe we come up w/ some workaround or other way to handle people who _can't_ list nodes.

thallgren: I would like to get this info from the traffic-manager (I even have a branch where I'm testing it), but that in turn requires an RBAC setup for the manager. A default setup won't let me use kates at all. While I think that's something that we really will want, I though it was beyond the scope of the work I was doing. But perhaps that's a necessary step.

donnyyung: Ah okay that makes sense (and I agree that shouldn't be tied up into this work).  I'll chat more with you once I wake up tomorrow, but I don't think this is a dealbreaker then!

LukeShu:
"Remove subnet package. It's no longer needed" is doing a lot more than you would think by the commit message!

pkg/client/connector/traffic_manager.go
```diff
@@ -499,3 +511,102 @@ func (tm *trafficManager) uninstall(c context.Context, ur *rpc.UninstallRequest)
...
+		},
+	}
+	var kubeDNS net.IP
+	for _, dnsService := range dnsServices {
+		svc := kates.Service{TypeMeta: metav1.TypeMeta{Kind: "Service"}, ObjectMeta: dnsService}
+		if err := tm.client.Get(c, &svc, &svc); err == nil {
+			kubeDNS = iputil.Parse(svc.Spec.ClusterIP)
+			break
+		}
+	}
```
LukeShu, line 540: Part of me thinks we should just grab the IP from the `nameserver` line in `/etc/resolv.conf`.

thallgren: Yes, let's do that when we move the subnet discovery to the traffic-manager service.

thallgren:
I'm looking at commit 401efe8d7058ebcd6cd8eab87aaa5ff8678a1fdf. What's in that commit that you wouldn't expect from its commit message?

pkg/client/connector/traffic_manager.go
thallgren: I re-requested your review because you are requesting changes. If it's related to this then I suggest we fix this when we move the subnet discovery over to the traffic-manager service.

LukeShu:
> I'm looking at commit [401efe8](https://github.com/telepresenceio/telepresence/commit/401efe8d7058ebcd6cd8eab87aaa5ff8678a1fdf). What's in that commit that you wouldn't expect from its commit message?

... good question.

I think I meant to say "R.I.P. all watchers except the one for namespaces"?  And in that case, the answer is: the new `getOutboundInfo` method.

LukeShu (approved):

PR #1690, opened 2021-04-27
Improve license support

donnyyung:
The follow-up my recent license support PR. As usual, recommend a commit-by-commit review, but this PR isn't too complicated, most of the changes are gRPC / auto-generated.

## Checklist

 - [x] My change is adequately tested. -- Still need to add tests to validate license improvements


pkg/client/connector/install.go
```diff
@@ -1030,6 +1030,8 @@ func (ki *installer) managerDeployment(env client.Env, addLicense bool) *kates.D
 	if env.SystemAPort != "" {
 		containerEnv = append(containerEnv, corev1.EnvVar{Name: "SYSTEMA_PORT", Value: env.SystemAPort})
 	}
+	clusterID := ki.getClusterId(c)
+	containerEnv = append(containerEnv, corev1.EnvVar{Name: "CLUSTER_ID", Value: clusterID})
```
donnyyung, line 1034: I'm open to better ideas about how to get the cluster ID known by the agent, my first thought was to have the agent query it directly, but that would require a service account for the agent to use the kubernetes api, which seemed... pretty heavy / difficult for users? 

But if that's something we think would be more ideal + handled w/ documentation, let me know!

cmd/traffic/cmd/manager/envconfig.go
```diff
@@ -12,6 +12,7 @@ type Env struct {
 	ServerPort  string `env:"SERVER_PORT,default=8081"`
 	SystemAHost string `env:"SYSTEMA_HOST,default=app.getambassador.io"`
 	SystemAPort string `env:"SYSTEMA_PORT,default=443"`
+	ClusterId   string `env:"CLUSTER_ID,default="`
```
LukeShu, line 15: nit: style: `ID`, not `Id`.

rpc/manager/manager.proto
```diff
@@ -166,6 +166,9 @@ message InterceptInfo {
   string id = 5;
   SessionInfo client_session = 6;
 
+  // api_key # add real comment
```
LukeShu, line 169: add real comment

pkg/client/connector/intercept.go
```diff
@@ -214,13 +214,18 @@ func (tm *trafficManager) addIntercept(c context.Context, ir *rpc.CreateIntercep
 		}()
 	}
 
+	apiKey, err := tm.getAPIKey(c, "agent", false)
```
LukeShu, line 217: Consider including the agent name (e.g. `"agent-"+mechanism` or something) in this string?

pkg/client/connector/service_proxy.go
```diff
@@ -64,6 +64,15 @@ func (p *mgrProxy) CanConnectAmbassadorCloud(ctx context.Context, arg *empty.Emp
...
 	return client.CanConnectAmbassadorCloud(ctx, arg, callOptions...)
 }
+
+func (p *mgrProxy) GetCloudConfig(ctx context.Context, arg *empty.Empty) (*managerrpc.AmbassadorCloudConfig, error) {
+	client, callOptions, err := p.get()
+	if err != nil {
+		return nil, err
+	}
+	return client.GetCloudConfig(ctx, arg, callOptions...)
+}
```
LukeShu: Does this ever get called? I know it's implementing the interface.  I almost want to say "just have it always return an error, it should never be called because the client should already have the config."

LukeShu (approved):
Woo!

pkg/client/connector/service_proxy.go
donnyyung: I'll add a comment about investigating it's potential removal.  I think when I did it, I was mostly just seeing that pretty much every function was this way, so I followed the pattern.  Which I should have dug in a little further to actually ensure that it was necessary.

PR #1695, opened 2021-04-28
[v2] Fix `telepresence login` when there's an expired login

LukeShu:
This came up during smoke-testing of 2.2.1-rc.0.

### DIFF
--- a/pkg/client/connector/auth/login.go
+++ b/pkg/client/connector/auth/login.go
@@ -22,7 +22,6 @@ import (
 	"github.com/datawire/dlib/dgroup"
 	"github.com/datawire/dlib/dhttp"
 	"github.com/datawire/dlib/dlog"
-	"github.com/telepresenceio/telepresence/rpc/v2/connector"
 	"github.com/telepresenceio/telepresence/v2/pkg/client"
 	"github.com/telepresenceio/telepresence/v2/pkg/client/cache"
 	"github.com/telepresenceio/telepresence/v2/pkg/client/connector/auth/authdata"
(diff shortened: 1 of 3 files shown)


pkg/client/connector/auth/login.go
```diff
@@ -108,19 +107,6 @@ func NewLoginExecutor(
 	return ret
 }
 
-// EnsureLoggedIn will check if the user is logged in and if not initiate the login flow.
-func EnsureLoggedIn(ctx context.Context, executor LoginExecutor) (connector.LoginResult_Code, error) {
```
donnyyung, line 112: Okay so I understand this issue bc EnsureLoggedIn would see that the token file existed and would return OLD_LOGIN_REUSED, regardless of whether that token was expired or not.

LukeShu: Yeah.

pkg/client/connector/service.go
```diff
@@ -237,11 +237,13 @@ func (s *service) UserNotifications(_ *empty.Empty, stream rpc.Connector_UserNot
 
 func (s *service) Login(ctx context.Context, _ *empty.Empty) (*rpc.LoginResult, error) {
 	ctx = s.callCtx(ctx, "Login")
-	resultCode, err := auth.EnsureLoggedIn(ctx, s.loginExecutor)
-	if err != nil {
+	if _, err := s.loginExecutor.GetToken(ctx); err == nil {
```
donnyyung, line 240: Ah okay I was confused about where the refresh was actually happening but found it happens here: https://github.com/telepresenceio/telepresence/blob/release/v2/pkg/client/connector/auth/refresh.go#L52 

and this PR makes it so we actually check the token + refresh if necessary instead of just checking for the presence of `tokens.json`

donnyyung (approved):

LukeShu:
Doing smoke tests with a build from this commit, It works up until kubeception gives me a bum ambassador.ambassador load balancer.

Issue #1697, opened 2021-04-30
[v2] intercept not woking with --namespace


xavidop:
Hello,

I have notices that when we have the same helm chart deployed in multiple namespaces when I try to intercept a service of one namespace using the --namespace, the intercept works but the env variables inject is from the same service/deployment but from a different namespace.

I tried with the --namespace and --mapped-namespaces with the same result. The services from the other namespaces have installed the traffic-agent but are not intercepted.

I am using telepresence 2.2.1 and running telepresence on EKS.

Maybe it is not only the env vars, it is injecting the same workload but from another namespace.

Regards,
Xavi

xavidop:
Adding more information to this issue, when I try to uninstall the agent in one service that has that strange behaviour I get this error:
```bash
Launching Telepresence Daemon v2.2.1 (api v3)
Connecting to traffic manager...
Connected to context arn:aws:eks:us-east-1:168387678261:cluster/vf-dev-0 (https://F9A0562D4733620902567FBC85ED2845.gr7.us-east-1.eks.amazonaws.com)
Disconnecting...done
Telepresence Daemon quitting...done
telepresence: error: Deployment name="creator-app" namespace="dev-xavi": unable to locate port "tm-app" in container "app"
```

I just run `telepresence uninstall -d <workload> -n <namespace>`. It seems like the traffic agents are in bad state.

frankgu968:
Should link the pull request: https://github.com/telepresenceio/telepresence/pull/1700

khussey:
This is fixed in Telepresence 2.2.2, which is now available.

PR #1700, opened 2021-05-01
fix: multiple agents running getting wrong agent when intercepts

xavidop:
---

## Helpful information

Regarding the issue https://github.com/telepresenceio/telepresence/issues/1697 there is a problem when trying to intercept a workload. That workload exists in multiple namespaces. When a workload is intercepted and there is another interception done, the new interceptions gets the env vars in a wrong way, getting the env vars from other namespace.


thallgren (approved):
Good catch. Thanks!

PR #1702, opened 2021-05-03
Improve smoke tests

donnyyung:
This PR removes edge stack from the smoke tests which makes it faster,
as well as doesn't make us depend on a loadbalancer type service to
smoke test telepresence.  Additionally, this matches what is used in our
quick start, so all-in-all a much better smoke test

---
## Checklist

 - [x] I made sure to update `./CHANGELOG.md`. - just a smoke test change so no need to add to changelog.


LukeShu (approved):
Woo!

PR #1704, opened 2021-05-03
stalebot: Don't close issues or PRs, just mark them stale

LukeShu:
I was looking for an outside-contributor PR, and none existed.  That didn't seem right!

Having stalebot close issues and PRs is just a jerk move.  Have it flag them stale, but don't close them.


donnyyung (approved):
LGTM! We may want to figure out some policy for closing issues that we aren't / can't work on if the queue gets too long but hopefully we can keep up with the issues and never encounter that problem 😬

PR #1714, opened 2021-05-07
Translate tp1 commands to tp2

donnyyung:
Here's my PR for translating tp1 to tp2 commands. It's not exhaustive, but it tries to capture most of the commands (--swap-deployment, --run, --docker-run), as well as the various flags that can be used with those commands.  

---
## Checklist

 - [x] I made sure to either submit a docs PR, or tell Matt about the necessary documentation changes. - No docs PR for this needed. I do want to make a doc of tp1 -> tp2 but that's a separate work item

### DIFF
--- a/pkg/client/cli/command.go
+++ b/pkg/client/cli/command.go
@@ -97,7 +97,23 @@ func OnlySubcommands(cmd *cobra.Command, args []string) error {
 // user typos a command and types something invalid.
 func RunSubcommands(cmd *cobra.Command, args []string) error {
 	cmd.SetOut(cmd.ErrOrStderr())
-	cmd.HelpFunc()(cmd, args)
+
+	// determine if --help was explicitly asked for
+	var usedHelpFlag bool
+	for _, arg := range args {
+		if arg == "--help" {
+			usedHelpFlag = true
+		}
+	}
+	// If there are no args or --help was used, then it's not a legacy
+	// Telepresence command so we return the help text
+	if len(args) == 0 || usedHelpFlag {
+		cmd.HelpFunc()(cmd, args)
+		return nil
+	}
+	if err := checkLegacyCmd(cmd, args); err != nil {
+		return err
+	}
 	return nil
 }
 
@@ -114,7 +130,6 @@ func Command(ctx context.Context) *cobra.Command {
 		Short:              myName,
 		Long:               help,
 		RunE:               RunSubcommands,
-		PostRunE:           checkLegacyCmd,
 		SilenceErrors:      true, // main() will handle it after .ExecuteContext() returns
 		SilenceUsage:       true, // our FlagErrorFunc will handle it
 		DisableFlagParsing: true, // Bc of the legacyCommand parsing, see legacy_command.go
--- a/pkg/client/cli/command.go
+++ b/pkg/client/cli/command.go
@@ -3,7 +3,6 @@ package cli
 import (
 	"context"
 	"fmt"
-	"os"
 	"runtime"
 	"strings"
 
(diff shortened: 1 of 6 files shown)


pkg/client/cli/command.go
```diff
@@ -96,34 +107,41 @@ func Command(ctx context.Context) *cobra.Command {
...
 
-		Short:         myName,
-		Long:          help,
-		RunE:          RunSubcommands,
-		SilenceErrors: true, // main() will handle it after .ExecuteContext() returns
-		SilenceUsage:  true, // our FlagErrorFunc will handle it
+		Short:              myName,
+		Long:               help,
+		RunE:               RunSubcommands,
+		PostRunE:           checkLegacyCmd,
```
donnyyung: I added it as a PostRun so that this logic didn't get gross (in case we want to remove this at some point in the future).  But that being said, when users use telepresence 1 commands, the help text will be printed to the terminal (from RunSubCommands) and then my command will run. 

Example output:
```
...
  other Telepresence flags:
      --no-report   turn off anonymous crash reports and log submission on failure

Use "telepresence [command] --help" for more information about a command.

You used a telepresence 1 command that roughly translates to the following:
telepresence intercept echo-easy --port 9090 -- sleep 1
running...
Launching Telepresence Daemon v2.2.2-14-g56008474-1620355231 (api v3)
Connecting to traffic manager...
...
```
I considered this to not be a negative since I think while we want tp1 commands to work, we do want people to eventually migrate to using tp2 commands so the help text will maybe push them to explore.  But if you think it's messy, I can probably merge it into the RunSubcommands function.

pkg/client/cli/legacy_command.go
```diff
@@ -218,12 +218,15 @@ func checkLegacyCmd(cmd *cobra.Command, args []string) error {
...
 	_ = scout.Report(cmd.Context(), "Used legacy syntax")
 
-	if msg != "" {
-		fmt.Fprintln(cmd.OutOrStderr(), msg)
-	}
-
+	// Separate the help output from the tp1 -> tp2 translation
 	if tp2Cmd != "" {
-		fmt.Fprintf(cmd.OutOrStderr(), "\nYou used a telepresence 1 command that roughly translates to the following:\ntelepresence %s\n", tp2Cmd)
+		fmt.Fprintf(cmd.OutOrStderr(), "\nLegacy telepresence command used\n")
```
thallgren: This is really cool. Especially this part:
```
Legacy telepresence command used
Command roughly translates to the following in Telepresence 2:
telepresence intercept echo-easy --port 9090 -- python3 -m http.server 9090
running...
```
but does it have to be prefixed by the whole usage-blob? In my (very personal opinion), that makes the message less obvious.

pkg/client/cli/legacy_command_test.go
```diff
@@ -24,13 +24,13 @@ func Test_legacyCommands(t *testing.T) {
 			name:               "swapDeploymentMethod",
 			inputLegacyCommand: "telepresence --swap-deployment myserver --method inject-tcp --expose 9090 --run python3 -m http.server 9090",
 			outputTP2Command:   "intercept myserver --port 9090 -- python3 -m http.server 9090",
-			msg:                "Telepresence 2 doesn't have methods. You can use --docker-run for container, otherwise tp2 works similarly to vpn-tcp",
+			msg:                "Telepresence 2 doesn't have methods. You can use --docker-run for container, otherwise it works similarly to vpn-tcp\n",
```
thallgren: Suggest using the term "proxying methods" instead of just "methods" since that term is very generic and somewhat close to "mechanisms".

thallgren (approved):
Looks great! I made a couple of suggestions. Feel free to ignore.

pkg/client/cli/legacy_command.go
donnyyung: Yeah I'm going to take the help text out. I thought it was _maybe_ useful for pointing people at the help info so they could learn the new syntax w/o crutching on the legacy Telepresence commands for too long. 

But a few people have mentioned not liking it and it obscures the actual message we want. I might add to the text above something like: "For more information about the new commands, use telepresence --help" to point people to the help text to learn more. What do you think about that? Or do you think `--help` is obvious enough that we don't need to draw attention to it. Kenn also suggested I ask Brian his opinion so I'll do that in parallel and add a comment with what he suggested.

--- edit ---
Brian agreed with this sentiment so I have removed the help text from showing when running a legacy Telepresence command.

PR #1735, opened 2021-05-18
Add Helm Chart for Server Components

iNoahNothing:
Create a helm chart for the traffic-manager and other server-side components.

To-do:

- [x] Update the NOTES.txt
- [x] Update Chart.yaml
- [ ] Add CHANGELOG
- [ ] Add README

### DIFF
--- a/charts/telepresence/Chart.yaml
+++ b/charts/telepresence/Chart.yaml
@@ -1,23 +1,11 @@
 apiVersion: v2
 name: telepresence
-description: A Helm chart for Kubernetes
-
-# A chart can be either an 'application' or a 'library' chart.
-#
-# Application charts are a collection of templates that can be packaged into versioned archives
-# to be deployed.
-#
-# Library charts provide useful utilities or functions for the chart developer. They're included as
-# a dependency of application charts to inject those utilities and functions into the rendering
-# pipeline. Library charts do not define any templates and therefore cannot be deployed.
+description: A chart for deploying the server-side components of Telepresence
 type: application
-
-# This is the chart version. This version number should be incremented each time you make changes
-# to the chart and its templates, including the app version.
-# Versions are expected to follow Semantic Versioning (https://semver.org/)
 version: 0.1.0
 
-# This is the version number of the application being deployed. This version number should be
-# incremented each time you make changes to the application. Versions are not expected to
-# follow Semantic Versioning. They should reflect the version the application is using.
+# Note: This is the version of the Traffic Manager that will be installed by 
+# this chart. The telepresence CLI will always attempt to update the Traffic 
+# Manager if it is not the same version as the CLI so ensure you are keeping
+# these in sync.
 appVersion: 2.2.2
--- a/charts/telepresence/Chart.yaml
+++ b/charts/telepresence/Chart.yaml
@@ -1,0 +1,23 @@
+apiVersion: v2
+name: telepresence
+description: A Helm chart for Kubernetes
+
+# A chart can be either an 'application' or a 'library' chart.
+#
+# Application charts are a collection of templates that can be packaged into versioned archives
+# to be deployed.
+#
+# Library charts provide useful utilities or functions for the chart developer. They're included as
+# a dependency of application charts to inject those utilities and functions into the rendering
+# pipeline. Library charts do not define any templates and therefore cannot be deployed.
+type: application
+
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.1.0
+
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
+appVersion: 2.2.2
--- a/charts/telepresence/templates/deployment.yaml
+++ b/charts/telepresence/templates/deployment.yaml
@@ -39,6 +39,12 @@ spec:
           resources:
             {{- toYaml . | nindent 12 }}
           {{- end }}
+          {{- if .Values.licenseKey.create }}
+          volumeMounts:
+          - name: license
+            mountPath: /home/telepresence
+            readOnly: true
+          {{- end }}
       {{- with .Values.nodeSelector }}
       nodeSelector:
         {{- toYaml . | nindent 8 }}
@@ -50,4 +56,11 @@ spec:
       {{- with .Values.tolerations }}
       tolerations:
         {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- if .Values.licenseKey.create }}
+      volumes:
+      - name: license
+        secret: 
+          defaultMode: 420
+          secretName: {{ .Values.licenseKey.secret.name }}
       {{- end }}
--- a/charts/telepresence/templates/deployment.yaml
+++ b/charts/telepresence/templates/deployment.yaml
@@ -2,6 +2,7 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: {{ include "telepresence.fullname" . }}
+  namespace: {{ include "telepresence.namespace" . }}
   labels:
     {{- include "telepresence.labels" . | nindent 4 }}
 spec:
--- a/charts/telepresence/templates/deployment.yaml
+++ b/charts/telepresence/templates/deployment.yaml
@@ -1,0 +1,53 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ include "telepresence.fullname" . }}
+  labels:
+    {{- include "telepresence.labels" . | nindent 4 }}
+spec:
+  replicas: {{ .Values.replicaCount }}
+  selector:
+    matchLabels:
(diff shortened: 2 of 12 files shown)


charts/telepresence/values.yaml
```diff
@@ -0,0 +1,45 @@
+# The Traffic Manager only support running with one replica at the moment.
+# Configuring the replicaCount will be added in future versions of Telepresence
+# replicaCount: 1
+
+image:
+  repository: docker.io/datawire/tel2
+  pullPolicy: IfNotPresent
+  # Overrides the image tag whose default is the chart appVersion.
+  tag: ""
```
donnyyung, line 21: This isn't so much a problem with the chart, but if a user has a different version of the cli, it will try to update the image to the version of their cli. I think we have an issue to improve behavior around this, but just might be worth something to tell people who are using the chart, that it's important that users have the correct version of the CLI + have their `TELEPRESENCE_REGISTRY=image.repository` if they change that value.

charts/telepresence/templates/deployment.yaml
```diff
@@ -0,0 +1,53 @@
...
+        {{- toYaml . | nindent 8 }}
+    {{- end }}
+      labels:
+        {{- include "telepresence.selectorLabels" . | nindent 8 }}
+    spec:
+      {{- with .Values.image.imagePullSecrets }}
+      imagePullSecrets:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      securityContext:
```
donnyyung, line 26: When users are in an air-gapped environment, (thus needing a license) they use `telepresence license -f <file>` it outputs a secret like so: 
```
apiVersion: v1
data:
  hostDomain: <auth.datawire.io (encoded in base64)>
  license: <base64 encoded jwt>
kind: Secret
metadata:
  creationTimestamp: null
  name: systema-license
  namespace: ambassador
  ```
  The only value above that will change for users is the license, since the hostDomain only other suitable value is for staging.
  
  In the yaml when the systema-license secret is found in the ambassador namespace, we add the following:
  ```
          volumes:
        - name: license
          secret:
            defaultMode: 420
            secretName: systema-license
```

and:
```
          volumeMounts:
          - mountPath: /home/telepresence/
            name: license
            readOnly: true
```
The name of the secret can probably be changed if the user wants but the `mountPath` + `name` in the volume mount are explicitly used in the traffic-manager so they need to remain those values.

let me know if any of this doesn't make sense.. more info on the license workflow can be found at the bottom of this page: https://www.getambassador.io/docs/telepresence/latest/reference/cluster-config/

charts/telepresence/Chart.yaml
```diff
@@ -0,0 +1,23 @@
...
+
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.1.0
+
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
+appVersion: 2.2.2
```
donnyyung, line 11: In my mind, the telepresence binary installing the traffic-agent is still our preferred way of having the traffic-manager installed.  So I'm wondering who this chart is for and what is our responsibility with this chart? 

e.g. Do we have a helm repo we are going to be pushing this chart to (it's been a while since I've used helm so if helm charts no longer exist, apologies for my ignorance)? 
Do we need to bump it whenever we have a new app version or should we just make changes when we change the yaml? 
Is it possible to only bump the appVersion and publish a new chart? or do we need to bump the chart whenever we bump the appVersion?  If the latter, would it make sense to version them together since the cli + traffic-manager are expected to be the same version?
Should we be adding linting this to our CI?

Not opposed just want to make sure we have a shared understanding of the purpose of this chart and what our responsibilities are.

donnyyung: Discussed with Kenn so have answers around the responsibility (e.g. we should own the chart)

charts/telepresence/values.yaml
donnyyung: I think the answer is to document this for now, and very soon we'd like to make the go code use the helm chart for deploying the traffic-manager or have the go code generate the helm chart (but that's work for our team, and probably me) so for now I'd just make sure we have that documented wherever we add docs about the helm chart.

donnyyung (changes requested):
Looks like a good first pass. Just left some questions to ensure we are on the same page. Thanks for this!

charts/telepresence/templates/deployment.yaml
iNoahNothing: Hmmm okay so it sounds like you are suggesting that we create this `volumeMount` by default in the chart and have the user add the license later if they need to manually.

Is that correct?

charts/telepresence/Chart.yaml
iNoahNothing: This chart is for people who do not wish to/cannot installing software into their cluster that they do not control the YAML for. This can be a common need for users whose environments are locked down by a centralized Ops team and people outside of that Ops team has not access to create cluster-side resources like this. 

This helm chart just creates a way to have an ops team deploy the server-side components using the same mechanisms they do for other things in their environment.

charts/telepresence/templates/deployment.yaml
```diff
@@ -0,0 +1,66 @@
+apiVersion: apps/v1
```
donnyyung, line 1: I'm sorry this wasn't obvious from the files we had in k8s/*.yaml and I should have caught it on the first review, but could we have these default to the `ambassador` namespace for all the things deployed to the cluster? Currently, the cli requires the service + deployment be in the ambassador namespace or else it will install another traffic-manager there.

charts/telepresence/templates/licenseSecret.yaml
```diff
@@ -0,0 +1,11 @@
+{{- if and .Values.licenseKey.create .Values.licenseKey.secret.create }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ .Values.licenseKey.secret.name }}
+  labels:
+    {{- include "telepresence.labels" . | nindent 4 }}
+data:
+  hostDomain: {{ printf "auth.datawire.io" | b64enc | quote }}
+  license: {{ .Values.licenseKey.value | quote | b64enc | quote }}
```
donnyyung: I think what you want is `{{ .Values.licenseKey.value | b64enc | quote }}`. I compared a helm template with the secret that telepresence generated and it wasn't giving me the correct value bc it was adding the quotes to the secret.

charts/telepresence/templates/deployment.yaml
```diff
@@ -0,0 +1,66 @@
...
+    spec:
+      {{- with .Values.image.imagePullSecrets }}
+      imagePullSecrets:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      securityContext:
+        {{- toYaml .Values.podSecurityContext | nindent 8 }}
+      containers:
+        - name: {{ .Chart.Name }}
+          securityContext:
```
donnyyung, line 30: We also have the following environment variables: 
```
        - name: LOG_LEVEL
          value: debug
        - name: SYSTEMA_HOST
          value: app.getambassador.io
        - name: SYSTEMA_PORT
          value: "443"
        - name: CLUSTER_ID
          value: <uid from default namespace>
```
LOG_LEVEL we should definitely add (use debug as the default), the middle ones are already the default so that's not a huge deal.  The last one you can get from `telepresence current-cluster-id` or `kubectl get ns default -o yaml` which I think we should put in the docs for now.. I want to deprecate it as an env var once we move the installer into the traffic-manager (and the traffic-manager can actually talk to the k8s cluster and get this info itself).. but for now I say we add it to the chart and add a comment above the value about how the user can get it.

charts/telepresence/templates/deployment.yaml
iNoahNothing: So setting the `metadata.namespace` would be very anti to the typical helm patterns.

Instead I just created a check that makes the install fail if you try to use any other namespace 🙂

charts/telepresence/templates/licenseSecret.yaml
iNoahNothing: Good catch. I fixed the issue that caused me to need that by just adding the "" to the values file

charts/telepresence/templates/deployment.yaml
iNoahNothing: Thanks! I did not see these in `k8s/manager.yaml` so I omitted them.

I have added the `--set clusterID=$(kubectl get ns default -o jsonpath='{.metadata.uid'})` to the README

charts/telepresence/templates/tests/test-connection.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v1
```
donnyyung, line 1: I think this also needs to be `ambassador` namespace (probably the easiest solution) or change the args to be `traffic-manager.ambassador` so that it's always finds the traffic-manager in the right namespace, regardless of whichever namespace it's deployed to.  But I think i'd go with deploying it in the ambassador namespace just for ease.

donnyyung (approved):

PR #1743, opened 2021-05-21
Adding helm chart to CI

donnyyung:
This PR does a few things:

- Makes the helm chart lintable
- Adds that lint to our CI process
- Keeps the helm chart in sync with the latest GA version of telepresence.
- Adds a script for publishing the helm chart
- Creates CI for publishing helm chart versions.

Here are the proposed changes to the release process, additions are in green: https://www.notion.so/datawire/Proposal-with-Helm-Chart-To-Release-Telepresence-2-x-x-130753a3fa7a4691bee3d47b767ff64e

I have ran the script for publishing to the `charts-dev` path without trouble, so I believe I have tested it to the best of my ability. Obviously if the first production release does not work, I will debug and get it fixed. 

## Checklist

 - [x] I made sure to either submit a docs PR, or tell Matt about the necessary documentation changes. -- todo


build/go.mk
```diff
@@ -107,6 +107,10 @@ lint: lint-deps ## (Lint) Run the linters (golangci-lint and protolint)
 	$(tools/golangci-lint) run --timeout 2m ./...
 	$(tools/protolint) lint rpc
 
+.PHONY: lint-chart
+lint-chart:
+	helm lint charts/telepresence --set isCI=true
+
```
LukeShu: Is there a good reason this isn't part of the main `make lint`?

.circleci/config.yml
```diff
@@ -160,6 +177,18 @@ jobs:
...
+    executor: vm-linux
+    steps:
+    - checkout
+    - install-helm
+    - run: make lint-chart
+    # For now, I don't think we need to build rc charts (nor do I think
+    # we have a need for rc charts), but explicitly calling them out here
+    # in case we change our mind
+    - run: |
+        [[ $CIRCLE_TAG == "*rc*" ]] || BUCKET_DIR="charts" packaging/push_chart.sh
```
LukeShu: 1. Drop the quotes around `"*rc*"`; because of the double quotes, it'll match `*rc*` literally rather than as a glob.  If this is surprising to you, then it may be helpful to know that despite the visual similarity to the `[` command, the `[[ … ]]` construct is a special piece of Bash syntax, rather than a regular command.
2. I'd be more comfortable excluding _all_ prerelease tags, by having it by `[[ $CIRCLE_TAG == *-* ]]`.

LukeShu: Actually: See my next comment below.

.circleci/config.yml
```diff
@@ -191,6 +220,13 @@ workflows:
               only: /^v.*/
             branches:
               ignore: /.*/
+      - release-chart:
+          context: telepresence2-release
+          filters:
+            tags:
+              only: /^chart-v.*/
```
LukeShu: 1. Do we expect to be doing chart releases that don't coincide with a Telepresence release?  IMO, this would be a good opportunity to version them in lockstep.
2. It would be good to just modify the regex here instead of having the `[[ $CIRCLE_TAG == … ]]` check in the job.  I believe `/^chart-v(.(?!rc))*$/` would do what you want (the Circle [docs](https://circleci.com/docs/2.0/workflows/#using-regular-expressions-to-filter-tags-and-branches) say that it uses [`java.util.regex` syntax](https://docs.oracle.com/javase/7/docs/api/java/util/regex/Pattern.html)).
3. So using my suggestion of excluding all prerelease tags, the regex would be `/^chart-v[^-]*$/`

build/go.mk
donnyyung: Not sure if a _good_ reason, but in case somebody wanted to _only_ lint the chart or _only_ do the go lint, having them separate would be easier... but the commands themselves aren't that complicated so maybe it's cleaner to combine them and then someone can look the commands up if they only want to do one or the other.

.circleci/config.yml
donnyyung: 1. So in an ideal world, no. and I had originally planned to keep the versions in-sync. But after chatting with people on edge stack, they noted the possibility that sometimes you might need to only change the chart (e.g. an error in the yaml when a user uses some value, which isn't used in our smoke tests).  Which maybe for edge stack that is a problem since they release less frequently, but we could do another release shortly after the problem was discovered "x.y.z+1" with a changelog that's like, "Bugfix: Fixes issue with helm chart".  So if we aren't opposed to doing something like that, then I think for our own sanity it makes sense to keep the versions in lockstep.

2/3. Good idea, will do if ^ isn't a good workflow and we can't keep them in lockstep.

LukeShu: 1. Given the goal of keeping telepresence releases light, I have no problem "Bugfix: Helm chart" releases.  So in my mind, that means not having separate `chart-v*` tags.

.circleci/config.yml
LukeShu: The `[[ … ]]` check is still broken.

.circleci/config.yml
donnyyung: Sounds good, I will remove the `chart-v*` and fold helm chart publish into our already existing release steps then.

.circleci/config.yml
```diff
@@ -160,6 +176,18 @@ jobs:
       - run: |
           [[ $CIRCLE_TAG == *-* ]] || TELEPRESENCE_VERSION=$CIRCLE_TAG make promote-to-stable
 
+  "release-chart":
+    executor: vm-linux
+    steps:
+    - checkout
+    - install-helm
+    - run: make lint
```
LukeShu: Why have a `make lint` here? We don't do any other checks in the release jobs.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ### 2.3.3 (TBD)
 
 - Feature: `telepresence intercept` now supports a `--to-pod` flag that can be used to port-forward sidecars' ports from an intercepted pod
+- Feature: The helm chart is now published as part of our CI.
```
LukeShu: Nit: Capitalize "Helm"

charts/telepresence/README.md
```diff
@@ -27,30 +27,45 @@ The following tables lists the configurable parameters of the Ambassador chart a
...
+| image.name         | The name of the image to use for the traffic-manager                                                                      | `tel2`                                                                                            |
 | image.pullPolicy         | How the `Pod` will attempt to pull the image.                                                                           | `IfNotPresent`                                                                                    |
 | image.tag                | Override the version of the Traffic Manager to be installed.                                                            | `""` (Defined in `appVersion` Chart.yaml)                                                         |
 | image.imagePullSecrets   | The `Secret` storing any credentials needed to access the image in a private registry.                                  | `[]`                                                                                              |
+| ports            | The ports the Traffic Manager `Deployment` will expose. **Do not change.**                               | `[{"name":"api","port":8081},{"name": "https", "port": 8443]` |
```
LukeShu: What's the point in having a setting if we tell people not to set it?

packaging/push_chart.sh
```diff
@@ -0,0 +1,76 @@
...
+echo "CHART_PACKAGE is here: ${CHART_PACKAGE}"
+
+bucket_dir=
+if [[ -n "${BUCKET_DIR}" ]]; then
+    bucket_dir="${BUCKET_DIR}"
+else
+    bucket_dir="charts-dev"
+fi
+
+if [ -z "$AWS_BUCKET"] ; then
```
LukeShu: You definitely need a space between the `"` and the `]`.

On that note, running `shellcheck` on our shell scripts from `make lint` would be a good enhancement.

smoke-tests/run_smoke_test.sh
```diff
@@ -437,6 +471,13 @@ finish_step
 #### Step 12 - licensed uninstall everything          ####
 ##########################################################
 
+# First we uninstall the helm chart if it was used
+if [ ! -z $USE_CHART ]; then
```
LukeShu: Either use double `[[ brackets ]]` or put `"quotes"` around the variable.

.circleci/config.yml
```diff
@@ -59,6 +59,21 @@ commands:
...
+    steps:
+      - run:
+          name: "Install Helm << parameters.version >>"
+          command: |
+            set -x
+            curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/master/scripts/get-helm-3
+            chmod 700 get_helm.sh
+            ./get_helm.sh --version << parameters.version >>
+            rm ./get_helm.sh
+
```
LukeShu: Since we're relying on `helm lint` in `make lint`, which is something we expect developers to run, I think I'd prefer to have this in `tools.mk`

.circleci/config.yml
donnyyung: I was being too defensive since I wanted to ensure the chart linted prior to pushing, but since we ensure lint on merge to master, then there's really no _real_ reason to lint again. So I'll remove it.

charts/telepresence/README.md
donnyyung: I saw that this is what we did with service.ports, but I think it doesn't really make sense / is kinda confusing, so removing the "do not changes" I added, as well as the pre-existing ones and will have it hard-coded in the yaml.

LukeShu (approved):

PR #1744, opened 2021-05-22
Let sshfs allow root on volume mounts

thallgren:
## Description

Docker runs as root so root must be allowed to access the telepresence
remote mount that sshfs creates.

## Checklist


donnyyung (approved):
LGTM

PR #1747, opened 2021-05-24
Remove the need for extra connections between agent and manager.

thallgren:
## Description

Replaces the forwarding of intercepted connections from agent to the
manager with a ConnTunnel so that no extra connections have to be made
to ports that aren't exposed in the manager's deployment manifest.

Also ensures that all outbound traffic from the client to the
traffic-manager is routed to intercepted agents so that those requests
originates from a pod in the currently intercepted namespace.

A welcome side-effect is that the agent no longer needs to export ports
externally in order to serve remote VolumeMounts since those connections
too originates in the agent.

## Checklist


LukeShu:
One day I will find out the reason why v4 ICMP and v6 ICMP have different numbers.

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+### 2.3.0 TBD
```
LukeShu: 2.3.1

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
...
+
+- Change: Outbound connections are now routed to intercepted agents.
+- Change: Inbound connection from intercepted agent now uses existing gRPC instead of new connection to manager port between 6000-7999
```
LukeShu: This sentence reads super weird to me.  Perhaps:

> - Change: Inbound connections from an intercepted agent are now tunneled to the manager over the existing gRPC connection, instead of establishing a new connection to the manager for each inbound connection.

LukeShu:
If I were the one doing this, I'd squash the last 3 commits together.

cmd/traffic/cmd/agent/forwarder.go
```diff
@@ -91,34 +116,74 @@ func (f *Forwarder) Target() (string, int32) {
...
 	f.mu.Unlock()
+	if tunnel != nil {
+		return f.interceptConn(ctx, clientConn, intercept, tunnel)
+	}
 
 	targetAddr, err := net.ResolveTCPAddr("tcp", fmt.Sprintf("%s:%d", targetHost, targetPort))
 	if err != nil {
-		dlog.Infof(ctx, "Error on resolve(%s:%d): %+v", targetHost, targetPort, err)
-		return
+		return fmt.Errorf("error on resolve(%s:%d): %+v", targetHost, targetPort, err)
```
LukeShu: Use `%w` when wrapping errors.

cmd/traffic/cmd/agent/forwarder.go
```diff
@@ -131,8 +196,7 @@ func (f *Forwarder) ForwardConn(clientConn *net.TCPConn) {
 
 	targetConn, err := net.DialTCP("tcp", nil, targetAddr)
 	if err != nil {
-		dlog.Infof(ctx, "Error on dial: %+v", err)
-		return
+		return fmt.Errorf("error on dial: %+v", err)
```
LukeShu: Likewise.

cmd/traffic/cmd/agent/forwarder.go
```diff
@@ -157,8 +221,83 @@ func (f *Forwarder) ForwardConn(clientConn *net.TCPConn) {
 	for i := 0; i < 2; i++ {
 		select {
 		case <-ctx.Done():
-			return
 		case <-done:
 		}
 	}
```
LukeShu: I almost posted a comment about how the original code was wrong, but that was me mis-interpretting the code.

The original code would have (IMO) been more clearly written as
```go
	// Wait for both sides to close the connection
	for numClosed := 0; numClosed < 2; {
		select {
		case <-ctx.Done():
			return
		case <-done:
			numClosed++
		}
	}
```

That is:
 1. In the "happy" path, we don't want to terminate until `<-done` has been read twice, or
 2. In the "sad" path, we want to bail early if `ctx.Done()` gets closed.
 
With that in mind, I'm not sure how it's correct to remove the bail-early `return`.

But I also don't see how it's correct to orphan the goroutines as it did when bailing early.  It seems to me that the reason `done` is a channel at all (instead of a `sync.WaitGroup`) is in order to support context cancellation.

LukeShu:
I get the feeling that we're acquiring a lot of duplicate code around pumping forwarded connections (manager/systema, manager/agent, agent/app...). I might be adding "deduplicate all that" to my cooldown list.

LukeShu:
I guess part of that is that in my mind the connpool thing works that way, but really it doesn't, since it's handling multiplexing.  Which is a little silly, because the HTTP/2 stream that gRPC is using can already handle multiplexing.

LukeShu:
Minor concern/question about cancellation in the agent.  A couple of nitpicks.  Otherwise, looks good.

LukeShu (approved):
Hit it!

PR #1757, opened 2021-05-27
Avoid calling withConnector twice when logging in.

thallgren:
## Description

Avoid calling withConnector twice when logging in.

## Checklist

### DIFF
--- a/pkg/client/cli/intercept_state.go
+++ b/pkg/client/cli/intercept_state.go
@@ -177,7 +177,7 @@ func (ii *interceptInfo) intercept(cmd *cobra.Command, args []string) error {
 				canConnect = resp.CanConnect
 			}
 			if canConnect {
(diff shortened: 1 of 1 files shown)


donnyyung (approved):
Nice catches, sorry I missed this the first time.

LukeShu (changes requested):
The `cliutil` stuff has the advantage over the `connector_state.go` stuff in that has a pipe for the connector to show messages to the user.  This is important for the login flow, which might want to print messages on the terminal.  For the login flow is actually the forcing reason that I implemented that functionality.  If you're going not use `cliutil` for login, you'll need to call `connector.UserNotifications()` yourself.

That said, I think it should be fine to have `withConnector()` and `cliutil.WithConnector()` both called, so I'm not sure the PR is necessary.

PR #1758, opened 2021-05-27
[v2] Fix a number of race conditions

LukeShu:
## Description

1. The connector's `tm.managerErr` and `tm.managerClient` was set and read from multiple goroutines without proper synchronization.
2. Fix a race condition with the daemon socket existing but having the wrong permissions.
3. (this is the important one) Fix a race condition in the connector with it having a stale dead gRPC connection to the daemon when started without the daemon.

I don't believe that any of these are regressions in 2.3.0, but I absolutely can believe that aspects of timing changed in 2.3.0 such that race condition 3 is now more likely to trigger.  Based on the analysis (below) I don't believe that users are any less likely to trigger this than our smoke tests are.

Other changes made along the way:

- Include more context around errors in `connector.log`
- Slightly improve startup logging in `daemon.log`
- Have the local log files include sub-second precision in their timestamps
- Upgrade a number of dependencies, allowing the tests to once again run on my laptop (cgroups v1 was already "legacy" 6 years ago, Arch Linux just recently finally turned it off, and surprise, dtest doesn't work without it)
- Avoid using the global `logrus.StandardLogger()`

----

Discussion of race condition 3:

At startup, the connector `grpc.Dial`s the daemon.  `grpc.Dial` is non-blocking (by default), and will retry connections.  This both means that connection errors are hidden until much later, and meant that it was OK if the daemon wasn't yet running, because it it would retry.

Well, there are several ways that the connector can be started without starting the daemon, `telepresence login` being the simplest.  So you'd `telepresence login` the connector would start up, try to hit the daemon, and set the connection error to "file does not exist", and retry periodically.  Then later you'd run a command that starts the daemon, waits for it to start up, then calls `connector.Connect`.  That `connector.Connect` call asks the connector to talk to the daemon.  Whether that call from the connector to the daemon errors depends on whether the connector happened to retry the connection in the time between the daemon becoming ready and the connector's `initGrpc` trying to talk to the daemon.

So, pass the flag to `grpc.Dial` to have it block, and don't call `grpc.Dial` until we need it.

## Checklist

 - [x] My change is adequately tested. - I've been running the reproducer in a loop, and smoke tests should handle everything else
 - [x] I updated `DEVELOPING.md` with any any special dev tricks I had to use to work on this code efficiently. - no tricks, I don't think


pkg/client/connector/traffic_manager.go
```diff
@@ -178,17 +188,19 @@ func (tm *trafficManager) initGrpc(c context.Context, portIf interface{}) (err e
...
 	if err != nil {
+		tm.managerClient = nil
 		return fmt.Errorf("getOutboundInfo: %w", err)
 	}
 	if _, err := tm.daemon.SetOutboundInfo(c, outboundInfo); err != nil {
+		tm.managerClient = nil
 		return fmt.Errorf("daemon.SetOutboundInfo: %w", err)
 	}
 
+	close(tm.startup)
```
donnyyung, line 199: This is mostly for my own edification bc I still find channels a little hazy at times... So could you just check that I understand correctly? 

All the `<- tm.startup`s you added are blocking until they can pull something off the channel.  However, I don't see anything like ` tm.startup <- true`.  So closing the channel also tell the places where it is waiting to pull something off of the channel, that they can continue.  Is that right or am I missing something?

pkg/client/sockets.go
```diff
@@ -66,6 +66,10 @@ func SocketURL(socket string) string {
...
 
 // DialSocket dials the given unix socket and returns the resulting connection
-func DialSocket(c context.Context, socketName string) (*grpc.ClientConn, error) {
-	return grpc.DialContext(c, SocketURL(socketName), grpc.WithInsecure(), grpc.WithNoProxy())
+func DialSocket(ctx context.Context, socketName string) (*grpc.ClientConn, error) {
+	return grpc.DialContext(ctx, SocketURL(socketName),
+		grpc.WithInsecure(),
+		grpc.WithNoProxy(),
+		grpc.WithBlock(),
+	)
```
LukeShu, line 74: The only thing that actually changed in this block is the addition of `grpc.WithBlock()`.

pkg/client/connector/service.go
```diff
@@ -404,11 +403,26 @@ func (s *service) connectWorker(c context.Context, cr *rpc.ConnectRequest, k8sCo
...
 	}
 
+	// establish a connection to the daemon gRPC service
+	dlog.Info(c, "Connecting to daemon...")
+	conn, err := client.DialSocket(c, client.DaemonSocketName)
+	if err != nil {
+		dlog.Errorf(c, "unable to connect to daemon: %+v", err)
+		s.cancel()
+		return &rpc.ConnectInfo{
+			Error:     rpc.ConnectInfo_CLUSTER_FAILED,
```
donnyyung: Should this be the newly added `ConnectInfo_DAEMON_FAILED`?

donnyyung (approved):
I think there's one tiny thing that needs to be fixed but other than that it looks good to me!

pkg/client/connector/traffic_manager.go
LukeShu: Correct, the only "value" that will ever be read from `tm.startup` is EOF.

pkg/client/connector/service.go
LukeShu: Oops, yeah it should.

PR #1759, opened 2021-05-27
[v2] Make logging in user- and root daemon configurable.

thallgren:
## Description

This PR adds a new `logLevels` object to the Telepresence config file so that the level for the user- and root daemon can be configured. The default for the root daemon is now "info" to avoid network traffic logging. The PR also changes the default logging for the traffic-manager to info.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 - Change: The traffic-agent no longer has a sshd running. Remote volume mounts use sshfs in slave mode, talking directly to sftp.
 - Change: The local DNS now routes the name lookups to intercepted agents or traffic-manager.
 - Feature: Telepresence is now installable via brew
-- Change: The default log-level for the traffic-manager and the root-daemon was changed to "info".
+- Change: The default log-level for the traffic-manager and the root-daemon was changed from "debug' to "info".
 
 ### 2.2.2 (May 17, 2021)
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 - Change: The traffic-agent no longer has a sshd running. Remote volume mounts use sshfs in slave mode, talking directly to sftp.
 - Change: The local DNS now routes the name lookups to intercepted agents or traffic-manager.
 - Feature: Telepresence is now installable via brew
+- Change: The default log-level for the traffic-manager and the root-daemon was changed to "info".
```
LukeShu: Perhaps say "changed from 'XXX' to 'info'" to say what it was before?

LukeShu (approved):
The YAML parsing seems a little over-complicated to me (why not define a wrapper type around `logrus.LogLevel` that parses strings, and use `sigs.k8s.io/yaml`?), but lgtm.  And +1 for using the names userDaemon and rootDaemon :)

thallgren:
Regarding the parsing. The reason I like this approach is that when parsing a user-facing file, it's great that the parser is capable of telling the user exactly what's wrong and also the exact position in the parsed file. I'm familiar with the yaml parser used here and it's already in our dependency tree. I've never used `sigs.k8s.yaml`. I'll check it out.

PR #1761, opened 2021-05-28
Restore functionality of overriding DNS resolver

thallgren:
## Description

This PR contains a couple of fixes related to startup/shutdown and some improvements to the DNS resolver. The overriding DNS resolver in particular didn't work as expected when using `intercept --local-only` because it relied on that the search path was dealt with somewhere else (which is normally is, but not for that resolver).

This PR only affects things that hasn't been released yet and doesn't contain anything that deserves a mention in the changelog/release notes.

The PR should be tested with Linux + resolveconf in particular, since that's where most prominent fix is. I tested it using an ubuntu box with resolved disabled.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - Change: The default log-level for the traffic-manager and the root-daemon was changed from "debug" to "info".
 - Bugfix: Using --docker-run no longer fail to mount remote volumes when docker runs as root.
 - Bugfix: Fixed a number of race conditions.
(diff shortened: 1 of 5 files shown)


LukeShu:
Code looks good to my eyes, I'm testing it now...

LukeShu:
But FWIW, you have a typo in a commit message ("lookkups" has too many "k"s), and I'd squash the last 2 commits together.

LukeShu:
First run-through of the smoke tests went fine, I'll run it a few more times.

LukeShu (approved):
I successfully ran through the smoke tests 3 times with this, so I think it's good!

PR #1762, opened 2021-05-28
build-sys: Statically link the CLI

LukeShu:
## Description

We always did this with `teleproxy`, somehow we lost it along the way to Telepresence 2.  Bring it back!

Well, we only did it for GOOS=linux with `teleproxy`, and not for GOOS=darwin because it needed CGo for github.com/datawire/pf.  But we don't use that anymore now that we have the TUN device, so we can do it on GOOS=darwin now too!

## Checklist

 - [x] I made sure to update `./CHANGELOG.md`. - yep
 - [x] I made sure to either submit a docs PR, or tell Matt about the necessary documentation changes. - no docs nescessary
 - [x] My change is adequately tested. - I tested that it builds fine for GOOS=linux and GOOS=darwin


donnyyung (approved):
LGTM and also ran smoke tests on my mac to just double-check the new cli build worked.

PR #1764, opened 2021-05-31
Limit the DNS requests that result in cluster-side lookups.

thallgren:
## Description

This commit adds a set of very common top-level domains for which our
resolver will return not found (or revert to fallback, in case of the
overriding resolver).

The overriding resolver is also tweaked so that it no longer uses the
search path when the last name of a qualified name is equal to a
namespace or "svc".

The max time for the LookupHost is now hard-coded to 4 seconds.


pkg/client/daemon/outbound.go
```diff
@@ -184,9 +209,8 @@ func (o *outbound) resolveInCluster(c context.Context, qType uint16, query strin
 		}
 	}
 
-	// Give the cluster lookup a reasonable timeout
-	tos := client.GetConfig(c).Timeouts
-	c, cancel := tos.TimeoutContext(c, client.TimeoutTrafficManagerAPI)
+	// Give the cluster lookup a reasonable timeout.
+	c, cancel := context.WithTimeout(c, 4*time.Second)
```
donnyyung, line 213: What's the rationale for making this timeout hard-coded to 4 seconds? Is it just so we don't get hung if the cluster lookup fails and resort to the fallback?

thallgren: Yes, I think we want a shorter timeout here than the default that is used in resolv.conf (5 seconds). We should make this timeout configurable in the future, but I didn't want to start yet another cycle of release notes and docs just for that.

donnyyung (approved):
I think this LGTM. We'll want Luke to test it out on his machine since he was seeing the issues that required this PR, but I was able to run smoke tests on mac successfully, so it at least seems okay for mac os :)

LukeShu:
- I'm testing it now.
- But I don't see what actually fixed the issue?  d9a3763ef5582a4f8a604f2b065092b19bdf52e8 looks to me like a mitigation to not also bork non-cluster DNS, but cluster DNS was borked too, and I don't see anything in here fixing that.
- Regarding letting the DNS search path be handled by the agent containers: does this actually vary from pod-to-pod except for the namespace?  It seems to me that the code is increasingly complex as it tries to reason about other things handling the search path, and that it would simplify things to let it happen in fewer places.

thallgren:
I haven't found any problems with the cluster DNS. The only time I ran into problems in step 8 was when I was using the wrong proprietary agent (i forgot setting `TELEPRESENCE_AGENT_IMAGE` to the latest one).
The agent containers may have custom DNS resolution configured, so yes, it may vary.

LukeShu:
I'm having to log in (click things in my browser) during step 6.  I never had to log in during step 6 before?

LukeShu:
I've seen it pass twice.  I've seen it fail a bunch of different ways:

```
Intercept id was present and it shouldn't be. Failed step: 10
```

```
Step 5 success!

Using docker.io/datawire/ambassador-telepresence-agent:1.7.0-rc.0 as the agent for remainder of tests
telepresence: error: rpc error: code = Unknown desc = EOF
Proprietary traffic agent image wasn't used and it should be
```

```
Okay one more thing. Please login to System A in the window that pops up
telepresence: error: connector.Connect: the port-forward connection to the traffic manager timed out.  The current timeout 20s can be configured as "timeouts.trafficManagerConnect" in "/home/lukeshu/.config/telepresence/config.yml"
telepresence: error: connector.Connect: the port-forward connection to the traffic manager timed out.  The current timeout 20s can be configured as "timeouts.trafficManagerConnect" in "/home/lukeshu/.config/telepresence/config.yml"
Failed in step: 
> Output wasn't supposed to be empty and it was
```

```
Intercept id wasn't present and it should be. Failed step: 8
```

donnyyung:
I believe the first + last error Luke mentioned (around the interceptID being present or not when the opposite was expected, is the issue that we've been seeing and Jose root caused).

LukeShu:
I think the middle error corresponds with hiccups in the port-forward?

LukeShu:
I'm getting this about every 3rd run
```
telepresence: error: Intercept "b6991f80-54de-4414-b514-8409fcbde851:dataprocessingservice": Failed to establish intercept: intercept in error state NO_AGENT: No agent found for "dataprocessingservice"
Failed in step: 11
> Output wasn't supposed to be empty and it was
```

LukeShu:
So I'm thinking actually drop d9a3763?

PR #1768, opened 2021-06-02
Make the cluster-side DNS resolver configurable

### LINKED ISSUE #1775: [v2] Failed to connect to OpenShift 3 ( using Minishift ) - getOutboundInfo: failed to get IP of cluster's DNS service
Hello,

I am receiving the following error when trying to connect:

```
$ telepresence connect --insecure-skip-tls-verify=true

Launching Telepresence Daemon v2.3.0 (api v3)
Connecting to traffic manager...
Disconnecting...done
Telepresence Daemon quitting...done
telepresence: error: connector.Connect: getOutboundInfo: failed to get IP of cluster's DNS service
```

Here are more details on how I install my Minishift and connect to it:
```
$ minishift delete --force --clear-cache
...
$ minishift start --vm-driver=virtualbox
...
$ oc login -u system:admin
...
$ telepresence connect --insecure-skip-tls-verify=true

Launching Telepresence Daemon v2.3.0 (api v3)
Connecting to traffic manager...
Disconnecting...done
Telepresence Daemon quitting...done
telepresence: error: connector.Connect: getOutboundInfo: failed to get IP of cluster's DNS service
```

Information about the system:
```
$ docker --version
Docker version 20.10.6, build 370c289

$ kubectl version
Client Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.1", GitCommit:"5e58841cce77d4bc13713ad2b91fa0d961e69192", GitTreeState:"clean", BuildDate:"2021-05-12T14:11:29Z", GoVersion:"go1.16.3", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"11+", GitVersion:"v1.11.0+d4cacc0", GitCommit:"d4cacc0", GitTreeState:"clean", BuildDate:"2021-05-12T14:13:55Z", GoVersion:"go1.10.8", Compiler:"gc", Platform:"linux/amd64"}
WARNING: version difference between client (1.21) and server (1.11) exceeds the supported minor version skew of +/-1

$ oc version
Client Version: v4.2.0-alpha.0-657-g51011e4
Kubernetes Version: v1.11.0+d4cacc0

$ kubectl cluster-info
Kubernetes control plane is running at https://192.168.99.102:8443

$ telepresence version
Client: v2.3.0 (api v3)
Root Daemon: not running
User Daemon: not running
```

Thank you

thallgren:
## Description

This PR contains improvements that enables configuration of the local and remote IP address of the DNS resolver, and limit the number of `LookupHost()` requests that the client sends to the cluster. The configuration is an extension to the kubeconfig's Cluster definition. The improvements are mainly in five areas.

1. The IP of the local resolver can now be set in the cluster configuration and will become the default unless `--dns` is passed on the command line. This setting is only relevant to Linux workstations that aren't configured to use `systemd.resolved`.
2. The IP of the remote resolver can be configured. This can be useful for systems for which Telepresence doesn't have automatic DNS discovery.
3. The overriding DNS resolver, used on Linux without `systemd-resolved` and in Docker containers, receives all DNS requests that are issued on the client. This is very different from the other resolvers which configure the search paths, domains, and routes, that they want to handle. Since we don't want to send all DNS requests over to the cluster, this resolver now checks if the last name in a QN is a namespace, or if the QN ends with ".svc" or ".cluster.local". When this is the case, the name will not be subjected to concatenations with the search path.
4. The DNS resolvers on all platforms are now configured with an `exclude-suffixes` list by default. This list contains a set of very well known top level domains such as ".com", ".org", and a few more. QNs that have those suffixes will not be dealt with by the resolver. Instead it will return not found, or in case of the overriding resolver, use the fallback. The list is configurable under the new `DNS` heading in the Telepresence config file.
5. There's also an `include-suffixes` list. This list has higher priority than the `exclude-suffixes` list and is intended to be used when it's essential that some suffixes are sent to the cluster resolver that normally wouldn't be. In addition to overriding the `exclude-suffixes`, an entry in this list will also become a "route" in the `systemd-resolved` case and a `domain` on MacOS. This to ensure that the DNS request reaches the resolver in the first place. This solution is primarily intended for deployments that use custom DNS configurations, and should be seen as a workaround for the fact that we cannot automatically cater for what domains such configurations may have.

## Checklist


pkg/client/daemon/outbound.go
```diff
@@ -37,7 +37,7 @@ type outbound struct {
 
 	// Namespaces, accessible using <service-name>.<namespace-name>
 	namespaces map[string]struct{}
-	domains    map[string]iputil.IPs
+	domains    map[string]struct{}
```
thallgren, line 40: The `domains` map was actually a leftover from 2.2 but is now needed by the MacOS resolver for a slightly different purpose.

pkg/client/daemon/outbound.go
```diff
@@ -187,9 +207,8 @@ func (o *outbound) resolveInCluster(c context.Context, qType uint16, query strin
 		}
 	}
 
-	// Give the cluster lookup a reasonable timeout
-	tos := client.GetConfig(c).Timeouts
-	c, cancel := tos.TimeoutContext(c, client.TimeoutTrafficManagerAPI)
+	// Give the cluster lookup a reasonable timeout.
+	c, cancel := context.WithTimeout(c, 4*time.Second)
```
donnyyung: I think we should make this timeout configurable, add a TODO comment saying "we * should * make this configurable at some point", or provide a reason why this should be hardcoded to 4 seconds.

pkg/client/connector/k8s_config.go
```diff
@@ -1,17 +1,47 @@
...
+	// ExcludeSuffixes are suffixes for which the DNS resolver will always fail (or fallback
+	// in case of the overriding resolver).
+	ExcludeSuffixes []string `json:"exclude-suffixes,omitempty"`
+
+	// IncludeSuffixes are suffixes for which the DNS resolver will always attempt to do
+	// a lookup. Includes have higher priority than excludes.
+	IncludeSuffixes []string `json:"include-suffixes,omitempty"`
+}
+
+// kubeconfigExtension is an extension read from the selected kubeconfig Cluster.
```
donnyyung, line 42: Is the reasoning for putting this configuration in the kubeconfig because users could potentially have different values for different clusters?  

I guess I'm torn because I'd prefer for all of our config to be in one place (e.g. `telepresence/config.yml`) because once configuration is spread across multiple files, I feel like it can get confusing to figure out what you configure where.  

I think I still need to do some more thinking about this, but just wanted to check + ensure I understood the reasoning for this decision.

pkg/client/daemon/outbound.go
thallgren: That's an oversight. This timeout should definitely be added.

pkg/client/connector/k8s_config.go
thallgren: I had this in the `config.yml` initially. Then @LukeShu suggested that we should use the kubeconfig extension mechanism for the `also-proxy` stuff because it's bound to be cluster specific and I really liked that idea. Since the DNS resolver resides in the cluster nowadays, it's also cluster specific (besides the `local-ip`, but separating that out seemed worse than keeping the config together). The include-suffixes for instance, is partly aimed at solving the problem with pods using custom DNS-resolvers.

donnyyung: Ah okay that makes a lot of sense. I was on the fence initially, but after thinking more about it and listening to our reasoning, I think it makes more sense for cluster-specific config to be added via a kubeconfig extension and the `config.yml` can be for more "global" telepresence config. I think as long as we document that well, then it shouldn't be confusing for our users.

pkg/client/daemon/outbound.go
```diff
@@ -135,18 +137,40 @@ func (o *outbound) routerServerWorker(c context.Context) (err error) {
...
 const tel2SubDomain = "tel2-search"
-const dotTel2SubDomain = "." + tel2SubDomain
-const dotKubernetesZone = "." + kubernetesZone
+const dotTel2SubDomain = "." + tel2SubDomain + "."
+const dotKubernetesZone = "." + kubernetesZone + "."
 
 var localhostIPv6 = []net.IP{{0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1}}
 var localhostIPv4 = []net.IP{{127, 0, 0, 1}}
 
+func (o *outbound) doClusterLookup(query string) bool {
```
LukeShu: Based on the name of this function, I was expecting it to actually _do_ the lookup.  Perhaps `shouldDoClusterLookup`?

pkg/client/daemon/outbound.go
```diff
@@ -135,18 +137,40 @@ func (o *outbound) routerServerWorker(c context.Context) (err error) {
...
+	if strings.HasSuffix(query, dotKubernetesZone) && strings.Count(query, ".") < 4 {
```
LukeShu, line 147: Why `strings.Count(query, ".") < 4`?

pkg/client/daemon/outbound.go
```diff
@@ -135,18 +137,40 @@ func (o *outbound) routerServerWorker(c context.Context) (err error) {
...
+		if strings.HasSuffix(query, sfx) {
+			return false
+		}
+	}
+	return true
+}
+
 func (o *outbound) resolveInCluster(c context.Context, qType uint16, query string) []net.IP {
-	query = query[:len(query)-1]
 	query = strings.ToLower(query) // strip of trailing dot
```
LukeShu: I think that `// strip of trailing dot` had been intended for the previous deleted line, and therefore should also be deleted?

pkg/client/daemon/outbound.go
LukeShu: I guess that's sorta answered by the old code
```patch
-	// We don't care about queries to the kubernetes zone unless they have at least two additional elements.
-	if strings.HasSuffix(query, dotKubernetesZone) && strings.Count(query, ".") < 3 {
```
but I still don't know why we don't care unless they have at least two additional elements.

pkg/client/daemon/outbound_linux.go
```diff
@@ -36,30 +36,51 @@ func (o *outbound) dnsServerWorker(c context.Context) error {
...
-	ips := o.resolveInCluster(c, qType, query)
-	if len(ips) == 0 && len(o.search) > 0 {
-		// Apply search path
-		q := query[:len(query)-1]
-		q = strings.ToLower(q)
-		q = strings.TrimSuffix(q, dotTel2SubDomain)
-		if strings.HasSuffix(q, ".local") {
-			return nil
+// applySearch returns true if search path should be applied
+func (o *outbound) applySearch(query string) bool {
```
LukeShu: Again; confusing name.  Perhaps `shouldApplySearch`?

pkg/client/daemon/outbound_linux.go
```diff
@@ -36,30 +36,51 @@ func (o *outbound) dnsServerWorker(c context.Context) error {
...
-			ips = o.resolveInCluster(c, qType, q+s)
-			if len(ips) > 0 {
-				break
+			if ips := o.resolveInCluster(c, qType, query+s); len(ips) > 0 {
+				return ips
 			}
 		}
 	}
-	return ips
+	return o.resolveInCluster(c, qType, query)
```
LukeShu: I don't get why we have to loop over the search path here.  I understood why we _used to_ have to loop over it; in the pre-`resolveInCluster` days.  But why are we still doing this client-side, when we can just trim off the trailing `.` and let the agent apply its search path when responding to the LookupHost request?

rpc/daemon/daemon.proto
```diff
@@ -47,6 +47,25 @@ message IPNet {
   int32 mask = 2;
 }
 
+// DNS configuration for the local DNS resolver
+message DNS {
```
LukeShu: Naming this just `DNS` seems a little... oblique.  Perhaps `DNSConfig`?

pkg/client/connector/k8s_config.go
```diff
@@ -1,17 +1,50 @@
...
+	// to the first line of /etc/resolv.conf
+	LocalIP iputil.IPKey `json:"local-ip,omitempty"`
+
+	// RemoteIP is the address of the cluster's DNS service. It will default
+	// to the IP of the kube-dns.kube-system or the dns-default.openshift-dns service.
+	RemoteIP iputil.IPKey `json:"remote-ip,omitempty"`
+
+	// ExcludeSuffixes are suffixes for which the DNS resolver will always fail (or fallback
+	// in case of the overriding resolver).
+	ExcludeSuffixes []string `json:"exclude-suffixes,omitempty"`
```
LukeShu: What does "fail" mean in this context? Returning NXDOMAIN?

pkg/client/connector/k8s_config.go
```diff
@@ -1,17 +1,50 @@
...
 	"github.com/spf13/pflag"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/rest"
 
 	"github.com/datawire/ambassador/pkg/kates"
+	"github.com/telepresenceio/telepresence/v2/pkg/iputil"
 )
 
+// The dns config part of the kubeconfigExtension that is read from the kubeconfig
+type dns struct {
```
LukeShu: Again, unclear name. Perhaps `dnsConfig`?

LukeShu: Actually, `dnsConfig` is almost just as bad.  How about `kubeconfigDNS`?

pkg/client/connector/k8s_config.go
```diff
@@ -21,6 +54,8 @@ type k8sConfig struct {
 	config      *rest.Config
 }
 
+const configExtension = "telepresence.getambassador.io"
```
LukeShu: This should probably be `telepresence.io`? Since it's CNCF and not Ambassador Labs?

pkg/client/daemon/outbound.go
```diff
@@ -135,18 +137,40 @@ func (o *outbound) routerServerWorker(c context.Context) (err error) {
...
+			return true
+		}
+	}
+
+	// Skip configured excludeSuffixes
+	for _, sfx := range o.dnsConfig.ExcludeSuffixes {
+		if strings.HasSuffix(query, sfx) {
+			return false
+		}
+	}
```
LukeShu, line 165: Is giving `includeSuffixes` precedence over `excludeSuffixes` the right call?

LukeShu:
I don't suppose we could get that `DNS` gRPC object to be included in the `telepresence status` output?

pkg/client/connector/k8s_config.go
```diff
@@ -1,17 +1,50 @@
...
+	// ExcludeSuffixes are suffixes for which the DNS resolver will always fail (or fallback
+	// in case of the overriding resolver).
+	ExcludeSuffixes []string `json:"exclude-suffixes,omitempty"`
+
+	// IncludeSuffixes are suffixes for which the DNS resolver will always attempt to do
+	// a lookup. Includes have higher priority than excludes.
+	IncludeSuffixes []string `json:"include-suffixes,omitempty"`
+
+	// The maximum number of seconds to wait for a cluster side host lookup.
+	LookupTimeout int32 `json:"lookup-timeout,omitempty"`
```
LukeShu: Can we use a [`metav1.Duration`](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#Duration) instead of an int, so that it doesn't need to be a whole-number of seconds?

In the YAML it'd be a string:
```yaml
lookup-timeout: 4s
```

LukeShu (changes requested):
On the whole: Looks good, yay!

I do have some concerns, though.  From highest priority to lowest priority:

- Change Request: I think it should be using `telepresence.io` instead of `telepresence.getambassador.io`, since it's a CNCF project not an Ambassador Labs project.
- Change Request: Use `metav1.Duration` to configure the timeout, rather than an `int` number of seconds.
- Question: I don't get why we still need to loop over the search-path client-side.
- Question: One of the comments mentions "always fail", but it's not obvious what exactly "fail" means here.  That should be clarified in the comment. ("always return NXDOMAIN"?)
- Question: It is unclear to my why we don't send `SINGLE.cluster.local` queries to the cluster and only send them if there are multiple lables before the `.cluster.local` (other than "that's what it's always done")
- Nitpick: There are some confusing names that could be improved.
- Nitpick: The obsolete `// strip of trailing dot` comment should be deleted.

Probably not in this PR (or in 2.3.1), but if you're up for it:

- Feature request: Include the `DNS` config object in the `telepresence status` output.

pkg/client/connector/k8s_config.go
thallgren: We use `telpresence.getambassador.io` in other places. The annotation in the workloads for instance, is `telepresence.getambassador.io/actions`. I like the consistency, but perhaps we should consider changing the annotation too?

pkg/client/daemon/outbound.go
thallgren: The reason for making that call is that there's a list of default `excludeSuffixes`. The idea is that you might want to keep most of them but override a few. The `includeSuffixes` has no default list.

pkg/client/connector/k8s_config.go
thallgren: I thought about that but I'm concerned that users will forget the 's' and run into problems that may be difficult to understand. Do you think it makes sense to configure this timeout in fractions of seconds?

pkg/client/daemon/outbound.go
thallgren: To my knowledge there are no names of interest that would be just `xxx.cluster.local`. You would have at least `xxx.svc.cluster.local`, `xxx.pod.cluster.local` or similar.

pkg/client/daemon/outbound.go
thallgren: Yup. That comment should not be there.

pkg/client/daemon/outbound_linux.go
thallgren: This logic is only used by the overriding resolver. And yes, in most cases it would work to just pass things over. The exception is for local-only intercepts. It should be also be noted that there's a coming limitation that will limit the search path length to one, namely the intercepted namespace so there will be no repeated search.

pkg/client/connector/k8s_config.go
thallgren: Yes, it will return NXDOMAIN.

pkg/client/connector/k8s_config.go
thallgren: This is a subtype in the`kubeconfigExtension`. I think `kubconfigDNS` in `kubeconfigExtension` is a bit of an overkill. Hence just `dns`.

pkg/client/daemon/outbound.go
thallgren: ... and we are using `cluster.local` as the domain which means that queries like `xxx.cluster.local` do arrive quite often.

pkg/client/daemon/outbound_linux.go
thallgren: I'm working on a PR to impose the limitation to only be able to intercept one namespace at a time. In that PR, the use of  a search path  is replaced by passing the optional local-only intercepted namespace.

pkg/client/connector/k8s_config.go
thallgren: Thought about it some more. Duration it is.

pkg/client/connector/k8s_config.go
LukeShu: Hmm, so we are.  I guess as long as that's what we're already doing, then it's good enough.

LukeShu (approved):
Thanks for making those changes. Hit it!

PR #1771, opened 2021-06-03
Add agent2systema rpcs for RemoveIntercept

donnyyung:
This adds a gRPC which enables the agent to inform Ambassador Cloud
(System A) that an intercept has been left.

---

## Helpful information

    ```shell
    make lint check-local
    ```

    ```shell
    ```


rpc/systema/agent2systema.proto
```diff
@@ -30,3 +33,7 @@ message ReviewInterceptResponse {
   // allowed to become active.
   string msg = 2;
 }
+
+message LeftIntercept {
```
donnyyung: I wasn't sure what to name this... Not sure if `LeftInterceptInfo` would be better... naming is hard so open to suggestions.

rpc/systema/agent2systema.proto
```diff
@@ -2,19 +2,22 @@ syntax = "proto3";
 
 package telepresence.systema;
 
+import "google/protobuf/empty.proto";
 import "rpc/manager/manager.proto";
 
 option go_package = "github.com/telepresenceio/telepresence/rpc/v2/systema";
 
-// How Calls from the agent to System A are authenticated is TBD.
+// How Calls from the agent to System A are authenticated is via apikey.
```
LukeShu: This sentence is now awkwardly worded.  The sentence was there for symmetry with the multiple comments in `manager2systema.proto`:
```
// Calls from the X to Y are authenticated via RFC 6750
// Bearer tokens.
```
But as I paste that, I realize that comment is outdated; it isn't using RFC 6750 Bearer tokens any more.  So probably update those comments too?

So maybe "Calls from the X to Y are authenticated via API keys per https://www.notion.so/datawire/Telepresence-System-A-API-Key-Integration-c13c4676e9944abfb02713b8517051e3 (see also: `pkg/client/connector/auth/apikey.go`, `cmd/traffic/cmd/manager/systema.go:systemaCredentials`)."? I know it's a faux pas to link to non-public things, but... no one outside of Ambassador Labs is going to be hacking on auth with Ambassador Cloud, and it's the best reference for us when we do hack on it.

rpc/systema/agent2systema.proto
LukeShu: `Left` sounds weird.  Perhaps `RemovedIntercept`?  Or `InterceptRemoval`? I think I like `InterceptRemoval` best.

LukeShu:
The general strategy seems good to me.

donnyyung:
I have went back-and-forth on this all day, but I think it makes sense to do this in the manager for a few reasons:
1. It was a bunch of extra work to see if an intercept had been removed instead of just hooking into the RemoveIntercept Call.
2. When an agent's pod restarts, the intercept is still listed as active (which might be a bug but I think is the behavior we want), so if i put this logic in the agent, then it would say the intercept was removed when it really wasn't.
3. I think my hesitancy to do it in the manager was bc managing the API Keys might be more complicated, but I think it's better to pick the right location for the solution than pick whichever is easier.

PR #1772, opened 2021-06-04
Add RemoveIntercept RPCs

donnyyung:
This adds RemoveIntercept RPCs so that System A is updated when intercepts are left. This will enable Ambassador Cloud to display information about all intercepts that a user creates while logged in, instead of just intercepts with a Preview URL.  It does this by making this call whenever intercepts that contain an APIKey are deleted (during garbage collection). 

I recommend a commit-by-commit review since they detail my thought process and how I landed on the final solution.

## Checklist

 - [ ] I made sure to update `./CHANGELOG.md`. -- I don't know that this needs a changelog bump since it's mainly just for System A to use.


cmd/traffic/cmd/manager/internal/state/state.go
```diff
@@ -97,22 +97,24 @@ type State struct {
 	//     `intercept.ClientSession.SessionId`)
 	//  6. `intercepts` needs to be pruned in-sync with `agents` (based on
 	//     `agent.Name == intercept.Spec.Agent`)
-	port         uint16
-	intercepts   watchable.InterceptMap
-	agents       watchable.AgentMap                   // info for agent sessions
-	clients      watchable.ClientMap                  // info for client sessions
-	sessions     map[string]SessionState              // info for all sessions
-	listeners    map[string]connpool.Handler          // listeners for all intercepts
-	agentsByName map[string]map[string]*rpc.AgentInfo // indexed copy of `agents`
```
donnyyung: Many of the contents of this commit are reverted in https://github.com/telepresenceio/telepresence/pull/1772/commits/5c9ab850915ac9c2d5d17b0dd81509a02ac1aaaf ... The reason for that was bc I originally thought we should call "RemoveIntercept" when the user left an intercept, but through later development realized we'd need to add the same thing to the garbage-collection go routine, and so thought it unnecessary to do it two times.  

That being said, I did want to leave this commit up in case y'all feel it is worth doing it twice or prefer this method to just the garbage collection.

cmd/traffic/cmd/manager/service.go
```diff
@@ -405,13 +405,34 @@ func (m *Manager) RemoveIntercept(ctx context.Context, riReq *rpc.RemoveIntercep
...
 
+	interceptID := fmt.Sprintf("%s:%s", sessionID, name)
 	dlog.Debugf(ctx, "RemoveIntercept called: %s", name)
 
 	if m.state.GetClient(sessionID) == nil {
 		return nil, status.Errorf(codes.NotFound, "Client session %q not found", sessionID)
 	}
-
-	if !m.state.RemoveIntercept(sessionID + ":" + name) {
+	interceptAPIkey := m.state.GetInterceptAPIKey(interceptID)
```
donnyyung: This goes away in the final commit since I opted to only do this in the garbage collection go routine, so feel free to ignore this, unless you think it should also be done here.

cmd/traffic/cmd/manager/internal/state/state.go
```diff
@@ -553,6 +560,13 @@ func (s *State) GetIntercept(interceptID string) *rpc.InterceptInfo {
 	return intercept
 }
 
+func (s *State) GetInterceptAPIKey(interceptID string) string {
+	if apiKey, ok := s.interceptAPIkeys[interceptID]; ok {
```
thallgren: I believe this one needs `s.mu.Lock()`  / `s.mu.Unlock()`.

cmd/traffic/cmd/manager/internal/state/state.go
```diff
@@ -99,20 +99,24 @@ type State struct {
...
-	clients          watchable.ClientMap                  // info for client sessions
-	sessions         map[string]SessionState              // info for all sessions
-	interceptAPIkeys map[string]string                    // map of APIKeys to intercepts
-	listeners        map[string]connpool.Handler          // listeners for all intercepts
-	agentsByName     map[string]map[string]*rpc.AgentInfo // indexed copy of `agents`
+	agents           watchable.AgentMap      // info for agent sessions
+	clients          watchable.ClientMap     // info for client sessions
+	sessions         map[string]SessionState // info for all sessions
+	interceptAPIKeys map[string]string       // InterceptIDs mapped to the APIKey used to create them
+	// We need this extra state because the garbage collector
```
thallgren: This comment is misplaced (has nothing to do with `listeners`.

cmd/traffic/cmd/manager/internal/state/state.go
```diff
@@ -555,16 +556,28 @@ func (s *State) RemoveIntercept(interceptID string) bool {
 	return didDelete
 }
 
-func (s *State) GetIntercept(interceptID string) *rpc.InterceptInfo {
-	intercept, _ := s.intercepts.Load(interceptID)
-	return intercept
+// GetInterceptAPIKeys returns the s.interceptAPIKeys map.
```
thallgren: This feels dangerous. The map needs to be protected by a mutex but the comment doesn't mention that. Perhaps create a slice that just contains the values within mutex lock and return that instead?

cmd/traffic/cmd/manager/internal/state/state.go
```diff
@@ -555,16 +556,28 @@ func (s *State) RemoveIntercept(interceptID string) bool {
...
 
-func (s *State) GetInterceptAPIKey(interceptID string) string {
-	if apiKey, ok := s.interceptAPIkeys[interceptID]; ok {
-		return apiKey
+// RemoveInterceptAPIKey removes the associated APIKey for an Intercept ID
+// Only call on an intercept that has been deleted.
+func (s *State) RemoveInterceptAPIKey(interceptID string) bool {
+	// If the APIKey isn't present, then we return false since we didn't remove
+	// anything since no APIKey was associated with that intercept.
+	if _, ok := s.interceptAPIKeys[interceptID]; !ok {
```
thallgren, line 577: Missing mutex.

thallgren (changes requested):

cmd/traffic/cmd/manager/service.go
```diff
@@ -545,3 +546,33 @@ func (m *Manager) WatchLookupHost(session *rpc.SessionInfo, stream rpc.Manager_W
...
+// reapDomain informs SystemA that an intercept with a domain has been garbage collected
+func (m *Manager) reapDomain(ctx context.Context, sa systema.SystemACRUDClient, interceptUpdate watchable.InterceptMapUpdate) error {
+	dlog.Debugf(ctx, "systema: removing domain: %q", interceptUpdate.Value.PreviewDomain)
+	_, err := sa.RemoveDomain(ctx, &systema.RemoveDomainRequest{
+		Domain: interceptUpdate.Value.PreviewDomain,
+	})
+	if err != nil {
+		return err
+	}
+	return nil
```
josecv: we could probably just `return err` here; same for `reapIntercept`

donnyyung: Yeah I went back and forth on that, my concern was that it'd potentially be confusing to read cuz at a quick glance you might think it's always returning an `err`, and felt it was more clear to only return `err` when `err != nil`... but looking I guess we do do that in other places so I suppose I'm incorrect.

cmd/traffic/cmd/manager/internal/state/state.go
```diff
@@ -548,6 +555,34 @@ func (s *State) RemoveIntercept(interceptID string) bool {
...
+// which apiKey is used.
+func (s *State) GetInterceptAPIKeys() []string {
+	s.mu.Lock()
+	apiKeys := make([]string, 0, len(s.interceptAPIKeys))
+	for _, key := range s.interceptAPIKeys {
+		apiKeys = append(apiKeys, key)
+	}
+	s.mu.Unlock()
+	return apiKeys
+}
```
LukeShu, line 569: Probably easier to just return the first non-empty API key?  Simplify the code here, simplify the code in systema.go, avoid a few allocations.

cmd/traffic/cmd/manager/internal/state/state.go
```diff
@@ -97,22 +97,27 @@ type State struct {
...
-	port         uint16
-	intercepts   watchable.InterceptMap
-	agents       watchable.AgentMap                   // info for agent sessions
-	clients      watchable.ClientMap                  // info for client sessions
-	sessions     map[string]SessionState              // info for all sessions
-	listeners    map[string]connpool.Handler          // listeners for all intercepts
-	agentsByName map[string]map[string]*rpc.AgentInfo // indexed copy of `agents`
+	//  7. `interceptAPIKeys` needs to persist past intercepts since they are used by the garbage
+	//      collector, after the intecept has been deleted from the map, if there are no active
+	//      sessions in the traffic-manager
```
LukeShu: This comment doesn't really tell us why `interceptAPIKeys` needs to be part of what the mutex protects (which is what this comment block is about, if you scroll up).

Perhaps:
> `interceptAPIKeys` entries need to be created and updated in-sync with `intercepts` (but not deleted in-sync with `intercepts`; that happens separately, in `RemoveInterceptAPIKey()`)

?

cmd/traffic/cmd/manager/service.go
```diff
@@ -545,3 +546,27 @@ func (m *Manager) WatchLookupHost(session *rpc.SessionInfo, stream rpc.Manager_W
 func (m *Manager) expire() {
 	m.state.ExpireSessions(m.clock.Now().Add(-15 * time.Second))
 }
+
+// reapDomain informs SystemA that an intercept with a domain has been garbage collected
+func (m *Manager) reapDomain(ctx context.Context, sa systema.SystemACRUDClient, interceptUpdate watchable.InterceptMapUpdate) error {
```
LukeShu: Why did you define these here in `service.go` instead of `manager.go` (where it seems to me like they belong?--everything in `service.go` deals directly with responding to gRPC requests).

cmd/traffic/cmd/manager/service.go
```diff
@@ -545,3 +546,27 @@ func (m *Manager) WatchLookupHost(session *rpc.SessionInfo, stream rpc.Manager_W
...
+	dlog.Debugf(ctx, "systema: removing domain: %q", interceptUpdate.Value.PreviewDomain)
+	_, err := sa.RemoveDomain(ctx, &systema.RemoveDomainRequest{
+		Domain: interceptUpdate.Value.PreviewDomain,
+	})
+	return err
+}
+
+// reapIntercept informs SystemA that an intercept has been garbage collected
+func (m *Manager) reapIntercept(ctx context.Context, sa systema.SystemACRUDClient, interceptUpdate watchable.InterceptMapUpdate) error {
```
LukeShu: I like that you defined these to take MapUpdates as arguments, instead of having them take the `MapUpdate.Value` as the argument.  It encodes the assumptions about "when is this safe to call" in to the type signature.  I wouldn't have thought to do that.

cmd/traffic/cmd/manager/internal/state/state.go
donnyyung: Yeah this was mostly me fighting with go formatting to add a longer comment to really explain why I was adding it, but I will improve the messaging to better fit with the comment block.

cmd/traffic/cmd/manager/service.go
donnyyung: This was bad pattern matching since I saw all the methods in that file and didn't see any in `manager.go` but I will move it.

cmd/traffic/cmd/manager/service.go
donnyyung: Yeah although now that I'm looking at it again, I should probably check in this function that `interceptUpdate.Delete` is true... I know I do it before the function is called, but we don't want to call this function unless the intercept was indeed deleted, so I think I should code that into the functions in case they ever get used elsewhere.

cmd/traffic/cmd/manager/service.go
LukeShu: I tend to prefer the explicit form like Donny originally did.  I have two reasons for this preference:

- It makes coverage testing easier, as you can see in the coverage report whether you covered both the error and non-error cases.  The single `return err` hides this info.
- Because nil structs get silently promoted to non-nil interfaces, having the separate returns is defensive programming in that it catches a class of bugs.  For example:

Consider the code

```go
func innerFunc() error {
	…
}

func outerFunc() error {
	err := innerFunc()
	return err
}
```

In this version, that `return err` is fine.  However, what if we changed `innerFunc`'s signature changed to `func innerFunc() *MyErrorType` (where `*MyErrorType` implements `error`)?  It seems like an innocuous change, and the compiler would happily accept it without complaining.  However, the nil `*MyErrorType` would get promoted to a non-nil `error`, and suddenly `outerFunc` is returning non-nil when it should be returning nil!

And things like that give me the willies, so I prefer to write code in a style where that can't happen.

LukeShu:
> I think the better longterm solution is to have a System A heartbeat RPC
> that informs System A of intercepts that are still "live", but since we
> are a little strapped for time right now and that would require a
> cross-team effort, I've added a way to keep APIKeys associated wither
> intercepts in the state of the manager and only delete them * after *
> the garbage collection has informed System A that the intercept + domain
> has been removed.

So there's already an RPC where Ambassador Cloud can ask the manager for the complete list of intercepts.  The problem with it is that it requires the manager to be connected to Ambassador Cloud, but the manager doesn't keep a persistent connection unless there's an active preview URL.  Another solution might be to have the manager connect as long as it has an API key to connect with?

LukeShu (approved):
A couple of minor nitpicks and suggestions that you can act on if you want.  On the whole: Looks good!

> I recommend a commit-by-commit review since they detail my thought process and how I landed on the final solution.

FWIW, since the fix ended up removing so much of the earlier code, I actually did find it easier to review it all squashed together.

I guess I think a good rule of thumb would be to not ask the reviewer to review code that a later commit deletes?

cmd/traffic/cmd/manager/service.go
josecv: > Because nil structs get silently promoted to non-nil interfaces, having the separate returns is defensive programming in that it catches a class of bugs.

Waat I had no idea that could happen, makes sense to keep it the way Donny had it in that case. The point about code coverage is valid too

donnyyung:
Yeah I had thought about using that RPC but then ran into the problem you mentioned... I perhaps should have tried harder to make it work, which I will probably do during cooldown.  I think this is a net-improvement as is, and should cover most of the scenarios and I don't want to make any huge changes this late in the cycle, but thanks for giving me good ideas to further improve it in the future.

PR #1778, opened 2021-06-07
smoke-tests: change intercept ID parsing to match new output format

josecv:


donnyyung (approved):
LGTM! Also validated smoke tests passed on my machine just to be extra safe :)

PR #1782, opened 2021-06-08
Fix race between reading from the cluster and closing the connection

josecv:
Basically when transferring a large file from the cluster, sometimes the
daemon would process a FIN from the remote end before all packets that
came before the FIN had been routed.

---

## Helpful information

    ```shell
    make lint check-local
    ```

    ```shell
    ```

### DIFF
--- a/cmd/traffic/logger.go
+++ b/cmd/traffic/logger.go
@@ -12,11 +12,11 @@ import (
 func makeBaseLogger() dlog.Logger {
 	logrusLogger := logrus.New()
 	logrusFormatter := &logrus.TextFormatter{
-		TimestampFormat: "2006-01-02 15:04:05",
+		TimestampFormat: "2006-01-02 15:04:05.0000",
 		FullTimestamp:   true,
 	}
 	logrusLogger.SetFormatter(logrusFormatter)
-	logrusLogger.SetReportCaller(false)
+	logrusLogger.SetReportCaller(true)
 
 	const defaultLogLevel = logrus.InfoLevel
 
--- a/pkg/tun/tcp/handler.go
+++ b/pkg/tun/tcp/handler.go
@@ -139,6 +139,11 @@ type handler struct {
 	sendLock      sync.Mutex
 	sendCondition *sync.Cond
 
+	// isClosing indicates whether Close() has been called on the handler
+	isClosing int32
+	// readyToFin will be closed once it is safe for the handler to send a FIN packet and terminate the connection
+	readyToFin chan interface{}
+
 	// random generator for initial sequence number
 	rnd *rand.Rand
 }
@@ -163,6 +168,7 @@ func NewHandler(
 		rcvWnd:            maxReceiveWindow,
 		wfState:           stateIdle,
 		rnd:               rand.New(rndSource),
+		readyToFin:        make(chan interface{}),
 	}
 	h.sendCondition = sync.NewCond(&h.sendLock)
 	return h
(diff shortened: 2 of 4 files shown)


cmd/traffic/logger.go
```diff
@@ -12,11 +12,11 @@ import (
 func makeBaseLogger() dlog.Logger {
 	logrusLogger := logrus.New()
 	logrusFormatter := &logrus.TextFormatter{
-		TimestampFormat: "2006-01-02 15:04:05",
+		TimestampFormat: "2006-01-02 15:04:05.0000",
 		FullTimestamp:   true,
 	}
 	logrusLogger.SetFormatter(logrusFormatter)
-	logrusLogger.SetReportCaller(false)
+	logrusLogger.SetReportCaller(true)
```
thallgren, line 19: Adding milliseconds is fine (and very much needed) but do we really need the caller here?

pkg/tun/tcp/handler.go
```diff
@@ -303,12 +313,19 @@ func (h *handler) sendSyn(ctx context.Context, ackNumber uint32, ack bool) {
...
 func (h *handler) writeToTunLoop(ctx context.Context) {
+	defer close(h.readyToFin)
+	timer := time.NewTimer(100 * time.Millisecond)
 	for {
 		var cm connpool.Message
 		select {
 		case <-ctx.Done():
 			return
 		case cm = <-h.fromMgr:
+		case <-timer.C:
```
thallgren: From the looks of it, this timer will only make difference on very short-lived connections. A timer will only fire once. Perhaps you intended to use a `Ticker` or: `case <-time.After(100 * time.Milliecond):`?

pkg/tun/tcp/handler.go
```diff
@@ -139,6 +139,11 @@ type handler struct {
 	sendLock      sync.Mutex
 	sendCondition *sync.Cond
 
+	// isClosing indicates whether Close() has been called on the handler
+	isClosing bool
```
thallgren: This variable is written and read from different goroutines so it needs a mutex or be atomic. Perhaps you can use an `int32` instead of a `bool` and then `atomic.StoreInt32()` and `atomic.LoadInt32()` to store and test.

thallgren (changes requested):
I requested some minor changes. Besides that it looks good. I'm very happy that someone besides me dives into this code.

cmd/traffic/logger.go
josecv: I wanted to make it consistent with the daemon logs, plus I found the caller really useful for tracking down what `pkg/connpool` is doing -- are there any concerns around adding it?

pkg/tun/tcp/handler.go
josecv: Ah yes this should definitely be a `Ticker`

cmd/traffic/logger.go
thallgren: The reason I asked was actually for consistency. I don't see that this is enabled in the daemon logs.

thallgren (approved):
Looks good now!

cmd/traffic/logger.go
josecv: hmm, i see it enabled [here](https://github.com/telepresenceio/telepresence/blob/cb198902a5497b97e66bb31d71c7ab5d17ac791d/pkg/client/logging/initcontext.go#L30)

PR #1783, opened 2021-06-09
Add an agent injecting mutating webhook to the traffic manager.

thallgren:
## Description

This PR adds a mutating webhook that injects a traffic-agent to
pods that have the following annotation:

    telepresence.getambassador.io/traffic-agent: enabled

The installer will henceforth ignore deployments that have this
annotation and consider them to already have an agent installed.

The webhook is using `kates.List()` to list Service objects which
means that the cluster must be configured to allow the traffic-manager
to do that. A sample `k8s/agent-injector-rbac.yaml` is included that
provides the needed credentials.

A lot of code used by the agent installer is reused in the injector
and was therefore moved into the `agentinstall` package that they
both now share.
## Checklist


k8s/agent-injector-rbac.yaml
```diff
@@ -0,0 +1,27 @@
+# This should be the minium RBAC that is needed for users (not operators) to
```
donnyyung: I think this rbac is actually something that is needed for the operator who will be deploying the traffic-manager and not users (who won't need this role attached to their account and shouldn't need these permissions).

thallgren: Doh, that comment is off. I'll change it.

thallgren: Hmm, it's already changed. You need to refresh your browser :-)

LukeShu:
If `envconfig.go` is moving from `manager/` to `manager/managerutil/`, then `envconfig_test.go` should probably move too?

donnyyung:
I'm reliably getting: ```telepresence: error: Intercept "3b3c482d-4de9-4145-a343-f7d8fff0e3aa:echo-easy": Failed to establish intercept: intercept in error state AGENT_ERROR: no valid license & cannot connect to Ambassador Cloud for license check``` when using this branch with the Smart Agent.

Looking at the config, 
```
time="2021-06-10 15:25:43" level=info msg="config = {Config:{Name:echo-easy Namespace:default PodIP:10.42.0.45 AgentPort:9900 AppMounts:/tel_app_mounts AppPort:8080 ManagerHost:traffic-manager.ambassador ManagerPort:8081} EnvoyLogLevel:info EnvoyBaseID:1 UseHTTP2:true TerminatingTLSSecret: OriginatingTLSSecret: License: LicenseHost: LicensePublicKeys:{KeyMap:map[auth.datawire.io:-----BEGIN PUBLIC KEY----------END PUBLIC KEY----- beta-auth.datawire.io:-----BEGIN PUBLIC KEY----------END PUBLIC KEY-----]} ClusterID: CanConnectCloud:false SystemAHost: SystemAPort:}" func=main.Main file="github.com/datawire/telepresence2/proprietary/cmd/app-sidecar/sidecar.go:178" PID=1
```

Looking at the config, SystemAHost and SystemAPort (which the agent talks to the manager to get), aren't being correctly populated. I'll keep looking but wanted to raise the issue here in case you might know what in the changes might have affected this.

-- Edit --
I added a comment with what I believe to be the cause of the issue.

cmd/traffic/cmd/manager/service.go
```diff
@@ -47,11 +48,10 @@ func (wall) Now() time.Time {
 	return time.Now()
 }
 
-func NewManager(ctx context.Context, env Env) *Manager {
+func NewManager(ctx context.Context) *Manager {
 	ret := &Manager{
 		ctx:   ctx,
 		clock: wall{},
-		env:   env,
```
donnyyung, line 54: I think this is the issue with the smart agent bc this function uses the config to get the System A info: https://github.com/telepresenceio/telepresence/blob/release/v2/cmd/traffic/cmd/manager/service.go#L100-L102

LukeShu:
Why not straight-up use the existing install "actions" system?

LukeShu:
"Add an agent injecting mutating webhook to the traffic manager" is difficult to review... it should probably be multiple commits.

cmd/traffic/cmd/manager/internal/mutator/agent_injector.go
```diff
@@ -0,0 +1,234 @@
+package mutator
+
+import (
+	"context"
+	"fmt"
+	"strconv"
+	"strings"
+
+	admission "k8s.io/api/admission/v1"
+	v1 "k8s.io/api/core/v1"
```
LukeShu: IMO, importing things as `v1` is an anti-pattern.  Say what it's v1 of! i.e. `corev1`

pkg/install/container.go
```diff
@@ -0,0 +1,147 @@
...
+	"strings"
+
+	v1 "k8s.io/api/core/v1"
+
+	"github.com/datawire/ambassador/pkg/kates"
+)
+
+const envPrefix = "TEL_APP_"
+
+func AgentContainer(name, imageName string, appContainer *v1.Container, port v1.ContainerPort, appPort int, managerNs string) v1.Container {
```
LukeShu: That's a lot of args.  Maybe it should take a single struct?

At least wrap this to be 1 arg per line.

cmd/traffic/cmd/manager/managerutil/envconfig.go
```diff
@@ -0,0 +1,42 @@
...
+	"context"
+
+	"github.com/sethvargo/go-envconfig"
+)
+
+type Env struct {
+	User        string `env:"USER,default="`
+	AgentImage  string `env:"TELEPRESENCE_AGENT_IMAGE,default="`
+	AgentPort   int32  `env:"TELEPRESENCE_AGENT_PORT,default=9900"`
+	Namespace   string `env:"MANAGER_NAMESPACE,default="`
```
LukeShu: I don't like having short envconfig field names like this.  If we thought it was worth it to put `MANAGER_` in the env-var name, then surely it's worth it to put `Manager` in the field name.

cmd/traffic/cmd/manager/internal/mutator/agent_injector.go
```diff
@@ -0,0 +1,234 @@
...
+			if svcPort.Name != "" {
+				return svcPort.Name
+			}
+			return strconv.Itoa(int(svcPort.Port))
+		}(), refPodName)
+
+	agentImage := env.AgentImage
+	if agentImage == "" {
+		agentImage = env.Registry + "/tel2:" + strings.TrimPrefix(version.Version, "v")
+	}
```
LukeShu: Consider doing this in `LoadEnv`

build/admission_controller_tls.go
```diff
@@ -0,0 +1,71 @@
+package main
```
LukeShu, line 1: So... when does this need to get run?  Because nothing's running it right now.

cmd/traffic/cmd/manager/systema.go
```diff
@@ -8,6 +8,8 @@ import (
 	"net"
 	"sync"
 
+	"github.com/telepresenceio/telepresence/v2/cmd/traffic/cmd/manager/managerutil"
+
 	"google.golang.org/grpc"
```
LukeShu: This should go further down, with the rest of the other telepresenceio imports.

cmd/traffic/cmd/manager/service.go
```diff
@@ -8,6 +8,8 @@ import (
 	"sort"
 	"time"
 
+	"github.com/telepresenceio/telepresence/v2/cmd/traffic/cmd/manager/managerutil"
+
 	"github.com/google/uuid"
```
LukeShu: This should go further down, with the rest of the telepresence.io imports.

LukeShu:
I want to say: I'm making a lot of style-related comments.  The point of them isn't to nitpick the style; I'm commenting when it's something that makes review harder and is costing me time.  For instance, I wanted to know what changed between `install_actions.go` and `install/container.go`, but that turned out to be needlessly difficult because `corev1` was renamed to `v1` and a bunch of brackets changed.  These things are style nitpicks that  on their own wouldn't be worth mentioning.  But they make the diff be 90% noise, so it's difficult to see what's actually changing.

pkg/client/config.go
```diff
@@ -338,7 +338,7 @@ var defaultConfig = Config{
 		PrivateIntercept:             5 * time.Second,
 		PrivateProxyDial:             5 * time.Second,
 		PrivateTrafficManagerAPI:     5 * time.Second,
-		PrivateTrafficManagerConnect: 20 * time.Second,
+		PrivateTrafficManagerConnect: 60 * time.Second,
```
LukeShu, line 341: This should be called out as a "change" in the changelog (with a note about why it changed).

pkg/client/envconfig.go
```diff
@@ -19,8 +19,9 @@ type Env struct {
 	LoginClientID      string `env:"TELEPRESENCE_LOGIN_CLIENT_ID,default=telepresence-cli"`
 	UserInfoURL        string `env:"TELEPRESENCE_USER_INFO_URL,default=https://${TELEPRESENCE_LOGIN_DOMAIN}/api/userinfo"`
 
-	Registry   string `env:"TELEPRESENCE_REGISTRY,default=docker.io/datawire"`
-	AgentImage string `env:"TELEPRESENCE_AGENT_IMAGE,default="`
+	ManagerNamespace string `env:"TELEPRESENCE_MANAGER_NAMESPACE,default=ambassador"`
```
LukeShu: Supporting this is probably worth mentioning in the changelog.

LukeShu (changes requested):
Please make this easier to review.  I spent all morning on it, and didn't get through it.

pkg/client/envconfig.go
thallgren: Perhaps, but it isn't a change really. We've always supported that env using an `os.GetEnv()` in the installer.

build/admission_controller_tls.go
thallgren: This utility was created before I automated all of it in the installer. I thought it worth keeping around because it might help if we need to generate certs from the helm-chart.

thallgren:
@LukeShu there are two reasons why I didn't use the installer "actions":

1. There's no need to `do`/`undo`. This is always just `do`
2. This is all about json patches. The actions actually modify manifests.

One could argue that the actions should be rewritten as json patches, but that didn't feel like something that needed to be done in this PR. Perhaps we should do that when we move the agent installer to the traffic-manager.

cmd/traffic/cmd/manager/systema.go
thallgren: So why does the linter _sometimes_ complain about this? It would be great if it captured all.

pkg/client/envconfig.go
thallgren: I removed this change as it's not relevant to this PR anyway.

pkg/install/resource/tm_cluster_role_binding.go
```diff
@@ -0,0 +1,56 @@
...
+	return cr
+}
+
+func (ri tmClusterRoleBinding) Create(ctx context.Context) error {
+	clb := ri.roleBinding()
+	clb.Subjects = []rbac.Subject{
+		{
+			Kind:      "ServiceAccount",
+			Name:      "traffic-manager",
+			Namespace: "ambassador",
```
thallgren: This should not be hardcoded to "ambassador". A fix will arrive in subsequent commits.

pkg/install/tls.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+	caBytes, err := x509.CreateCertificate(cryptorand.Reader, caCert, caCert, &caPrivKey.PublicKey, caPrivKey)
+	if err != nil {
+		return nil, nil, nil, fmt.Errorf("failed to generate CA certificate: %w", err)
+	}
+	if caPem, err = ToPEM("ca.pem", "CERTIFICATE", caBytes); err != nil {
+		return nil, nil, nil, err
+	}
+
+	dnsNames := []string{"agent-injector", "agent-injector.ambassador", "agent-injector.ambassador.svc"}
```
thallgren: The hardcoded "ambassador" will be replaced in a subsequent commit.

pkg/install/resource/instance.go
```diff
@@ -0,0 +1,138 @@
...
+	"github.com/telepresenceio/telepresence/v2/pkg/client"
+)
+
+// An Instance exposes CRUD operations for a k8s resource such as
+// a Service.
+type Instance interface {
+	Create(context.Context) error
+	Exists(context.Context) (bool, error)
+	Delete(context.Context) error
+	Update(context.Context) error
```
LukeShu: CRUD, not CRDU! :P

pkg/install/resource/instance.go
```diff
@@ -0,0 +1,138 @@
...
+
+// The scope is available through the context
+type scopeKey struct{}
+
+func withScope(ctx context.Context, scope *scope) context.Context {
+	return context.WithValue(ctx, &scopeKey{}, scope)
+}
+
+func getScope(ctx context.Context) *scope {
+	if sc, ok := ctx.Value(&scopeKey{}).(*scope); ok {
```
LukeShu: Those should just be `scopeKey{}`, not `&scopeKey{}`.

Honestly, I'm shocked it seems to work at all with the `&`.

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ### 2.3.1 (TBD)
 
+- Feature: Agents can now be installed using a mutator webhook
+- Change: The `trafficManagerConnect` timout default has changed from 20 seconds to 60 seconds, in order to facilitate TODO.
```
LukeShu: So I didn't actually finish writing that sentence.

pkg/install/resource/tm_deployment.go
```diff
@@ -0,0 +1,183 @@
...
+	"github.com/telepresenceio/telepresence/v2/pkg/install"
+)
+
+const managerLicenseName = "systema-license"
+
+type tmDeployment struct {
+	found *kates.Deployment
+}
+
+var TrafficManagerDeployment Instance = &tmDeployment{}
```
LukeShu: So this is a global variable with state.  Which isn't a blocker, but... globals are a smell.

pkg/install/resource/tm_namespace.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"context"
+
+	"github.com/datawire/ambassador/pkg/kates"
+)
+
+type nsResource int
+
+var TrafficManagerNamespace Instance = nsResource(0)
+var TrafficManagerNamespaceKeep Instance = nsResource(1)
```
LukeShu: Can we change these from `var` to `const`?

pkg/install/resource/tm_svc.go
```diff
@@ -0,0 +1,59 @@
...
+
+	"k8s.io/apimachinery/pkg/util/intstr"
+
+	"github.com/datawire/ambassador/pkg/kates"
+	"github.com/telepresenceio/telepresence/v2/pkg/install"
+)
+
+type tmSvc int
+
+var TrafficManagerSvc Instance = tmSvc(0)
```
LukeShu: Likewise, can this be `const`?

pkg/install/resource/tm_deployment.go
LukeShu: Maybe we avoid that by having `GetTrafficManagerResources()` say `&tmDeployment{}` instead of `TrafficManagerDeployment`?

thallgren: True that. Bad design. I'll fix it. There's no need for this one to be a variable.

pkg/install/resource/tm_namespace.go
thallgren: No, because they are of type `Instance` which is an interface. We could of course cover that with an anonymous var instead.

pkg/install/resource/instance.go
thallgren: Not intended. I'll remove that ampersand. IIRC, it works but the behavior is undefined in the spec.

cmd/traffic/cmd/manager/internal/mutator/service.go
```diff
@@ -0,0 +1,182 @@
...
+		w.WriteHeader(http.StatusOK)
+	})
+
+	server := &http.Server{
+		Addr:     ":" + strconv.Itoa(install.MutatorWebhookPortHTTPS),
+		ErrorLog: dlog.StdLogger(ctx, dlog.LogLevelError),
+		Handler:  mux}
+
+	dlog.Infof(ctx, "Mutating webhook service is listening on %v", server.Addr)
+	err := dutil.ListenAndServeHTTPSWithContext(ctx, server, certPath, keyPath)
```
LukeShu: If you use `dhttp` instead of the old deprecated `dutil`, then you don't have to configure `ErrorLog`.

cmd/traffic/cmd/manager/internal/mutator/service.go
```diff
@@ -0,0 +1,182 @@
...
+	dlog.Info(ctx, "Mutating webhook service stopped")
+	return nil
+}
+
+// Skip mutate requests in these namespaces
+func isNamespaceOfInterest(ctx context.Context, ns string) bool {
+	for _, skippedNs := range []string{
+		metav1.NamespacePublic,
+		metav1.NamespaceSystem,
+		"kube-node-lease",
```
LukeShu: You can type `"kube-node-lease"` as `corev1.NamespaceNodeLease` if you want.

CHANGELOG.md
thallgren: I'll replace the `TODO` with `the extended time it takes to apply everything needed for the mutator webhook.`

pkg/install/resource/tm_deployment.go
```diff
@@ -0,0 +1,183 @@
...
+		{
+			Name:      "license",
+			ReadOnly:  true,
+			MountPath: "/home/telepresence/",
+		},
+		{
+			Name:      "tls",
+			ReadOnly:  true,
+			MountPath: "/var/run/secrets/tls",
+		},
```
LukeShu, line 99: You added this in "Move install, update, and delete of traffic-manager to resource package" but probably didn't mean to add it until "Add a mutator webhook service to the traffic-manager".  You don't need to change it, but I got thrown through a loop for a minute.

pkg/install/resource/tm_namespace.go
thallgren: I'll replace with:
```
const TrafficManagerNamespace = nsResource(0)
const TrafficManagerNamespaceKeep = nsResource(1)

var _ Instance = TrafficManagerNamespace
```
and do the same thing for the other integer base types.

cmd/traffic/cmd/manager/internal/mutator/agent_injector.go
```diff
@@ -0,0 +1,228 @@
...
+		// It is very probable the pod was not yet assigned a namespace,
+		// in which case we should use the AdmissionRequest namespace.
+		podNamespace = req.Namespace
+	}
+	podName := pod.Name
+	if podName == "" {
+		// It is very probable the pod was not yet assigned a name,
+		// in which case we should use the metadata generated name.
+		podName = pod.ObjectMeta.GenerateName
+	}
```
LukeShu, line 55: That... can't be right?  The Pod name is `GenerateName` plus a random suffix.
```
$ kubectl get pod -o yaml verylargejavaservice-6b696cb89f-blfdd |y2j|jq '{name: .metadata.name, generateName: .metadata.generateName}'
{
  "name":         "verylargejavaservice-6b696cb89f-blfdd",
  "generateName": "verylargejavaservice-6b696cb89f-"
}
```

pkg/install/resource/injector_svc.go
```diff
@@ -0,0 +1,58 @@
...
+			{
+				Name: "https",
+				Port: 443,
+				TargetPort: kates.IntOrString{
+					Type:   intstr.String,
+					StrVal: "https",
+				},
+			},
+		},
+	}
```
LukeShu, line 45: Is there a reason we don't just shove the port in to the existing manager service? Is there a reason to choose one approach over the other?

LukeShu (changes requested):
That was _so much_ easier to review, thank you!

I have 2 minor changes to request, and a question I want clarified.

- Change requested: the context key change
- Change requested: the changelog TODO
- Question: The Pod generateName thing.  Is it OK because the podName only gets used for user-messages?

Other than that, looks solid!  I mentioned a number of nitpicks and minor suggestions that you can act on or ignore.

LukeShu:
After you make the changes, feel free to merge without waiting for me to re-review.

cmd/traffic/cmd/manager/internal/mutator/agent_injector.go
thallgren: Copy pasta from old impl. I'll remove it.

pkg/install/resource/injector_svc.go
thallgren: The webhook mechanism is hardcoded to port 443 and I didn't want to dedicate that port for this since it uses a generated TLS that is useless for other purposes. The traffic-manager service has `ClusterIP: None` so it's headless and cannot do proxying and I wanted a service that proxies 443 to  8443 in the manager. I'm sure there are other ways to do this. Let's revisit.

LukeShu (approved):
Hit it!

PR #1784, opened 2021-06-09
Allow the user's GOROOT to be set in unit tests.

josecv:
This is needed to make the unit tests more robust with brew-installed
ko

---

## Helpful information

    ```shell
    make lint check-local
    ```

    ```shell
    ```

### DIFF
--- a/pkg/client/cli/telepresence_test.go
+++ b/pkg/client/cli/telepresence_test.go
@@ -810,7 +810,7 @@ func (ts *telepresenceSuite) publishManager() error {
 		"TELEPRESENCE_VERSION=" + ts.testVersion,
 		"TELEPRESENCE_REGISTRY=" + dtest.DockerRegistry(ctx),
 	}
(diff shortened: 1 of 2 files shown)


LukeShu:
wut.  I'm actually confused at what lead to this.

1. It shouldn't be necessary to manually install `ko`, the Makefile should automatically install it at `./tools/bin/ko`
2. Even if you do manually install `ko`, it shouldn't matter because it should still use the one at `./tools/bin/ko`
3. *how in the world* is `GOROOT` getting messed up!?  `GOROOT` should be totally automatic since like Go 1.9 (2017).  That suggests to me that Brew's `ko` is just fundamentally broken.

LukeShu:
I think the various `*_test.go` files *should* be shelling out to the Makefile to ensure that `./tools/bin/ko` is OK?  But maybe that isn't happening, and so it's broken when you `go test` instead of `make test`?

josecv:
closing this PR as it turns out the issues I was seeing with `ko` were stemming from me upgrading go without rebuilding the `ko` in `./tools/bin/ko`; I'll add a note into `DEVELOPING.md` explaining this

PR #1788, opened 2021-06-11
Fix a race between closing a connection to the agent and writing to it.

josecv:
## Description

This was causing issues when an intercepted service transfers a large
file; the last packet wouldn't always make it to the traffic agent, since
the manager would get the instruction to close the connection from the
daemon and act upon it immediately, without allowing the last packet to
be written to the agent.

## Checklist

 - [x] I made sure to update `./CHANGELOG.md` (updated as part of #1782).

### DIFF
--- a/pkg/connpool/dialer.go
+++ b/pkg/connpool/dialer.go
@@ -33,14 +33,15 @@ type TunnelStream interface {
 
 // The dialer takes care of dispatching messages between gRPC and UDP connections
 type dialer struct {
-	id         ConnID
-	release    func()
-	bidiStream TunnelStream
-	incoming   chan Message
-	conn       net.Conn
-	idleTimer  *time.Timer
-	idleLock   sync.Mutex
-	connected  int32
+	id            ConnID
+	release       func()
+	bidiStream    TunnelStream
+	incoming      chan Message
+	conn          net.Conn
+	idleTimer     *time.Timer
+	idleLock      sync.Mutex
+	connected     int32
+	writerClosing chan struct{}
 }
 
 // NewDialer creates a new handler that dispatches messages in both directions between the given gRPC server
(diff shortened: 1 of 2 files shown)


pkg/connpool/dialer.go
```diff
@@ -161,7 +169,7 @@ func (h *dialer) sendTCD(ctx context.Context, code ControlCode) {
 func (h *dialer) readLoop(ctx context.Context) {
 	defer func() {
 		// allow write to drain and perform the close of the connection
-		dtime.SleepWithContext(ctx, 200*time.Millisecond)
+		dtime.SleepWithContext(ctx, 100*time.Millisecond)
```
josecv, line 181: walking this back to its previous value as i think the new change is actually the right way to ensure the write drains

LukeShu (commented):
This seems wrong to me... one end of the socket is allowed to stay open arbitrarily long after the other end is closed.  This looks to me like it breaks that.

josecv:
hmm i'm not sure i understand the concern -- if the connection [is closed upon receiving the `ReadClose`](https://github.com/telepresenceio/telepresence/blob/c634f29f2f2a9e65994c846bd8d695c0353ebfb4/pkg/connpool/dialer.go#L120-L121) wouldn't the net effect be the same? [the write loop would still exit here](https://github.com/telepresenceio/telepresence/blob/c634f29f2f2a9e65994c846bd8d695c0353ebfb4/pkg/connpool/dialer.go#L218-L225), i think

unless `Write()`ing to a closed connection doesn't result in an error? I guess in that case something different needs to happen, maybe having a lock around writing a payload so that close can't be called until the write loop is done processing

LukeShu (commented):
I guess if closing was already the behavior, then I guess I don't object to your change.  But we should look in to that behavior around half-closed sockets.

But also, I really don't like polling the atomic.  Can you change `writerClosing` to a `chan struct{}` that gets closed?  Either in a `sync.Once` or with
```go
select {
case <-h.writerClosing:
    // already closed, nothing to do
default:
    close(h.writerClosing)
}
```
(Is the select safe? Can `handleControl` be called from 2 different goroutines?)

josecv:
> Can `handleControl` be called from 2 different goroutines?

I really don't think so -- the only way to get there is through [`HandleMessage`](https://github.com/telepresenceio/telepresence/blob/release/v2/pkg/connpool/dialer.go#L132), which is only called from the [`DialLoop`](https://github.com/telepresenceio/telepresence/blob/release/v2/pkg/connpool/stream.go#L58-L86), which is in turn reading one message at a time from the `msgCh`

I'll change `writerClosing` to a channel, that's way nicer than the atomic

LukeShu (approved):
Hit it!

PR #1797, opened 2021-06-16
Implement a --to-pod flag to forward ports from an intercepted pod

josecv:
## Description

## Checklist


pkg/client/connector/intercept.go
```diff
@@ -127,6 +129,41 @@ func (tm *trafficManager) workerPortForwardIntercepts(ctx context.Context) error
...
+func (tm *trafficManager) portForward(ctx context.Context, wg *sync.WaitGroup, fk forwardKey, port int32) {
+	defer wg.Done()
+	// Using kubectl port-forward here would require the pod name to be either fetched from the API server, or threaded
+	// all the way through from the intercept request to the agent and into the WatchIntercepts. Instead, we re-use
+	// the forwarder from the agent, and dial the pod's IP directly.
+	addr := net.TCPAddr{
+		IP:   net.IPv4(127, 0, 0, 1),
+		Port: int(port),
+	}
+	f := agent.NewForwarder(&addr, fk.PodIP, port)
```
thallgren: This feels like a violation of implicit package structure rules. A `pkg` should not access things under `cmd`. If the `Forwarder` no longer is agent specific, then please move it to a package under `pkg`.

pkg/client/connector/intercept.go
```diff
@@ -127,6 +129,41 @@ func (tm *trafficManager) workerPortForwardIntercepts(ctx context.Context) error
...
+		pfCtx := dgroup.WithGoroutineName(ctx, fmt.Sprintf("/%s:%d", fk.PodIP, port))
+		wg.Add(1)
+		go tm.portForward(pfCtx, wg, fk, port)
+	}
+	return started
+}
+
+func (tm *trafficManager) portForward(ctx context.Context, wg *sync.WaitGroup, fk forwardKey, port int32) {
+	defer wg.Done()
+	// Using kubectl port-forward here would require the pod name to be either fetched from the API server, or threaded
```
thallgren: ... and it would create additional connections that we need to recover in case of failure. Using the Forwarder is a much better choice because it will reuse the TUN-device and the already existing port-forward.

pkg/client/cli/intercept_state.go
```diff
@@ -407,6 +411,14 @@ func (is *interceptState) createRequest() (*connector.CreateInterceptRequest, er
 		}
 	}
 
+	if is.toPod != "" {
+		port, err := parsePort(is.toPod)
+		if err != nil {
+			return nil, err
+		}
+		spec.ExtraPorts = []int32{int32(port)}
```
thallgren: I would love to see this in plural all the way from how the flag is declared.

pkg/client/connector/intercept.go
```diff
@@ -127,6 +129,41 @@ func (tm *trafficManager) workerPortForwardIntercepts(ctx context.Context) error
...
+		started = true
+		mntCtx := dgroup.WithGoroutineName(ctx, fmt.Sprintf("/%s:%d", fk.PodIP, sftpPort))
+		wg.Add(1)
+		go tm.workerMountForwardIntercept(mntCtx, mountForward{fk, sftpPort}, wg)
+	}
+	for _, port := range extraPorts {
+		started = true
+		pfCtx := dgroup.WithGoroutineName(ctx, fmt.Sprintf("/%s:%d", fk.PodIP, port))
+		wg.Add(1)
+		go tm.portForward(pfCtx, wg, fk, port)
```
thallgren: The two names `workerMountForwardIntercept()` and `portForward()` aren't great as siblings here.

thallgren (changes requested):

thallgren (approved):
Looks great. Love the fact that you include a regression test.

PR #1801, opened 2021-06-17
Move the discovery of the cluster's DNS server and subnets to the traffic-manager

thallgren:
## Description

Moves the discovery of the cluster's DNS server and subnets to the traffic-manager and improves the way its done. The subnet package is reintroduced (by reverting the commit that removed it) and all review comments that were made on that package were accepted and applied here instead.

The move changes the RBAC for both the client and the traffic-manager. Details in the "Adjust RBAC rules" commit.

## Checklist


CHANGELOG.md
```diff
@@ -2,7 +2,13 @@
 
 ### 2.3.2 (TBD)
 
-- Bug: On a MacOS, files generated under /etc/resolver as the result of using include-suffixes in the cluster config, are now properly removed on quit.
+- Change: The list of subnets that the virtual network interface will route is configured dynamically and will follow changes in the cluster.
```
donnyyung, line 5: is there a docs PR w/ release notes for all these changes + any other docs changes? Or did you / will you not have time for that and need help?

donnyyung: Answered offline, Thomas plans to do that tomorrow!

pkg/client/daemon/tunrouter.go
```diff
@@ -83,6 +84,15 @@ type tunRouter struct {
...
 	dnsLocalAddr *net.UDPAddr
 
+	// Cluster subnets reported by the traffic-manager
+	clusterSubnets []*net.IPNet
+
+	// Subnets configured by the user
+	alsoProxySubnets []*net.IPNet
+
+	// Subnets that the router is currently configured with
+	subnets []*net.IPNet
```
LukeShu: Would `curSubnets` be a better name?

pkg/client/daemon/tunrouter.go
```diff
@@ -92,37 +102,94 @@ type tunRouter struct {
...
+		}
+		for _, routedSn := range t.subnets {
+			if subnet.Equal(desiredSn, routedSn) {
+				ln--
+				desiredSubnets[i] = desiredSubnets[ln]
+				break
+			}
+		}
+	}
+	desiredSubnets = desiredSubnets[:ln]
```
LukeShu: I'm not asking you to change it, and maybe it's just me, but I found this code hard to read.  I'm not sure what I'd do about that without making it less efficient. :shrug:

pkg/client/daemon/tunrouter.go
```diff
@@ -92,37 +102,94 @@ type tunRouter struct {
...
+			}
+			ln--
+			t.subnets[i] = t.subnets[ln]
+		}
+	}
+	t.subnets = t.subnets[:ln]
+
+	for _, sn := range desiredSubnets {
+		if err := t.dev.AddSubnet(ctx, sn); err != nil {
+			return fmt.Errorf("failed to remove subnet %s: %w", sn, err)
```
LukeShu: Surely that should say "failed to add subnet".

pkg/client/daemon/tunrouter.go
LukeShu: Perhaps also the comment should say that it should *only* be used within `refreshSubnets()`?

CHANGELOG.md
```diff
@@ -2,7 +2,13 @@
...
+- Change: The client no longer need permission to list nodes.
```
LukeShu: needs

pkg/client/daemon/tunrouter.go
donnyyung: I guess maybe I'm a little confused so maybe adding a comment would be helpful...

but could you check my understanding of the inner for loop.  If the desiredSn (at index i), is already in t.subnets, then we want to "trim" it.  So we decrement the length of the desired subnets by 1 since we are going to "remove" one.  

But in  `desiredSubnets[i] = desiredSubnets[ln]`, we set the ith index to be the new decremented lnth element, but since that element is now in position `i`, will we never check if desiredSubnets[ln] is already included right?

But I see this type of loop is also used in the Unique function below, so I must just be missing something.

pkg/client/connector/service_proxy.go
```diff
@@ -188,6 +188,10 @@ func (p *mgrProxy) AgentLookupHostResponse(ctx context.Context, arg *managerrpc.
...
 }
 
-func (p *mgrProxy) WatchLookupHost(info *managerrpc.SessionInfo, server managerrpc.Manager_WatchLookupHostServer) error {
+func (p *mgrProxy) WatchLookupHost(_ *managerrpc.SessionInfo, server managerrpc.Manager_WatchLookupHostServer) error {
 	return errors.New("WatchLookupHost is not implemented by the mgrProxy")
 }
+
+func (p *mgrProxy) WatchClusterInfo(_ *managerrpc.SessionInfo, server managerrpc.Manager_WatchClusterInfoServer) error {
+	return errors.New("WatchClusterInfo is not implemented by the mgrProxy")
+}
```
LukeShu, line 201: I saw this and got real confused. "Won't the daemon need that!?"  I hadn't realized that the daemon hits the port-forward directly.  We should maybe be changing that.

LukeShu (changes requested):
- Change request: Typo in the changelog
- Change request: Copy/pasteod error message
- Minor suggestion: Naming and comment on `.subnets`

But on the whole :heart:

pkg/client/daemon/tunrouter.go
thallgren: So "ln" is short for "length". When something that should be removed is discovered at position `i`, then that element is overwritten with the last element in the list (effectively `ln - 1` , but `ln` is decreased on the preceding line instead of after). The decrease of `ln` is necessary because the last element is now a duplicate (it's also at position `i`).

At the end of the loop, the slice is truncated using `[:ln]` which effectively removes all the duplicates.

thallgren: I'll try and come up with comments that explains what goes on.

thallgren:
The force push contains the following:
1. Rewrite of the handling of the subnets to make the code easier to read.
2. Fixed typo on the CHANGELOG.md
3. Squashed changes into relevant commits
4. Rebased on top of release/v2 to get rid of merge conflict in changelog.

LukeShu (approved):
`Partition` is a definite readability improvement!

LukeShu:
It looks like CI failed because it ended up with a bum cluster? That shouldn't be happening with k3s, but whatever; I restarted it.

LukeShu:
Woo, CI passed, hit it!

PR #1811, opened 2021-06-23
Wait for daemon shutdown signal when the connector hangs up on streams

josecv:
This is to prevent the daemon from dying too early when the connector
goes away.

## Description

## Checklist

### DIFF
--- a/pkg/client/daemon/tunrouter.go
+++ b/pkg/client/daemon/tunrouter.go
@@ -2,6 +2,7 @@ package daemon
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"math/rand"
 	"net"
@@ -210,7 +211,14 @@ func (t *tunRouter) setOutboundInfo(ctx context.Context, mi *daemon.OutboundInfo
 		}
 
 		dgroup.ParentGroup(ctx).Go("watch-cluster-info", func(ctx context.Context) error {
-			return t.watchClusterInfo(ctx, kubeDNS)
+			err := t.watchClusterInfo(ctx, kubeDNS)
+			var recvErr *client.RecvEOF
+			if errors.As(err, &recvErr) {
+				// If the remote end, which is the connector, has hung up mid-stream, that usually means that
+				// the daemon will be shutting down soon too.
+				<-ctx.Done()
+			}
+			return err
 		})
 	}
 	return nil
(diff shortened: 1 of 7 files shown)


pkg/client/daemon/tunrouter.go
```diff
@@ -210,7 +211,13 @@ func (t *tunRouter) setOutboundInfo(ctx context.Context, mi *daemon.OutboundInfo
 		}
 
 		dgroup.ParentGroup(ctx).Go("watch-cluster-info", func(ctx context.Context) error {
-			return t.watchClusterInfo(ctx, kubeDNS)
+			err := t.watchClusterInfo(ctx, kubeDNS)
+			if errors.Is(err, &client.RecvEOF{}) {
```
LukeShu: `errors.Is` is for error values; `errors.As` is for error types.

So for example, you'd use `errors.Is(err, os.ErrExist)` but `errors.As(err, &os.PathError{})`.

So I'm thinking that this should be `errors.As`.

pkg/client/daemon/tunrouter.go
```diff
@@ -332,7 +339,11 @@ func (t *tunRouter) run(c context.Context) error {
 		}
 		t.connStream = connpool.NewStream(tunnel)
 		dlog.Debug(c, "MGR read loop starting")
-		return t.connStream.DialLoop(c, &t.closing, t.handlers)
+		err = t.connStream.DialLoop(c, &t.closing, t.handlers)
+		if errors.Is(err, &client.RecvEOF{}) {
```
LukeShu: Likewise

pkg/client/stream_error.go
```diff
@@ -0,0 +1,44 @@
...
+func (e *RecvEOF) Unwrap() error {
+	return e.err
+}
+
+// Is implements errors.Is for the RecvEOF type. Any error of type RecvEOF will be reported
+// as being the same as any other.
+func (e *RecvEOF) Is(target error) bool {
+	_, ok := target.(*RecvEOF)
+	return ok
+}
```
LukeShu: If you use `errors.As`, then you you shouldn't need to implement an `Is` or `As` method.

LukeShu (changes requested):
Minor change regarding `errors.Is` vs `errors.As`.  Otherwise: Woohoo!

LukeShu (approved):
Hit it!

PR #1812, opened 2021-06-23
Build system clean-up (including: shellcheck)

LukeShu:
## Description

1. Get rid of the old `file-pr.go` script
2. Clean up/organize the Makefiles
3. Lint the shell scripts with `shellcheck`

As usual, I suggest a commit-by-commit review.  GitHub's webui is having trouble with some of the file renames.

## Checklist


build-aux/tools.mk
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Install tools used by the build
+# This file deals installing programs used by the build.
```
donnyyung: nit: deals with*

build-aux/main.mk
```diff
@@ -12,15 +12,15 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
```
donnyyung, line 14: I know this is the "main" file, but like you've added in the other *.mk files, I think a blurb about the purpose of this file would be helpful.. something like: "This file deals with the linting, testing, building, publishing, and releasing of Telepresence artifacts". 

 I think it's good to have a specified description for each file so if some new _thing_ comes along we know whether it makes sense to put it in here or a different *.mk file.. this will also prevent main.mk from being a catch-all location.

donnyyung (approved):
Just a few tiny comment-related nits, but feel free to take it or leave it.  Thanks for adding this and introducing me to shellcheck!

donnyyung:
also, I tested the smoke tests a few times on macOS, as well as various make commands that I use when developing / testing and encountered no issues or unexpected behavior.

LukeShu:
I also ran the smoke tests.

PR #1817, opened 2021-06-24
telepresence_test: set up kubeconfig in main goroutine

josecv:
## Description

This seems to eliminate a bug where, for some reason, the
buildExecutable goroutine destroys the telepresence-test-user from the
kubeconfig.

## Checklist

### DIFF
--- a/pkg/client/cli/telepresence_test.go
+++ b/pkg/client/cli/telepresence_test.go
@@ -102,38 +102,12 @@ func (ts *telepresenceSuite) SetupSuite() {
 		ts.NoError(err)
 	}()
 
-	wg.Add(1)
-	go func() {
-		defer wg.Done()
-
-		kubeconfig := dtest.Kubeconfig(ctx)
-		os.Setenv("DTEST_KUBECONFIG", kubeconfig)
-		os.Setenv("KUBECONFIG", kubeconfig)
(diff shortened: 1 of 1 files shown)


pkg/client/cli/telepresence_test.go
```diff
@@ -102,39 +102,10 @@ func (ts *telepresenceSuite) SetupSuite() {
...
-		ts.NoError(err)
-
-		// We start with the default context, and will switch to the
-		// telepresence-test-developer user later in the tests
-		err = run(ctx, "kubectl", "config", "use-context", "default")
-		ts.NoError(err)
-	}()
 	wg.Wait()
 
+	ts.setupKubeConfig(ctx)
```
donnyyung, line 110: I would add a comment here: "We set up the KubeConfig outside of the goroutines above because there seems to be a bug where buildExecutable sometimes destroys the test user that is created in this function". 

Just so that someone in the future isn't like, "oh we should be doing this in parallel with the other setup steps" and then causes the issue to happen again.

donnyyung (approved):
LGTM just left a suggestion to add a comment so people in the future know why we did this!.

pkg/client/cli/telepresence_test.go
josecv: ah, definitely, i'll add it

PR #1820, opened 2021-06-25
Run unit tests via richgo so the output is colourized

josecv:

## Description

## Checklist


build-aux/main.mk
```diff
@@ -153,8 +153,8 @@ format: $(tools/golangci-lint) $(tools/protolint) ## (QA) Automatically fix lint
 	$(tools/protolint) lint --fix rpc || true
 
 .PHONY: check
-check: $(tools/ko) ## (QA) Run the test suite
-	go test ./...
+check: $(tools/ko) $(tools/richgo) ## (QA) Run the test suite
```
donnyyung, line 156: I don't have the strongest opinions on this, but since we do lose some information (e.g. the times in the output), maybe it's worth adding this as "check-colorized", so when you are running it locally you can choose which one you want?  But I think maybe Luke or Thomas will have stronger opinions, so I'll defer to them.

josecv:
going to close this as it doesn't really add all that much value and it brings in new dependencies

PR #1822, opened 2021-06-25
Add nightly build

donnyyung:
## Description

Adding nightly builds to our telepresence ci. This PR depends on this change in saas_app (https://github.com/datawire/saas_app/pull/1364) which will enable the nightly build to be consumed.

## Checklist

 - [ ] I made sure to either submit a docs PR, or tell Matt about the necessary documentation changes. - TBD

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -176,6 +176,27 @@ jobs:
     - checkout
     - run: |
         [[ $CIRCLE_TAG == *-* ]] || BUCKET_DIR="charts" make push-chart
+  
+  "tag-nightly":
+    executor: vm-linux
+    steps:
+    - checkout
+    - install-go
+    - run:
+        name: "Tag nightly"
+        command: |
+          existingTags=$(git tag --points-at HEAD)
+          if [[ -n $existingTags ]]; then
+              echo "Tag for the HEAD of release/v2 already exists:"
+              echo $existingTags
+              exit 0
+          else
+              lastTag=$(git describe --tags --match="v*" --abbrev=0
+              time=$(date +"%s")
+              newTag="$lastTag-nightly-$time"
+              make prepare-release TELEPRESENCE_VERSION=$newTag
+              git push origin {,rpc/}$time
+          fi
 
 #  "release-finalize":
 #    executor: vm-linux
@@ -215,6 +236,17 @@ workflows:
               only: /^v.*/
             branches:
               ignore: /.*/
+
+  'Nightly':
+    triggers:
+    - schedule:
+        cron: "0 0 * * 1-5"
+        filters:
+          branches:
+            only:
+              - release/v2
+    jobs:
+      - tag-nightly
       #- release-finalize:
       #    context: telepresence2-release
       #    requires:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -159,6 +159,7 @@ jobs:
       - run: TELEPRESENCE_VERSION=$CIRCLE_TAG make push-images push-executable
       - run: |
           [[ $CIRCLE_TAG == *-* ]] || TELEPRESENCE_VERSION=$CIRCLE_TAG make promote-to-stable
+          [[ $CIRCLE_TAG == *-nightly-* ]] && TELEPRESENCE_VERSION=$CIRCLE_TAG make promote-nightly
 
   "release-macos":
     executor: vm-macos
@@ -169,6 +170,7 @@ jobs:
       - run: TELEPRESENCE_VERSION=$CIRCLE_TAG make push-executable
       - run: |
           [[ $CIRCLE_TAG == *-* ]] || TELEPRESENCE_VERSION=$CIRCLE_TAG make promote-to-stable
+          [[ $CIRCLE_TAG == *-nightly-* ]] && TELEPRESENCE_VERSION=$CIRCLE_TAG make promote-nightly
 
   "release-chart":
     executor: vm-linux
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -159,6 +160,28 @@ jobs:
...
+        command: |
+          existingTags=$(git tag --points-at HEAD)
+          if [[ -n $existingTags ]]; then
+              echo "Tag for the HEAD of release/v2 already exists:"
+              echo $existingTags
+              exit 0
+          else
+              lastTag=$(git describe --tags --match="v*" --abbrev=0
+              time=$(date +"%s")
+              newTag="$lastTag-nightly-$time"
```
LukeShu: 1. Isn't this going to recurse?  On the first night it works correct, and tags `v2.3.2-nightly-1624775772`.  Then on the second day it tags `v2.3.2-nightly-1624775772-nightly-1624862232`, and so on?
2. I'd prefer to have yyyy.mm.dd rather than seconds since the epoch.  Make it easier for people to talk about which one they have.
3. That tag format isn't semver.  `vX.Y.Z-P` indicates something *before* X.Y.Z, but we're actually tagging something *after* it.

build-aux/main.mk
```diff
@@ -132,6 +132,18 @@ ifeq ($(GOHOSTOS), darwin)
 	packaging/homebrew-package.sh $(patsubst v%,%,$(TELEPRESENCE_VERSION))
 endif
 
+# Prerequisites:
+# The awscli command must be installed and configured with credentials to upload
+# to the datawire-static-files bucket.
+.PHONY: promote-nightly
+promote-nightly: ## (Release) Update stable.txt in S3
```
LukeShu: Doc `## comment` needs to be edited to say `nightly.txt`

.circleci/config.yml
```diff
@@ -149,6 +149,7 @@ jobs:
       - run: TELEPRESENCE_VERSION=$CIRCLE_TAG make push-images push-executable
       - run: |
           [[ $CIRCLE_TAG == *-* ]] || TELEPRESENCE_VERSION=$CIRCLE_TAG make promote-to-stable
+          [[ $CIRCLE_TAG == *-nightly-* ]] && TELEPRESENCE_VERSION=$CIRCLE_TAG make promote-nightly
```
LukeShu: I have mixed feelings about creating actual Git tags, instead of just putting the shorthash in the docker tag.

.circleci/config.yml
LukeShu: I'm thinking
```console
$ git describe --tags --match 'v*' --exclude='*-*'|sed "s/-/-nightly.$(date +%Y.%m.%d)-/"
v2.3.2-nightly.2021.06.27-19-ge9b3c27f
```
but with the Z bumped.  Probably reuse `genversion.go`?

.circleci/config.yml
LukeShu: Using `&&` chains make me nervous.  Because `set -e` is weird, they won't cause it to bail, but if the condition doesn't trigger it will set `$?` to non-zero, which I think will cause CircleCI to consider the step to have failed.

So I guess change this to `[[ $CIRCLE_TAG != *-nightly-* ]] ||`.

.circleci/config.yml
```diff
@@ -159,6 +160,28 @@ jobs:
       - run: TELEPRESENCE_VERSION=$CIRCLE_TAG make push-executable
       - run: |
           [[ $CIRCLE_TAG == *-* ]] || TELEPRESENCE_VERSION=$CIRCLE_TAG make promote-to-stable
+          [[ $CIRCLE_TAG == *-nightly-* ]] && TELEPRESENCE_VERSION=$CIRCLE_TAG make promote-nightly
```
LukeShu: Likewise.

LukeShu (changes requested):
.

.circleci/config.yml
donnyyung: 1. I think I misunderstood how --abbrev=0 worked, so I believe you are right.
2. Noted + made that change.
3. Ah at last job we used vX.Y.Z-P to be something after vX.Y.Z and before vX.Y.Z+1, so I guess we were using semver wrong, thanks for letting me know.

I originally didn't use `genversion.go` since the format of the tags was different, but it probably makes sense to include all our version-related things there, so have updated it.

.circleci/config.yml
donnyyung: I have removed the tagging and put building + pushing + promoting of artifacts in the "publish-*-nightly" jobs for linux + macOS.

build-aux/genversion.go
```diff
@@ -39,9 +39,16 @@ func Main() error {
 	}
 	gitDescVer.Patch++
 
-	_, err = fmt.Printf("v%s-%d\n", gitDescVer, time.Now().Unix())
-	if err != nil {
-		return err
+	// If an additional arg has been used, we include it in the tag
+	if len(os.Args) >= 2 {
+		t := time.Now()
+		date := fmt.Sprintf("%d.%d.%d", t.Year(), t.Month(), t.Day())
```
LukeShu: This would be better written as `date := t.Format("2006.01.02")`

build-aux/genversion.go
```diff
@@ -39,9 +39,16 @@ func Main() error {
...
+		_, err = fmt.Printf("v%d.%d.%d-%s-%s-%s\n", gitDescVer.Major, gitDescVer.Minor, gitDescVer.Patch, os.Args[1], date, gitDescVer.Pre[0].String())
```
LukeShu: `gitDescVer.Pre[0]` is the number of commits since the previous tag, and `gitDescVer.Pre[1]` is the short hash of the commit (with a `g` prefix), right?

I'd think it would be sufficiently useful to include the shorthash, so it's easier to figure out which commit a nightly version was built from.

LukeShu (approved):
I've got some nitpicks about `genversion.go` that you can ignore if you want.

On the whole, lgtm!

build-aux/genversion.go
donnyyung: This is good to know for the future, but I'm going to remove the "date" since it's sort of irrelevant when the git commit hash is there (and actually means something)... and why add extra bloat to the tags.

PR #1827, opened 2021-06-29
Support configuring the traffic manager's namespace

josecv:

## Description

Support configuring the traffic manager's namespace

## Checklist

### DIFF
--- a/pkg/client/connector/k8s_config.go
+++ b/pkg/client/connector/k8s_config.go
@@ -12,6 +12,7 @@ import (
 	"k8s.io/client-go/rest"
 
 	"github.com/datawire/ambassador/pkg/kates"
+	"github.com/telepresenceio/telepresence/v2/pkg/client"
 	"github.com/telepresenceio/telepresence/v2/pkg/iputil"
 )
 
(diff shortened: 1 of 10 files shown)


pkg/client/connector/k8s_config.go
```diff
@@ -15,6 +16,14 @@ import (
 	"github.com/telepresenceio/telepresence/v2/pkg/iputil"
 )
 
+// this is modified in tests
+var defaultManagerNamespace = func() string {
+	if ns := os.Getenv("TELEPRESENCE_MANAGER_NAMESPACE"); ns != "" {
```
thallgren: This ENV variable and it's default setting should be moved into `client.Env`.

thallgren (changes requested):
Looks good overall, but it would be nice if we could also get rid of the `os.Getenv()` call.

thallgren (approved):

thallgren:
Looks good now. I would prefer though, if squashing commits was done after I've got a chance to review the requested changes so that I don't need to review all of it once again.

thallgren (approved):

PR #1831, opened 2021-06-30
Make intercept recover from deletion of intercepted pod.

thallgren:
## Description

This PR will make intercepts survive when the pod that is currently intercepted is deleted. The intercept will be dormant until a new pod is ready to accept responsibility for it (which may be immediately in case the replica count was greater than 1). The PR contains necessary changes to all binaries, client, manager, and agent. Everything is divided into concise logical and fairly small pieces and a commit by commit review is strongly advised.

## Checklist

 - [ ] I made sure to submit a docs PR.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -59,6 +59,15 @@ commands:
             curl -L --fail -o /tmp/kubectl https://storage.googleapis.com/kubernetes-release/release/v<< parameters.version >>/bin/$(uname -s | tr A-Z a-z)/amd64/kubectl
             sudo install /tmp/kubectl /usr/local/bin/kubectl
 
+  "install-sshfs":
+    steps:
+      - run:
+          name: "Install sshfs"
+          command: |
+            sudo apt-get update
+            sudo apt-get install -y sshfs
+            sudo sh -c 'echo user_allow_other >> /etc/fuse.conf'
+
   "dirty-check":
     steps:
       - run:
@@ -106,6 +115,7 @@ jobs:
       - install-go
       - run: make build
       - install-kubectl
+      - install-sshfs
       - run:
           command: make test
           # Both CircleCI and `go test` itself time out after 10m by
--- a/pkg/client/cli/telepresence_test.go
+++ b/pkg/client/cli/telepresence_test.go
@@ -978,13 +978,19 @@ func (is *interceptedSuite) TestE_StopInterceptedPodOfMany() {
 	}, 5*time.Second, time.Second)
 
 	// Verify response from intercepting client
-	hc := http.Client{Timeout: time.Second}
-	resp, err := hc.Get("http://hello-0")
-	require.NoError(err)
-	defer resp.Body.Close()
-	body, err := ioutil.ReadAll(resp.Body)
-	assert.NoError(err)
-	assert.Equal("hello-0 from intercept at /", string(body))
+	assert.Eventually(func() bool {
+		hc := http.Client{Timeout: time.Second}
+		resp, err := hc.Get("http://hello-0")
+		if err != nil {
+			return false
+		}
+		defer resp.Body.Close()
+		body, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			return false
+		}
+		return "hello-0 from intercept at /" == string(body)
+	}, 5*time.Second, time.Second)
 
 	// Verify that volume mount is restored
 	st, err := os.Stat(filepath.Join(is.mountPoint, "var"))
--- a/pkg/client/cli/telepresence_test.go
+++ b/pkg/client/cli/telepresence_test.go
@@ -348,7 +348,11 @@ func (cs *connectedSuite) ns() string {
 func (cs *connectedSuite) SetupSuite() {
 	require := cs.Require()
 	c := dlog.NewTestContext(cs.T(), false)
-	cs.NoError(cs.tpSuite.kubectl(c, "config", "use-context", "telepresence-test-developer"))
+
+	cs.Eventually(func() bool {
+		return run(c, "kubectl", "config", "use-context", "telepresence-test-developer") == nil
+	}, 5*time.Second, time.Second)
+
 	stdout, stderr := telepresence(cs.T(), "connect")
 	require.Empty(stderr)
 	require.Contains(stdout, "Connected to context")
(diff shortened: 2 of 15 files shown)


.circleci/config.yml
```diff
@@ -54,11 +54,20 @@ commands:
         default: "1.19.3"
     steps:
       - run:
-          name: "Install kubectl << parameters.version >>"
+          name: "Install sshfs << parameters.version >>"
```
donnyyung: This should still say `install kubectl` right?

donnyyung:
Thomas I did a once over of this tonight but I want to do some local testing and digging deeper into the meatier commits.  I'll do that first thing in the morning so you can get feedback asap.

donnyyung:
Okay two things from my manual testing: 
1. I know this is covered in the tests and it passes, but when I scale the deployment from 1 -> 0 - > 1, telepresence list still reports `NO_AGENT` even after the pod comes up. But what is weird is if I curl the service, it does still appear as if it's intercepted.

2. If I have a deployment that has two replicas and I intercept it, all requests go to intercepted service. When I `kubectl delete` one of the pod, 50% of requests go to intercepted service and the other half go to the service in the cluster. When I `kubectl delete` the other pod, all requests go to the service in the cluster.

I'm now going to do another run through of the PR if I can see where this is happening, but just wanted to note what I'm seeing.

pkg/client/cli/telepresence_test.go
```diff
@@ -859,6 +871,115 @@ func (is *interceptedSuite) TestB_ListingActiveIntercepts() {
...
+				if zp == currentPod {
+					return false
+				}
+			}
+			return true
+		}, 5*time.Second, time.Second)
+
+	// Verify that intercept is still active
+	assert.Eventually(func() bool {
+		stdout, _ := telepresence(is.T(), "--namespace", is.ns(), "list", "--intercepts")
```
donnyyung, line 974: One thing I'd add to this test is to curl the service to validate that you are getting the intercepted payload and not what's in the cluster (E.g. in case 2 of my comment on the PR, it says "ACTIVE" for the intercept but all requests are going to the service in the cluster).

donnyyung:
Screen Shot 2021-07-01 at 10 32 30 AM
One other weird thing, I don't seem to be able to do 2 local intercepts anymore, the 2nd always times out and then when I try to leave them, it says that I left (subsequent calls to `telepresence leave` fail saying the intercept doesn't exist) but the cli doesn't update with that information.

pkg/client/connector/intercept.go
```diff
@@ -154,11 +269,7 @@ func (tm *trafficManager) addIntercept(c context.Context, ir *rpc.CreateIntercep
 	}
 
 	<-tm.startup
-	intercepts, err := actions.ListMyIntercepts(c, tm.managerClient, tm.session().SessionId)
-	if err != nil {
-		return nil, err
-	}
-	for _, iCept := range intercepts {
+	for _, iCept := range tm.getCurrentIntercepts() {
```
donnyyung, line 277: Just to check my understanding.. in the past we limited the intercepts we iterated through to be the ones associated with the session ID for the client. But since in the loop we are checking if the names are the same (which implicitly includes the sessionID) or that the targetPort + targetHost are already in use (implicitly checking for the client the call is coming from), we can replace this with iterating through _all_ the intercepts without issue?

donnyyung:
The most recent commit fixed issue 2.. issues 1 and screenshot are due to sshfs not ending, making the watch hang.

pkg/client/connector/intercept.go
thallgren: The `WatchInterecps()`, when called from a client, will only include those intercepts that are associated with that client. The traffic-manager ensures that all others are rejected [by this filter](https://github.com/telepresenceio/telepresence/blob/thallgren/intercept-survival/cmd/traffic/cmd/manager/service.go#L224-L227).

thallgren:
Changes since @donnyyung last reviewed:
```patch
diff --git a/pkg/client/cli/cliutil/daemon.go b/pkg/client/cli/cliutil/daemon.go
index b0ae8902..6657dbaf 100644
--- a/pkg/client/cli/cliutil/daemon.go
+++ b/pkg/client/cli/cliutil/daemon.go
@@ -61,8 +61,14 @@ func launchDaemon(ctx context.Context, dnsIP string) error {
                // use `sudo`'s `--prompt` flag for this because (1) we don't want it to be
                // re-displayed if they typo their password, and (2) it might be ignored anyway
                // depending on `passprompt_override` in `/etc/sudoers`.  So we'll do a pre-flight
-               // `sudo --non-interactive --validate` to decide whether to display it.
-               needPwCmd := dexec.CommandContext(ctx, "sudo", "--non-interactive", "--validate")
+               // `sudo --non-interactive true` to decide whether to display it.
+               //
+               // Note: Using `sudo --non-interactive --validate` does not work well in situations
+               // where the user has configured `myuser ALL=(ALL:ALL) NOPASSWD: ALL` in the sudoers
+               // file. Hence the use of `sudo --non-interactive true`. A plausible cause can be
+               // found in the first comment here:
+               // https://unix.stackexchange.com/questions/50584/why-sudo-timestamp-is-not-updated-when-nopasswd-is-set
+               needPwCmd := dexec.CommandContext(ctx, "sudo", "--non-interactive", "true")
                needPwCmd.DisableLogging = true
                if err := needPwCmd.Run(); err != nil {
                        fmt.Printf("Need root privileges to run: %s\n", logging.ShellString(args[0], args[1:]))
diff --git a/pkg/client/cli/telepresence_test.go b/pkg/client/cli/telepresence_test.go
index c639985a..b03e4671 100644
--- a/pkg/client/cli/telepresence_test.go
+++ b/pkg/client/cli/telepresence_test.go
@@ -977,6 +977,15 @@ func (is *interceptedSuite) TestE_StopInterceptedPodOfMany() {
                return false
        }, 5*time.Second, time.Second)
 
+       // Verify response from intercepting client
+       hc := http.Client{Timeout: time.Second}
+       resp, err := hc.Get("http://hello-0")
+       require.NoError(err)
+       defer resp.Body.Close()
+       body, err := ioutil.ReadAll(resp.Body)
+       assert.NoError(err)
+       assert.Equal("hello-0 from intercept at /", string(body))
+
        // Verify that volume mount is restored
        st, err := os.Stat(filepath.Join(is.mountPoint, "var"))
        require.NoError(err, "Stat on <mount point>/var failed")
```

pkg/client/cli/telepresence_test.go
thallgren: Added.

donnyyung (approved):
No longer seeing any of the issues I mentioned :) Thanks for debugging with me and good work!

PR #1835, opened 2021-07-02
Changing configuration for agent image + registry

donnyyung:
## Description
I'm having trouble runnings tests locally so just PR-ing this so I can see how the tests run in circle, don't review this yet cuz I gotta clean up some of the commits! 

## Checklist

### DIFF
--- a/pkg/client/config.go
+++ b/pkg/client/config.go
@@ -344,6 +344,7 @@ func (ll *LogLevels) merge(o *LogLevels) {
 type Images struct {
 	Registry          string `json:"registry,omitempty"`
 	AgentImage        string `json:"agentImage,omitempty"`
+	WebhookRegistry   string `json:"webhookRegistry,omitempty"`
 	WebhookAgentImage string `json:"webhookAgentImage,omitempty"`
 }
 
@@ -366,6 +367,8 @@ func (img *Images) UnmarshalYAML(node *yaml.Node) (err error) {
 			img.Registry = v.Value
 		case "agentImage":
 			img.AgentImage = v.Value
+		case "webhookRegistry":
+			img.WebhookRegistry = v.Value
 		case "webhookAgentImage":
 			img.WebhookAgentImage = v.Value
 		default:
@@ -386,6 +389,9 @@ func (i *Images) merge(o *Images) {
 	}
 	if o.Registry != "" {
 		i.Registry = o.Registry
+	}
+	if o.WebhookRegistry != "" {
+		i.WebhookRegistry = o.WebhookRegistry
 	}
 }
 
@@ -446,6 +452,7 @@ var defaultConfig = Config{
 	},
 	Images: Images{
 		Registry:          "docker.io/datawire",
+		WebhookRegistry:   "docker.io/datawire",
 		AgentImage:        "",
 		WebhookAgentImage: "",
 	},
--- a/pkg/client/config.go
+++ b/pkg/client/config.go
@@ -58,19 +58,16 @@ func (c *Config) UnmarshalYAML(node *yaml.Node) (err error) {
 			if err != nil {
 				return err
 			}
-			continue
 		case kv == "logLevels":
 			err := ms[i+1].Decode(&c.LogLevels)
 			if err != nil {
 				return err
 			}
-			continue
 		case kv == "images":
 			err := ms[i+1].Decode(&c.Images)
 			if err != nil {
 				return err
 			}
-			continue
 		case parseContext != nil:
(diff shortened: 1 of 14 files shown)


pkg/install/resource/tm_deployment.go
```diff
@@ -43,12 +43,13 @@ func (ri *tmDeployment) desiredDeployment(ctx context.Context) *kates.Deployment
...
 
 	sc := getScope(ctx)
+	imgConfig := client.GetConfig(ctx).Images
 	var containerEnv = []corev1.EnvVar{
 		{Name: "LOG_LEVEL", Value: "info"},
 		{Name: "SYSTEMA_HOST", Value: sc.env.SystemAHost},
 		{Name: "SYSTEMA_PORT", Value: sc.env.SystemAPort},
 		{Name: "CLUSTER_ID", Value: sc.clusterID},
-		{Name: "TELEPRESENCE_REGISTRY", Value: sc.env.Registry},
+		{Name: "TELEPRESENCE_REGISTRY", Value: imgConfig.Registry},
```
thallgren: This (and the agent image further down) is somewhat problematic from a life-cycle perspective. If the manager is installed and a user's config is modified, then that modification will have no effect because it doesn't cause the manager deployment to change. I know that this problem was already there when modifying the env vars but that was not something that we made apparent to the user the way we do now.

I don't have a good solution for this but I think it's important that there's a caveat and an explanation in the docs.

pkg/client/config.go
```diff
@@ -50,21 +52,26 @@ func (c *Config) UnmarshalYAML(node *yaml.Node) (err error) {
...
 			return err
 		}
-		if kv == "timeouts" {
+		switch {
+		case kv == "timeouts":
 			err := ms[i+1].Decode(&c.Timeouts)
 			if err != nil {
 				return err
 			}
 			continue
```
thallgren:
```suggestion
```

pkg/client/config.go
```diff
@@ -50,21 +52,26 @@ func (c *Config) UnmarshalYAML(node *yaml.Node) (err error) {
...
-		}
-		if kv == "logLevels" {
+		case kv == "logLevels":
 			err := ms[i+1].Decode(&c.LogLevels)
 			if err != nil {
 				return err
 			}
 			continue
```
thallgren:
```suggestion
```

pkg/client/config.go
```diff
@@ -50,21 +52,26 @@ func (c *Config) UnmarshalYAML(node *yaml.Node) (err error) {
...
-		}
-		if parseContext != nil {
+		case kv == "images":
+			err := ms[i+1].Decode(&c.Images)
+			if err != nil {
+				return err
+			}
+			continue
```
thallgren:
```suggestion
```

pkg/client/config.go
```diff
@@ -343,6 +392,10 @@ var defaultConfig = Config{
 	LogLevels: LogLevels{
 		UserDaemon: logrus.DebugLevel,
 		RootDaemon: logrus.InfoLevel,
+	},
+	Images: Images{
+		Registry:   "docker.io/datawire",
+		AgentImage: "",
```
thallgren: Shouldn't these defaults be initialized from the current environment variables? I think it's either that or remove them from the env.

thallgren (changes requested):

pkg/tun/device_darwin.go
```diff
@@ -143,9 +143,9 @@ type addrIfReq6 struct {
 	addr           unix.RawSockaddrInet6
 	dest           unix.RawSockaddrInet6
 	mask           unix.RawSockaddrInet6
-	flags          int32 // nolint:structcheck
-	expire         int64 // nolint:structcheck
-	preferred      int64 // nolint:structcheck
+	flags          int32 //nolint:structcheck
+	expire         int64 //nolint:structcheck
+	preferred      int64 //nolint:structcheck
```
thallgren: You can skip those three. This fix is already in `release/v2`.

pkg/client/config.go
donnyyung: Ah yeah sorry about that, I hadn't gotten to the point where I removed the environment variables.  Will be in the next commit.

pkg/install/resource/tm_deployment.go
donnyyung: I think I have tackled this in a future commit by separating the agentImage key into agentImage and webhookAgentImage so it's not one overloaded variable that is easier to distinguish between in the documentation.

Although revisiting this now, I might need to break up imgConfig.Registry into two variables as well.

Anyways, let me know what you think of this solution in the later commits.

thallgren (approved):
All looks good.

PR #1837, opened 2021-07-05
Fix garbage in log statement caused by %s on []byte.

thallgren:
This fixes a regression in a PR that hasn't been merged but not
released yet so no CHANGELOG statement is needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+### 2.3.4 (TBD)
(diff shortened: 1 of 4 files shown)


LukeShu (approved):
I'd just cast it before printing it, instead of introducing another variable.
```go
dlog.Infof(c, "Automatically set -dns=%s", net.IP(o.dnsConfig.LocalIp))
```
Either way, lgtm

PR #1839, opened 2021-07-06
fix the out-of-range issue

johnlanni:
## Description

this panic will happen when multiple services match the deployment.

image

## Checklist

### DIFF
--- a/pkg/install/discovery.go
+++ b/pkg/install/discovery.go
@@ -96,8 +96,8 @@ func FindMatchingService(c context.Context, client *kates.Client, portNameOrNumb
 	portRef := ""
 	if len(matchingSvcs) > 0 {
 		svcNames := make([]string, 0, len(matchingSvcs))
-		for i, svc := range matchingSvcs {
-			svcNames[i] = svc.Name
+		for _, svc := range matchingSvcs {
+			svcNames = append(svcNames, svc.Name)
 		}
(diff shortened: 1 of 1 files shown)


pkg/install/discovery.go
```diff
@@ -96,8 +96,8 @@ func FindMatchingService(c context.Context, client *kates.Client, portNameOrNumb
 	portRef := ""
 	if len(matchingSvcs) > 0 {
 		svcNames := make([]string, 0, len(matchingSvcs))
-		for i, svc := range matchingSvcs {
-			svcNames[i] = svc.Name
+		for _, svc := range matchingSvcs {
+			svcNames = append(svcNames, svc.Name)
```
thallgren: Using `append` when the size is know isn't ideal since it creates a new slice instance for each iteration. The `for` loop is correct. The mistake is on the previous line which should't have a middle zero argument. It should just say:
```
svcNames := make([]string, len(matchingSvcs))
```

thallgren (changes requested):

pkg/install/discovery.go
johnlanni: Thanks for your guidance.

thallgren (approved):

thallgren:
Thank's @johnlanni for fixing this. Contributions like this are very much appreciated.

PR #1843, opened 2021-07-06
Improve debug logging of intercept filter.

thallgren:
This logging was helpful when debugging intercept reconnect issues so
I thought I'd leave it in.

### DIFF
--- a/cmd/traffic/cmd/manager/service.go
+++ b/cmd/traffic/cmd/manager/service.go
@@ -205,20 +205,22 @@ func (m *Manager) WatchIntercepts(session *rpc.SessionInfo, stream rpc.Manager_W
 		filter = func(id string, info *rpc.InterceptInfo) bool {
 			// Don't return intercepts for different agents.
 			if info.Spec.Namespace != agent.Namespace || info.Spec.Agent != agent.Name {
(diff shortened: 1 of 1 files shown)


cmd/traffic/cmd/manager/service.go
```diff
@@ -205,20 +205,22 @@ func (m *Manager) WatchIntercepts(session *rpc.SessionInfo, stream rpc.Manager_W
...
 			}
 			// Don't return intercepts that aren't in a "agent-owned" state.
 			switch info.Disposition {
-			case rpc.InterceptDispositionType_WAITING:
-			case rpc.InterceptDispositionType_ACTIVE:
-			case rpc.InterceptDispositionType_AGENT_ERROR:
+			case rpc.InterceptDispositionType_WAITING,
+				rpc.InterceptDispositionType_ACTIVE,
+				rpc.InterceptDispositionType_AGENT_ERROR:
 				// agent-owned state: continue along
```
donnyyung: Since we're moving the return, maybe change this comment to be `// agent-owned state: include the intercept` since we're returning.

donnyyung (approved):
One nit, feel free to take it or leave it

PR #1844, opened 2021-07-06
Fix bug where sshfs would be terminated prematurely

josecv:

## Description

## Checklist

### DIFF
--- a/pkg/client/connector/intercept.go
+++ b/pkg/client/connector/intercept.go
@@ -88,15 +88,14 @@ func (lpf livePortForwards) start(ctx context.Context, tm *trafficManager, ii *m
 		Name:  ii.Spec.Name,
 		PodIP: ii.PodIp,
 	}
-	if _, isLive := lpf.live[fk]; !isLive {
-		pfCtx, pfCancel := context.WithCancel(ctx)
-		livePortForward := &livePortForward{cancel: pfCancel}
-		forwarded := tm.startForwards(pfCtx, &livePortForward.wg, fk, ii.SftpPort, ii.Spec.ExtraPorts)
-		if forwarded {
-			lpf.snapshot[fk] = struct{}{}
+	if tm.shouldForward(ii) {
+		lpf.snapshot[fk] = struct{}{}
+		if _, isLive := lpf.live[fk]; !isLive {
+			pfCtx, pfCancel := context.WithCancel(ctx)
+			livePortForward := &livePortForward{cancel: pfCancel}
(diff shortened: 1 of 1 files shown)


pkg/client/connector/intercept.go
```diff
@@ -88,15 +88,14 @@ func (lpf livePortForwards) start(ctx context.Context, tm *trafficManager, ii *m
...
+	if tm.shouldForward(ii) {
+		lpf.snapshot[fk] = struct{}{}
+		if _, isLive := lpf.live[fk]; !isLive {
+			pfCtx, pfCancel := context.WithCancel(ctx)
+			livePortForward := &livePortForward{cancel: pfCancel}
+			tm.startForwards(pfCtx, &livePortForward.wg, fk, ii.SftpPort, ii.Spec.ExtraPorts)
+			dlog.Debugf(ctx, "Started forward for %+v", fk)
 			lpf.live[fk] = livePortForward
-		} else {
-			pfCancel()
```
donnyyung, line 99: Just to check my understanding.. this was the main cause of the bug bc we would cancel the portForward before the intercept was removed.  So now we only cancel portforwards in the `cancelUnwanted` method that only closes portForwards that don't have an associated intercept.

josecv: @donnyyung ah good question -- not quite; so basically the `livePortForwards` struct carries a list of [desired](https://github.com/telepresenceio/telepresence/blob/f1e38614a71de5a522b70984134b92a98f4e6755/pkg/client/connector/intercept.go#L78) and [live](https://github.com/telepresenceio/telepresence/blob/f1e38614a71de5a522b70984134b92a98f4e6755/pkg/client/connector/intercept.go#L75) port-forwards

The root of the issue is that the list of desired intercepts is [wiped](https://github.com/telepresenceio/telepresence/blob/f1e38614a71de5a522b70984134b92a98f4e6755/pkg/client/connector/intercept.go#L177) every time we receive a set of snapshots from the traffic manager. However, it was only being populated [when the intercept wasn't already live](https://github.com/telepresenceio/telepresence/blob/release/v2/pkg/client/connector/intercept.go#L91-L97). So if an intercept called for a portforward, but was already live, it wouldn't be registered in the list of desired intercepts, and so `cancelUnwanted` would cancel it

That's why this commit moves the `lpf.snapshot[fk] = struct{}{}` statement outside of the check for whether the port-forward is live. After all, the port-forward is still "desired" even if it's already live.

donnyyung (approved):
Thanks for this!

PR #1845, opened 2021-07-07
Allow installing multiple traffic managers in multiple namespaces

josecv:
## Description

## Checklist


thallgren:
Isn't there a need to also fix the manifests that the auto-installer creates and maintains? I'm thinking about everything under `pkg/install/resource`.

josecv:
@thallgren so I think we'd decided on making this possible only through the helm chart, since an ops team that wants to manage multiple traffic-manager instances would probably want to do it through some kind of configuration as code anyway.

So this leaves the auto-installer's manifests to only ever create cluster-wide traffic managers, and when they notice that they conflict with a helm-managed manifest, they ignore it and move on. I did make some changes to those resources just to make sure they match the names of the resources in the helm chart

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ### 2.3.4 (July 9, 2021)
 
+- Feature: The Telepresence Helm chart now supports installing multiple
```
donnyyung, line 9: nit: Make sure we move this to 2.3.5 (TBD)

charts/telepresence/templates/trafficManagerRbac/namespace-scope.yaml
```diff
@@ -1,10 +1,10 @@
+{{- if .Values.managerRbac.create }}
 # This contains the serviceAccount used for the traffic-manager deployment,
-# as well as the various roles + bindings that the traffic-manager needs.
+# as well as the various namespace-scoped roles + bindings that the traffic-manager needs.
 # This will likely expand over time as we move more things from the clients
 # domain into the traffic-manager.  But the good news there is that it will
 # require less permissions in clientRbac.yaml
 
-{{- if .Values.managerRbac.create }}
 apiVersion: v1
```
donnyyung: Just a suggestion: 

I know you didn't touch this section, but since this serviceAccount is used in the `namespace-scope.yaml` and the `cluster-scope.yaml`, maybe we should move the serviceAccount into its own yaml? Just so it doesn't seem like this is _only_ used for the objects in namespace-scope.yaml.

charts/telepresence/templates/clientRbac/namespace-scope.yaml
```diff
@@ -0,0 +1,41 @@
+# These are the namespace-scoped rbac roles + bindings that will be used by users
+# who want to use telepresence once its components have been set
+# up in the cluster.
+{{- if .Values.clientRbac.create }}
+{{- if .Values.clientRbac.namespaced }}
+
```
donnyyung, line 6: Should we also add the check here to ensure that the user has subjects set in their values.yaml?

.circleci/config.yml
```diff
@@ -68,6 +68,19 @@ commands:
             sudo apt-get install -y sshfs
             sudo sh -c 'echo user_allow_other >> /etc/fuse.conf'
 
+  "install-helm":
```
donnyyung: So I too once added an "install-helm" to circle, but Luke pointed out since it might be something users run from their box, it'd be better to include installation of helm in the Makefile, which we have here:https://github.com/telepresenceio/telepresence/blob/release/v2/build-aux/tools.mk#L73-L85

So I'd suggest making helm a dependency of `make check` and then you can remove this part. Does that make sense / seem reasonable?

donnyyung (changes requested):
A few nits / suggestions and one _real_ suggestion which is why I'm marking request changes. 

But all in all this looks _super_ great and I'm particularly hype about the helm tests you added to the test suite (which I probably should have done with the helm work, so sorry that fell to you 🤦 )

charts/telepresence/templates/clientRbac/namespace-scope.yaml
josecv: i don't think we need to duplicate it -- the check in `cluster-scope.yaml` [isn't conditional](https://github.com/telepresenceio/telepresence/pull/1845/files#diff-4ea4fc391ecffb9d40967c748d5dc01bfa589f6061b2786f8ad5d738bacaa9bcR7-R11) so we should hit it regardless of whether `clientRbac.namespaced` is set

.circleci/config.yml
josecv: Aah! Yes that makes a lot of sense 🙂

charts/telepresence/templates/clientRbac/namespace-scope.yaml
donnyyung: ah good point :)

pkg/client/connector/install_actions.go
```diff
@@ -362,6 +362,11 @@ func (ata *addTrafficAgentAction) Do(obj kates.Object) error {
 		return install.ObjErrorf(obj, "unable to find app container %q in", ata.containerName)
 	}
 
+	// Under some odd circumstances, the agent volume can be left over after an uninstall.
+	// Drop it if we get here and it's present, since it'll cause errors.
+	// We ignore the error from this since we don't care if the volume isn't already present
+	_ = ata.dropAgentAnnotationVolume(obj, tplSpec)
```
donnyyung, line 368: I think these odd circumstances are only in tests and using the smart agent bc the test version is something like this: `v2.0.0-gotest.53272` which means system a gives us the *pinned* agent image: docker.io/datawire/ambassador-telepresence-agent:1.2.0 , and since our current version of telepresence is not backwards compatible with that version, it's not surprising that weirdness ensues.

So I'm wondering if we should change the GenGo to create a v2.y.z that is closer to reality so we don't run into this bug.. but I think we are fine w/ leaving this function in, just wanted to explain why I think it was happening since it just happened to me.

donnyyung (approved):
Thanks so much for addressing all my comments, this LGTM!

donnyyung (approved):
New changes LGTM!

PR #1852, opened 2021-07-08
fix the traffic manager's cluster role

### LINKED ISSUE #1840: The daemon did not get the correct service subnet
When I was using telepresence intercept, I found that it is possible to access pods' IPs locally, but not to access some services' IPs. I saw this from the daemon.log:

image

Only added this service subnet: `10.96.0.0/16`

Then I read the code in `cmd/traffic/cmd/manager/internal/cluster/info.go` and found that the way to obtain the service CIDR is like this: https://stackoverflow.com/a/61685899

However, the result of running the below command locally is: `10.96.0.0/14`

```bash
echo '{"apiVersion":"v1","kind":"Service","metadata":{"name":"tst"},"spec":{"clusterIP":"1.1.1.1","ports":[{"port":443}]}}' | kubectl apply -f - 2>&1 |sed 's/.*valid IPs is //'
```

Why the subnet is not correct ?

johnlanni:
## Description

fix #1840 

the traffic manager also need the create permission to create the bogus service to parse the accurate service subnet.

## Checklist

### DIFF
--- a/charts/telepresence/templates/trafficManagerRbac.yaml
+++ b/charts/telepresence/templates/trafficManagerRbac.yaml
@@ -24,6 +24,7 @@ rules:
   verbs:
   - get
   - list
+  - create
 - apiGroups:
   - ""
   resources:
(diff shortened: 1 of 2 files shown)


thallgren:
While addresses the problem, it's the wrong fix. We don't want to give the `traffic-manager` cluster wide permissions to create services. We created a `Role` that permits the `traffic-manager` to create a service only in the "ambassador" namespace. The reason it doesn't work is that the `traffic-manager` attempts to create the service in the "default" namespace.

PR #1853, opened 2021-07-08
Ensure that traffic-manager creates dummy service in correct namespace.

thallgren:
## Description

The `traffic-manager` only has permissions to create services in the
namespace that it resides in (typically "ambassador"). Before this
commit it would make an attemtp to create a dummy service in the
"default" namespace to provoke an error message that contains the
service subnet. This would sometimes fail on a permission error because
the `traffic-manager` isn't allowed to create services in the "default"
namespace.

## Checklist


josecv (approved):
works for me!

PR #1857, opened 2021-07-10
Drop dependency on kubectl

LukeShu:
## Description

- The first 4 commits are generic minor fixes.
- The next 4 commits are getting rid of `kubectl port-forward`
- The next 2 commits are modifying to `teleperesence current-cluster-id` to not use `kubectl`
- The final commit is removing the `kubectl` version check, as nothing is using kubectl anymore.

This is failing CI (it did pass once, but then I fussed with the changelog) because the new code is faster and so exposes a pre-existing race condition. So I'm thinking that this work can go ahead and be reviewed/approved, but block merging it on a separate PR that fixes the race?

## Checklist

 - [x] I made sure to either submit a docs PR, or tell Matt about the necessary documentation changes. - yes, there is a commit in the `cycle5-2021` branch
 - [x] My change is adequately tested. - More unit tests would be nice, but yeah


pkg/client/daemon/tunrouter.go
```diff
@@ -191,10 +191,7 @@ func (t *tunRouter) setOutboundInfo(ctx context.Context, mi *daemon.OutboundInfo
 		defer cancel()
 
 		var conn *grpc.ClientConn
-		conn, err = grpc.DialContext(tc, fmt.Sprintf("127.0.0.1:%d", mi.ManagerPort),
-			grpc.WithInsecure(),
-			grpc.WithNoProxy(),
-			grpc.WithBlock())
+		conn, err = client.DialSocket(tc, client.ConnectorSocketName)
```
thallgren, line 194: This introduces one extra hop for all communication between the TUN-device and the cluster. Why would we want that?

LukeShu: Well, in that commit it does, but in the PR as a whole it doesn't.  Previously it was going through 1 hop: `kubectl port forward`.  Now it's going through 1 hop: the connector.

- before: daemon → kpf → cluster
- middle (8f59b2af8ecb39dba17a13851ad001927f06cf93): daemon → connector → kpf → cluster
- after (7cbbdf703a22191c9c1335c11d3bb6ef43bfe2c8): daemon → connector → cluster

LukeShu: Now, there is an argument to be made that `service_proxy.go` operating at L7 isn't going to perform as well as `kubectl port-forward` operating at L4.  Maybe.  And we can improve this if we want to get hacky treat requests to the manager service as opaque requests instead of structured data.

thallgren: Would it be possible to have the daemon do embedded port-forwards too so that we get daemon → cluster ?

LukeShu: Because there might be auth extensions that get shelled out to by the kubeconfig, we really want communication with the kube stuff to happen as the normal user and not as root.

pkg/client/connector/service_proxy.go
```diff
@@ -168,12 +169,51 @@ func (p *mgrProxy) ReviewIntercept(ctx context.Context, arg *managerrpc.ReviewIn
...
+	}
+	grp := dgroup.NewGroup(ctx, dgroup.GroupConfig{})
+	grp.Go("manager->daemon", func(ctx context.Context) error {
+		for {
+			payload, err := fhManager.Recv()
+			if err != nil {
+				if err == io.EOF {
+					return nil
+				}
+				return err
```
thallgren, line 191: Perhaps check if `ctx.Error() != nil` here and return `nil` if that's true. The rpc error caused by a cancelled context is just confusing.

pkg/client/connector/service_proxy.go
```diff
@@ -193,9 +233,28 @@ func (p *mgrProxy) AgentLookupHostResponse(ctx context.Context, arg *managerrpc.
...
+	if err != nil {
+		return err
+	}
+	for {
+		info, err := cli.Recv()
+		if err != nil {
+			if err == io.EOF {
+				return nil
+			}
+			return err
```
thallgren, line 254: Same here.

thallgren (approved):
Looks great. My only concern is some error returns that I think can be nil when the context is done.

pkg/client/daemon/tunrouter.go
thallgren: Ah, yes. That makes sense.

LukeShu:
v2:
- In the proxy, ignore gRPC errors if the context is canceled
- Fix a race condition with the proxy being used before proxy.SetClient is called

Diff from v1:
```patch
diff --git a/pkg/client/connector/intercept.go b/pkg/client/connector/intercept.go
index 7867f5a9..82bc4308 100644
--- a/pkg/client/connector/intercept.go
+++ b/pkg/client/connector/intercept.go
@@ -329,7 +329,7 @@ func (tm *trafficManager) addIntercept(c context.Context, ir *rpc.CreateIntercep
 		}()
 	}
 
-	apiKey, err := tm.getAPIKey(c, "agent-"+spec.Mechanism, false)
+	apiKey, err := tm.callbacks.GetAPIKey(c, "agent-"+spec.Mechanism, false)
 	if err != nil {
 		dlog.Errorf(c, "error getting apiKey for agent: %s", err)
 	}
diff --git a/pkg/client/connector/service.go b/pkg/client/connector/service.go
index 9c01fea0..7134db49 100644
--- a/pkg/client/connector/service.go
+++ b/pkg/client/connector/service.go
@@ -414,7 +414,14 @@ func (s *service) connectWorker(c context.Context, cr *rpc.ConnectRequest, k8sCo
 	connectStart := time.Now()
 
 	dlog.Info(c, "Connecting to traffic manager...")
-	tmgr, err := newTrafficManager(c, s.env, s.cluster, s.scoutClient.Reporter.InstallID(), s.getCloudAPIKey)
+	tmgr, err := newTrafficManager(c,
+		s.env,
+		s.cluster,
+		s.scoutClient.Reporter.InstallID(),
+		trafficManagerCallbacks{
+			GetAPIKey: s.getCloudAPIKey,
+			SetClient: s.managerProxy.SetClient,
+		})
 	if err != nil {
 		dlog.Errorf(c, "Unable to connect to TrafficManager: %s", err)
 		// No point in continuing without a traffic manager
@@ -440,7 +447,6 @@ func (s *service) connectWorker(c context.Context, cr *rpc.ConnectRequest, k8sCo
 			ErrorText: err.Error(),
 		}
 	}
-	s.managerProxy.SetClient(tmgr.managerClient)
 
 	// Start k8s-watches
 	s.clusterRequest <- cluster
diff --git a/pkg/client/connector/service_proxy.go b/pkg/client/connector/service_proxy.go
index c14c70f0..6cac9ca1 100644
--- a/pkg/client/connector/service_proxy.go
+++ b/pkg/client/connector/service_proxy.go
@@ -116,7 +116,7 @@ func (p *mgrProxy) WatchAgents(arg *managerrpc.SessionInfo, srv managerrpc.Manag
 	for {
 		snapshot, err := cli.Recv()
 		if err != nil {
-			if err == io.EOF {
+			if err == io.EOF || srv.Context().Err() != nil {
 				return nil
 			}
 			return err
@@ -138,7 +138,7 @@ func (p *mgrProxy) WatchIntercepts(arg *managerrpc.SessionInfo, srv managerrpc.M
 	for {
 		snapshot, err := cli.Recv()
 		if err != nil {
-			if err == io.EOF {
+			if err == io.EOF || srv.Context().Err() != nil {
 				return nil
 			}
 			return err
@@ -185,7 +185,7 @@ func (p *mgrProxy) ClientTunnel(fhDaemon managerrpc.Manager_ClientTunnelServer)
 		for {
 			payload, err := fhManager.Recv()
 			if err != nil {
-				if err == io.EOF {
+				if err == io.EOF || ctx.Err() != nil {
 					return nil
 				}
 				return err
@@ -199,7 +199,7 @@ func (p *mgrProxy) ClientTunnel(fhDaemon managerrpc.Manager_ClientTunnelServer)
 		for {
 			payload, err := fhDaemon.Recv()
 			if err != nil {
-				if err == io.EOF {
+				if err == io.EOF || ctx.Err() != nil {
 					return nil
 				}
 				return err
@@ -248,7 +248,7 @@ func (p *mgrProxy) WatchClusterInfo(arg *managerrpc.SessionInfo, srv managerrpc.
 	for {
 		info, err := cli.Recv()
 		if err != nil {
-			if err == io.EOF {
+			if err == io.EOF || srv.Context().Err() != nil {
 				return nil
 			}
 			return err
diff --git a/pkg/client/connector/traffic_manager.go b/pkg/client/connector/traffic_manager.go
index bd274d43..fb90c289 100644
--- a/pkg/client/connector/traffic_manager.go
+++ b/pkg/client/connector/traffic_manager.go
@@ -29,11 +29,16 @@ import (
 	"github.com/telepresenceio/telepresence/v2/pkg/install"
 )
 
+type trafficManagerCallbacks struct {
+	GetAPIKey func(context.Context, string, bool) (string, error)
+	SetClient func(client manager.ManagerClient, callOptions ...grpc.CallOption)
+}
+
 // trafficManager is a handle to access the Traffic Manager in a
 // cluster.
 type trafficManager struct {
 	*installer // installer is also a k8sCluster
-	getAPIKey  func(context.Context, string, bool) (string, error)
+	callbacks  trafficManagerCallbacks
 
 	// local information
 	env         client.Env
@@ -74,7 +79,7 @@ func newTrafficManager(
 	env client.Env,
 	cluster *k8sCluster,
 	installID string,
-	getAPIKey func(context.Context, string, bool) (string, error),
+	callbacks trafficManagerCallbacks,
 ) (*trafficManager, error) {
 	userinfo, err := user.Current()
 	if err != nil {
@@ -96,7 +101,7 @@ func newTrafficManager(
 		installID:   installID,
 		startup:     make(chan bool),
 		userAndHost: fmt.Sprintf("%s@%s", userinfo.Username, host),
-		getAPIKey:   getAPIKey,
+		callbacks:   callbacks,
 	}
 
 	return tm, nil
@@ -160,7 +165,7 @@ func (tm *trafficManager) run(c context.Context) error {
 		InstallId: tm.installID,
 		Product:   "telepresence",
 		Version:   client.Version(),
-		ApiKey:    func() string { tok, _ := tm.getAPIKey(c, "manager", false); return tok }(),
+		ApiKey:    func() string { tok, _ := tm.callbacks.GetAPIKey(c, "manager", false); return tok }(),
 	})
 	if err != nil {
 		return client.CheckTimeout(tc, fmt.Errorf("manager.ArriveAsClient: %w", err))
@@ -168,9 +173,14 @@ func (tm *trafficManager) run(c context.Context) error {
 	tm.managerClient = mClient
 	tm.sessionInfo = si
 
+	// Gotta call mgrProxy.SetClient before we call daemon.SetOutboundInfo which tells the
+	// daemon to use the proxy.
+	tm.callbacks.SetClient(tm.managerClient)
+
 	// Tell daemon what it needs to know in order to establish outbound traffic to the cluster
 	if _, err := tm.daemon.SetOutboundInfo(c, tm.getOutboundInfo()); err != nil {
 		tm.managerClient = nil
+		tm.callbacks.SetClient(nil)
 		return fmt.Errorf("daemon.SetOutboundInfo: %w", err)
 	}
 
@@ -401,7 +411,7 @@ func (tm *trafficManager) remain(c context.Context) error {
 		case <-ticker.C:
 			_, err := tm.managerClient.Remain(c, &manager.RemainRequest{
 				Session: tm.session(),
-				ApiKey:  func() string { tok, _ := tm.getAPIKey(c, "manager", false); return tok }(),
+				ApiKey:  func() string { tok, _ := tm.callbacks.GetAPIKey(c, "manager", false); return tok }(),
 			})
 			if err != nil {
 				if c.Err() != nil {
```

pkg/client/connector/service_proxy.go
LukeShu: Done

pkg/client/connector/traffic_manager.go
```diff
@@ -192,17 +165,22 @@ func (tm *trafficManager) initGrpc(c context.Context, portStr string) (err error
...
 	})
 	if err != nil {
 		return client.CheckTimeout(tc, fmt.Errorf("manager.ArriveAsClient: %w", err))
 	}
 	tm.managerClient = mClient
 	tm.sessionInfo = si
 
+	// Gotta call mgrProxy.SetClient before we call daemon.SetOutboundInfo which tells the
+	// daemon to use the proxy.
+	tm.callbacks.SetClient(tm.managerClient)
```
LukeShu, line 178: The fix for the race is to do this here...

pkg/client/connector/service.go
```diff
@@ -451,7 +447,6 @@ func (s *service) connectWorker(c context.Context, cr *rpc.ConnectRequest, k8sCo
 			ErrorText: err.Error(),
 		}
 	}
-	s.managerProxy.SetClient(tmgr.managerClient)
```
LukeShu, line 454: ... instead of here

thallgren (approved):
New changes that fixes the race looks good.

A general reflection is that there's a lot of things copied from `service` to the `trafficManager` now. Perhaps we should consider letting the `trafficManager` have a pointer to the `service` at some point?

PR #1860, opened 2021-07-12
Clean up the connector's flow control, fix a crash

LukeShu:
## Description

99% of this is trying to clean up the connector's flow control so that I could reason about it to figure out how to fix the race condition that https://github.com/telepresenceio/telepresence/pull/1857 was hitting.  I had thought that it was a pre-existing race condition that this PR would be fixing, but instead since the race was introduced in #1857 I went ahead and fixed it in #1857.  But I was still left with a bunch of cleanup that is probably worth landing.

And while I was looking at all of the connector code to clean it up, I noticed a possible crash during shutdown.  So go ahead and fix that.

The connector has a bunch of objects for its worker goroutines, and they're all in the same package and so can reach in to eachothers struct members directly.  It makes it really hard to reason about when this or that gets set.  So, the theme of this PR is to move each of them in to its own package, and reduce the surface area where they interact.  So the PR is actually pretty small, but the diff is big because of a lot of renaming files and changing whether things are public/private.

## Checklist

 - [x] I made sure to either submit a docs PR, or tell Matt about the necessary documentation changes. - no docs changes besides release notes, which will get synced from changelog
 - [x] My change is adequately tested. - yes


pkg/client/connector/command.go
```diff
@@ -286,9 +280,9 @@ func (s *service) connectWorker(c context.Context, cr *rpc.ConnectRequest, k8sCo
 
 	ret := &rpc.ConnectInfo{
 		Error:          rpc.ConnectInfo_UNSPECIFIED,
-		ClusterContext: s.cluster.K8sConfig.Context,
-		ClusterServer:  s.cluster.K8sConfig.Server,
-		ClusterId:      s.cluster.GetClusterId(c),
+		ClusterContext: s.sharedState.GetClusterNonBlocking().K8sConfig.Context,
+		ClusterServer:  s.sharedState.GetClusterNonBlocking().K8sConfig.Server,
+		ClusterId:      s.sharedState.GetClusterNonBlocking().GetClusterId(c),
```
thallgren: Please assign the result of `GetClusterNonBlocking()` to a variable at line 221 instead of calling it repeatedly 6 times.

pkg/client/connector/command.go
```diff
@@ -218,11 +219,11 @@ func (s *service) connectWorker(c context.Context, cr *rpc.ConnectRequest, k8sCo
 	connectStart := time.Now()
 
 	dlog.Info(c, "Connecting to traffic manager...")
-	tmgr, err := newTrafficManager(c,
+	tmgr, err := worker_trafficmanager.NewTrafficManager(c,
```
thallgren: The naming of the package and the constructor seems a bit off. Instead of repeated names in package and constructor, how about `trafficmanager.NewWorker()` ? If we want to have `worker` in the package structure, then perhaps use `pkg/client/connector/worker/trafficmanager`, `pkg/client/connector/worker/grpc`, and `client/connector/worker/cluster`.

thallgren (changes requested):
Looks good overall. I like the split and encapsulation. I have two concerns:
1. Unnecessary repeated calls of a function should be avoided.
2. The names of packages and constructors can be improved.

thallgren (approved):
All good.

LukeShu:
Resolved logical merge conflicts. No other changes. Will merge once CI turns green.

PR #1862, opened 2021-07-13
Fix macOS DNS glitch

thallgren:
## Description

DNS queries that ended with ".tel2-search." were incorrectly stripped,
not just from that suffix, but also from the trailing dot. On systems
with recursive DNS (e.g. running the cluster locally), this resulted
in sequences of queries like:
```
QTYPE[1] localhost.tel2-search. -> NOT FOUND
QTYPE[1] localhos.tel2-search. -> NOT FOUND
QTYPE[1] localho.tel2-search. -> NOT FOUND
```

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 - Bugfix: Fixed and improved several error messages, to hopefully be
   more helpful.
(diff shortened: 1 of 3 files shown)


josecv (approved):
makes sense

PR #1864, opened 2021-07-13
Native windows port of the telepresence CLI

### LINKED ISSUE #92: Windows support
It would be good to support Windows.

Options:

* Windows Subsystem for Linux.
* `torsocks` equivalent.
* Other implementation strategies, e.g. #76.

thallgren:
## Description

This PR makes it possible to compile and run the telepresence CLI on Windows. This build system is not yet ported so tests that require that will not run on Windows yet.

In order to build and run, you will need to do the following (I didn't put this in DEVELOPING.md because I think it will be very short lived):
1. Determine a telepresence version, e.g. `v2.3.4` and ensure that the cluster that you want to connect to have access to a traffic-manager and traffic agent of that version from a docker repository.
2. Build or install the windows binary using `go build` or `go install` with an included `-ldflags` to inject the chosen version:
```
go install -trimpath -ldflags='-X=github.com/telepresenceio/telepresence/v2/pkg/version.Version=v2.3.4
```
3. Ensure that the `wintun.dll` downloaded from https://www.wintun.net/ is available to the dll loader (for instance, put it adjacent to the telepresence binary).
4. Enjoy ;-)

## Checklist


pkg/client/cli/cliutil/exec.go
```diff
@@ -3,75 +3,61 @@ package cliutil
...
 		if sig == nil {
 			return
 		}
 		_ = cmd.Process.Signal(sig)
 	}()
+
 	s, err := cmd.Process.Wait()
 	if err != nil {
-		return fmt.Errorf("%s: %w", logging.ShellString(exe, args), err)
+		return fmt.Errorf("%s: %v", logging.ShellString(exe, args), err)
```
josecv: hmm why not `%w`? what kind of errors can we get from cmd.Process.Wait()?

pkg/client/cli/cliutil/exec_windows.go
```diff
@@ -0,0 +1,44 @@
...
+		verb = "open"
+	}
+	return shellExec(ctx, verb, exe, args)
+}
+
+func shellExec(_ context.Context, verb, exe string, args []string) error {
+	cwd, _ := os.Getwd()
+	verbPtr, _ := windows.UTF16PtrFromString(verb)
+	exePtr, _ := windows.UTF16PtrFromString(exe)
+	cwdPtr, _ := windows.UTF16PtrFromString(cwd)
```
josecv, line 31: It looks like these can only error if the string contains a NUL byte somewhere, so it makes sense to ignore the error -- should we add a short comment explaining this?

pkg/client/sockets_windows.go
```diff
@@ -0,0 +1,64 @@
...
+)
+
+// The Windows IPC between the CLI and the user and root daemons is based on named pipes rather than
+// unix sockets.
+const (
+	// ConnectorSocketName is the path used when communicating to the connector process
+	ConnectorSocketName = `\\.\pipe\telepresence-connector`
+
+	// DaemonSocketName is the path used when communicating to the daemon process
+	DaemonSocketName = `\\.\pipe\telepresence-daemon`
```
josecv: it seems like these aren't paths so much as just pipe names, is that right? If so maybe we should update the comment to reflect that, and perhaps add a link to [the windows docs](https://docs.microsoft.com/en-us/windows/win32/ipc/pipe-names)

pkg/client/daemon/service.go
```diff
@@ -190,19 +188,10 @@ func run(c context.Context, loggingDir, configDir, dns string) error {
 
 		// Listen on unix domain socket
 		dlog.Debug(c, "gRPC server starting")
-		origUmask := unix.Umask(0)
-		grpcListener, err = net.Listen("unix", client.DaemonSocketName)
-		unix.Umask(origUmask)
+		grpcListener, err = client.ListenSocket(c, processName, client.DaemonSocketName)
```
josecv: how does the cleanup work on windows? Will [these lines](https://github.com/telepresenceio/telepresence/pull/1864/commits/1cb5724deccca2c75c9c07e5ed6f2b53c2e49edd#diff-f541fcdb8230f0df59b1ecefcedb4056040fb4d4e68b15acb6906daaeef8e575R152-R154) actually clean anything up when using a named pipe?

pkg/client/sockets_windows.go
```diff
@@ -0,0 +1,64 @@
...
+		grpc.WithBlock(),
+		grpc.FailOnNonTempDialError(true),
+		grpc.WithContextDialer(func(c context.Context, s string) (net.Conn, error) {
+			conn, err := winio.DialPipeContext(c, socketName)
+			return conn, err
+		}))
+	return conn, err
+}
+
+const AllowEveryone = "S:(ML;;NW;;;LW)D:(A;;0x12019f;;;WD)"
```
josecv: This is... something else -- is there a reference doc or something where we can decode this magic string?

pkg/client/sockets_windows.go
```diff
@@ -0,0 +1,64 @@
...
+	return winio.ListenPipe(socketName, config)
+}
+
+// SocketExists returns true if a socket is found at the given path
+func SocketExists(path string) bool {
+	uPath, err := windows.UTF16PtrFromString(path)
+	if err != nil {
+		return false
+	}
+	h, err := windows.CreateFile(uPath, windows.GENERIC_READ|windows.GENERIC_WRITE, 0, nil, windows.OPEN_EXISTING, windows.FILE_FLAG_OVERLAPPED, 0)
```
josecv, line 70: this system call is very badly named -- could we add a comment stating that the `OPEN_EXISTING` disposition will make `CreateFile` fail if the file doesn't already exist? At first glance it looks like "Create a file or open it if it's already there"

pkg/client/logging/sysinfo.go
```diff
@@ -2,18 +2,18 @@ package logging
...
 //
 // Unix:
 //  info.Sys().(*syscall.Stat_t)
 //
 // Windows:
 //  info.Sys().(*syscall.Win32FileAttributeData)
-type sysinfo interface {
-	birthtime() time.Time
+type SysInfo interface {
+	Birthtime() time.Time
```
josecv: nit: should this be called `BirthTime`?

pkg/client/logging/sysinfo_windows.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+	err := api.GetNamedSecurityInfo(
+		wi.path,
+		api.SE_FILE_OBJECT,
+		api.OWNER_SECURITY_INFORMATION,
+		&wi.owner,
+		&wi.group,
+		&wi.dacl,
+		&wi.sacl,
+		&wi.secDesc,
```
josecv: the [internet says](https://docs.microsoft.com/en-us/windows/win32/api/aclapi/nf-aclapi-getnamedsecurityinfoa) that this buffer needs to be freed by the user -- does that apply here? do we need to do something special to garbage collect `secDesc`?

pkg/client/logging/sysinfo_windows.go
```diff
@@ -0,0 +1,98 @@
...
+	)
+	if err != nil {
+		panic(err)
```
josecv: i don't know how I should feel about this -- should we at least rename this function to `MustGetSysInfo` to denote that it can panic? What kinds of errors can we get out of `GetNamedSecurityInfo`?

pkg/dpipe/dpipe_windows.go
```diff
@@ -0,0 +1,25 @@
...
+	"time"
+
+	"github.com/datawire/dlib/dexec"
+)
+
+func waitCloseAndKill(ctx context.Context, cmd *dexec.Cmd, peer io.Closer, closing *int32, killTimer **time.Timer) {
+	<-ctx.Done()
+
+	// A process is sometimes not terminated gracefully by the SIGTERM, so we give
+	// it a second to succeed and then kill it forcefully.
```
josecv: these comments don't seem to apply here

pkg/client/daemon/dns/flush_windows.go
```diff
@@ -0,0 +1,9 @@
+package dns
+
+import (
+	"context"
+)
+
+func Flush(c context.Context) {
+	// TODO
```
josecv: should this be implemented? I guess `setDNS` actually flushes, so maybe this comment should just be updated to explain why this isn't implemented?

josecv (commented):
A lot of new code, and I don't really feel qualified to pass judgement on the big picture, but it all seems to make sense to me.

Mostly just have questions and suggestions for comments; my biggest concern is around cleaning up the sockets, as I'm not sure if it's possible to simply `os.Remove` a named pipe

pkg/client/sockets_windows.go
thallgren: But ... isn't that self evident? ;-)

pkg/client/logging/sysinfo.go
thallgren: It should. Compound words in English bugs me. You use "namespace" as one word but "birth time" as two. Weird. In Swedish, we make it simple. Always concatenate :-)

thallgren: Or "timestamp" versus "time span". What?!

pkg/client/daemon/service.go
thallgren: The "these lines" link doesn't make much sense to me.

thallgren: Oh, you're referring to:
```
	defer func() {
		if grpcListener != nil {
			_ = os.Remove(grpcListener.Addr().String())
		}
	}()
```
No, that won't do on Windows.

pkg/client/logging/sysinfo_windows.go
thallgren: Good catch!

pkg/client/daemon/dns/flush_windows.go
thallgren: I don't think `setDNS()` performs a complete flush. It will just flush the device. Windows may have other caches in other places and we should find out where and implement a proper flush here. I'll update the TODO comment.

thallgren:
@josecv I think I've addressed all you comments now. I also rebased the PR on top of the current release/v2.

pkg/client/daemon/dns/flush_windows.go
thallgren: Doh. You're right. The `setDNS()` calls `ipconfig /flushdns`.

Euthpic:
I am puzzled about this step(newbie to go):
```
Build or install the windows binary using go build or go install with an included -ldflags to inject the chosen version:
go install -trimpath -ldflags='-X=github.com/telepresenceio/telepresence/v2/pkg/version.Version=v2.3.4
```
I installed go sdk and set all the GOPATH/GOBIN/GOROOT,but this command still doesn't work correctly:
```
package .: no Go files in D:\download
```
Then I download the realease from https://github.com/telepresenceio/telepresence/archive/refs/tags/v2.3.4.zip.
And run this command in /d/download/telepresence-2.3.4/cmd/telepresence,it still faied:
```
# github.com/telepresenceio/telepresence/v2/pkg/dpipe
..\..\pkg\dpipe\dpipe.go:46:27: undefined: unix.SIGKILL
..\..\pkg\dpipe\dpipe.go:50:26: undefined: unix.SIGTERM
# github.com/telepresenceio/telepresence/v2/pkg/tun
..\..\pkg\tun\device.go:14:18: undefined: Device
..\..\pkg\tun\device.go:15:9: undefined: openTun
..\..\pkg\tun\device.go:20:10: undefined: Device
..\..\pkg\tun\device.go:26:10: undefined: Device
..\..\pkg\tun\device.go:31:10: undefined: Device
..\..\pkg\tun\device.go:37:10: undefined: Device
..\..\pkg\tun\device.go:43:10: undefined: Device
..\..\pkg\tun\device.go:47:10: undefined: Device
..\..\pkg\tun\device.go:52:13: undefined: unix.Socket
..\..\pkg\tun\device.go:52:33: undefined: unix.SOCK_DGRAM
..\..\pkg\tun\device.go:52:33: too many errors
# github.com/telepresenceio/telepresence/v2/pkg/connpool
..\..\pkg\connpool\connid.go:58:22: undefined: unix.IPPROTO_TCP
..\..\pkg\connpool\connid.go:83:22: undefined: unix.IPPROTO_TCP
..\..\pkg\connpool\connid.go:106:7: undefined: unix.IPPROTO_TCP
..\..\pkg\connpool\connid.go:112:7: undefined: unix.IPPROTO_UDP
..\..\pkg\connpool\connid.go:137:10: undefined: unix.IPPROTO_TCP
..\..\pkg\connpool\connid.go:139:10: undefined: unix.IPPROTO_UDP
..\..\pkg\connpool\connid.go:141:10: undefined: unix.IPPROTO_ICMP
..\..\pkg\connpool\connid.go:143:10: undefined: unix.IPPROTO_ICMPV6
..\..\pkg\connpool\connid.go:151:7: undefined: unix.IPPROTO_ICMP
..\..\pkg\connpool\connid.go:153:7: undefined: unix.IPPROTO_TCP
..\..\pkg\connpool\connid.go:153:7: too many errors
# github.com/telepresenceio/telepresence/v2/pkg/client/logging
..\..\pkg\client\logging\rotatingfile.go:146:17: undefined: getSysInfo
..\..\pkg\client\logging\rotatingfile.go:208:10: undefined: dupToStd
..\..\pkg\client\logging\rotatingfile.go:246:15: undefined: getSysInfo
..\..\pkg\client\logging\rotatingfile.go:261:38: undefined: getSysInfo
# github.com/telepresenceio/telepresence/v2/pkg/tun/ip
..\..\pkg\tun\ip\header.go:128:26: undefined: unix.IPPROTO_UDP
```

Which step was wrong?

pkg/ipproto/ipproto_windows.go
```diff
@@ -0,0 +1,10 @@
+package ipproto
+
+import "syscall"
+
+const (
+	TCP    = syscall.IPPROTO_TCP
+	UDP    = syscall.IPPROTO_UDP
+	ICMP   = 0x1
+	ICMPV6 = 0x3a
+)
```
LukeShu: The numbers are the same on all platforms (by definition)... if we can't just grab them all from the same syscall-ish package, I'd just write the raw numbers and throw in a comment pointing at https://www.iana.org/assignments/protocol-numbers/protocol-numbers.xhtml

pkg/client/cli/cliutil/exec.go
```diff
@@ -0,0 +1,63 @@
...
+// Run will run the given executable with given args and env, wait for it to terminate, and return
+// the result. The run will dispatch signals as appropriate for the given platform (SIGTERM and SIGINT on Unix platforms
+// and os.Interrupt on Windows).
+func Run(ctx context.Context, sCmd SafeCobraCommand, exe string, args []string, env map[string]string) error {
+	cmd := exec.CommandContext(ctx, exe, args...)
+	cmd.Stdout = sCmd.OutOrStdout()
+	cmd.Stderr = sCmd.ErrOrStderr()
+	cmd.Stdin = sCmd.InOrStdin()
+	for k, v := range env {
+		cmd.Env = append(os.Environ(), k+"="+v)
```
LukeShu: Isn't this going to discard all but the last item in `env`?

Shouldn't this be
```go
cmd.Env = os.Environ()
for k, v := range env {
	cmd.Env = append(cmd.Env, k+"="+v)
```

pkg/client/cli/cliutil/exec.go
```diff
@@ -0,0 +1,63 @@
...
+)
+
+// Run will run the given executable with given args and env, wait for it to terminate, and return
+// the result. The run will dispatch signals as appropriate for the given platform (SIGTERM and SIGINT on Unix platforms
+// and os.Interrupt on Windows).
+func Run(ctx context.Context, sCmd SafeCobraCommand, exe string, args []string, env map[string]string) error {
+	cmd := exec.CommandContext(ctx, exe, args...)
+	cmd.Stdout = sCmd.OutOrStdout()
+	cmd.Stderr = sCmd.ErrOrStderr()
+	cmd.Stdin = sCmd.InOrStdin()
```
LukeShu, line 23: Here's something that slipped in recently that you may not have noticed: The tests don't do `cmd.SetOut` or anything anymore; they fork the actual process.  So it shouldn't be necessary to explicitly pass around stdio anymore (whether via a cobra command or via separate stdin/out/err arguments).

pkg/client/cli/cliutil/exec_unix.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/datawire/dlib/dcontext"
+	"github.com/datawire/dlib/dexec"
+	"github.com/telepresenceio/telepresence/v2/pkg/client/logging"
+	"github.com/telepresenceio/telepresence/v2/pkg/proc"
+)
+
+func background(ctx context.Context, exe string, args []string) error {
+	// The context should not kill it if cancelled
+	ctx = dcontext.WithoutCancel(ctx)
+	cmd := exec.CommandContext(ctx, exe, args...)
```
LukeShu: If you're using stdlib `exec` instead of dlib `dexec`, the *only* thing that the ctx given to `CommanContext` does is cancel it, so if you're using `WithoutCancel`, might as well just use `dexec.Command` (no Context).

pkg/client/cli/cliutil/exec_unix.go
```diff
@@ -0,0 +1,74 @@
...
+			pwCmd := dexec.CommandContext(ctx, "sudo", "true")
+			pwCmd.DisableLogging = true
+			if err := pwCmd.Run(); err != nil {
+				return err
+			}
+		}
+		args = append([]string{"--non-interactive", "--preserve-env", exe}, args...)
+		exe = "sudo"
+	}
+	return Background(ctx, exe, args)
```
LukeShu: I'm not a fan of this business with having `exe` be separate than just `args[0]`.

pkg/client/cli/cliutil/daemon.go
```diff
@@ -53,53 +46,7 @@ func launchDaemon(ctx context.Context, dnsIP string) error {
...
-
-	if err := cmd.Start(); err != nil {
-		return fmt.Errorf("%s: %w", logging.ShellString(args[0], args[1:]), err)
-	}
-	if err := cmd.Process.Release(); err != nil {
-		return fmt.Errorf("%s: %w", logging.ShellString(args[0], args[1:]), err)
-	}
-
-	return nil
+	return BackgroundAsRoot(ctx, client.GetExe(), []string{"daemon-foreground", logDir, configDir, dnsIP})
```
LukeShu: I'm already not a fan of having the `exe` be separate from `args[0]`, but declaring it as `args []string` instead of `args ...string` just makes this clunky when it could be `return BackgroundAsRoot(ctx, client.GetExe(), "daemon-foreground", logDir, configDir, dnsIP)`.

pkg/client/cli/cliutil/exec.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+
+	var err error
+	if err = cmd.Start(); err != nil {
+		return fmt.Errorf("%s: %w", logging.ShellString(exe, args), err)
+	}
+
+	// Ensure that interrupt is propagated to the child process
+	sigCh := signalNotifications()
+	defer close(sigCh)
```
LukeShu: This is panicy.  The `os/signal.Notify` machinery will panic if it tries to write to this closed signal.  Need to call `signal.Stop(sigCh)`.

pkg/client/cli/cliutil/exec_unix.go
josecv: This is the only comment I didn't really address for two reasons:

1. I think it's pretty reasonable, especially in code like this where the executable is being modified. IMHO it'd look a little bit weird to do `args[0] = "sudo"` instead of `exe = "sudo"`
2. This is consistent with the `dexec.Command*` APIs

josecv:
@Euthpic we'll be releasing a Developer Preview of the windows client this week 🙂 I suggest waiting until we do that so you don't have to go through the motions of building it yourself

pkg/client/cli/cliutil/exec.go
josecv: So removing these lines actually resulted in [a test failure](https://app.circleci.com/pipelines/github/telepresenceio/telepresence/2245/workflows/b4878a93-5976-4110-8e9a-f1c210f19aff/jobs/11744?invite=true#step-108-240) :(

LukeShu: Not removing them, but just using `cmd.Stdout = os.Stdout` instead; no need to take the `sCmd` argument.

pkg/client/cli/cliutil/daemon.go
```diff
@@ -53,53 +46,7 @@ func launchDaemon(ctx context.Context, dnsIP string) error {
...
-			// it with Setpgid=true, so do a pre-flight `sudo --validate` to read the
-			// password, and then enforce that being re-used by passing
-			// `--non-interactive`.
-			pwCmd := dexec.CommandContext(ctx, "sudo", "--validate")
-			pwCmd.DisableLogging = true
-			if err := pwCmd.Run(); err != nil {
-				return err
-			}
-		}
-		args = append([]string{"sudo", "--non-interactive", "--preserve-env"}, args...)
```
LukeShu, line 85: @josecv wrote
>  IMHO it'd look a little bit weird to do args[0] = "sudo" instead of exe = "sudo"

I don't think this looked weird.  I'd argue that it, in fact, read much easier.

pkg/client/cli/cliutil/daemon.go
```diff
@@ -53,53 +46,7 @@ func launchDaemon(ctx context.Context, dnsIP string) error {
...
-		// use `sudo`'s `--prompt` flag for this because (1) we don't want it to be
-		// re-displayed if they typo their password, and (2) it might be ignored anyway
-		// depending on `passprompt_override` in `/etc/sudoers`.  So we'll do a pre-flight
-		// `sudo --non-interactive true` to decide whether to display it.
-		//
-		// Note: Using `sudo --non-interactive --validate` does not work well in situations
-		// where the user has configured `myuser ALL=(ALL:ALL) NOPASSWD: ALL` in the sudoers
-		// file. Hence the use of `sudo --non-interactive true`. A plausible cause can be
-		// found in the first comment here:
-		// https://unix.stackexchange.com/questions/50584/why-sudo-timestamp-is-not-updated-when-nopasswd-is-set
```
LukeShu, line 70: This comment went away... for no real reason?  IMO it was a very helpful comment.

pkg/client/cli/cliutil/exec.go
josecv: OOH okay my bad

pkg/client/cli/cliutil/daemon.go
```diff
@@ -53,53 +46,7 @@ func launchDaemon(ctx context.Context, dnsIP string) error {
...
-		needPwCmd := dexec.CommandContext(ctx, "sudo", "--non-interactive", "true")
-		needPwCmd.DisableLogging = true
-		if err := needPwCmd.Run(); err != nil {
-			fmt.Printf("Need root privileges to run: %s\n", logging.ShellString(args[0], args[1:]))
-			// `sudo` won't be able to read the password from the terminal when we run
-			// it with Setpgid=true, so do a pre-flight `sudo --validate` to read the
-			// password, and then enforce that being re-used by passing
-			// `--non-interactive`.
-			pwCmd := dexec.CommandContext(ctx, "sudo", "--validate")
```
LukeShu, line 79: These 2 `--validate`s were changed to `true`s.  Was that intentional?  There was a lot shoved in to that one commit.

pkg/client/cli/cliutil/exec.go
```diff
@@ -0,0 +1,69 @@
...
+
+// Background will start the given exe as a detach process.
+func Background(ctx context.Context, exe string, args ...string) error {
+	return background(ctx, exe, args...)
+}
+
+// BackgroundAsRoot will start the given exe as a detach process with elevated privileges.
+func BackgroundAsRoot(ctx context.Context, exe string, args ...string) error {
+	return backgroundAsRoot(ctx, exe, args...)
+}
```
LukeShu, line 69: These aren't great function names... where's the verb?

pkg/client/cli/cliutil/exec_unix.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+	return Background(ctx, exe, args...)
+}
+
+func signalNotifications() chan os.Signal {
+	// Ensure that SIGINT and SIGTERM are propagated to the child process
+	sigCh := make(chan os.Signal, 1)
+	signal.Notify(sigCh, unix.SIGINT, unix.SIGTERM)
+	return sigCh
+}
```
LukeShu, line 72: Why repeat this business several times, instead of just having each platform declare a `signalsToForward` variable?

LukeShu:
I didn't look at the `dpipe` changes at all, since I know it's getting rewritten in @josecv's PR.

pkg/client/cli/cliutil/exec.go
LukeShu: Also note that if we don't need to pass in `sCmd`, then there's no reason to move SafeCobraCommand in to cliutil.

.circleci/config.yml
```diff
@@ -26,7 +26,7 @@ commands:
     parameters:
       version:
         type: string
-        default: "1.15"
+        default: "1.16"
```
LukeShu, line 29: The only thing that actually ended up using Go 1.16 was one place ware it said `io.Discard` instead of the older `ioutil.Discard`.

LukeShu:
Having the `Background` and `BackgroundAsRoot` functions in `cliutil` just seems a little wonky to me.  I'd be inclined to put them in `proc` instead.

go.mod
```diff
@@ -18,11 +21,13 @@ require (
 	github.com/spf13/cobra v1.1.1
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.7.0
-	github.com/telepresenceio/telepresence/rpc/v2 v2.3.7
-	golang.org/x/net v0.0.0-20210226172049-e18ecbb05110
+	github.com/telepresenceio/telepresence/rpc/v2 v2.3.6
```
LukeShu, line 24: *Downgrading* `rpc/v2` is a little weird... I suspect it's the artifact of a bad rebase?

PR #1867, opened 2021-07-14
Make max size for messages received over gRPC configurable.

thallgren:
## Description

Adds a `grpc` entry to the `config.yml` that contains `maxReceiveSize`
entry of type [resource Quantity](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/scheduling/resources.md#resource-quantities) which enables entry of strings like "10MiB".

## Checklist

 - [ ] I made sure to either submit a docs PR.

### DIFF
--- a/pkg/client/connector/traffic_manager.go
+++ b/pkg/client/connector/traffic_manager.go
@@ -133,7 +133,8 @@ func (tm *trafficManager) run(c context.Context) error {
 		fmt.Sprint(install.ManagerPortHTTP))
 
 	// First check. Establish connection
-	tos := &client.GetConfig(c).Timeouts
+	clientConfig := client.GetConfig(c)
+	tos := &clientConfig.Timeouts
 	tc, cancel := tos.TimeoutContext(c, client.TimeoutTrafficManagerAPI)
 	defer cancel()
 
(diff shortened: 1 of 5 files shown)


pkg/client/connector/traffic_manager.go
```diff
@@ -150,11 +151,17 @@ func (tm *trafficManager) run(c context.Context) error {
...
-	conn, err = grpc.DialContext(tc, grpcAddr,
-		grpc.WithContextDialer(grpcDialer),
+	opts := []grpc.DialOption{grpc.WithContextDialer(grpcDialer),
 		grpc.WithInsecure(),
 		grpc.WithNoProxy(),
-		grpc.WithBlock())
+		grpc.WithBlock()}
+
+	if mxRecvSize := clientConfig.Grpc.MaxReceiveSize; mxRecvSize != nil {
+		if mz, ok := mxRecvSize.AsInt64(); ok {
```
donnyyung, line 160: I think it would maybe make sense to do this check when we load the config (and have MaxReceiveSize be an Int64 instead of a quantity). Just so we don't need to do this check every time we use this value.. and then can also throw an error when loading the config and inform the user then.. instead of having it load successfully into the config but then not be used (Bc it's a decimal or something) which might be confusing to the user.

donnyyung (approved):

donnyyung:
Approving and merging so we can get into the RC. Let's circle back on my comment later / when you have time @thallgren

pkg/client/connector/traffic_manager.go
thallgren: Yep. That makes a lot of sense.

PR #1873, opened 2021-07-15
fix a nil pointer bug

johnlanni:
## Description

image

It seems that the `ClientTunnelServer` of the client session may not have been initialized when the client call the traffic-manager's `AgentTunnel` method.

## Checklist

### DIFF
--- a/cmd/traffic/cmd/manager/internal/state/state.go
+++ b/cmd/traffic/cmd/manager/internal/state/state.go
@@ -688,6 +688,9 @@ func (s *State) AgentTunnel(ctx context.Context, clientSessionInfo *rpc.SessionI
 		if !ok {
 			return nil, nil, status.Errorf(codes.NotFound, "client session %q not found", clientSessionID)
 		}
+		if cs.ClientTunnelServer == nil {
+			return nil, nil, status.Errorf(codes.NotFound, "client session %q has no tunnel server", clientSessionID)
+		}
 		return as, cs, nil
 	}()
 	if err != nil {


cmd/traffic/cmd/manager/internal/state/state.go
```diff
@@ -688,6 +688,9 @@ func (s *State) AgentTunnel(ctx context.Context, clientSessionInfo *rpc.SessionI
 		if !ok {
 			return nil, nil, status.Errorf(codes.NotFound, "client session %q not found", clientSessionID)
 		}
+		if cs.ClientTunnelServer == nil {
```
thallgren, line 691: @johnlanni can you please describe how you ended up with `cs.ClientTunnelServer` being `nil`  at this point? I can't see how that could happen. The `AgentTunnel` is created during an intercept and an intercept cannot be made unless the `ClientTunnelServer` is active.

johnlanni: @thallgren I'm also confused about this, but seeing this line caused panic, the only possibility is that `ClientTunnelServer` is nil. The `cs` and `msg` variables cannot be nil pointers.
```go
 if err = cs.ClientTunnelServer.Send(msg.TunnelMessage()); err != nil {
```

johnlanni: After I upgraded my telepresence to 2.3.6, this problem has not happened yet.

thallgren:
Closing this as the tunneling was completely rewritten in 2.4.5 and the support for `AgentTunnel` was finally dropped in `2.5.0`.

PR #1877, opened 2021-07-16
Misc cleanup, especially DEVELOPING.md

LukeShu:
## Description

This is some cleanup work that was going to end up in my `--apikey` PR, but I figured was worth PRing separately for review purposes.

A large part of this PR is that the `TELEPRESENCE_REGISTRY` change broke my dev workflow, and so I looked at `DEVELOPING.md`, and became aware of how much stuff in there was outdated.

There's also some code cleanup too.

## Checklist

 - [x] My change is adequately tested. - yeah, if CI passes
 - [x] I updated `DEVELOPING.md` with any any special dev tricks I had to use to work on this code efficiently. - that's like 90% of what this is

### DIFF
--- a/cmd/telepresence/main.go
+++ b/cmd/telepresence/main.go
@@ -8,6 +8,8 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/telepresenceio/telepresence/v2/pkg/client/cli"
+	"github.com/telepresenceio/telepresence/v2/pkg/client/connector"
+	"github.com/telepresenceio/telepresence/v2/pkg/client/daemon"
 	"github.com/telepresenceio/telepresence/v2/pkg/filelocation"
 )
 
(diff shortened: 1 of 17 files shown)


pkg/version/version.go
```diff
@@ -12,13 +16,58 @@ var Version string
...
 	// Prefer version number inserted at build using --ldflags, but if it's not set...
 	if Version == "" {
-		if v := os.Getenv("TELEPRESENCE_VERSION"); v != "" {
-			Version = v
-		} else if i, ok := debug.ReadBuildInfo(); ok {
+		if i, ok := debug.ReadBuildInfo(); ok {
 			// Fall back to version info from "go get"
 			Version = i.Main.Version
 		} else {
 			Version = "(unknown version)"
```
thallgren, line 23: Why not default to `"(devel)"` here and get rid of the additional tests?

pkg/client/connector/userd_trafficmgr/install_test.go
```diff
@@ -94,7 +94,7 @@ func removeManager(t *testing.T, kubeconfig, managerNamespace string) {
 func TestE2E(t *testing.T) {
 	ctx := dlog.NewTestContext(t, false)
 	kubeconfig := dtest.Kubeconfig(ctx)
-	registry := dtest.DockerRegistry(ctx)
```
thallgren: This will invalidate a subsequent change that I made which is now merged. That change is:
```
	os.Setenv("DTEST_REGISTRY", registry) // Prevent calls to dtest.RegistryUp() which may panic
```

cmd/telepresence/main.go
```diff
@@ -19,6 +21,8 @@ func main() {
 	}
 
 	cmd := cli.Command(ctx)
+	cmd.AddCommand(daemon.Command())
```
thallgren: This is likely to conflict with what's on release/v2 where I added special handling for the root daemon.

thallgren (approved):
I left some comments but overall I think this looks great. It needs a rebase which will result in a couple of conflicts but other than that, it's good to go.

PR #1880, opened 2021-07-16
Scripts to build on windows

josecv:
## Description

## Checklist


thallgren (approved):
Assuming it all works, this looks good to me. I have a hard time parsing those for loops ;-)

thallgren:
Oh, and you'll need to gpg-sign and sign-off on the commits.

donnyyung (approved):
This LGTM and since I had a PR that was built on top of this PR, I anecdotally feel that the tests are very thorough and consistent.

.circleci/config.yml
```diff
@@ -108,6 +176,51 @@ commands:
...
+          platform: windows
+      - run:
+          command: |
+            DTEST_KUBECONFIG="$HOME/kubeconfig" PATH="/c/:/c/Program Files/SSHFS-Win/bin:$PATH" DTEST_REGISTRY="docker.io/datawire" make test
+          no_output_timeout: 25m
+      - cleanup-kluster:
+          platform: windows
+      - save-logs
+
   "build-and-test":
```
donnyyung: I would maybe change this to "build-and-test-linux" but we can always do that at a later point

pkg/client/logging/dup_unix.go
```diff
@@ -21,3 +23,23 @@ func dupToStd(file *os.File) (err error) {
...
+	return stdoutFd, stderrFd, err
+}
+
+func restoreStd(stdoutFd, stderrFd int) error {
+	err := unix.Dup2(stdoutFd, 1)
+	if err != nil {
+		return err
+	}
+	return unix.Dup2(stderrFd, 2)
+}
```
LukeShu: Since these are only used by the tests, I'd prefer for them to be in `dup_{unix,windows}_test.go` files, rather than in the main `dup_unix.go` and `sysinfo_windows.go`.

pkg/client/logging/sysinfo_windows.go
```diff
@@ -16,6 +16,14 @@ func dupToStd(_ *os.File) interface{} {
...
 }
 
+func dupStd() (int, int, error) {
+	return -1, -1, errors.New("dupStd() is not implemented on windows")
+}
+
+func restoreStd(int, int) error {
+	return errors.New("restoreStd() is not implemented on windows")
+}
+
```
LukeShu: Rather than having `initcontext_test.go` wrap these calls in `if runtime.GOOS != "windows" { … }`, why not just make these be safe no-ops on Windows?  And the instead of having them return ints, return a cleanup function.

So on unix something like

```go
func dupStd() (func(), error) {
	stdoutFd, err := unix.Dup(1)
	if err != nil {
		return nil, err
	}
	stderrFd, err := unix.Dup(2)
	if err != nil {
		return nil, err
	}
	return func() {
		_ = unix.Dup2(stdoutFd, 1)
		_ = unix.Dup2(stderrFd, 2)
	}, nil
}
```

and on Windows

```go
func dupStd() (func(), error) {
	return func() {}, nil
}
```

josecv: that is definitely much better

pkg/client/logging/rotatingfile.go
```diff
@@ -240,48 +240,48 @@ func (rf *RotatingFile) openNew(prevInfo SysInfo) (err error) {
 	var newFile *os.File
 	if rf.file == nil {
 		if newFile, err = createFile(fullPath, rf.fileMode); err != nil {
-			return err
+			return fmt.Errorf("Failed to createFile %s: %w", fullPath, err)
```
LukeShu: Error messages usually start lowercase

winmake.bat
```diff
@@ -0,0 +1,46 @@
+echo off
```
LukeShu: I'd encourage you to put files like this in the `build-aux/` directory.

Dockerfile.winbuild
```diff
@@ -0,0 +1,17 @@
+FROM golang:1.16.5
```
LukeShu: I'd encourage you to put files like this in the `build-aux/` directory.

pkg/client/cli/telepresence_test.go
```diff
@@ -171,11 +173,13 @@ func (ts *telepresenceSuite) TearDownSuite() {
...
-		ts.Empty(stderr)
-		ts.Contains(stdout, fmt.Sprintf("Client: %s", client.DisplayVersion()))
-	})
+	if goRuntime.GOOS != "windows" {
+		ts.Run("Version", func() {
+			stdout, stderr := telepresence(ts.T(), "version")
+			ts.Empty(stderr)
+			ts.Contains(stdout, fmt.Sprintf("Client: %s", client.DisplayVersion()))
+		})
+	}
```
LukeShu: Why can't this test run on Windows?

pkg/client/cli/telepresence_test.go
```diff
@@ -303,6 +307,11 @@ func (ts *telepresenceSuite) TestA_WithNoDaemonRunning() {
 		ctx = filelocation.WithAppUserLogDir(ctx, tmpDir)
 		_, stderr := telepresenceContext(ctx, "connect")
 		require.Empty(stderr)
+		if goRuntime.GOOS == "windows" {
+			// Windows needs some time to get its bearings
+			time.Sleep(10 * time.Second)
+		}
```
LukeShu, line 311: This at least needs a "FIXME" or a "BUG" comment; `telepresence connect` should block until it is set up.

pkg/client/cli/telepresence_test.go
josecv: it actually can now, we just had to update the stable.txt file; good catch, i'll re-enable this test

pkg/client/cli/telepresence_test.go
```diff
@@ -619,6 +637,10 @@ func (cs *connectedSuite) TestJ_ListOnlyMapped() {
 }
 
 func (cs *connectedSuite) TestK_DockerRun() {
+	// This test only runs on linux as it requires docker, and CI can't run linux docker containers inside non-linux runners
+	if goRuntime.GOOS != "linux" {
+		return
+	}
```
LukeShu, line 641: Prefer to use `cs.SkipNow()` (`cs.T().SkipNow()`?) instead of just returning early, so that the summary shows that it was skipped.

Does CI have `docker` installed?  If so, we should prefer to make this decision based on that rather than coding CI details in to the tests.

pkg/client/cli/telepresence_test.go
```diff
@@ -1391,17 +1441,22 @@ func (ts *telepresenceSuite) kubectlOut(ctx context.Context, args ...string) (st
...
 	// Go sets a lot of variables that we don't want to pass on to the ko executable. If we do,
 	// then it builds for the platform indicated by those variables.
 	cmd.Env = []string{
 		"TELEPRESENCE_VERSION=" + ts.testVersion,
 		"TELEPRESENCE_REGISTRY=" + dtest.DockerRegistry(ctx),
 	}
-	includeEnv := []string{"HOME=", "PATH=", "LOGNAME=", "TMPDIR=", "MAKELEVEL="}
+	includeEnv := []string{"HOME=", "PATH=", "Path=", "LOGNAME=", "TMPDIR=", "MAKELEVEL="}
 	for _, env := range os.Environ() {
+		dlog.Infof(ctx, "I found variable %s", env)
```
LukeShu: Not OK! This will leak CI secrets if the test fails!

pkg/client/cli/telepresence_test.go
LukeShu: And even then, it would be better to call `.SkipNow()` inside of the test.

pkg/client/logging/initcontext_test.go
```diff
@@ -98,24 +99,27 @@ func TestInitContext(t *testing.T) {
 		check.Contains(string(bs), fmt.Sprintf("%s\n%s\n", infoMsg, errMsg))
 	})
 
-	t.Run("captures output of builtin functions", func(t *testing.T) {
-		ctx, _, logFile := testSetup(t)
-		check := require.New(t)
+	// This will fail on Windows
+	if runtime.GOOS != "windows" {
+		t.Run("captures output of builtin functions", func(t *testing.T) {
+			ctx, _, logFile := testSetup(t)
```
LukeShu: It would be better to `.SkipNow()` inside of the test.

pkg/client/cli/telepresence_test.go
josecv: OMG this is very stupid on my part; I believe Circle should [mask](https://circleci.com/docs/2.0/env-vars/#secrets-masking) the secrets, and I haven't seen them be printed, but I'll definitely remove this and go check that it hasn't  been printed anywhere

pkg/client/cli/telepresence_test.go
josecv: So CI has `docker` installed, it's just that pulling the image will fail because there's no windows version available.

I'll make the skipNow change

.circleci/config.yml
```diff
@@ -7,9 +7,13 @@
 #     + DOCKERHUB_PASSWORD
 #     + AWS_ACCESS_KEY_ID (for S3)
 #     + AWS_SECRET_ACCESS_KEY (for S3)
+#     + TELEPRESENCE_KUBECEPTION_TOKEN
```
LukeShu: Variables that Telepresence isn't using at runtime should probably have `DEV_` in them.

pkg/client/cli/telepresence_test.go
```diff
@@ -1526,6 +1584,9 @@ func telepresence(t testing.TB, args ...string) (string, string) {
 // telepresence executes the CLI command in-process
 func telepresenceContext(ctx context.Context, args ...string) (string, string) {
 	var stdout, stderr strings.Builder
+	// The cmd object does not log with timestamps, so dump out the telepresence command here.
+	// That way we have a timestamped record of when it ran, which is useful for correlating with the daemon logs
+	dlog.Debug(ctx, "telepresence invoked with", args)
```
LukeShu, line 1586: Bwuuu? `cmd` is using dlog, same as this is?

.circleci/config.yml
```diff
@@ -108,22 +176,74 @@ commands:
...
     steps:
       - checkout
       - install-go
       - run: make build
       - install-kubectl
       - install-sshfs
+      - attach_workspace:
+          at: /tmp/workspace
+      - prepare-kluster:
+          platform: linux
```
LukeShu, line 235: Part of me doesn't like that we're doing this even when unnecessary (i.e. on GOOS=linux where dtest can spin up a local k3s cluster just fine).  But another part of me hates dtest and is excited that this gets us closer to killing it.

pkg/tun/device_windows.go
```diff
@@ -102,7 +112,40 @@ func (t *Device) setDNS(ctx context.Context, server net.IP, domains []string) (e
...
+}
+$job | Receive-Job
+`, t.interfaceIndex, domain)
+	cmd := dexec.CommandContext(ctx, "powershell.exe", "-NoProfile", "-NonInteractive", pshScript)
+	cmd.DisableLogging = true // disable chatty logging
+	dlog.Debugf(ctx, "Calling powershell's SetDNSDomain %q", domain)
+	if err := cmd.Run(); err != nil {
+		// Log the error, but don't actually fail on it: This is all just a fallback for SetDNS, so the domains might actually be working
+		dlog.Errorf(ctx, "Failed to set NetworkAdapterConfiguration DNS Domain: %v. Will proceed, but namespace mapping might not be functional.", err)
+	}
```
LukeShu, line 148: gross gross gross :)

I'm not looking at this too closely, but it would be good to include references in those comments ("apparently well-known"), and maybe be more specific that "some systems".

josecv: > would be good to include references in those comments ("apparently well-known"),

Good point, I'll add some links to the comment

> and maybe be more specific that "some systems".

This I can't do :( unfortunately "some systems" means "CircleCI but not Jose's windows computer"

LukeShu (changes requested):
A couple of minor changes.

PR #1881, opened 2021-07-16
Minimize the amount of info that the client requests from the cluster.

### LINKED ISSUE #1739: [v2] Telepresence list is too slow
I have a namespace `xx` with 41 deployments.
When I run `telepresence list -n xx` it takes 80 seconds to respond.

When I run it against my `default` namespace (8 deployments) it takes 3 seconds.

```
$ telepresence version
Client v2.2.1 (api v3)
Daemon v2.2.1 (api v3)
```

thallgren:
## Description

This PR aims to remove a lot of redundant requests that were made by the client in order to obtain agents and various other information about resources in the cluster. The focus was to make `telepresence list` faster, but this PR is likely to also speed up other things.

## Checklist


pkg/client/connector/userd_k8s/k8s_cluster.go
```diff
@@ -101,37 +101,56 @@ func (kc *Cluster) check(c context.Context) error {
...
+	objs := make([]kates.Object, len(statefulSets))
+	for i, ss := range statefulSets {
+		objs[i] = ss
+	}
+	return objs, nil
 }
 
-// PodNames returns the names of all replica sets found in the given Namespace
+// PodNames returns the names of all pods found in the given Namespace
 func (kc *Cluster) PodNames(c context.Context, namespace string) ([]string, error) {
```
donnyyung, line 144: I think we could make a `func (kc *Cluster) Pods` like you did for the other objects, since it's only called in `refreshReplicaSet`, we could delete this function, as well as func `FindPod`

And then refreshReplicaSet would be something like this (snippet has not been tested, it might work but think of it as just pseudocode to get my point across):
```
  353 func (ki *installer) refreshReplicaSet(c context.Context, namespace string, rs *kates.ReplicaSet) error {
  354     pods, err := ki.Pods(c, namespace)
  355     if err != nil {
  356         return err
  357     }
  358     for _, pod := range pods {
  359         for _, ownerRef := range pod.OwnerReferences {
  360             if ownerRef.UID == rs.UID {
  361                 dlog.Infof(c, "Deleting pod %s owned by rs %s", pod.Name, rs.Name)
  362                 pod := &kates.Pod{
  363                     TypeMeta: kates.TypeMeta{
  364                         Kind: "Pod",
  365                     },
  366                     ObjectMeta: kates.ObjectMeta{
  367                         Namespace: pod.Namespace,
  368                         Name:      pod.Name,
  369                     },
  370                 }
  371                 if err := ki.Client().Delete(c, pod, pod); err != nil {
  372                     return err
  373                 }
  374             }
  375         }
  376     }
  377     return nil
  378 }
  ```

donnyyung (approved):
LGTM, I left one comment about a way I think we could further optimize but I don't think your PR needs to be blocked on that... sorry that most of this non-optimal code was introduced by me a few months ago 🤦

PR #1886, opened 2021-07-18
Avoid adding all subcommands except "daemon-foreground" for root daemon.

thallgren:
## Description

Attempting to load extensions or provide other options to commands in
the root daemon is pointless. This commit ensures that if the subcommand
is `daemon-foreground`, then only that subcommand is added.

Among other things, this restores the logging functionality in the
daemon which was otherwise initialized too early which cause the config
to be read from the roots config directory rather than the users.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Bugfix: Fixed an issue that could cause the user daemon to crash
   during shutdown.
 - Bugfix: The Traffic Manager's RoleBinding now correctly appoint the "traffic-manager" role so that service subnet discovery works correctly.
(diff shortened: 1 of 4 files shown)


LukeShu (approved):
OK. I'd rather do this in `cmd/telepresence/main.go` to be able to be extra-sure that no cli stuff is happening.  But this is fine.

thallgren:
@LukeShu I did move the check to `cmd/telepresence/main.go` because there were indeed some cli stuff happening (check for legacy commands for instance). Given the urgency of getting this out, I didn't want to wait for you to re-review so I merged it.

PR #1887, opened 2021-07-18
Fixed traffic-manager unable to detect service IP range

akharlamov:
## Description

The PR fixes issue with an auto-installed traffic-manager unable to detect services IP range on the cluster.
The root cause was incorrectly named role traffic-manager-ambassador instead of traffic-manager. 

The patch is proved to fix the issue but was not completely tested.


thallgren:
Thanks @akharlamov, but #1884 beat you to it I'm afraid.

akharlamov:
Oh, well. It is fixed. That's good.

PR #1899, opened 2021-07-20
Improve logging and tests.

thallgren:
## Description

This PR fixes the logging of the `PortForwardDialer`, increases the number of log files that can rotate when running tests, fixes capture of the `traffic-manager` logs during tests, and fixes a glitch in a test where helm uninstalled the `traffic-manager` without first quitting the client (this is what made all subsequent calls to `Remain()` fail).

## Checklist

 - [x] I made sure to update `./CHANGELOG.md`. (no entries needed. This is all about our test framework).

### DIFF
--- a/pkg/client/cli/telepresence_test.go
+++ b/pkg/client/cli/telepresence_test.go
@@ -1258,6 +1258,7 @@ func (hs *helmSuite) TestF_MultipleInstalls() {
 		hs.Contains(stdout, "with-probes: intercepted")
 	})
 	hs.Run("Uninstalls successfully", func() {
+		telepresence(hs.T(), "quit")
 		hs.NoError(run(ctx, "kubectl", "config", "use-context", "default"))
 		defer func() { hs.NoError(run(ctx, "kubectl", "config", "use-context", "telepresence-test-developer")) }()
 		hs.NoError(run(ctx, "helm", "uninstall", "traffic-manager", "-n", hs.managerNamespace2))
--- a/pkg/client/cli/telepresence_test.go
+++ b/pkg/client/cli/telepresence_test.go
@@ -18,6 +18,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/datawire/dlib/dtime"
+
 	"github.com/stretchr/testify/suite"
 	k8sruntime "k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/tools/clientcmd"
@@ -51,6 +53,7 @@ type telepresenceSuite struct {
 	testVersion          string
 	namespace            string
 	managerTestNamespace string
+	logCapturingPods     sync.Map
 }
 
 func (ts *telepresenceSuite) SetupSuite() {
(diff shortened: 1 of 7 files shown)


pkg/dnet/kpfconn.go
```diff
@@ -27,6 +27,7 @@ import (
 
 type k8sPortForwardDialer struct {
 	// static
+	logCtx          context.Context
```
LukeShu, line 30: I'm not sure adding this is right.  9 times out of 10, storing a context in a struct is the wrong thing.  Since we're only using it for logging, it doesn't matter if it's canceled; and so storing a persistent logCtx only serves to obscure where `Dial` is being called from (not that we call it from very many places).

LukeShu (changes requested):
Other than `logCtx`, I love it.

pkg/client/cli/telepresence_test.go
```diff
@@ -1251,6 +1258,7 @@ func (hs *helmSuite) TestF_MultipleInstalls() {
 		hs.Contains(stdout, "with-probes: intercepted")
 	})
 	hs.Run("Uninstalls successfully", func() {
+		telepresence(hs.T(), "quit")
```
josecv, line 1261: this makes lots of sense, should we also add it to `TestZ_Uninstall`? It's not necessary right now, but if we ever add more tests between G and Z it could be

pkg/dnet/kpfconn.go
thallgren: I don't love it either, but I couldn't find any other way to make a logger configured context available to the dialer. The context passed in `DialContext()` is NOT propagated to the dialer. So I added this and specifically named it `logCtx` to indicate that this context is indeed only intended for logging. Do you have a better proposal?

josecv (changes requested):
+1 on Luke's comment, also one small suggestion :)

pkg/dnet/kpfconn.go
LukeShu: Wait, really?

thallgren: To clarify, this change has nothing to do with cancelled contexts. It's needed because the context that the dialer gets, is not configured for logging. It's created here: https://github.com/datawire/grpc-go/blob/master/clientconn.go#L147

thallgren: I guess that `context.Background()` is a perfect candidate for a `dcontext.WithoutCancel(ctx)`, but perhaps a depencendy to `dlib` is undesired here?

LukeShu: Wow, really.  Huh. OK.  I guess that's another PR that I should file against the gRPC library :/

pkg/client/cli/telepresence_test.go
thallgren: Added.

LukeShu (approved):

josecv (approved):

PR #1903, opened 2021-07-20
Adding rpc for cloud messages

donnyyung:

## Description

This is the rpc I'm suggesting we create for: https://github.com/datawire/saas_app/issues/1367 . I made it a little open-ended in case we havre more messages in the future, but fine to restrict it if you think I over-engineered this.

## Checklist


rpc/systema/cli2systema.proto
```diff
@@ -0,0 +1,26 @@
...
+option go_package = "github.com/telepresenceio/telepresence/rpc/v2/systema";
+
+// Calls from the cli to Ambassador Cloud (System A) are unauthenticated
+// This is because this rpc is (currently) only used to get messages from
+// the cloud that we want to raise to our users.
+service SystemACli {
+  // GetMessages is used by the cli to get messages from the cloud.
+  // Telepresence should cache these messages, since they won't change
+  // too frequently.
+  rpc GetMessages(google.protobuf.Empty) returns (MessageResponse);
```
thallgren: If this is something that we want to use when the user issues a command, why return a map? Why not:
```
message StringMessage {
  string text = 1;
}
```
and then
```
GetUnauthenticatedResponse(StringMessage) StringMessage
```
Or are there situations when we want to show more than one command at a time?

donnyyung: So for now, we only are going to be raising a message when a user does a global intercept (while not logged in).  Max mentioned that we might potentially have more messages for other commands in the future, so I went with a map so we could get the messages for all commands at one time and then cache it in a file on the users machine (so we don't do a call to Ambassador Cloud every time we run a command to see if there's a message).

So I was thinking this call wouldn't necessarily be done per-command, but whenever a user issues a command, it would:

- Check to see if the file exists, if not, do the call to ambassador cloud to populate the file
- If the file exists, ensure that it was fetched somewhat recently (how recent is TBD), if not recent, fetch messages again
- Check to see if there is a message associated with the command, if there is, render it to the user
- Update the file that the message was rendered so that we don't annoy the user by repeating the same messages over and over again.

Does that make sense?

thallgren: Yes, if the intent is to cache the messages, it makes a lot of sense. I would suggest using more descriptive names for both method and return value though. The term `Message` is extremely generic and we might want to expand the API with more "messages" in the future.

donnyyung: Yeah good point, I think I made it too vague to support some _potential_ future use-case.. but I think it's better to start with something that is more clear to the user and we can make it more complicated as the need arises.. so I've pushed an update that has improved the names, as well as, removed the map for a CommandMessageResponse, with one entry `string intercept`, which contains the message to be given when a user does an intercept. 

If we end up needing more messages, say to `quit`, we could add `string quit = 2;`. I think this will be simpler and more clear than one map for all possible commands.

thallgren (approved):

rpc/systema/cli2systema.proto
thallgren: One thing to consider is that telepresence must be able to detect when it's time to refresh the cache. Perhaps it's as simple as storing the telepresence's version in the cache and the just make the call when the version changes.

donnyyung: Yup that'll be handled in the follow-up PR that I'm going to start working on now!

PR #1904, opened 2021-07-21
More efficient Pod deletion in refreshReplicaSet() function.

thallgren:
## Description

Let the `refreshReplicaSet()` function fetch Pods directly instead of
first fetching their names and then do a proper fetch.

## Checklist

 - [x] I made sure to update `./CHANGELOG.md`. Not needed, this is just a minor optimization.


donnyyung (approved):
Thank you for taking this suggestion!

PR #1905, opened 2021-07-21
Close dialer when get ConnectReject

donnyyung:

## Description

This adds a case to the dialer to close it when it gets "ConnectReject".  This is encountered when you do an intercept and your local process isn't running, if you try to curl that endpoint it will hand indefinitely:
```
(⎈ |default:default)➜  ~/go/telepresenceio git:(donnyyung/no-process) ✗ telepresence intercept echo-easy --port 1235
Using Deployment echo-easy
intercepted
    Intercept name    : echo-easy
    State             : ACTIVE
    Destination       : 127.0.0.1:1235
    Volume Mount Point: /var/folders/cp/2r22shfd50d9ymgrw14fd23r0000gp/T/telfs-249111851
(⎈ |default:default)➜  ~/go/telepresenceio git:(donnyyung/no-process) ✗ curl echo-easy
^C
```

And starting a process on that port still doesn't result in the request going through, so this PR correctly handles ConnectReject by closing it:
```
(⎈ |default:default)➜  ~/go/telepresenceio git:(donnyyung/no-process) ✗ telepresence intercept echo-easy --port 1235
Using Deployment echo-easy
intercepted
    Intercept name    : echo-easy
    State             : ACTIVE
    Destination       : 127.0.0.1:1235
    Volume Mount Point: /var/folders/cp/2r22shfd50d9ymgrw14fd23r0000gp/T/telfs-744925586
(⎈ |default:default)➜  ~/go/telepresenceio git:(donnyyung/no-process) ✗ curl echo-easy
curl: (52) Empty reply from server
```
## Checklist

### DIFF
--- a/pkg/connpool/dialer.go
+++ b/pkg/connpool/dialer.go
@@ -117,7 +117,7 @@ func (h *dialer) handleControl(ctx context.Context, cm Control) {
 	case ConnectOK:
 		go h.writeLoop(ctx)
 		go h.readLoop(ctx)
-	case Disconnect, ConnectReject:
+	case Disconnect:
 		h.Close(ctx)
 		h.sendTCD(ctx, DisconnectOK)
 	case ReadClosed:
@@ -134,7 +134,7 @@ func (h *dialer) handleControl(ctx context.Context, cm Control) {
 				close(h.writerClosing)
 			}
 		}
-	case WriteClosed:
+	case WriteClosed, ConnectReject:
 		h.Close(ctx)
 	case KeepAlive:
 		h.resetIdle()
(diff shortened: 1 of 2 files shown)


pkg/connpool/dialer.go
```diff
@@ -117,7 +117,7 @@ func (h *dialer) handleControl(ctx context.Context, cm Control) {
 	case ConnectOK:
 		go h.writeLoop(ctx)
 		go h.readLoop(ctx)
-	case Disconnect:
+	case Disconnect, ConnectReject:
```
thallgren: I think this is more similar to a `WriteClosed` in that there's no point in sending a `DisconnectOK` back.

thallgren (changes requested):

pkg/connpool/dialer.go
donnyyung: Yeah that makes sense... and "Disconnect" implies a connection, which we don't have if we get "ConnectReject" so I have made the change!

thallgren (approved):

Issue #1907, opened 2021-07-21
Telepresence v2.3.x mutating webhook annotation results in ReplicaSet controller error


frankgu968:
When the [mutating webhook](https://www.getambassador.io/docs/telepresence/latest/reference/cluster-config/) is enabled with the annotation `telepresence.getambassador.io/inject-traffic-agent: enabled`, the cluster `replicaset-controller` throws error due to incorrect key: 

```
Conditions:
  Type             Status  Reason
  ----             ------  ------
  ReplicaFailure   True    FailedCreate
Events:
  Type     Reason        Age                 From                   Message
  ----     ------        ----                ----                   -------
  Warning  FailedCreate  14m (x20 over 53m)  replicaset-controller  Error creating: Internal error occurred: replace operation does not apply: doc is missing path: /spec/containers/0//livenessProbehttpGet/port/name: missing value
```

It is clear that there is a string substitution problem in the resource path. 

Upon closer inspection of the source, the problem can be narrowed down to the `hidePorts` function in `cmd/traffic/manager/internal/mutator/agent_injector.go`. Specifically, 2 problems exist: 
1. The slash on the livenessProbe path is in the wrong position
2. k8s livenessProblem named ports use the `ports: <name>` notation rather than having a `/port/name` path. 
The correct k8s resource path should be `/spec/containers/0/livenessProbe/httpGet/port` in this case. [Reference](/spec/containers/0//livenessProbehttpGet/port)

thallgren:
Resolved in 2.3.7. Thanks @frankgu968 for the fix!

PR #1908, opened 2021-07-21
[v2] Fix incorrect hidePorts k8s resource path

frankgu968:
## Description

This PR closes issue https://github.com/telepresenceio/telepresence/issues/1907

No documentation changes are necessary and no special dev tricks were used. 
## Checklist

### DIFF
--- a/cmd/traffic/cmd/manager/internal/mutator/agent_injector.go
+++ b/cmd/traffic/cmd/manager/internal/mutator/agent_injector.go
@@ -197,20 +197,20 @@ func hidePorts(pod *corev1.Pod, cn *corev1.Container, portName string, patches [
 	hidePort := func(path string) {
 		patches = append(patches, patchOperation{
 			Op:    "replace",
-			Path:  fmt.Sprintf("%s/%s/name", containerPath, path),
+			Path:  fmt.Sprintf("%s/%s", containerPath, path),
 			Value: hiddenPortName,
 		})
 	}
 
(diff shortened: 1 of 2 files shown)


frankgu968:
^ not sure what the specific test error is here...

cmd/traffic/cmd/manager/internal/mutator/agent_injector.go
```diff
@@ -197,7 +197,7 @@ func hidePorts(pod *corev1.Pod, cn *corev1.Container, portName string, patches [
 	hidePort := func(path string) {
 		patches = append(patches, patchOperation{
 			Op:    "replace",
-			Path:  fmt.Sprintf("%s/%s/name", containerPath, path),
+			Path:  fmt.Sprintf("%s/%s", containerPath, path),
```
thallgren, line 200: While this change fixes the probe ports, it will no longer work on the container port. For the container it is indeed the value of the port's "name" key that should be replaced. I think that can be fixed by appending the `/name` suffix at 207.

thallgren (changes requested):

cmd/traffic/cmd/manager/internal/mutator/agent_injector.go
frankgu968: Good catch! Fixed according to your suggestion

thallgren (approved):
All good now. Many thanks for providing this @frankgu968 .

PR #1909, opened 2021-07-22
Improve systemd-resolved based DNS resolver stability and handover to the overriding one

### LINKED ISSUE #1754: Telepresence V2 : My DNS is disabled once telepresence connect
I installed Telepresence V2.2 then after connect By : 

 telepresence connect

I connect to remote cluster but My DNS is overridden by telepresence and not able to see it

thallgren:
## Description

This PR deals with initialization of the `systemd-resolved` DNS resolver and if it fails, how telepresence defaults to the overriding resolver. The first three commits should be considered more of a clean-up and consistency nature. Please review one commit at a time.

## Checklist

 - [x I made sure to update `./CHANGELOG.md`.

### DIFF
--- a/pkg/client/daemon/dbus/resolved.go
+++ b/pkg/client/daemon/dbus/resolved.go
@@ -1,6 +1,7 @@
 package dbus
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"net"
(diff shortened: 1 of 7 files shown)


pkg/client/daemon/dbus/resolved.go
```diff
@@ -1,21 +1,19 @@
...
 	"strings"
 	"syscall"
 
 	"github.com/godbus/dbus/v5"
+
+	"github.com/datawire/dlib/dlog"
 )
 
 type (
-	// A ResolveD is a `dbus.Conn` with methods to communicate with the ResolveD daemon.
```
donnyyung, line 14: I think this is mostly bc of my ignorance and lack of experience with resolveD, but are we not using a persistent connection to increase stability (e.g. and not have some failure if the connection is closed unexpectedly) and bc the number of calls are infrequent enough that it's not problematic to make a new connection each time we need to communicate with the dbus?

donnyyung (approved):
I don't have the strongest understanding of resolveD / don't work on a machine that uses it, but I don't see any red flags here (my comment was mostly just  asking for a better understanding of our pivot to not having a long-running dbus connection) and I think it's worth it to merge to get a nightly that our users can test before 2.3.7

pkg/client/daemon/dbus/resolved.go
thallgren: The calls are very infrequent, yes. Only happens when we update dns (intercept a new namespace) or when the set of IP subnet changes (which in reality almost never happens), so I saw no reason why the connection should be kept open.

PR #1910, opened 2021-07-23
Surface logs on error

rhs:
## Description

Display a summary of the daemon and connector logs to the user whenever telepresence exits with an error. This will help advertise the location of the logs and also give users a better chance of diagnosing/fixing issues.

## Checklist

### DIFF
--- a/cmd/telepresence/main.go
+++ b/cmd/telepresence/main.go
@@ -61,8 +61,8 @@ func summarizeLogs(ctx context.Context, cmd *cobra.Command) {
 		fmt.Fprintf(cmd.ErrOrStderr(), "%s: error: %+v\n", cmd.CommandPath(), err)
 	}
 
-	fmt.Fprintf(cmd.ErrOrStderr(), "\n%s", daemonLogs)
 	if daemonLogs != "" {
+		fmt.Fprintf(cmd.ErrOrStderr(), "\n%s", daemonLogs)
 	}
 	if connectorLogs != "" {
 		fmt.Fprintf(cmd.ErrOrStderr(), "\n%s", connectorLogs)
--- a/cmd/telepresence/main.go
+++ b/cmd/telepresence/main.go
@@ -10,6 +10,7 @@ import (
 	"github.com/telepresenceio/telepresence/v2/pkg/client/cli"
 	"github.com/telepresenceio/telepresence/v2/pkg/client/connector"
 	"github.com/telepresenceio/telepresence/v2/pkg/client/daemon"
+	"github.com/telepresenceio/telepresence/v2/pkg/client/logging"
 	"github.com/telepresenceio/telepresence/v2/pkg/filelocation"
 )
 
@@ -45,6 +46,25 @@ func main() {
 
 	if err := cmd.ExecuteContext(ctx); err != nil {
 		fmt.Fprintf(cmd.ErrOrStderr(), "%s: error: %v\n", cmd.CommandPath(), err)
+		summarizeLogs(ctx, cmd)
 		os.Exit(1)
+	}
+}
+
+func summarizeLogs(ctx context.Context, cmd *cobra.Command) {
+	daemonLogs, err := logging.SummarizeLog(ctx, daemon.ProcessName)
+	if err != nil {
+		fmt.Fprintf(cmd.ErrOrStderr(), "%s: error: %+v\n", cmd.CommandPath(), err)
+	}
+	connectorLogs, err := logging.SummarizeLog(ctx, connector.ProcessName)
+	if err != nil {
+		fmt.Fprintf(cmd.ErrOrStderr(), "%s: error: %+v\n", cmd.CommandPath(), err)
+	}
+
+	fmt.Fprintf(cmd.ErrOrStderr(), "\n%s", daemonLogs)
+	if daemonLogs != "" {
+	}
+	if connectorLogs != "" {
+		fmt.Fprintf(cmd.ErrOrStderr(), "\n%s", connectorLogs)
 	}
 }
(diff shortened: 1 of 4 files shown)


cmd/telepresence/main.go
```diff
@@ -30,6 +33,25 @@ func main() {
...
+		fmt.Fprintf(cmd.ErrOrStderr(), "%s: error: %+v\n", cmd.CommandPath(), err)
+	}
+	connectorLogs, err := logging.SummarizeLog(ctx, connector.ProcessName)
+	if err != nil {
+		fmt.Fprintf(cmd.ErrOrStderr(), "%s: error: %+v\n", cmd.CommandPath(), err)
+	}
+
+	fmt.Fprintf(cmd.ErrOrStderr(), "\n%s", daemonLogs)
+	if daemonLogs != "" {
+	}
```
thallgren, line 72: Missing printout of the daemonLogs.

pkg/client/logging/initcontext.go
```diff
@@ -63,3 +66,61 @@ func InitContext(ctx context.Context, name string) (context.Context, error) {
...
+		// XXX: is there a better way to detect error lines?
+		parts := strings.Fields(text)
+		if len(parts) > 2 && parts[2] == "error" {
+			errors = append(errors, text)
+			if len(errors) > 10 {
+				errors = errors[len(errors)-10:]
+			}
+		}
+		tail = append(tail, text)
+		if len(tail) > 10 {
```
thallgren: Not sure how useful a `tail -n10` will be.  It's mostly messages about cleaning up and shutting down go-routines.

cmd/telepresence/main.go
```diff
@@ -30,6 +33,25 @@ func main() {
 
 	if err := cmd.ExecuteContext(ctx); err != nil {
```
thallgren, line 47: There are some situations where it doesn't make sense to print the logs:
1. This is not the  CLI process (it's the `connector` or `daemon` executing in the background).
2. This is the CLI process, but the error is a usage error (not logged).
3. The error is expected, but not anything that has info in the logs. E.g. user attempts to logout but is not logged in.

thallgren (changes requested):
I'm not sure I'm on board with this. I do think it's great that the user is made aware of the location of the log files but the errors there can be both unrelated or otherwise confusing when presented out of context. I think presenting logs up front is somewhat brutal and might scare users away.

Suggestion: Inform the user about the location of the log files and that information pertaining to the error _might_ be found there. Don't print out info from actual logs.

cmd/telepresence/main.go
rhs: Thanks! Fixed.

rhs:
> I'm not sure I'm on board with this. I do think it's great that the user is made aware of the location of the log files but the errors there can be both unrelated or otherwise confusing when presented out of context. I think presenting logs up front is somewhat brutal and might scare users away.
> 
> Suggestion: Inform the user about the location of the log files and that information pertaining to the error _might_ be found there. Don't print out info from actual logs.

I think it's helpful to split up users into two categories: demo users (or uncommitted users), and committed users. The primary goal of this change is to help make it possible for a committed user to be successful in using the product, however I do believe it helps with uncommitted users as well. If you are a demo or uncommitted user, then any error has a decent chance of scaring you away, however the worst possible (most frustrating) kind of error is a silent error, and at the moment there seems to be a decently large number of those. I believe on balance this will help even with demo users since it will eliminate a good number of those silent errors.

As a committed user I wrote up my experience trying to use the product here:

https://docs.google.com/document/d/1_M_AdEWBpd-x0cW99ekeNWzXorXStng7OG0o5vnSuVc/edit?usp=sharing

TL;DR with respect to this change is that there were multiple occasions where vital info was in the logs. Now I totally agree it would be better if that vital info was surfaced in a friendlier way, however this seems like a necessary stopgap until we do a better job of surfacing those issues.

I'm open to ditching the tail aspect and only reporting errors, or perhaps filtering the tail. I do think anything the client is doing that involves the cluster is meaningful to report to users. There are a decent number of users who will be able to see a kubectl command and realize it won't work because of e.g. rbac. Do you think there is a reasonable way to filter the tail down a bit?

cmd/telepresence/main.go
rhs: Is there a way to detect if you are the connector or daemon? I added this here because @LukeShu recommended this as the place to start, but I'm happy to make this more specific if that's possible.

rhs:
Had a chat with @thallgren over zoom and came up with the following path forward:

I will break up this PR into two separate phases:

Phase 1:
  - Fix it so only the CLI displays extra info.
  - Only display the location of the logs rather than displaying the contents.
  - If possible try to figure out a simple heuristic for figuring out whether the bubbled up error is high quality or low quality and only display the additional info for low quality errors.
  
Phase 2:
  - I will experiment with some ways to filter down/condense/extract useful info from the logs so we can boost the signal to noise ratio.

pkg/client/logging/initcontext.go
```diff
@@ -63,3 +66,34 @@ func InitContext(ctx context.Context, name string) (context.Context, error) {
...
+	}
+
+	filename := filepath.Join(dir, name+".log")
+	file, err := os.Open(filename)
+	if err != nil {
+		return "", err
+	}
+	scanner := bufio.NewScanner(file)
+
+	errors := []string{}
```
thallgren, line 83: Why not just use an integer?

thallgren (approved):
Looks good to me aside from my little nit-pick. Feel free to ignore.

pkg/client/logging/initcontext.go
rhs: Only because I expect to make it fancier it a later PR. ;-)

PR #1921, opened 2021-07-26
Prevent subnet updates when subnets haven't changed.

thallgren:
## Description

The `traffic-manager` listens to `Nodes` or `Pods` to determine what
pod-subnets that are used. The `Nodes` or `Pods` may change frequently
although the subnet set remains the same, causing a lot of unnecessary
information to be sent from the `traffic-manager` to connected clients.
Especially o larger clusters.

This commit ensures that changes that don't alter the actual subnets
are ignored and not propagated to the root daemon.

## Checklist

### DIFF
--- a/cmd/traffic/cmd/manager/internal/cluster/info.go
+++ b/cmd/traffic/cmd/manager/internal/cluster/info.go
@@ -13,7 +13,6 @@ import (
 	"github.com/datawire/ambassador/pkg/kates"
 	"github.com/datawire/dlib/derror"
 	"github.com/datawire/dlib/dlog"
-	"github.com/telepresenceio/telepresence/rpc/v2/manager"
 	rpc "github.com/telepresenceio/telepresence/rpc/v2/manager"
 	"github.com/telepresenceio/telepresence/v2/cmd/traffic/cmd/manager/managerutil"
 	"github.com/telepresenceio/telepresence/v2/pkg/iputil"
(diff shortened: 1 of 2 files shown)


cmd/traffic/cmd/manager/internal/cluster/info.go
```diff
@@ -178,15 +180,38 @@ func (oi *info) watchSubnets(ctx context.Context, name, kind string, retriever f
...
+					continue
+				}
+			}
+			changed = true
+			break
+		}
+	}
+	if changed {
+		oi.podCIDRMap = makeCIDRMap(podCIDRs)
+		oi.PodSubnets = toRPCSubnets(podCIDRs)
```
LukeShu, line 210: Having these two different views of the same data seems funny to me.  I'd almost rather just have `.podCIDRMap` and call `toRPCSubnets` in the RPC handler.

LukeShu (approved):
Looks good. One nitpick that you can ignore if you want.

cmd/traffic/cmd/manager/internal/cluster/info.go
thallgren: I was thinking of performance (fast comparison using hashmap) and scalability. The data is kept for all clients, which in a big installation might be hundreds. Why trigger hundreds of conversions of the same map in the RPC handler?

PR #1922, opened 2021-07-26
Make references in agent's environment work

thallgren:
## Description

The `traffic-agent` would receive a copy of the intercepted containers
environment where the names were prefixed with "TEL_APP_" to distinguish
those names from the names found in the agent's own environment. That
was problematic, given that when Kubernetes starts a container, it
interpolates references to other variables in the form "$(NAME)".

This commit removes the prefixes so that the names in the environment
copy are left intact. Instead, it's ensured that the names used in the
agent's own environment are prefixed in such a way that a name collision
is highly unlikely.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,11 @@
 # Changelog
 
 ### 2.3.8 (TBD)
-- Bugfix: Initialisatialisation of systemd-resolved` based DNS sets
+- Bugfix: Initialization of systemd-resolved` based DNS sets
   routing domain to improve stability in non-standard configurations.
+
+- Bugfix: A `$(NAME)` reference in the agent's environment will now be
+  interpolated correctly.
 
 ### 2.3.7 (July 23, 2021)
 
--- a/cmd/traffic/cmd/agent/agent.go
+++ b/cmd/traffic/cmd/agent/agent.go
@@ -23,26 +23,26 @@ import (
 )
 
 type Config struct {
-	Name        string `env:"AGENT_NAME,required"`
-	Namespace   string `env:"AGENT_NAMESPACE,default="`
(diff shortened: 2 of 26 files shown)


CHANGELOG.md
```diff
@@ -1,9 +1,12 @@
 # Changelog
 
 ### 2.3.8 (TBD)
```
donnyyung, line 3: This is maybe changed in other PRs but this should be 2.4.0, but also something we can fix before building an rc so not a blocker.

cmd/traffic/cmd/agent/agent.go
```diff
@@ -113,14 +111,14 @@ func Main(ctx context.Context, args ...string) error {
...
-		if os.Getenv("AGENT_NAME") == "" {
-			os.Setenv("AGENT_NAME", "test-agent")
+		if os.Getenv("_TEL_AGENT_NAME") == "" {
+			os.Setenv("_TEL_AGENT_NAME", "test-agent")
 		}
-		if os.Getenv("APP_PORT") == "" {
-			os.Setenv("APP_PORT", "8080")
+		if os.Getenv("_TEL_AGENT_APP_PORT") == "" {
+			os.Setenv("_TEL_AGENT_APP_PORT", "8080")
 		}
```
donnyyung, line 124: I think these changes need to be duplicated in the Smart Agent.

donnyyung (approved):
One change for the changelog and then a comment on something that also needs to be changed on the Smart Agent, I will see if there is anything else there affected by these changes and ping you.

CHANGELOG.md
thallgren: Yeah, I think it says 2.3.8 on the current release/v2 branch too, so let's fix that later on.

PR #1927, opened 2021-07-27
Fixes an edge case when selecting a container by port number

### LINKED ISSUE #1926: Edge case error when targeting a container by port number
if your matching/target container is at containers list index 0, but if there is a container at index 1 with no ports, then the "no ports" container will end up the selected one

```
// Here we are using containerPortIndex <=0 instead of matchingContainer == nil because if a
// container has no ports, we want to use it but we don't want
// to break out of the loop looking at containers in case there
// is a better fit.  Currently, that is a container where the
// ContainerPort matches the targetPort in the service.
for ci := 0; ci < len(cns) && containerPortIndex <= 0; ci++ {
	cn := &cns[ci]
	if len(cn.Ports) == 0 {
		matchingServicePort = port
		matchingContainer = cn
		containerPortIndex = -1
	}
	for pi := range cn.Ports {
		if cn.Ports[pi].ContainerPort == portNum {
			matchingServicePort = port
			matchingContainer = cn
			containerPortIndex = pi
			break
		}
	}
}

```

0x6a77:
## Description

This fixes an edge case when selecting a container by port number. The solution was to make the logic much clearer and easier to read.

## Checklist

resolves #1926

### DIFF
--- a/pkg/install/discovery.go
+++ b/pkg/install/discovery.go
@@ -168,7 +168,7 @@ Please specify the Service port you want to intercept by passing the --port=loca
 		//     kubectl create deploy my-deploy --image my-image
 		//     kubectl expose deploy my-deploy --port 80 --target-port 8080
 		if matchingContainer == nil {
-			for ci := 0; ci < len(cns); ci++ {
+			for ci := range cns {
 				cn := &cns[ci]
 				if len(cn.Ports) == 0 {
 					matchingServicePort = port
--- a/pkg/install/discovery.go
+++ b/pkg/install/discovery.go
@@ -148,8 +148,8 @@ Please specify the Service port you want to intercept by passing the --port=loca
 			}
 		}
 	} else {
-		portNum := port.TargetPort.IntVal
 		// First see if we have a container with a matching port
+		portNum := port.TargetPort.IntVal
 	containerLoop:
 		for ci := range cns {
 			cn := &cns[ci]
@@ -158,7 +158,7 @@ Please specify the Service port you want to intercept by passing the --port=loca
 					matchingServicePort = port
 					matchingContainer = cn
 					containerPortIndex = pi
-					break
+					break containerLoop
 				}
 			}
 		}
@@ -175,7 +175,7 @@ Please specify the Service port you want to intercept by passing the --port=loca
 					matchingContainer = cn
 					containerPortIndex = -1
 					break
-				}	
+				}
 			}
 		}
 	}
--- a/pkg/install/discovery.go
+++ b/pkg/install/discovery.go
@@ -149,18 +149,9 @@ Please specify the Service port you want to intercept by passing the --port=loca
 		}
 	} else {
 		portNum := port.TargetPort.IntVal
-		// Here we are using containerPortIndex <=0 instead of matchingContainer == nil because if a
-		// container has no ports, we want to use it but we don't want
-		// to break out of the loop looking at containers in case there
-		// is a better fit.  Currently, that is a container where the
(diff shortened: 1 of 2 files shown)


pkg/install/discovery.go
```diff
@@ -149,18 +149,9 @@ Please specify the Service port you want to intercept by passing the --port=loca
...
 	} else {
 		portNum := port.TargetPort.IntVal
-		// Here we are using containerPortIndex <=0 instead of matchingContainer == nil because if a
-		// container has no ports, we want to use it but we don't want
-		// to break out of the loop looking at containers in case there
-		// is a better fit.  Currently, that is a container where the
-		// ContainerPort matches the targetPort in the service.
-		for ci := 0; ci < len(cns) && containerPortIndex <= 0; ci++ {
+		// First see if we have a container with a matching port
+		for ci := 0; ci < len(cns); ci++ {
```
thallgren:
```suggestion
	containerLoop:
		for ci := range cns {
```

pkg/install/discovery.go
```diff
@@ -170,6 +161,18 @@ Please specify the Service port you want to intercept by passing the --port=loca
 				}
 			}
 		}
+		// If no container matched, then use the first container with no ports at all
+		if matchingContainer == nil {
+			for ci := 0; ci < len(cns); ci++ {
```
thallgren:
```suggestion
			for ci := range cns {
```

pkg/install/discovery.go
```diff
@@ -170,6 +161,18 @@ Please specify the Service port you want to intercept by passing the --port=loca
 				}
 			}
 		}
+		// If no container matched, then use the first container with no ports at all
```
thallgren:
```suggestion
		// If no container matched, then use the first container with no ports at all. This
		// enables intercepts of containers that indeed do listen a port but lack a matching
		// port description in the manifest, which is what you get if you do:
		//     kubectl create deploy my-deploy --image my-image
		//     kubectl expose deploy my-deploy --port 80 --target-port 8080
```

pkg/install/discovery.go
thallgren, line 153: And then use `break containerLoop` instead of just `break` in the inner loop.

thallgren (changes requested):
Thank you very much for this fix @0x6a77. It looks good aside from my nits. After fixing them, please do the following:
1. Add a bugfix entry to `CHANGELOG.md`
2. `make format` (to get rid of the current lint error).
3. Commit with a "Sign-off" (will get rid of the DCO error).

pkg/install/discovery.go
thallgren: And then use `break containerLoop` instead of just `break` in the inner loop.

0x6a77:
that failing test doesn't seem like it's related. any ideas about what i should do to fix it?

thallgren (approved):
This looks great. Thanks. I agree that the test failure is unrelated, perhaps a temporary glitch at our download site. I've restarted that test.

josecv:
Hi @0x6a77, hope you are well!
Thank you for this PR. We'd like to invite you to our upcoming [meet the maintainers](https://www.getambassador.io/about-us/events/meet-the-maintainers-sessions/telepresence/081821/) event.
This will be an opportunity to chat with us and other members of the open source community, provide feedback, and ask questions on Telepresence. We will also be presenting features that were introduced to Telepresence in the past six weeks.
It will be taking place on August 18 at 10:00 AM Eastern Time.
Hope to see you there!

PR #1931, opened 2021-07-27
Test preview url in smoke-tests

donnyyung:

## Description

We should test the preview-url actually works in our smoke tests using the apikey work that Luke + Max landed, so this is doing that.

## Checklist

### DIFF
--- a/smoke-tests/run_smoke_test.sh
+++ b/smoke-tests/run_smoke_test.sh
@@ -97,24 +97,23 @@ get_preview_url() {
 # Puts workstation api key in a variable
 get_workstation_apikey() {
     local cache_file
-    uname=$(uname)
-    if [ "$uname" == "Darwin" ]; then
-        cache_file="$HOME/Library/Caches/telepresence/apikeys.json"
-    elif [ "$uname" == "Linux" ]; then
-        if [ -n "$XDG_CONFIG_HOME" ]; then
-            cache_file="$XDG_CONFIG_HOME/telepresence/apikeys.json"
-        else
-            cache_file="$HOME/.cache/telepresence/apikeys.json"
-        fi
-    fi
-    apikey=$(grep -o -E "\"telepresence:workstation\":\"[A-Za-z0-9=]+\"" "$cache_file" | awk -F: '{print $3}')
+    case "$(uname)" in
+	Darwin)
+		cache_file="$HOME/Library/Caches/telepresence/apikeys.json"
+		;;
+	Linux)
+		cache_file="${XDG_CONFIG_HOME:-$HOME/.cache}/telepresence/apikeys.json"
+		;;
+    *)
+        echo "OS is unknown by smoke-tests. Update get_workstation_apikey to include default config location for your OS"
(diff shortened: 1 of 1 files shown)


smoke-tests/run_smoke_test.sh
```diff
@@ -86,19 +86,37 @@ has_intercept_id() {
...
+    uname=$(uname)
+    if [ "$uname" == "Darwin" ]; then
+        cache_file="$HOME/Library/Caches/telepresence/apikeys.json"
+    elif [ "$uname" == "Linux" ]; then
+        if [ -n "$XDG_CONFIG_HOME" ]; then
+            cache_file="$XDG_CONFIG_HOME/telepresence/apikeys.json"
+        else
+            cache_file="$HOME/.cache/telepresence/apikeys.json"
+        fi
+    fi
```
LukeShu: A pair of style suggestions that you are free to ignore:
- I'd have written this as a switch/case
- use `:-` for fallbacks for unset variables: `${VARNAME:-DEFAULT_VALUE}`
```bash
case "$(uname)" in
	Darwin)
		cache_file="$HOME/Library/Caches/telepresence/apikeys.json"
		;;
	Linux)
		cache_file="${XDG_CONFIG_HOME:-$HOME/.cache}/telepresence/apikeys.json"
		;;
esac
```

smoke-tests/run_smoke_test.sh
```diff
@@ -86,19 +86,37 @@ has_intercept_id() {
...
+    apikey=$(grep -o -E "\"telepresence:workstation\":\"[A-Za-z0-9=]+\"" "$cache_file" | awk -F: '{print $3}')
+    if [[ -z $apikey ]]; then
+        echo "No apikey found"
+        exit 1
+    fi
+    # remove the leading and trailing "
+    apikey="${apikey%\"}"
+    apikey="${apikey#\"}"
+}
```
LukeShu, line 117: I'd be inclined to just use `jq` instead of parsing JSON with awk and Bash, but I guess we don't currently have a dependency on `jq`.

I'd have no objection to adding a dependency on `jq` for this.

```sh
jq -r '.[]|.["telepresence:workstation"]|strings' <"$cache_file"
```
aught to do it

LukeShu (approved):
Woohoo! I've got a couple of minor suggestions, but feel free to ignore them.

smoke-tests/run_smoke_test.sh
donnyyung: Ah okay I also wanted to use `jq` but didn't because it wasn't already a dependency, but I think it's worth it for cleaner code that is easier to read.

smoke-tests/run_smoke_test.sh
donnyyung: Ah yeah I should have looked up switch statements in bash, since this is an obvious candidate since the # of OS' is likely to grow (i.e. windows is imminent and we should probably make sure this can run on windows too.. will add that to my list of cooldown activities)

PR #1942, opened 2021-07-29
Windows installer

donnyyung:
## Description

v0 of our windows installer, still making improvements but putting it up so others can see + test.

Will likely replace this with an msi in the future, but this seems okay for a developer preview!

## Checklist

 - [x] I made sure to update `./CHANGELOG.md` - doesn't require a separate entry, "support for windows" should suffice.
 - [x] I made sure to either submit a docs PR, or tell Matt about the necessary documentation changes : https://github.com/telepresenceio/telepresence.io/pull/41


packaging/install-telepresence.ps1
```diff
@@ -0,0 +1,5 @@
+$current_directory = (pwd).path
+echo $current_directory
+Start-Process msiexec -Wait -verb runAs -Args "/i $current_directory\winfsp.msi /passive /qn /L*V winfsp-install.log"
+Start-Process msiexec -Wait -verb runAs -Args "/i $current_directory\sshfs-win.msi /passive /qn /L*V sshfs-win-install.log"
+[Environment]::SetEnvironmentVariable("Path", "C:\\;C:\\Program Files\\SSHFS-Win\\bin;$ENV:Path", "Machine")
```
donnyyung: I mostly just added this bc it was in the circle config, but wanted to check with Jose if he thinks it's necessary all the time.

Additionally, I figured I shouldn't add Telepresence to the PATH in the ps1 and just have them do it themselves, but if we think we should automatically add it to the path, I can add that here.

josecv: So I don't think we should add `C:\\` to the PATH here, that's there in Circle so that kubectl and other tools can be found. `sshfs-win` we definitely need, and the installer doesn't add it to the Path.

Other than that, I think if we're manipulating the Path we might as well add the Telepresence path, might make it a little easier for the user 🤷

packaging/windows-package.sh
```diff
@@ -0,0 +1,38 @@
...
+    echo "Could not create $ZIPDIR for windows package"
+    exit 1
+fi
+
+# Download sshfs-win.msi + winfsp.msi
+curl -L -o "${ZIPDIR}"/winfsp.msi https://github.com/billziss-gh/winfsp/releases/download/v1.9/winfsp-1.9.21096.msi
+curl -L -o "${ZIPDIR}"/sshfs-win.msi https://github.com/billziss-gh/sshfs-win/releases/download/v3.7.21011/sshfs-win-3.7.21011-x64.msi
+
+# Download wintun
+curl -L https://www.wintun.net/builds/wintun-0.12.zip | tar -vf - -Ox wintun/bin/amd64/wintun.dll > "${ZIPDIR}/wintun.dll"
```
josecv: Let's pull the versions into environment variables at the top of the file. e.g.

```shell
WINFSP_VERSION=1.9.21096
...
curl -L -o "${ZIPDIR}/winfsp.msi" "https://github.com/billziss-gh/winfsp/releases/download/v${WINFSP_VERSION%.*}/winfsp-${WINFSP_VERSION}.msi"
```

This'll make it a little easier to update the versions of dependencies

josecv (changes requested):
a few minor changes, but this looks great

packaging/install-telepresence.ps1
donnyyung: Yeah that sounds good, after sleeping on it, I also decided it made sense to include it. I'll add a comment here, and also add a comment in the docs that we are manipulating the PATH just so users know.

packaging/windows-package.sh
donnyyung: so obvious in hindsight but thank you!

packaging/windows-package.sh
```diff
@@ -23,11 +26,11 @@ if [[ ! "${ZIPDIR}" ]]; then
 fi
 
 # Download sshfs-win.msi + winfsp.msi
-curl -L -o "${ZIPDIR}"/winfsp.msi https://github.com/billziss-gh/winfsp/releases/download/v1.9/winfsp-1.9.21096.msi
-curl -L -o "${ZIPDIR}"/sshfs-win.msi https://github.com/billziss-gh/sshfs-win/releases/download/v3.7.21011/sshfs-win-3.7.21011-x64.msi
+curl -L -o "${ZIPDIR}"/winfsp.msi "https://github.com/billziss-gh/winfsp/releases/download/v1.9/winfsp-${WINFSP_VERSION}.msi"
```
josecv: You can add the 1.9 by doing `${WINFSP_VERSION%.*}` -- that will chomp off the last . and everything after it

josecv (approved):
minor comment around winfsp but otherwise looks good

packaging/windows-package.sh
donnyyung: Ah thank you! you learn something new in bash every day

PR #1946, opened 2021-07-30
Remove inital INFO log statment when config.yml file is missing

thallgren:
## Description

The PR also contains some clean-up and bug-fixes related to how commands are parsed and some improvements around how log files are reported when something goes wrong.

## Checklist

### DIFF
--- a/pkg/client/config.go
+++ b/pkg/client/config.go
@@ -537,7 +537,6 @@ func GetConfig(c context.Context) *Config {
 		var err error
 		config, err = loadConfig(c)
 		if err != nil {
-			dlog.Info(c, "No config found. Using default")
 			config = &defaultConfig
 		}
 	})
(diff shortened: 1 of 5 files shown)


josecv (approved):
lgtm

pkg/client/config.go
```diff
@@ -537,7 +537,6 @@ func GetConfig(c context.Context) *Config {
 		var err error
 		config, err = loadConfig(c)
 		if err != nil {
-			dlog.Info(c, "No config found. Using default")
```
donnyyung, line 540: Is there any way for us to log to the `connector.log` when no config is found and that we're using the default?  I feel like this might still be useful for the user bc say they get the path slightly wrong, without this message they might be confused why the config isn't working... and they might still be confused if we give them this error message but at least they know the config isn't being picked up.

thallgren: We do log "Here's the config" in the connect.log during an intercept, but other than that, no. There's no way. I think we should add commands to manage the config at some point (similar to kubectl config), but that's beyond the scope of this PR.

PR #1949, opened 2021-08-02
Uninstall agent even if the objects are in an inconsistent state

josecv:

## Description

## Checklist

### DIFF
--- a/pkg/install/util.go
+++ b/pkg/install/util.go
@@ -1,11 +1,13 @@
 package install
 
 import (
+	"errors"
 	"fmt"
 
-	corev1 "k8s.io/api/core/v1"
-
 	"github.com/datawire/ambassador/pkg/kates"
+
+	"github.com/hashicorp/go-multierror"
+	corev1 "k8s.io/api/core/v1"
 )
 
 func GetPodTemplateFromObject(obj kates.Object) (*kates.PodTemplateSpec, error) {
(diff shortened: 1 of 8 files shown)


pkg/install/util.go
```diff
@@ -40,3 +42,39 @@ func ObjErrorf(obj kates.Object, format string, args ...interface{}) error {
...
+
+func (e *AlreadyUndone) Unwrap() error {
+	return e.err
+}
+
+func NewAlreadyUndone(err error, msg string) error {
+	return &AlreadyUndone{err, msg}
+}
+
+// IsAlreadyUndone returns whether the given error -- possibly a multierror -- indicates that all actions have been undone.
```
LukeShu, line 64: My high school journalism teacher would have yelled at you for putting spaces round the `--` em-dashes :P

pkg/install/util.go
```diff
@@ -40,3 +42,39 @@ func ObjErrorf(obj kates.Object, format string, args ...interface{}) error {
...
+	if !errors.As(err, &multi) {
+		return false
+	}
+	for _, err := range multi.WrappedErrors() {
+		if !errors.As(err, &undone) {
+			return false
+		}
+	}
+	return true
+}
```
LukeShu, line 80: Looking at `go-multierror`, I'm like 90% sure that this is unnecessary and that you can just use `errors.As(err, &undone)`

LukeShu (approved):
A nitpick with `IsAlreadyUndone` that you can choose to ignore or tell me I'm wrong.

Resolve the merge conflicts, and it looks good!

pkg/install/util.go
josecv: I don't think so -- this is slightly different. My understanding of `go-multierror` is that `errors.As` will return "is _any_ error in this multierror an AlreadyUndone", whereas here we want _all_ errors in the multierror to eventually resolve back to an `AlreadyUndone`

PR #1950, opened 2021-08-02
Support sending messages from Cloud

donnyyung:
## Description

This PR enables us to have the cloud send messages to users when they perform certain commands. This will enable us to send messages to users to help enhance their experience using Telepresence.

## Checklist

 - [ ] I made sure to either submit a docs PR, or tell Matt about the necessary documentation changes. - still have to do
 - [x] My change is adequately tested. - I want to write more tests but this is a start

### DIFF
--- a/pkg/client/cli/cloud_messages.go
+++ b/pkg/client/cli/cloud_messages.go
@@ -116,6 +116,13 @@ func raiseCloudMessage(cmd *cobra.Command, _ []string) error {
 			return nil
 		}
 	}
+
+	// If the user has specified they are in an air-gapped cluster,
+	// we shouldn't try to get messages
+	if client.GetConfig(cmd.Context()).Cloud.SkipLogin {
+		return nil
+	}
+
 	// The command is the first word of cmd.Use
 	cmdUsed := strings.Split(cmd.Use, " ")[0]
 
--- a/pkg/client/cli/cloud_messages.go
+++ b/pkg/client/cli/cloud_messages.go
@@ -1,0 +1,106 @@
+package cli
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"strings"
+	"time"
+
+	"github.com/datawire/dlib/dtime"
+
+	"github.com/spf13/cobra"
+
+	"github.com/telepresenceio/telepresence/rpc/v2/systema"
+	"github.com/telepresenceio/telepresence/v2/pkg/client/cache"
+)
+
+const messagesCacheFilename = "cloud-messages.json"
+
+type cloudMessageCache struct {
+	NextCheck         time.Time       `json:"next_check"`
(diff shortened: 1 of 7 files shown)


pkg/client/cli/cloud_messages.go
```diff
@@ -0,0 +1,157 @@
...
+		_ = cache.SaveToUserCache(ctx, cmc, messagesCacheFilename)
+	}
+	return cmc, nil
+}
+
+// getCloudMessages communicates with Ambassador Cloud and stores those messages
+// in the cloudMessageCache.
+// For now this function will just return a fake result from the cloud
+// this will be replaced with the code to talk to Ambassador Cloud
+// via GetUnauthenticatedCommandMessages in a later commit
```
josecv: this comment seems like it's out of date

pkg/client/cli/cloud_messages.go
```diff
@@ -0,0 +1,157 @@
...
+			// We try again in an hour since we encountered an error
+			cmc.NextCheck = dtime.Now().Add(1 * time.Hour)
+		} else {
+			cmc.updateCacheMessages(ctx, resp)
+		}
+	}
+
+	// Get the message from the cache that should be delivered to the user
+	msg := cmc.getMessageFromCache(ctx, cmdUsed)
+	fmt.Fprintf(cmd.OutOrStdout(), "%s\n", msg)
```
josecv: nit: maybe only print `if msg != ""`? otherwise this'll print a newline every time

pkg/client/cli/cloud_messages.go
```diff
@@ -0,0 +1,157 @@
...
+	"github.com/telepresenceio/telepresence/v2/pkg/client/cache"
+	"github.com/telepresenceio/telepresence/v2/pkg/client/cli/cliutil"
+)
+
+const messagesCacheFilename = "cloud-messages.json"
+
+type cloudMessageCache struct {
+	NextCheck         time.Time       `json:"next_check"`
+	Intercept         string          `json:"intercept"`
+	MessagesDelivered map[string]bool `json:"messagess_delivered"`
```
josecv: hmm it seems like this could easily be a `map[string]struct{}` and instead of setting the value to false the key could just be deleted. I guess this way lets us distinguish between a message that's never been delivered and one that has changed, but it doesn't look like that's something we need right now

josecv (changes requested):

pkg/client/cli/cloud_messages.go
donnyyung: 🤦

pkg/client/cli/cloud_messages.go
donnyyung: Good point, I made this more complicated than it needed to be!

josecv (approved):

PR #1951, opened 2021-08-02
Ensure grpc max size is propagated to daemon server 

josecv:

## Description

## Checklist

### DIFF
--- a/pkg/client/cli/telepresence_test.go
+++ b/pkg/client/cli/telepresence_test.go
@@ -2,9 +2,11 @@ package cli_test
 
 import (
 	"bufio"
+	"bytes"
 	"context"
 	"encoding/base64"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"log"
 	"net"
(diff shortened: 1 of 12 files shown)


pkg/client/cli/telepresence_test.go
```diff
@@ -428,21 +430,33 @@ func (cs *connectedSuite) SetupSuite() {
...
 	// traffic-manager is installed
-	_, stderr := telepresence(cs.T(), "connect")
+	configDir := cs.T().TempDir()
+	registry := dtest.DockerRegistry(c)
+	configYml := fmt.Sprintf(`
+logLevels:
+  rootDaemon: debug
+images:
+  registry: %s
+timeouts:
```
donnyyung, line 442: I don't think a merge blocker, but if we need these defaults for a bunch of tests, I wonder if our defaults are wrong and we should bump them to these values. imo, the tests should run with as many default settings as possible, only changing them if we are specifically testing that config option.

donnyyung:
all-in-all this seems sane, good work! Will approve once we have passing tests.

I think we should chat sometime next week and get on the same page about the ideal way of doing this, I think maybe a configmap that the helm chart installs, but would also be easy enough for users to add directly if they use `telepresence connect` to install components in the cluster.

donnyyung (approved):
LGTM now that tests are passing!

PR #1952, opened 2021-08-03
On demand regenerate certificate used for mutating webhook

witdur:
## Description

Now helm chart has new value `agentInjector.certificate.regenerate` to control whether certificate used for mutating webhook should be regenerate or not, before every `helm upgrade` regenerate it and mutating webhook stops working (bad certificate) when `traffic-manager` was not restarted.

## Checklist

 - [ ] I made sure to either submit a docs PR, or tell Matt about the necessary documentation changes - no need
 - [ ] I updated `DEVELOPING.md` with any any special dev tricks I had to use to work on this code efficiently - no need


charts/telepresence/templates/agentInjectorWebhook.yaml
```diff
@@ -14,7 +15,11 @@ webhooks:
   {{- toYaml . | nindent 2 }}
 {{- end }}
   clientConfig:
-    caBundle: {{ $ca.Cert | b64enc }}
+{{- if and ($secretData) (eq .Values.agentInjector.certificate.regenerate false) }}
```
josecv: just out of curiosity (not something that you need to change) is there any reason not to use `not .Values.agentInjector.certificate.regenerate` instead of `eq .Values.agentInjector.certificate.regenerate false`?

josecv (changes requested):
thank you for submitting this pull request! One minor request: could you please add the new flag to the chart's readme under `charts/telepresence/README.md`?

josecv (approved):

PR #1975, opened 2021-08-10
Work around a race when writing large files to the cluster

josecv:

## Description

## Checklist

### DIFF
--- a/pkg/tun/tcp/handler.go
+++ b/pkg/tun/tcp/handler.go
@@ -313,14 +313,24 @@ func (h *handler) sendSyn(ctx context.Context, ackNumber uint32, ack bool) {
 // writeToTunLoop sends the packages read from the fromMgr channel to the TUN device
 func (h *handler) writeToTunLoop(ctx context.Context) {
 	defer close(h.readyToFin)
-	ticker := time.NewTicker(100 * time.Millisecond)
+	// We use a timer that we reset on each iteration instead of a ticker to prevent drift between
+	// the select loop and the ticker's interval. Otherwise, suppose we spend 499ms processing
+	// a message from fromMgr, a ticker would only give us a 1ms wait before checking h.isClosing
+	timer := time.NewTimer(500 * time.Millisecond)
+	defer timer.Stop()
 	for {
 		var cm connpool.Message
+		timer.Stop()
+		select {
+		case <-timer.C:
+		default:
+		}
+		timer.Reset(500 * time.Millisecond)
 		select {
 		case <-ctx.Done():
 			return
 		case cm = <-h.fromMgr:
-		case <-ticker.C:
+		case <-timer.C:
 			if atomic.LoadInt32(&h.isClosing) > 0 {
 				return
 			}
--- a/pkg/client/cli/telepresence_test.go
+++ b/pkg/client/cli/telepresence_test.go
@@ -448,7 +448,11 @@ grpc:
 	configYml = strings.TrimSpace(configYml)
 	c, err := client.SetConfig(c, configDir, configYml)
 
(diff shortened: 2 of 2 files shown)


pkg/tun/tcp/handler.go
```diff
@@ -313,14 +313,17 @@ func (h *handler) sendSyn(ctx context.Context, ackNumber uint32, ack bool) {
 // writeToTunLoop sends the packages read from the fromMgr channel to the TUN device
 func (h *handler) writeToTunLoop(ctx context.Context) {
 	defer close(h.readyToFin)
-	ticker := time.NewTicker(100 * time.Millisecond)
 	for {
 		var cm connpool.Message
 		select {
+		// FIXME(josecv): Prioritize draining the fromMgr queue over terminating; there's a subtle race
```
thallgren: If this actually makes a difference, then I think that is because the timer is increased from 100 to 500 ms. No prioritization is made because Go will shuffle the cases in a channel select on each call to ensure that order is insignificant.

I think we need a proper drain timeout. E.g. have a separate goroutine that only deals with the `h.fromMgr` channel and another that deals with `context.Done()` The latter will then close the `h.fromMgr` after some configurable delay. Perhaps that channel could also be closed at the writer end when the peer sends a close control message.

thallgren (changes requested):

pkg/tun/tcp/handler.go
josecv: > No prioritization is made because Go will shuffle the cases in a channel select on each call to ensure that order is insignificant.

oh wow TIL. It must be the timer change then that's fixed this, because I have not seen the tests fail on this branch.

> I think we need a proper drain timeout. E.g. have a separate goroutine that only deals with the h.fromMgr channel and another that deals with context.Done() The latter will then close the h.fromMgr after some configurable delay. Perhaps that channel could also be closed at the writer end when the peer sends a close control message.

Yeah I think that makes sense, my only worry about closing the channel is that we need to make sure the other end's context is cancelled too, or we'll get a panic when it tries to write to the channel. Right now I don't think that's the case, as I've seen the other end hang on a `fromMgr <- msg` operation after this end's context is finished.

pkg/tun/tcp/handler.go
```diff
@@ -313,14 +313,13 @@ func (h *handler) sendSyn(ctx context.Context, ackNumber uint32, ack bool) {
...
 	defer close(h.readyToFin)
-	ticker := time.NewTicker(100 * time.Millisecond)
 	for {
 		var cm connpool.Message
 		select {
 		case <-ctx.Done():
 			return
 		case cm = <-h.fromMgr:
-		case <-ticker.C:
+		case <-time.After(500 * time.Millisecond):
```
thallgren: Any special reason for not using a ticker here? Using a `time.After` instead will put more strain on the GC because it creates a new `time.Timer` on each call to `select`. Furthermore, the timer won't be GC:ed until it expires which means that there might be a fair amount of them in memory on a busy handler.

thallgren (changes requested):

pkg/tun/tcp/handler.go
josecv: So I was worried about drift between the ticker and the loop. At the end of the day, what we want is to say "if after X milliseconds there's nothing in the channel, see about stopping the loop". But the ticker is ticking whether or not we're in the select. So if, for example, we receive a message and spend 99ms processing it, then when the loop comes around we'll only wait 1ms before checking if we're closing.

I guess I could use a ticker and just use `Reset` right before the select, actually, I'll change it to do that

thallgren (approved):

PR #1978, opened 2021-08-10
Improve process around running tests for forked PRs

donnyyung:
We have secrets that are required for running our tests in circle, since
forked PRs don't have those secrets, we shouldn't run those tests on
forked PRs.

## Description

This PR is still a WIP but working on a better process so only the correct tests run on forked PRs and that there is a mechanism for maintainers to run the full set of tests on forked PRs.

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -374,10 +374,18 @@ workflows:
           context: telepresence2-release
           requires:
             - build-dev-image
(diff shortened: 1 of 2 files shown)


donnyyung:
We should ultimately fix this by moving to github actions, since they have a better secrets workflow than circle.

PR #1982, opened 2021-08-11
Log errors when port-forward fails

donnyyung:
## Description

Ideally we should error out completely if a user doesn't have permissions to port-forward, but for now adding this log line will be helpful so people see something like:

```
2021/08/11 17:41:31.8792 error   connector/background-manager : Error dialing portForward: pods "traffic-manager-7fd996577c-pb9bp" is forbidden: User "system:serviceaccount:default:telepresence-test-developer" cannot create resource "pods/portforward" in API group "" in the namespace "ambassador" (from pkg/dnet/kpfconn.go:209)
```
Instead of logs showing that we are retrying the port-forward over-and-over again without specifying the error we are running into.

## Checklist

### DIFF
--- a/pkg/client/connector/userd_trafficmgr/traffic_manager.go
+++ b/pkg/client/connector/userd_trafficmgr/traffic_manager.go
@@ -159,7 +159,8 @@ func (tm *trafficManager) Run(c context.Context) error {
 	opts := []grpc.DialOption{grpc.WithContextDialer(grpcDialer),
 		grpc.WithInsecure(),
 		grpc.WithNoProxy(),
-		grpc.WithBlock()}
+		grpc.WithBlock(),
+		grpc.WithReturnConnectionError()}
 
 	conn, err = grpc.DialContext(tc, grpcAddr, opts...)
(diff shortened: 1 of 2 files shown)


donnyyung:
I added a commit that includes the last connect error when we timeout, so now it'll look like this:
```
Launching Telepresence Root Daemon
Launching Telepresence User Daemon
Telepresence Root Daemon quitting... done
telepresence: error: connector.Connect: the traffic manager gRPC API timed out.  The current timeout 15s can be configured as "timeouts.trafficManagerAPI" in "/Users/donaldyung/Library/Application Support/telepresence/config.yml": connection error: desc = "transport: Error while dialing pods \"traffic-manager-5f8f54798f-c4lmb\" is forbidden: User \"system:serviceaccount:default:telepresence-test-developer\" cannot create resource \"pods/portforward\" in API group \"\" in the namespace \"ambassador\""
``` 
I'm still looking for a way to fail fast when this happens, but this makes it clear to the user at least what is wrong.

josecv (approved):

PR #1985, opened 2021-08-12
Do a helm install on telepresence connect

josecv:
## Description

## Checklist


charts/telepresence/templates/deployment.yaml
```diff
@@ -41,14 +41,23 @@ spec:
             value: "443"
           - name: CLUSTER_ID
             value: {{ .Values.clusterID }}
+          {{- if (and .Values.agentInjector.create .Values.agentInjector.agentImage.registry) }}
+          - name: TELEPRESENCE_REGISTRY
+            value: {{ .Values.agentInjector.agentImage.registry }}
+          {{- else }}
           - name: TELEPRESENCE_REGISTRY
```
donnyyung, line 48: I think we can / should remove this value so that we are only setting the environment variable once and that it's coming from the `.Values.agentInjector.agentImage.registry` value only.

charts/telepresence/templates/deployment.yaml
```diff
@@ -41,14 +41,23 @@ spec:
             value: "443"
           - name: CLUSTER_ID
             value: {{ .Values.clusterID }}
+          {{- if (and .Values.agentInjector.create .Values.agentInjector.agentImage.registry) }}
```
donnyyung: kind of a nit: since we have a default for `agentInjector.agentImage.registry`, I'd rather just always use that value here instead of doing an if/else and using `image.registry` in the `else` case... does that seem fair?

pkg/install/helm/install.go
```diff
@@ -0,0 +1,147 @@
...
+	install.Atomic = true
+	install.CreateNamespace = true
+	_, err := install.Run(chrt, getValues(ctx, clusterID))
+	return err
+}
+
+func upgradeExisting(ctx context.Context, chrt *chart.Chart, helmConfig *action.Configuration, namespace, clusterID string) error {
+	dlog.Info(ctx, "Existing Traffic Manager found, upgrading...")
+	upgrade := action.NewUpgrade(helmConfig)
+	upgrade.Timeout = 2 * time.Minute
```
donnyyung: Nit: Perhaps worth adding a key named something like `helmTimeout` to the timeouts block in config.yml and making this value configurable.. could then use this value for the timeout in `installNew` and `DeleteTrafficManager`.

donnyyung (approved):
Looks good! a few small nits, that you can take action on or not.

charts/telepresence/templates/deployment.yaml
```diff
@@ -39,14 +39,25 @@ spec:
             value: app.getambassador.io
           - name: SYSTEMA_PORT
             value: "443"
+          - name: CLUSTER_ID
```
donnyyung: so in this recent PR, I removed the CLUSTER_ID as an env variable and from the helm chart: https://github.com/telepresenceio/telepresence/pull/1986, so you should be able to remove everything in the helm chart that's related to CLUSTER_ID since that info is now gotten from the traffic-manager itself instead of passed in by the user or the cli on `connect`

donnyyung (changes requested):
see my comment about CLUSTER_ID

thallgren (approved):
Besides what Donny pointed out, this looks OK to me.

donnyyung (approved):
LGTM assuming tests pass, thanks for taking all the feedback :) and super excited for this change to get merged in!

PR #1986, opened 2021-08-13
Move get-cluster-id into traffic-manager

donnyyung:
## Description

The cluster ID was added as an environment variable to the traffic-manager deployment in the past, but since the traffic-manager now communicates with the kubeapi to get various things, it makes sense for the traffic-manager to do that itself, so that's what this PR does.

## Checklist

 - [x] I made sure to either submit a docs PR, or tell Matt about the necessary documentation changes. -- this was only mentioned in the helm chart, so doesn't require a docs change aside from releaseNote: https://github.com/telepresenceio/telepresence.io/pull/53

### DIFF
--- a/pkg/client/cli/telepresence_test.go
+++ b/pkg/client/cli/telepresence_test.go
@@ -1412,15 +1412,10 @@ func (hs *helmSuite) TestZ_Uninstall() {
 }
 
 func (hs *helmSuite) helmInstall(ctx context.Context, managerNamespace string, appNamespaces ...string) error {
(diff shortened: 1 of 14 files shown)


pkg/client/cli/telepresence_test.go
```diff
@@ -439,12 +439,13 @@ logLevels:
   rootDaemon: debug
 images:
   registry: %s
+  webhookRegistry: %s
```
donnyyung: I'm confused about how this test worked without this being set before.

thallgren (approved):
LGTM

PR #1991, opened 2021-08-16
Read namespace from kubeconfig on each call

thallgren:
## Description

Prior to this PR, the user daemon would initialize its notion of
the default namespace during `telepresence connect` and then leave it
unchanged until a `telepresence quit`. This PR ensures that the
namespace is initialized on each call.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - Feature: Helm chart has now feature to on demand regenerate certificate used for mutating webhook by setting value.
 `agentInjector.certificate.regenerate`
 
+- Bugfix: Telepresence will initialize the default namespace from the kubeconfig on each call instead of just doing it when connecting.
 
 ### 2.4.0 (August 4, 2021)
 
--- a/pkg/client/connector/command.go
+++ b/pkg/client/connector/command.go
@@ -89,7 +89,7 @@ func (s *service) connect(c context.Context, cr *rpc.ConnectRequest, dryRun bool
 	s.connectMu.Lock()
 	defer s.connectMu.Unlock()
 
-	Config, err := userd_k8s.NewConfig(cr.KubeFlags, s.env)
+	config, err := userd_k8s.NewConfig(cr.KubeFlags, s.env)
 	if err != nil && !dryRun {
 		return &rpc.ConnectInfo{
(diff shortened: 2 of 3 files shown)


donnyyung:
I think this is in general a good code change, the only issue I think it makes easier for a user to get into is say we have two namespaces: `test1` and `test2`, with an `api` deployment in each namespace. 

The user's kubeconfig is pointing at the `test1` namespace so they do a `telepresence intercept api --port 8000` and the `api` deployment in the `test1` namespace is intercepted.

Next the user changes their context to point at `test2`.  They want to then intercept the `api` deployment in that namespace so they do: 
`telepresence intercept api --port 8001` but they get the error message:
```
Intercept: Intercept with name "api" already exists
```
If they do a telepresence list, they won't see any active intercepts since the active intercept is in a different namespace.

So this might be a fully contrived example, but seems like something we are potentially more likely to bump into with this feature (which I think is good / better than requiring a user to log in + log out to change the default namespace). And I don't really think this needs to _block_ this PR, just wanted to raise it / get your opinion on it.

While I'm here... an easy fix might be a "telepresence list --all" flag or something like that, which gets all intercepts the traffic-manager knows about, so when a user does bump into the issue above they can be like, "oh I already have an intercept named `api` in the `test1` namespace, I need to give the one in `test2` a different name. 

a more complicated fix might be automatically appending the namespace to the intercept name, but then we run into the issue of the "name" not matching what the user specified, but maybe we could have the intercept ID be separate from the name in that scenario... I'd have to think on it more.

thallgren:
I think the decision we made in the past, to append the namespace to the intercept name in case a namespace was explicitly given, was a bit hasty, but since we did it that way, we now have two options:

1. Live with its consequences (the ambiguity that you mention).
2. Fix so that we always append the namespace to enforce unique names. That change should probably also be accompanied with a change to have `telepresence leave` accept a name without a namespace unless it's ambiguous.

I'm ambivalent about this since I think the ambiguity will be very uncommon and when people run into it after switching namespace while being connected, they will understand (and accept) why it happens.

donnyyung (approved):

donnyyung:
Yeah fair enough. Just wanted to get your thoughts but I think this is a good change, so I've approved it :)

I don't think we need to go with option 2 unless it becomes a pain point amongst users.

PR #1994, opened 2021-08-17
Increase the time to keep idle connections alive.

thallgren:
## Description

This PR increases the time to keep idle connections alive from
60 seconds to 7,200 seconds which is the same as the Linux default for
`tcp_keepalive_time`.

## Checklist


josecv (approved):
lgtm, eventually we should consider how to make this configurable though

PR #1999, opened 2021-08-19
Don't report metriton failures as errors

thallgren:
## Description

Metriton reporting failures must be logged as an info instead of error
because thanks to #1910, we now show error counts from the logs in order
to help the user locate problems and such failures aren't problems that
the user needs to be concerned about.

## Checklist


donnyyung (approved):
LGTM, thanks for cleaning this up!

PR #2000, opened 2021-08-19
Retry dial when socket exists after ungraceful termination

thallgren:
## Description

The daemons will sometimes leave a socket behind in the filesystem when
it is terminated ungracefully (kill -9 or similar). Telepresence will
think that the process is running and try to connect to that socket if
it exists and it will fail because there's no process alive that listens
to it.

Prior to this PR, telepresence would inform the user about the
failed attempt and exit but now, it will instead make an attempt to
remove the socket and if that succeeds, dial again.

## Checklist

### DIFF
--- a/pkg/client/sockets_unix.go
+++ b/pkg/client/sockets_unix.go
@@ -28,13 +28,26 @@ const (
 func dialSocket(ctx context.Context, socketName string, opts ...grpc.DialOption) (*grpc.ClientConn, error) {
 	ctx, cancel := context.WithTimeout(ctx, 5*time.Second) // FIXME(lukeshu): Make this configurable
 	defer cancel()
-	conn, err := grpc.DialContext(ctx, "unix:"+socketName, append([]grpc.DialOption{
-		grpc.WithInsecure(),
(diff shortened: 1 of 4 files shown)


pkg/client/sockets_unix.go
```diff
@@ -28,13 +28,26 @@ const (
...
+		}, opts...)...)
+		if err == nil {
+			return conn, nil
+		}
+
+		if firstTry && errors.Is(err, syscall.ECONNREFUSED) {
+			// Socket exists but doesn't accept connections. This usually means that the process
+			// terminated ungracefully. To remedy this, we make an attempt to remove the socket
+			// and dial again.
+			if rmErr := os.Remove(socketName); rmErr == nil {
```
josecv: It might be good to accumulate `rmErr` into `err` so that when we print out the failure we can say something about why we failed to remove it. Maybe something like:

`err = fmt.Errorf("%w (socket rm failed with %w)", err, rmErr)`

josecv (approved):
lgtm, one suggestion that i think is quite subjective so i'll leave it to your judgement 🙂

PR #2002, opened 2021-08-20
Ensure that getInstallIDFromFilesystem doesn't error on windows

thallgren:
## Description

Changes the `getInstallIDFromFilesystem` function so that it just omits
legacy ids that cannot be read rather than returning an error.

## Checklist

 - [x] I made sure to update `./CHANGELOG.md`. # No update needed since this bugfix doesn't affect users


donnyyung (approved):
LGTM! Nice work on this!

PR #2003, opened 2021-08-20
Revert "Don't report metriton failures as errors"

donnyyung:
I sadly think we can't do this because then the errors are exposed on stdout when users run intercepts and other commands:
```
(⎈ |default:default)➜  ~/go/telepresenceio git:(release/v2) ✗ telepresence intercept echo-easy --port 8000
Using Deployment echo-easy
intercepted
    Intercept name    : echo-easy
    State             : ACTIVE
    Destination       : 127.0.0.1:8000
    Volume Mount Point: /var/folders/cp/2r22shfd50d9ymgrw14fd23r0000gp/T/telfs-210508298
INFO[0005] scout report "intercept_success" failed: Post "https://metriton.datawire.io/scout": dial tcp 127.0.0.1:443: connect: connection refused  THREAD=/main
``` 
so I think it makes more sense to have them as errors and handle them that way, since this unfortunately makes the errors more apparent :/.

## Description

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,8 +26,6 @@
 - Bugfix: Telepresence will now remove a socket that is the result of an ungraceful termination and retry instead of printing
   an error saying "this usually means that the process has terminated ungracefully"
 
-- Change: Failure to report metrics is logged using loglevel info rather than error.
-
 ### 2.4.0 (August 4, 2021)
 
 - Feature: There is now a native Windows client for Telepresence.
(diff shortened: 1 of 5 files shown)


thallgren:
I disagree. Metriton report failures may occur from time to time (because our server is unreachable) but they must left the user completely unaffected. If that's not the case, then we're doing something wrong. If we let metriton errors have a meaning to the end-user (which is what we do when they surface), then that's just adding confusion.

If I understand it correctly, this metriton failure is issued by the CLI process. We should prevent that from happening, not revert my change.

donnyyung:
100% agreed and discussed offline. I'll close this PR.

PR #2004, opened 2021-08-21
Restore v1 behavior of mounting all directories from `/var/run/secrets`.

### LINKED ISSUE #1966: how to mount oidc tokens when using irsa in eks
any pods that do [this](https://aws.amazon.com/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/) cannot get the aws automatic rotating oidc web identity token mounted to the local machine. this issue serves two purposes:

1. share a workaround (that will show up in the google search index for anyone else up against this)
```
telepresence connect
telepresence intercept --namespace my_namespace --env-json telepresence.json my_workload
telepresence leave my_workload 
```

(now we know the agent is running)

```
kubectl exec -it -n my_namespace my_workload -c traffic-agent -- ln -s /var/run/secrets/eks.amazonaws.com /tel_app_mounts/var/run/secrets/
telepresence intercept --namespace my_namespace --env-json telepresence.json  my_workload
```

now the rotating token is locally mounted and updated at `$TELEPRESENCE_ROOT/var/run/secrets/eks.amazonaws.com/serviceaccount/token`. any project that uses boto3 or any aws sdk will "just work" because `AWS_ROLE_ARN` and `AWS_WEB_IDENTITY_TOKEN_FILE` will be in the env vars file (and point to the right place if you do the magic with local env vars)

2. propose a fix

the simplest fix is in `container.go` to treat `/var/run/secrets/eks.amazonaws.com` the same way as `/var/run/secrets/kubernetes.io`. a more advanced solution might be to do this for anything in `/var/run/secrets`, or maybe even just projected volume mounts in `/var/run/secrets`. eks seems to be the only service that uses a web token file for irsa, and eks seems to be the dominant player so maybe the simplest solution gets 90% of the benefit for 5% of the effort.

i probably can take on item 2, but it seems like it's enough of a change that it warrants discussion first

0x6a77:
resolves #1966

Signed-off-by: Jeff Williams <jeff.williams@gadgetworks.com>

## Description

This change supports other operators that may provide rotating tokens/secrets such as eks/irsa. It also restores v1 behavior.

## Checklist

### DIFF
--- a/cmd/traffic/cmd/agent/agent.go
+++ b/cmd/traffic/cmd/agent/agent.go
@@ -73,37 +73,61 @@ func AppEnvironment() map[string]string {
 	return fullEnv
 }
 
-// svcAccPath is the path where the ServiceAccount Admission Controller automatically provides its secrets.
-const svcAccPath = "/var/run/secrets/kubernetes.io"
 const tpMountsEnv = "TELEPRESENCE_MOUNTS"
 
-func (cfg *Config) HasMounts(ctx context.Context, env map[string]string) (bool, error) {
+func (cfg *Config) HasMounts(ctx context.Context, env map[string]string) bool {
 	tpMounts := env[tpMountsEnv]
-	stat, err := os.Stat(svcAccPath)
-	if err != nil || !stat.IsDir() {
-		return tpMounts != "", nil
+	retVal := false
+	split := strings.Split(tpMounts, ":")
+	for _, itemPath := range split {
+		dlog.Debugf(ctx, "agent mount path: %s", itemPath)
+		retVal = true
 	}
+	return retVal
+}
 
-	// This must be included in the shared mounts unless it's already provided
-	svcAccLink := filepath.Join(cfg.AppMounts, svcAccPath)
-	if stat, err = os.Stat(svcAccLink); err == nil && stat.IsDir() {
-		return true, nil
-	}
+// Add any token-rotating system secrets directories if they exist
+// e.g. /var/run/secrets/kubernetes.io or /var/run/secrets/eks.amazonaws.com
+func (cfg *Config) addSecretsMounts(ctx context.Context, env map[string]string) error {
+	var retErr error = nil
+	tpMounts := env[tpMountsEnv]
 
-	// Add a link to the kubernetes.io directory under {{.AppMounts}}/var/run/secrets
-	if err = os.MkdirAll(filepath.Dir(svcAccLink), 0700); err != nil {
-		return false, err
-	}
-	if err = os.Symlink(svcAccPath, svcAccLink); err != nil {
-		return false, err
-	}
-	if tpMounts == "" {
-		tpMounts = svcAccPath
-	} else {
-		tpMounts += ":" + svcAccPath
+	// This will attempt to handle all the secrets dirs, but will return the first error we encountered.
+	if secretsDir, retErr := os.Open("/var/run/secrets"); retErr == nil {
+		if fileInfo, retErr := secretsDir.Readdir(-1); retErr == nil {
+			secretsDir.Close()
+			for _, file := range fileInfo {
+				// Files that rotate, such as tokens found in /var/run/secrets, we need to symlink to the original
+				// (Stops at the first error found.)
+				if retErr == nil && file.IsDir() {
+					dirPath := "/var/run/secrets/" + file.Name()
+					dlog.Debugf(ctx, "checking agent secrets mount path: %s", dirPath)
+					if stat, retErr := os.Stat(dirPath); retErr == nil && stat.IsDir() {
+						dlog.Debugf(ctx, "discovered directory: %s", dirPath)
+						// Create this in shared mounts if not already provided
+						appMountsPath := filepath.Join(cfg.AppMounts, dirPath)
+						if _, retErr = os.Stat(appMountsPath); retErr != nil {
+							dlog.Debugf(ctx, "create appmounts directory: %s", appMountsPath)
+							// Add a link to the kubernetes.io directory under {{.AppMounts}}/var/run/secrets
+							if retErr = os.MkdirAll(filepath.Dir(appMountsPath), 0700); retErr == nil {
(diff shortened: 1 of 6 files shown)


cmd/traffic/cmd/agent/agent.go
```diff
@@ -73,37 +73,61 @@ func AppEnvironment() map[string]string {
...
-	stat, err := os.Stat(svcAccPath)
-	if err != nil || !stat.IsDir() {
-		return tpMounts != "", nil
+	retVal := false
+	split := strings.Split(tpMounts, ":")
+	for _, itemPath := range split {
+		dlog.Debugf(ctx, "agent mount path: %s", itemPath)
+		retVal = true
 	}
+	return retVal
```
thallgren: Splitting and looping here just go get debug logging seems a bit excessive. How about:
```go
	if tpMounts != "" {
		dlog.Debugf(ctx, "agent mount paths: %s", tpMounts)
		return true
	}
	return false
```

cmd/traffic/cmd/agent/agent.go
```diff
@@ -73,37 +73,61 @@ func AppEnvironment() map[string]string {
...
-	}
-	if err = os.Symlink(svcAccPath, svcAccLink); err != nil {
-		return false, err
-	}
-	if tpMounts == "" {
-		tpMounts = svcAccPath
-	} else {
-		tpMounts += ":" + svcAccPath
+	// This will attempt to handle all the secrets dirs, but will return the first error we encountered.
+	if secretsDir, retErr := os.Open("/var/run/secrets"); retErr == nil {
```
thallgren: Why not use something like:
```go
		fileInfos, err := ioutil.ReadDir("/var/run/secrets")
		if err == nil {
			return err
		}
		for _, file := range fileInfos {
```

cmd/traffic/cmd/agent/agent.go
```diff
@@ -73,37 +73,61 @@ func AppEnvironment() map[string]string {
...
+		if fileInfo, retErr := secretsDir.Readdir(-1); retErr == nil {
+			secretsDir.Close()
+			for _, file := range fileInfo {
+				// Files that rotate, such as tokens found in /var/run/secrets, we need to symlink to the original
+				// (Stops at the first error found.)
+				if retErr == nil && file.IsDir() {
```
thallgren: Please avoid deep nested if statements by using `continue` and `return` E.g.
```
if !file.isDir() {
	continue
}

File: cmd/traffic/cmd/agent/agent.go [hunk 4]
```diff
@@ -73,37 +73,61 @@ func AppEnvironment() map[string]string {
...
+					dirPath := "/var/run/secrets/" + file.Name()
+					dlog.Debugf(ctx, "checking agent secrets mount path: %s", dirPath)
+					if stat, retErr := os.Stat(dirPath); retErr == nil && stat.IsDir() {
```
thallgren: It's better to handle the unhappy path directly and provide some context to the error, e.g.
```
state, err := os.Stat(dirPath)
if err != nil {
	return fmt.Errorf("unable to stat %s: %w", err)
}
```

cmd/traffic/cmd/agent/agent.go
```diff
@@ -73,37 +73,61 @@ func AppEnvironment() map[string]string {
...
-		tpMounts += ":" + svcAccPath
+	// This will attempt to handle all the secrets dirs, but will return the first error we encountered.
+	if secretsDir, retErr := os.Open("/var/run/secrets"); retErr == nil {
+		if fileInfo, retErr := secretsDir.Readdir(-1); retErr == nil {
+			secretsDir.Close()
+			for _, file := range fileInfo {
+				// Files that rotate, such as tokens found in /var/run/secrets, we need to symlink to the original
+				// (Stops at the first error found.)
+				if retErr == nil && file.IsDir() {
+					dirPath := "/var/run/secrets/" + file.Name()
```
thallgren: Please use `filepath.Join()` consistently.

cmd/traffic/cmd/agent/agent.go
```diff
@@ -73,37 +73,61 @@ func AppEnvironment() map[string]string {
...
+					dlog.Debugf(ctx, "checking agent secrets mount path: %s", dirPath)
+					if stat, retErr := os.Stat(dirPath); retErr == nil && stat.IsDir() {
+						dlog.Debugf(ctx, "discovered directory: %s", dirPath)
+						// Create this in shared mounts if not already provided
+						appMountsPath := filepath.Join(cfg.AppMounts, dirPath)
+						if _, retErr = os.Stat(appMountsPath); retErr != nil {
```
thallgren: Use `if os.IsNotExists(retErr)` instead of  `retErr` != nil`. Other errors should yield an error return.

thallgren (changes requested):
Thanks Jeff, for providing this fix. Much appreciated. I think this looks good (my comments are mostly regarding style and making the code more readable).

cmd/traffic/cmd/agent/agent.go
0x6a77: yeah, good point. mostly this was left over from before splitting hasmounts(). will change

cmd/traffic/cmd/agent/agent.go
0x6a77: i thought this might get flagged. early returns are a thing i learned to dislike from way back in a previous clang life b/c someone would drop an early return and leave a nightmare resource leak bug. the inherent resource complexity is always there, but gets masked. this isn't my codebase so i should have stuck with the existing style. will fix. (thanks for being patient here)

thallgren:
@0x6a77 this looks good now, thanks. But I think you need to rebase against release/v2 HEAD. The PR currently duplicates my latest PR (which is already merged).

0x6a77:
weird, i did rebase against release/v2. looking

0x6a77:
ok, it should be good now

thallgren (approved):

thallgren:
Merged in #2006.

PR #2005, opened 2021-08-21
Discard logging when metriton fails in CLI app.

thallgren:
The CLI app writes directly to stdout/stderr instead of a log file so
any logs produced by failures in the metriton reporting must be removed.
Metriton errors doesn't affect normal operation. They are not a user
concern and will just cause confusion if printed on the console.


thallgren:
This PR superseds #2003 .

PR #2009, opened 2021-08-23
Fix deadlock in login handler

thallgren:
## Description

The access-token "refresh" routine has a select case that calls the
`loginExecutor.getToken` method which in turn would retrieve the `Token`
and potentially end up in a callback writing to the `refreshTimerReset`
channel that was read by another case in that same select, which lead
to a dead-lock because there would be nothing reading that channel.

This commit ensures that the `loginExecutor.getToken` method is called
in a separate goroutine to avoid blocking the for-loop containing the
select.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,9 @@
 
 - Change: Failure to report metrics is logged using loglevel info rather than error.
 
+- Bugfix: A potential deadlock situation is fixed that sometimes caused the user daemon to hang when the user
+  was logged in. 
+
 ### 2.4.0 (August 4, 2021)
 
 - Feature: There is now a native Windows client for Telepresence.
--- a/pkg/client/connector/userd_auth/login.go
+++ b/pkg/client/connector/userd_auth/login.go
@@ -99,8 +99,13 @@ func NewLoginExecutor(
 		// AFAICT, it's not possible to create a timer in a stopped state.  So we create it
 		// in a running state with 1 minute left, and then immediately stop it below with
 		// resetRefreshTimerUnlocked.
-		refreshTimer:      time.NewTimer(1 * time.Minute),
-		refreshTimerReset: make(chan time.Duration),
+		refreshTimer: time.NewTimer(1 * time.Minute),
+
+		// The refreshTimerReset channel must have a small buffer because it is potentially
+		// written and read by the same select. Deadlocks will happen if it is unbuffered.
+		// 3 slots should be plenty given that the writes are controlled by the above
+		// refreshTimer.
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -31,6 +31,9 @@
 
 - Change: Failure to report metrics is logged using loglevel info rather than error.
 
+- Bugfix: A potential deadlock situation is fixed that sometimes caused the user daemon to become a zombie process when the user
+  was logged in.
```
LukeShu, line 35: I normally understand the term "zombie process" in the unix sense: "a process that has exited but whose parent has not yet `wait()ed` on it", which I don't think is what's going on here.

pkg/client/connector/userd_auth/login.go
```diff
@@ -233,12 +233,17 @@ func (l *loginExecutor) Worker(ctx context.Context) error {
...
 			case <-l.refreshTimer.C:
-				dlog.Infoln(ctx, "refreshing access token...")
-				if token, err := l.getToken(ctx); err != nil {
-					dlog.Infof(ctx, "could not refresh assess token: %v", err)
-				} else if token != "" {
-					dlog.Infof(ctx, "got new access token")
-				}
+				// Call getToken() in separate goroutine because it may end up
+				// writing to the refreshTimerReset channel that this loop is
+				// the only reader of
```
LukeShu: The problem with that is that `l.getToken(ctx)` (may) mutate things in `l` without locking, so doing this in a separate goroutine can corrupt things if a second `<-l.refreshTimer.C` fires while the first goroutine is still running.

I think it would be better to just change the channel to be buffered?

```patch
--- a/pkg/client/connector/userd_auth/login.go
+++ b/pkg/client/connector/userd_auth/login.go
@@ -100,7 +100,10 @@ func NewLoginExecutor(
 		// in a running state with 1 minute left, and then immediately stop it below with
 		// resetRefreshTimerUnlocked.
 		refreshTimer:      time.NewTimer(1 * time.Minute),
-		refreshTimerReset: make(chan time.Duration),
+		// The "refresh" goroutine is the only reader of refreshTimerReset, and
+		// "refresh"->l.getToken()->l.tokenSource.Token()->l.tokenCB() may call
+		// l.resetRefreshTimer() to write to it, so we need a buffer size of at least 1.
+		refreshTimerReset: make(chan time.Duration, 1),
 	}
 	ret.oauth2ConfigMu.Lock()
 	ret.loginMu.Lock()
```

LukeShu (changes requested):
Good job tracking down the deadlock! But I'm not sold on the fix.

LukeShu (approved):
Hit it!

PR #2016, opened 2021-08-25
Ensure that images.repository is used with images.agentImage.

thallgren:
## Description

The configuration contains settings for both repository and image (for
both the CLI installer and the mutating webhook). Prior to this commit,
the `images.agentImage` was assigned verbatim to the container's
`image` attribute and the `repository` setting was ignored for the
agent. Now both are attributes are used.

## Checklist


josecv (approved):
lgtm, good to make this consistent with how the webhook image is configured

PR #2025, opened 2021-08-27
Bump dlib to 1.2.4.

josecv:
This'll include some perf improvements for log levels > debug

## Description

## Checklist

### DIFF
--- a/pkg/client/scout/os_metadata_windows.go
+++ b/pkg/client/scout/os_metadata_windows.go
@@ -6,6 +6,8 @@ import (
 	"context"
 	"strings"
 
+	"golang.org/x/sys/windows"
+
 	"github.com/datawire/dlib/dexec"
 	"github.com/datawire/dlib/dlog"
 )
@@ -13,6 +15,9 @@ import (
 func getOsMetadata(ctx context.Context) map[string]interface{} {
 	cmd := dexec.CommandContext(ctx, "systeminfo")
 	cmd.DisableLogging = true
+	cmd.SysProcAttr = &windows.SysProcAttr{
+		CreationFlags: windows.CREATE_NEW_PROCESS_GROUP,
+	}
 	r, err := cmd.Output()
 	osMeta := map[string]interface{}{}
 	if err != nil {
--- a/pkg/dpipe/dpipe_windows.go
+++ b/pkg/dpipe/dpipe_windows.go
@@ -23,6 +23,12 @@ type processInfo struct {
 	exe  string
 }
 
+func prepareCmd(cmd *dexec.Cmd) {
+	cmd.SysProcAttr = &windows.SysProcAttr{
+		CreationFlags: windows.CREATE_NEW_PROCESS_GROUP,
+	}
+}
+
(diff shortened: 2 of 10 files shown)


thallgren (approved):
Looks good to me. The `pkg/client/logging/discard.go` has moved to `pkg/log` though, so you'll need to resolve that.

pkg/dpipe/dpipe_windows.go
```diff
@@ -23,6 +23,12 @@ type processInfo struct {
 	exe  string
 }
 
+func prepareCmd(cmd *dexec.Cmd) {
```
thallgren: Given the somewhat thorough peruse of child processes and how they are killed in the right order on Windows, I'm surprised this would be needed.

pkg/tun/device_windows.go
```diff
@@ -151,7 +154,12 @@ $job | Receive-Job
 	dlog.Debug(ctx, "Calling ipconfig /flushdns")
 	cmd = dexec.CommandContext(ctx, "ipconfig", "/flushdns")
 	cmd.DisableLogging = true
-	_ = cmd.Run()
+	cmd.SysProcAttr = &windows.SysProcAttr{
```
thallgren, line 157: does `ipconfig` really start subprocesses? If not, why is this needed?

pkg/tun/device_windows.go
```diff
@@ -142,6 +142,9 @@ $job | Receive-Job
 `, t.interfaceIndex, domain)
 	cmd := dexec.CommandContext(ctx, "powershell.exe", "-NoProfile", "-NonInteractive", pshScript)
 	cmd.DisableLogging = true // disable chatty logging
+	cmd.SysProcAttr = &windows.SysProcAttr{
```
thallgren, line 145: Does `powershell.exe` start subprocesses? If not, then why is a process group significant?

pkg/client/scout/os_metadata_windows.go
```diff
@@ -6,13 +6,18 @@ import (
...
+	"golang.org/x/sys/windows"
+
 	"github.com/datawire/dlib/dexec"
 	"github.com/datawire/dlib/dlog"
 )
 
 func getOsMetadata(ctx context.Context) map[string]interface{} {
 	cmd := dexec.CommandContext(ctx, "systeminfo")
 	cmd.DisableLogging = true
+	cmd.SysProcAttr = &windows.SysProcAttr{
```
thallgren, line 18: Same here. Don't think the `systeminfo` starts additional processes, so I don't understand the need for a process group. On the contrary actually. Don't we want this to die if the parent dies?

thallgren (changes requested):
It's very possible that I've misunderstood something regarding Windows and how it handles process-groups, but it seems to me that adding specific behaviour to all (or most) invocations of a command to rectify a windows problem (or regression in dlib?) should belong in `dlib` rather than being repeated over and over again in our code.

thallgren:
Closing this since the upgrade of dlib was made some time ago.

PR #2031, opened 2021-09-01
Bump vm-linux to ubuntu-2004:202010-01

thallgren:


donnyyung (approved):
Seems fine to me as long as the tests pass

Issue #1970, opened 2021-08-09
telepresence uninstall --everything does not Uninstall Traffic Agents Injected via Mutating Webhook


iNoahNothing:
## Summary

`telepresence uninstall --everything` is supposed to remove all Telepresence components from the cluster and revert it back to a pre-Telepresence state. 

This works well when injecting the traffic agent via the normal injection mechanism but does not seem to correctly remove the traffic agent if it is installed via the [mutating webhook](https://www.getambassador.io/docs/telepresence/latest/reference/cluster-config/#mutating-webhook).

You just need to run
```
kubectl rollout restart deploy {{SERVICE}}
```
to revert the service back after uninstalling the traffic manager since the admissions controller has nothing telling it to inject the sidecar but that is a step that could/should be handled by telepresence.

## Replicator 

1. install telepresence and connect to the cluster

```
telepresence connect
```

2. Edit a `Deployment` that has not already been intercepted and add the `telepresence.getambassador.io/inject-traffic-agent: enabled` annotation to it
3. Watch the pods spin up and check to make sure that the traffic-agent container has been added
```
kubectl get pod
NAME                       READY   STATUS    RESTARTS   AGE
quote-5d785f4f6f-dg7h4     2/2     Running   0          8m7s
```
4. Run `telepresence uninstall --everything` and see that the quote pod is not removed
```
telepresence uninstall --everything

kubectl get pod -n ambassador
NAME                               READY   STATUS        RESTARTS   AGE
traffic-manager-75cc4c5b7b-z8xfm   0/1     Terminating   0          5m26s
kubectl get pod
NAME                       READY   STATUS    RESTARTS   AGE
quote-7d69cd6447-58jv4     2/2     Running   0          8m7s
```

5. Run `kubectl rolllout restart deploy {{SERVICE}}` to restart the deployment and uninstall the agent
```
kubectl rolllout restart deploy quote

kubectl get po
NAME                       READY   STATUS    RESTARTS   AGE
quote-758cfcdb67-6c4wx     1/1     Running   0          8m7s
```

efunk:
Resolved in Telepresence 2.4.3 GA.

PR #2046, opened 2021-09-03
draft

njayp:
# DRAFT PR
## Description

Uninstalls agents installed with agent-injector by rolling thier pods after agent-injector has been uninstalled.

## Issues

https://github.com/telepresenceio/telepresence/issues/1970

## Checklist

### DIFF
--- a/pkg/client/cli/telepresence_test.go
+++ b/pkg/client/cli/telepresence_test.go
@@ -792,7 +792,6 @@ func (cs *connectedSuite) TestM_AutoInjectedAgent() {
 
 	cs.Run("can be intercepted", func() {
 		defer telepresence(cs.T(), "leave", "echo-auto-inject-"+cs.ns())
-
 		require := cs.Require()
 		stdout, stderr := telepresence(cs.T(), "intercept", "--namespace", cs.ns(), "--mount", "false", "echo-auto-inject", "--port", "9091")
 		require.Empty(stderr)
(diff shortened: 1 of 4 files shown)


pkg/client/cli/telepresence_test.go
```diff
@@ -803,6 +802,37 @@ func (cs *connectedSuite) TestM_AutoInjectedAgent() {
...
+			2*time.Second,  // polling interval
+		)
+
+		stdout, stderr := telepresence(cs.T(), "uninstall", "-e")
+		require.Empty(stderr)
+		require.Contains(stdout, "Root Daemon quitting... done")
+
+		stdout, stderr = telepresence(cs.T(), "list", "--namespace", cs.ns())
+		require.Empty(stderr)
+		require.Contains(stdout, "echo-auto-inject: ready to intercept (traffic-agent not yet installed)")
```
donnyyung, line 831: My hunch is that maybe it's taking too long for all the agents to be uninstalled and when you call the list, the new traffic-manager is seeing the old `echo-auto-inject` agent before it's terminating.. 

One thing you could do to validate if that's the issue is between the uninstall and the list, do something like what we do here: https://github.com/telepresenceio/telepresence/blob/release/v2/pkg/client/cli/telepresence_test.go#L1161-L1181 to ensure there's only pod (and maybe even ensure there's only one pod with one container, so we know the agent has been removed)

donnyyung: Additionally, some of `TestZ_Uninstall` is affected by doing an `uninstall --everything` earlier in the tests, so perhaps it's worth moving this to the end of `TestZ_Uninstall` so we're still testing the individual uninstalls.. Or you could restore the necessary agents for those tests as part of the cleanup of this test, up to you!

njayp: Thanks for looking it over

PR #2047, opened 2021-09-04
Add error categorization to provide more concise info to the user.

### LINKED ISSUE #1995: Avoid pointless mention of daemon logs.
Output like this is pointless and hard to read:
```console
➜  telepresence leave hello
telepresence: error: rpc error: code = NotFound desc = Intercept named "hello" not found

See logs for details (3 errors found): "/Users/thhal/Library/Logs/telepresence/daemon.log"

See logs for details (1 error found): "/Users/thhal/Library/Logs/telepresence/connector.log"
➜  
```
Just writing the message would be a better user experience, e.g.
```console
➜  telepresence leave hello
Intercept named "hello" not found
➜  
```

thallgren:
## Description

There's little point in telling the user to look in the logs of the
user and root daemons when we know that there's no additional info
there. Some examples of this:

- The user makes an error (incorrect CLI options or syntax, misspelling
  the name of an intercept, etc.)
- The config.yml or kubeconfig file is corrupt in some way.
- The CLI `--docker-run` fails to start docker or create an environment file.

## Checklist


pkg/client/cli/cmds_intercept.go
```diff
@@ -294,11 +294,12 @@ func newInterceptState(
 	}
 }
 
-func interceptMessage(r *connector.InterceptResult) string {
+func interceptMessage(r *connector.InterceptResult) error {
 	msg := ""
+	cat := errcat.Unknown
```
donnyyung: Sort of a nit so take it or leave it but... This is how my mind works, but I think I'd prefer a var name like "category" or "errCategory" or "errCat" just because when I see "cat" I first think of the `cat` command, followed by a pet, and I don't think I'll see this and think it's short for category.

pkg/client/errcat/errors.go
```diff
@@ -0,0 +1,68 @@
...
+type categorized struct {
+	error
+	category Category
+}
+
+const (
+	OK       = Category(iota)
+	User     // User made an error
+	Config   // Errors in config.yml, extensions, or kubeconfig
+	OtherCLI // Error generated in the CLI process, so no use pointing logs
```
donnyyung: Again this is kind of a nit, but I think perhaps a better name for this would be `NoLogs`, it took me a bit to grok what `OtherCLI` meant, and I think it'd be even harder without the context of this PR.

pkg/client/errcat/errors.go
```diff
@@ -0,0 +1,68 @@
...
+	return ce.error
+}
+
+// GetCategory returns the error category for a categorized error, OK for nil, and
+// Unknown for other errors
+func GetCategory(err error) Category {
+	if err == nil {
+		return OK
+	}
+	for {
```
donnyyung, line 61: I think I'm being dense, but is this in a for loop so that we keep unwrapping the error until we get to a category or lackthereof?

donnyyung:
It's late so I'm going to do another scan of this in the morning. but so far it looks good and mostly just some nits around naming.  I'll ping you in the morning once I've done another read through

pkg/client/errcat/errors.go
thallgren: Yes, that is exactly right. The loop is needed because we don't know how many levels of `fmt.Errorf("...%w", err)` the argument have been exposed to.

donnyyung (approved):
On second look, this all seems good + sane. I think we'll have to use some extra eyes on future reviews to ensure people are using errcat instead of error in most cases.

PR #2037, opened 2021-09-02
Upgrade tools, make the linter stricter

LukeShu:
## Description

Now that we're on newer Go (https://github.com/telepresenceio/telepresence/pull/2036) I can upgrade golangci-lint and make a buncha changes that were blocked.
- Upgrade all of the tools that we use (golangci-lint, protoc, ...)
- Tighten the golangci-lint configuration up a bit; notably switch it back from goimports to gci, so that it's once again linting the grouping of imports.

v2: I split https://github.com/telepresenceio/telepresence/pull/2051 and https://github.com/telepresenceio/telepresence/pull/2052 out in to separate PRs that this is now on top of.

As usual, I suggest a commit-by-commit review.

## Checklist

 - [x] I made sure to update `./CHANGELOG.md`. - yep
 - [x] I made sure to either submit a docs PR, or tell Matt about the necessary documentation changes. - no changes necessary
 - [x] My change is adequately tested. - yep
 - [x] I updated `DEVELOPING.md` with any any special dev tricks I had to use to work on this code efficiently. - no tricks

PR #2052, opened 2021-09-05
pkg/client/logging: Implement btime on GOOS=linux

LukeShu:
## Description

We did some tricks to spoof the btime on GOOS=linux, since it isn't returned by the stat(2) system call.  However, it is returned by the statx(2) system call, so use that instead.

This is split out from https://github.com/telepresenceio/telepresence/pull/2037

## Checklist

 - [x] I made sure to update `./CHANGELOG.md`. - yep
 - [x] I made sure to either submit a docs PR, or tell Matt about the necessary documentation changes. - no docs nescessary
 - [x] My change is adequately tested. - yep, added tests


thallgren (approved):
Looks great. At some point (not in this PR) this should be move into a package of its own since it's completely generic.

PR #2056, opened 2021-09-07
Improve synchronization and close semantics for Telepresence gRPC tunnels

### LINKED ISSUE #1810: Regression: Unable to download large files via intercept after telepresence 2.3
Hello,

I have been using telepresence v2.2.2 with a GKE cluster and an nginx ingress for a few weeks, using telepresence to develop frontend against backend already in the cluster.

The frontend is using react and TypeScript, using yarn and `yarn.lock` file to lock all the dependencies explicitly. We're using webpack dev server with hot reload for development and that is how telepresence was used - original deployment of the prod version of the app gets intercepted and all requests for UI are routed to the webpack dev server on developer's computer.

The application in GKE is behind nginx-ingress (version 0.32.0) and ingress is exposing the application over https.

Things have worked fine for a while, including the hot reload feature. Recently I have updated to telepresence v2.3.2 and things stopped working. Specifically the UI never finishes loading. I can see some of webpack messages, but nothing finishes loading.

After a lot of trial and error with things that could possibly go wrong, I have ruled out:
* updates to the backend code - by rolling back to version that I am confident worked with telepresence and ensuring all backend components use a specific container image and settings (the project has a separate config repository for Kubernetes so it is easy to roll back to a specific version)
* updates to the frontend code - I have rolled back to commit I know to work with telepresence ; since all the npm packages are `yarn.lock`-ed, it should be a reproducible setup
* browser issues - I have run an older version of Chrome for a while to ensure the problem is not related to an update in the browser
* generic networking issues - I have replaced webpack dev server with `nginx` running on my machine and hosting a static directory, I was able to download small and large files without issues

After ruling out all the issues that could be caused by changes on my end, I have consistently tried all versions of telepresence and came up with `v2.2.2` working as expected and `v2.3.0` not working. I have also tried `v2.3.1` and `v2.3.2` and all of them fail to work and with `telepresence intercept` enabled.

I suspect this has to be some edge case related to many features webpack dev server is using - such as websockets it uses to inform the browser of updates in the remote code and need to reload.

Could anyone help determine what could be the issue? Is this a known problem with v2.3?

Are there some settings, additional configuration or changes to `traffic-manager` deployment I could make to get additional logs / information?

thallgren:
## Description

This PR refactors some of the logic related to the Telepresence gRPC tunnels so that the tunnel sends are throttled by a synchronization mechanism. The actual closing mechanism was also simplified to increase stability (and readability).

The first commit was triggered by frequent Helm timeouts during development.
Commit 2-4 are there to make things more readable and move some concerns into the tunnel itself.
Commit 5-7 is where the real changes are.

It's highly recommended to review this PR one commit at a time.
## Checklist


pkg/client/config.go
```diff
@@ -572,7 +572,7 @@ func GetDefaultConfig(c context.Context) Config {
 			PrivateProxyDial:             5 * time.Second,
 			PrivateTrafficManagerAPI:     15 * time.Second,
 			PrivateTrafficManagerConnect: 60 * time.Second,
-			PrivateHelm:                  12 * time.Second,
+			PrivateHelm:                  30 * time.Second,
```
josecv, line 575: Does this mean that we need to bump the `PrivateTrafficManagerAPI` timeout as well? I seem to recall doing 12s here because that one was set to 15

pkg/connpool/pool.go
```diff
@@ -35,15 +35,23 @@ func (p *Pool) release(id ConnID) {
 // HandlerCreator describes the function signature for the function that creates a handler
 type HandlerCreator func(ctx context.Context, release func()) (Handler, error)
 
-// Get finds a handler for the given id from the pool, or creates a new handler using the given createHandler func
+// Get finds a handler for the given id from the pool and returns it. Nil is returned if no such handler exisyd
```
josecv: typo :) `exisyd`

pkg/connpool/tunnel.go
```diff
@@ -18,33 +22,81 @@ type BidiStream interface {
...
+	if err := s.stream.Send(SyncRequestControl(ackSent).TunnelMessage()); err != nil {
+		return err
+	}
+	for ctx.Err() == nil {
+		lastAck := atomic.LoadUint32(&s.lastAck)
+		if ackSent <= lastAck+s.ackWindow {
+			dlog.Debugf(ctx, "Received Ack of %d", ackSent)
+			break
+		}
+		dtime.SleepWithContext(ctx, time.Millisecond)
```
josecv, line 113: hmm what's actually going to break us out of this? Is the idea that another goroutine is calling `Receive()` and so `lastAck` is gonna get updated over there?

Is there something we can do to make sure that we don't end up hung in here?

josecv: also, should the `SleepWithContext` maybe have an increasing backoff? i.e. sleep for 1ms, then 2ms, 4ms, etc

pkg/connpool/tunnel.go
```diff
@@ -18,33 +22,81 @@ type BidiStream interface {
...
+		if ctrl, ok := msg.(Control); ok {
+			switch ctrl.Code() {
+			case SyncRequest:
+				if err = s.stream.Send(SyncResponseControl(ctrl).TunnelMessage()); err != nil {
+					return nil, fmt.Errorf("failed to send sync response: %w", err)
+				}
+				continue
+			case SyncResponse:
+				atomic.StoreUint32(&s.lastAck, ctrl.AckNumber())
+				continue
```
josecv: hmm it makes me a little uncomfortable that we handle these in `Receive`, but expect the caller to handle all other control codes. Could we maybe make these two private? And possibly put the tunnel in a different package than e.g. the `dialer.go`

pkg/connpool/tunnel.go
```diff
@@ -156,23 +177,24 @@ func (s *tunnel) DialLoop(ctx context.Context, pool *Pool) error {
...
-			// Only Connect requested from peer may create a new instance at this point
-			return nil, nil
+	code := ctrl.Code()
+	dlog.Debugf(ctx, "<- MGR %s, code %s", id.ReplyString(), code)
+
+	if code != Connect {
+		if conn := pool.Get(id); conn != nil {
+			conn.HandleMessage(ctx, ctrl)
+		} else if code != Disconnect && code != DisconnectOK {
+			dlog.Error(ctx, "control packet lost because no connection was active")
```
josecv: could we include what type of control it was here?

josecv (commented):
lgtm, i have some minor suggestions and questions

pkg/client/config.go
thallgren: No, I don't think so. That one should be about calls that happen once the gRPC has become active. The `PrivateTrafficManagerConnect` however, must be larger than `PrivateHelm`.

pkg/connpool/tunnel.go
thallgren: You're right. It will hang unless an ack arrives. Will fix that.
I don't think there's much point in increasing the sleep given that the only thing the loop does is to do an atomic check of an integer, which is very cheap.

pkg/connpool/tunnel.go
thallgren: I'll make them private. Moving them to another package is a bit cumbersome given that the `ControlCode` is a proper type so then we'd have to break the `iota` sequence, which introduces a risk of missing a future overlap.

pkg/connpool/tunnel.go
thallgren: Oh, I already had a `for ctx.Err() == nil {` there so a context cancellation breaks the loop. So this is all good IMO.

josecv (approved):

PR #2060, opened 2021-09-08
Upgrade runc and containerd

### LINKED ISSUE #2059: Dependabot security vulnerabilities
Dependabot has identified some security vulnerabilities: https://github.com/telepresenceio/telepresence/security/dependabot

We need to review these to determine if we are impacted by the vulnerabilities. Regardless of impact, we should update the dependencies to remove the alerts.

josecv:

## Description

## Checklist

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -178,4 +178,12 @@ replace github.com/telepresenceio/telepresence/rpc/v2 => ./rpc
 // https://github.com/grpc/grpc-go/pull/4567
 replace google.golang.org/grpc v1.38.0 => github.com/datawire/grpc-go v1.38.0-dev.0.20210626184227-5ef87f395316
 
-replace github.com/opencontainers/runc => github.com/opencontainers/runc v1.0.0-rc95
+// Helm pulls both of these in to support the registry functionality (which telepresence doesn't leverage)
+// but it pulls in old, vulnerable, versions. These replaces will force it into pulling new versions.
+// If there's ever a helm version that depends on versions >= these two, we can drop this replace block.
+// We do this via replace because if we updated the require() //indirect block above, and then upgraded helm to a version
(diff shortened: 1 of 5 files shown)


go.mod
```diff
@@ -177,3 +177,7 @@ replace github.com/telepresenceio/telepresence/rpc/v2 => ./rpc
 
 // https://github.com/grpc/grpc-go/pull/4567
 replace google.golang.org/grpc v1.38.0 => github.com/datawire/grpc-go v1.38.0-dev.0.20210626184227-5ef87f395316
+
+replace github.com/opencontainers/runc => github.com/opencontainers/runc v1.0.0-rc95
```
donnyyung: Question mostly for my edification: Are we using these `replaces` bc runc and containerd are imported indirectly through a dependencies' dependency (potentially going even further), so instead of bumping the dependency to a version that uses an OK version of the dependencies (if that exists), we use the replace to force a certain version?

It might be worth adding a comment about when we think we'll be able to remove this replace line or why we need to use this (e.g. we use replace .... runc v1.0.0-rc95 bc go-cmp v0.5.5 uses an insecure version of runc, when we bump that library, we may be able to remove this), since in my uninformed opinion, it seems nice to have as little `replace`s as possible? If that thinking is incorrect, let me know.

josecv: yea so i made them `replace`s because the dependencies are being imported indirectly by something else -- to be totally honest, i'm not sure if there is a best practice around how this should be done, but i made them replaces because if we ever upgrade helm (which pulls both of those in), and the upgraded version still uses the bad versions, we might end up accidentally downgrading to the bad runc and containerd

josecv: i will definitely add a comment to this effect

donnyyung (approved):
Giving you an approval now so you can merge once you add your comment :)

PR #2066, opened 2021-09-10
Let the telepresence status command report correct DNS IP

### LINKED ISSUE #2061: nil local IP and nil remote IP
Is it normal or expected to have `<nil>` in the local and remote IPs?

```
LAMU02Y40MKJG5J:telepresence mgarriga$ tp connect
Launching Telepresence Root Daemon
Launching Telepresence User Daemon
Connected to context oea-dev (https://10.202.82.152:6443)
LAMU02Y40MKJG5J:telepresence mgarriga$ tp status
Root Daemon: Running
  Version   : v2.4.2 (api 3)
  DNS       :
    Local IP        : <nil>
    Remote IP       : <nil>
    Exclude suffixes: [.arpa .com .io .net .org .ru]
    Include suffixes: []
    Timeout         : 4s
  Also Proxy: (0 subnets)
User Daemon: Running
  Version           : v2.4.2 (api 3)
  Ambassador Cloud  : Logged out
  Status            : Connected
  Kubernetes server : https://10.202.82.152:6443
  Kubernetes context: oea-dev
  Telepresence proxy: ON (networking to the cluster is enabled)
  Intercepts        : 0 total
LAMU02Y40MKJG5J:telepresence mgarriga$ 
```

After I connect and try `curl -ik https://kubernetes.default` I get `curl: (7) Failed to connect to kubernetes.default port 443: Operation timed out`

thallgren:
## Description

The `telepresence status` would report that both local and remote
IP is "nil" on most platforms. The local IP is only used on Linux and
only by the overriding DNS resolver and should only be reported when
it's set. The remote IP is always set (but perhaps not in the config)
and the current value should be reported.

## Checklist

### DIFF
--- a/pkg/client/daemon/service.go
+++ b/pkg/client/daemon/service.go
@@ -48,7 +48,6 @@ to troubleshoot problems.
 // service represents the state of the Telepresence Daemon
 type service struct {
 	rpc.UnsafeDaemonServer
-	dns           string
 	hClient       *http.Client
 	outbound      *outbound
 	cancel        context.CancelFunc
@@ -149,7 +148,6 @@ func run(c context.Context, loggingDir, configDir, dns string) error {
 	dlog.Debug(c, "Listener opened")
 
 	d := &service{
-		dns: dns,
 		hClient: &http.Client{
 			Timeout: 15 * time.Second,
 			Transport: &http.Transport{
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,8 @@
 
 - Bugfix: The traffic manager will establish outbound connections in parallel instead of sequentially.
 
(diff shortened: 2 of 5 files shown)


pkg/client/daemon/service.go
```diff
@@ -48,7 +48,6 @@ to troubleshoot problems.
 // service represents the state of the Telepresence Daemon
 type service struct {
 	rpc.UnsafeDaemonServer
-	dns           string
```
donnyyung, line 51: This change doesn't seem directly related to the others, but was this just an extra value that we weren't using? I guess I'd expect more changes if we were removing a value from the struct, so just want to make sure I understand.

donnyyung (approved):
Left one question for my own knowledge but LGTM!

pkg/client/daemon/service.go
thallgren: Yes, it's a removal of a field that became obsolete some time ago.

PR #2070, opened 2021-09-12
Fix never released regression in the overriding DNS-resolver.

thallgren:
This commit fixes a regression in the overriding DNS-resolver used on
Linux systems that run without systemd-resolved. The regression was
introduced in pr #2038 and never released.


donnyyung (approved):
LGTM!

josecv (approved):

PR #2079, opened 2021-09-14
search "Application Support" folder

njayp:

## Description

The telepresence config file can be in either "\~/Library/ApplicationSupport" or "\~/Library/Application Support", so check both places

## Checklist

### DIFF
--- a/smoke-tests/run_smoke_test.sh
+++ b/smoke-tests/run_smoke_test.sh
@@ -168,7 +168,7 @@ get_config() {
 
     case "$(uname)" in
 	Darwin)
-        if [ test -f "$HOME/Library/ApplicationSupport/telepresence/config.yml" ]
+        if test -f "$HOME/Library/ApplicationSupport/telepresence/config.yml"
         then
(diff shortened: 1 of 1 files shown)


thallgren:
AFAICT, telepresence will never use the variant without a space. The `UserConfigDir()` function hardcodes the darwin folder to `home + "/Library/Application Support"`.  The same is true for `os.UserConfigDir()`.

@njayp when are you seeing the variant without a space?

thallgren:
@njayp can you please provide more info here or close the ticket?

njayp:
> @njayp can you please provide more info here or close the ticket?

sorry, I missed your first comment. It seems like the script currently has "~/Library/ApplicationSupport", without the space in line 171, so I changed it to have the space. At the same time, the script seems to work "as is" for everyone else, so I added logic to allow both locations

thallgren:
Ok, in that case I suggest that you simply change it to "~/Library/Application Support". If something, or someone, uses the name without space, then there's an error somewhere else that we don't want to hide.

PR #2088, opened 2021-09-16
Allow the webhook injector to inject into pods without named ports

josecv:
## Description

## Checklist


cmd/traffic/cmd/agent/client.go
```diff
@@ -187,7 +188,11 @@ func lookupHostWaitLoop(ctx context.Context, manager rpc.ManagerClient, session
 }
 
 func logLevelWaitLoop(ctx context.Context, logLevelStream rpc.Manager_WatchLogLevelClient) {
-	timedLevel := log.NewTimedLevel(os.Getenv("LOG_LEVEL"), log.SetLevel)
+	level, ok := os.LookupEnv("LOG_LEVEL")
+	if !ok {
+		level = os.Getenv(install.EnvPrefix + "LOG_LEVEL")
```
thallgren: I think `_TEL_AGENT_LOG_LEVEL` should have precedence. The `LOG_LEVEL`, if present, is either legacy (which is weird, if combined with a new agent image), or it's a copy from the intercepted container, in which case our agent should include it in the propagated environment (which it does), but not act on it since we now have our own TP-specific setting for that.

cmd/traffic/logger.go
```diff
@@ -15,7 +16,12 @@ func makeBaseLogger(ctx context.Context) context.Context {
 	logrusFormatter := log.NewFormatter("2006-01-02 15:04:05.0000")
 	logrusLogger.SetFormatter(logrusFormatter)
 
-	log.SetLogrusLevel(logrusLogger, os.Getenv("LOG_LEVEL"))
+	level, ok := os.LookupEnv("LOG_LEVEL")
+	if !ok {
+		level = os.Getenv(install.EnvPrefix + "LOG_LEVEL")
```
thallgren: The `install.EnvPrefix` is `_TEL_AGENT` and specific for the agent. This code is used by both the traffic-manager and the traffic-agent. Seems weird that the traffic-manager should attempt to set its log-level from the agent. Perhaps pass the level as an argument to `doMain()` and then to `makeBaseLogger()`?

cmd/traffic/cmd/agent/client.go
thallgren: The `_TEL_AGENT_LOG_LEVEL` must be included in the global `skipKeys` map.

pkg/client/connector/userd_trafficmgr/testdata/addAgentToWorkload/cur/deployment-initializer.output.yaml
```diff
@@ -76,6 +76,10 @@ deployment:
               - /bin/stat
               - /tmp/agent/ready
           resources: {}
+          securityContext:
+            runAsGroup: 7777
+            runAsNonRoot: true
+            runAsUser: 7777
```
thallgren, line 82: Will this user have permissions to mount everything from the intercepted container (I know we have a ticket up claiming that this happens already when running as root and I wouldn't like to see that problem getting worse)?

pkg/install/container.go
```diff
@@ -10,8 +10,10 @@ import (
 	"github.com/datawire/ambassador/pkg/kates"
 )
 
-const envPrefix = "_TEL_AGENT_"
+const EnvPrefix = "_TEL_AGENT_"
```
thallgren, line 13: This change belong in the previous commit. No need to change it though.

cmd/traffic/cmd/agentinit/agent_init_windows.go
```diff
@@ -0,0 +1,14 @@
+package agentinit
+
+import (
+	"context"
+	"fmt"
+)
+
+// This file needs to exist because agent_init.go imports iptables, which obviously doesn't exist on windows.
+// It really doesn't matter, since there's no support for windows-based containers of any sort, but it'll fail the build.
```
thallgren, line 9: This shouldn't be needed (but that's for another PR). There should never be an attempt to build stuff under `cmd/traffic` on platforms other than Linux.

cmd/traffic/cmd/agentinit/agent_init.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+	// Finally, any traffic heading out of the traffic agent should pass by unperturbed -- it should obviously not be
+	// redirected back into the agent, but it also should not pass through a mesh proxy.
+	err = iptables.Insert(nat, "OUTPUT", 1,
+		"-m", "owner", "--gid-owner", strconv.FormatInt(install.AgentUID, 10),
+		"-j", "RETURN")
+	if err != nil {
+		return fmt.Errorf("failed to insert --gid-owner rule in OUTPUT: %w", err)
+	}
+	return nil
```
thallgren, line 72: Great comments in this method!

pkg/install/discovery.go
```diff
@@ -19,13 +19,8 @@ import (
...
 	svcPorts := make([]*kates.ServicePort, 0)
 	ports := svc.Spec.Ports
-	var isName bool
 	validName := validation.IsValidPortName(nameOrNumber)
-	if len(validName) > 0 {
-		isName = false
-	} else {
-		isName = true
-	}
+	isName := len(validName) == 0
```
thallgren, line 23: Oh dear (blushing).

thallgren (changes requested):
Aside from a couple of nits, this looks really promising.

I see that the old behavior is retained when the ports are named. Is there any particular reason for that or could we use iptables there too? It might be easier to support just one solution.

pkg/client/connector/userd_trafficmgr/testdata/addAgentToWorkload/cur/deployment-initializer.output.yaml
josecv: That is a great question, let me go test it out

josecv:
> Is there any particular reason for that or could we use iptables there too?

There's nothing stopping us from using iptables there too --in fact when I was testing this, at one point I was using builds that only ever used iptables and they seemed to be okay. My only worry is accidentally breaking workloads that work today so IMO we should roll it out slowly and change that logic in a later release.
However, if you feel like we should just rip off the band-aid now I could also be convinced 🙂 

On a similar note, in theory this should also work to allow intercepting headless services, but when I tested that it didn't work at all --probably, some tweaking of the iptables would be necessary to get them working.

pkg/client/connector/userd_trafficmgr/testdata/addAgentToWorkload/cur/deployment-initializer.output.yaml
josecv: So i just tried out creating something like this:
```yaml
---
kind: PersistentVolumeClaim
apiVersion: v1
metadata:
  name: echo-pvc
spec:
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 500Gi
---
apiVersion: v1
kind: Service
metadata:
  name: echo-auto-inject
spec:
  type: ClusterIP
  selector:
    service: echo-auto-inject
  ports:
    - name: proxied
      port: 80
      targetPort: 8080
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: echo-auto-inject
  labels:
    service: echo-auto-inject
spec:
  replicas: 1
  selector:
    matchLabels:
      service: echo-auto-inject
  template:
    metadata:
      annotations:
        telepresence.getambassador.io/inject-traffic-agent: enabled
      labels:
        service: echo-auto-inject
    spec:
      volumes:
        - name: echo-vol
          persistentVolumeClaim:
            claimName: echo-pvc
      securityContext:
        fsGroup: 0
      containers:
        - name: echo-auto-inject
          image: jmalloc/echo-server
          volumeMounts:
          - name: echo-vol
            mountPath: /var/echo
          ports:
            - containerPort: 8080
          resources:
            limits:
              cpu: 50m
              memory: 128Mi
```

And then intercepting it and writing to the volume:

```bash
$ telepresence intercept echo-auto-inject --port 8080
Using Deployment echo-auto-inject
intercepted
    Intercept name    : echo-auto-inject
    State             : ACTIVE
    Destination       : 127.0.0.1:8080
    Volume Mount Point: /var/folders/j8/kzkn41mx2wsd_ny9hrgd66fc0000gp/T/telfs-1093182516

$ echo 'heyo' > /var/folders/j8/kzkn41mx2wsd_ny9hrgd66fc0000gp/T/telfs-1093182516/var/echo/fff
$ cat /var/folders/j8/kzkn41mx2wsd_ny9hrgd66fc0000gp/T/telfs-1093182516/var/echo/fff
heyo
```

It seems like it works well enough, but is there something specific I should try @thallgren?

josecv: Also tried with a non-root pod security context, and it works (whereas I think it didn't use to)

thallgren (approved):

Issue #2084, opened 2021-09-15
[v2] traffic-manager high CPU usage after some undetermined amount of usage


frankgu968:
The single `traffic-manager` instance deployed on our cluster suffers from pinned CPU usage. This causes all requests to fail and existing connections to time out. Our only recourse at the moment is to restart the `traffic-manager` and re-establish the intercepts. 

Our traffic volume: <20 intercepts and <80rps total; each request volume is small (kB). 

The `traffic-manager` starts off fine and shows steady CPU usage around 300mCores. However, after some undetermined amount of usage, the CPU usage shoots up to the configured `limit` (we tried 1 core, and increased it to 2 cores to no avail) and stays pinned. Note that the memory usage does not approach the configured `request` amount. 

During the cursory investigation with @josecv into this issue, I managed to pull a couple of flame graphs that show the `client.WatchSubnets` function starts off taking the majority of CPU time: 

![image](https://user-images.githubusercontent.com/9001524/133475376-24756f0e-c088-4b02-86a3-b1be7725663b.png)

At 1 hour of regular traffic: 
![image (1)](https://user-images.githubusercontent.com/9001524/133475405-7c1a1e02-c9a7-4bb9-86bc-f72e370155d2.png)

At 2 hours of regular traffic: 
![image (2)](https://user-images.githubusercontent.com/9001524/133475424-f18fb0ae-3cdc-47c9-be1d-09ce11a41f41.png)

A considerable amount of CPU is being spent on the `kates` client JSON marshaling and unmarshalling. This may be one facet of the `traffic-manager's` performance issue.

frankgu968:
Update: I managed to pull the pprof outputs from when the `traffic-manager` is at max CPU. It looks like there is some sort of race condition in the `watchable.(*AgentMap).coalesce` function that is deadlocking some goroutines: 

![Screenshot 2021-09-16 085615](https://user-images.githubusercontent.com/9001524/133646711-2576fec3-d959-44bc-a581-d4ab17f8664e.png)
![Screenshot 2021-09-16 085557](https://user-images.githubusercontent.com/9001524/133646739-d9259df4-0215-483f-84bb-222ba4f012f2.png)

![Screenshot 2021-09-16 090134](https://user-images.githubusercontent.com/9001524/133646749-b250fb5f-a973-46ec-ac7e-046623a961b5.png)
[...]

Issue #2023, opened 2021-08-27
Subnet retrieval should be more flexible


johnlanni:
Hi, when I enable the Calico CNI in k8s , the subnet of the pods’ IPs can’t be obtained by telepresence exactly.  I found some information about the Pod’s CIDR and Calico:

https://docs.projectcalico.org/networking/get-started-ip-addresses
https://github.com/projectcalico/calico/issues/2592#issuecomment-489233638

It seems we must enable the Host-local IPAM, then Calico will use the Node.Spec.PodCIDR for pod.
I think the logic of  `Deriving subnets from IPs of pods`  should be configureable, then we no need to enable the Host-local IPAM plugin.

![image](https://user-images.githubusercontent.com/6763318/131061920-8424e976-8371-458f-8448-d3d8e592aa09.png)

Or check whether Calico is deployed in the cluster, and obtain the CIDR from Calico’s blockaffinities, then adding to the pods' CIDRs.

![image](https://user-images.githubusercontent.com/6763318/131062065-8f6e059a-eecd-490a-9907-7bd189d71cf9.png)

efunk:
Resolved in Telepresence 2.4.4 GA

PR #2093, opened 2021-09-17
Improve traffic-managers podCIDR discovery

thallgren:
## Description

This PR contains two improvements to the pod CIDR discovery mechanism.
1. Performance improved by replacing the kates watchers with standard k8s APIs (#2084).
2. Make pod discovery strategy configurable in the Helm chart (#2023).

## Checklist


charts/telepresence/values.yaml
```diff
@@ -70,6 +70,20 @@ grpc: {}
...
+# podCIDRs is the verbatim list of CIDRs used when the podCIDRStrategy is set to environment
+podCIDRs: []
+
+# podCIDRStrategy controls what strategy the traffic-manager will use for finding out what
+# CIDRs the cluster is using for its pods. Valid values are:
+#
+#  nodePodCIDRs extract CIDRs from the podCIDR and podCIDRs field of the Node Spec.
+#  coverPodIPs  extract IPs from the podIP and podIPs field of the Pod Status and compute the CIDRs needed to cover those IPs.
+#  environment  use CIDRs listed in the space separated POD_CIDRS environment variable verbatim.
+#  auto         first tru nodePodCIDRs and if that fails, tru coverPodIPs
```
josecv: typo -- it says `tru` instead of `try`

cmd/traffic/cmd/manager/internal/cluster/nodewatcher.go
```diff
@@ -0,0 +1,249 @@
...
+	"github.com/datawire/dlib/dlog"
+	"github.com/telepresenceio/telepresence/v2/pkg/subnet"
+)
+
+type nodeWatcher struct {
+	lister   licorev1.NodeLister
+	informer cache.SharedIndexInformer
+	subnets  []*net.IPNet
+	changed  time.Time
+	lock     sync.Mutex // Protects all access to ipsMap
```
josecv: should this read `Protects all access to subnets`?

cmd/traffic/cmd/manager/internal/cluster/nodewatcher.go
```diff
@@ -0,0 +1,249 @@
...
+	if changed {
+		dlog.Infof(ctx, "Found %d subnets", len(w.subnets))
+		w.changed = time.Now()
+	} else {
+		dlog.Info(ctx, "No subnets found")
+	}
+	return changed
+}
+
+func (w *nodeWatcher) watch(ctx context.Context) {
```
josecv: maybe i'm just confused but it seems like watch is never invoked?

cmd/traffic/cmd/manager/internal/cluster/nodewatcher.go
```diff
@@ -0,0 +1,249 @@
...
+		}
+	}
+	if changed {
+		dlog.Infof(ctx, "Found %d subnets", len(w.subnets))
+		w.changed = time.Now()
+	} else {
+		dlog.Info(ctx, "No subnets found")
+	}
+	return changed
+}
```
josecv: it seems like this function combines initialization with a precondition check (i.e. "Are we actually able to watch subnets from nodes?") -- that's okay, but calling it twice in a row will make the second call return `false`, no? Maybe we should short-circuit return if `w.subnets` is already populated -- after all, if they're there then getting them is definitely viable

cmd/traffic/cmd/manager/internal/cluster/podwatcher.go
```diff
@@ -0,0 +1,247 @@
+package cluster
```
josecv, line 1: same comments apply here as for the nodewatcher :)

josecv (commented):
looks good but i'm somewhat confused about the `watch` function --just wanna make sure that it actually _is_ being invoked, otherwise none of this will work 🙃

thallgren:
The `watch` function was never called, but that only meant that future updates to podCIDRs after the initial scan where undetected. The initial scan happened anyway (in `viable`). Not good of course, but not a case of "none of this will work" either.

charts/telepresence/values.yaml
```diff
@@ -79,7 +79,7 @@ podCIDRs: []
 #  nodePodCIDRs extract CIDRs from the podCIDR and podCIDRs field of the Node Spec.
 #  coverPodIPs  extract IPs from the podIP and podIPs field of the Pod Status and compute the CIDRs needed to cover those IPs.
 #  environment  use CIDRs listed in the space separated POD_CIDRS environment variable verbatim.
-#  auto         first tru nodePodCIDRs and if that fails, tru coverPodIPs
+#  auto         first try nodePodCIDRs and if that fails, tru coverPodIPs
```
josecv, line 82: still says `tru coverPodIPs` :^)

josecv (approved):

Issue #2073, opened 2021-09-12
Confusing CLI behavior with Kubernetes Flags


NurlashKO:
After specifying any connection flag(s), for example:
```
$ telepresence connect --request-timeout 100
```
All those flags must be included in the proceeding calls:
```
$ telepresence list
telepresence: error: connector.Connect: Cluster configuration changed, please quit telepresence and reconnect
```
```
telepresence list --request-timeout 100    
xxx                        : ready to intercept (traffic-agent already installed)
yyy                        : ready to intercept (traffic-agent not yet installed)
```

It happens due to this check:
https://github.com/telepresenceio/telepresence/blob/release/v2/pkg/client/connector/userd_k8s/k8s_config.go#L150
```golang
// ContextServiceAndFlagsEqual determines if this instance is equal to the given instance with respect to context,
// server, and flag arguments.
func (kf *Config) ContextServiceAndFlagsEqual(okf *Config) bool {
	return kf != nil && okf != nil &&
		kf.Context == okf.Context &&
		kf.Server == okf.Server &&
		sliceEqual(kf.flagArgs, okf.flagArgs)
} 
```
It makes sense because of the next case:
```
$ telepresence connect --kubeconfig my_kubeconfig.txt

# will use default ~/.kube/config file or $KUBECONFIG
$ telepresence list 
``` 
However, developer experience still suffers because of the lots of flags.

Possible solution to this is pushing those flags that were used in initial `connect` statement to all following CLI calls.
Removing flag check from the function above is optional. 

If it sounds good to you, I would love to try implementing it.

NurlashKO:
Another, and probably even better solution, is pushing those flags that were used in initial `connect` statement to all following cli calls.

thallgren:
I think part of the problem is that `telepresence list`, `telepresence intercept`, and other commands, can be invoked without first doing a `telepresence connect`, in which case the connect is implicit. Passing flags to the command becomes ambiguous if flags from the initial connect are also in effect. Should those flags be combined or overridden? And if it's the latter, then how do I override so that I don't have any flags (because that's also a valid flag combo)?

I think it would be acceptable to make it so that only `telepresence connect` accept kubernetes flags and those flags are in effect until a `telepresence quit`. An implicit connect would always be made _without_ the kubernetes flags. In other words, in order to use kubernetes flags, it's required to do an explicit connect. What do you think?

NurlashKO:
@thallgren Thank you for your comment!
[...]

PR #2096, opened 2021-09-20
Move K8S commands from global to connect command spec. 

NurlashKO:
## Description

This PR:
- Downgrades the scope of k8s flags from `global` to `connect` command only.
- Removes flags check form validation function that runs whenever we run `traffic` command like `list` or `intercept`.
This change fixes issue described here in detail: https://github.com/telepresenceio/telepresence/issues/2073

## Checklist


pkg/client/cli/cmd.go
```diff
@@ -136,37 +136,26 @@ func Command(ctx context.Context) *cobra.Command {
...
 		{
-			Name: "Kubernetes flags",
+			Name: "Telepresence networking flags",
 			Flags: func() *pflag.FlagSet {
-				kubeFlags = pflag.NewFlagSet("", 0)
-				kubeConfig = kates.NewConfigFlags(false)
-				kubeConfig.Namespace = nil // some of the subcommands, like "connect", don't take --namespace
-				kubeConfig.AddFlags(kubeFlags)
-				return kubeFlags
+				netflags := pflag.NewFlagSet("", 0)
```
thallgren, line 141: It would make a lot of sense to move this to from global to connect too, and thereby get rid of hte globalFlagGroups altogether. The daemon will not change DNS-settings as long as it remains connected and I can't see any use-case where the user would like to switch the mapped namespaces while connected.

pkg/client/cli/cmds_misc.go
```diff
@@ -57,6 +58,15 @@ func connectCommand() *cobra.Command {
 			})
 		},
 	}
+	flags := cmd.Flags()
+	flags.AddFlagSet(func() *pflag.FlagSet {
+		kubeFlags = pflag.NewFlagSet("", 0)
+		kubeConfig = kates.NewConfigFlags(false)
+		kubeConfig.Namespace = nil // some of the subcommands, like "connect", don't take --namespace
```
thallgren, line 65: The comment is obsolet given that this in fact is the connect command.

thallgren (changes requested):
This looks promising but I have some concerns/suggestions:
1. Move the other two global options to connect too.
2. There should be no global `kubeConfig` variable when the flags aren't global. Its flags should be confined to the `connect` command that sets them.
3. The `ClusterIdCommand()` in `cmds_misc.go` will need the k8s flags too (but not global).

PR #2102, opened 2021-09-21
alternative way to drain watchable's coalesce

### LINKED ISSUE #2084: [v2] traffic-manager high CPU usage after some undetermined amount of usage

josecv:

## Description

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -230,7 +230,7 @@ jobs:
             # points to a broken IP
             echo "127.0.0.1 metriton.datawire.io" | sudo tee -a /etc/hosts
             DTEST_KUBECONFIG="$HOME/kubeconfig" PATH="/c/:/c/Program Files/SSHFS-Win/bin:$PATH" DTEST_REGISTRY="docker.io/datawire" make test
-          no_output_timeout: 27m
+          no_output_timeout: &testsTimeout 27m
       - cleanup-kluster:
           platform: windows
       - save-logs
@@ -257,7 +257,7 @@ jobs:
           # default.  `go test` gives us helpful output when that
           # happens, CircleCI doesn't.  So lengthen CircleCI's timeout
           # just a bit, so `go test`'s timeout output isn't hidden.
-          no_output_timeout: 25m
+          no_output_timeout: *testsTimeout
       - cleanup-kluster:
           platform: macos
       - save-logs
@@ -284,7 +284,7 @@ jobs:
           # default.  `go test` gives us helpful output when that
           # happens, CircleCI doesn't.  So lengthen CircleCI's timeout
           # just a bit, so `go test`'s timeout output isn't hidden.
-          no_output_timeout: 22m
+          no_output_timeout: *testsTimeout
       - save-logs
       - cleanup-kluster:
           platform: linux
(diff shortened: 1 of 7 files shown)


.circleci/config.yml
```diff
@@ -230,7 +230,7 @@ jobs:
             # points to a broken IP
             echo "127.0.0.1 metriton.datawire.io" | sudo tee -a /etc/hosts
             DTEST_KUBECONFIG="$HOME/kubeconfig" PATH="/c/:/c/Program Files/SSHFS-Win/bin:$PATH" DTEST_REGISTRY="docker.io/datawire" make test
-          no_output_timeout: 27m
+          no_output_timeout: &testsTimeout 27m
```
donnyyung, line 233: I had no idea you could do this in circle.

josecv: it's not so much a circle thing as just a [yaml feature](https://blog.daemonl.com/2016/02/yaml.html)

donnyyung:
I had a question offline that Jose answered for me, and so just want to include it in the PR bc it explained the issue really well:
```
no so the thing is that if the context is cancelled we don't return from the function -- instead we call that shutdown thing which closes the upstream which eventually returns us from the function (via !readOk)
but shutdown has to acquire a lock before closing the upstream, and while it's doing that, the for loop will loop again -- but since the ctx is already done, it'll just call shutdown again (a no-op after the first call), and loop again, and call shutdown again, etc
so eventually that for loop is just spinning (edited) 
```

donnyyung (approved):
LGTM!

cmd/traffic/cmd/manager/internal/watchable/generic.tmpl.go
```diff
@@ -348,10 +348,14 @@ func (tm *MAPTYPE) coalesce(
...
+		done := ctx.Done()
+		if ctx.Err() != nil {
+			shutdown()
+			done = context.Background().Done()
+		}
 		if snapshot.State == nil {
 			select {
-			case <-ctx.Done():
-				shutdown()
+			case <-done:
```
thallgren, line 358: In spite of tests passing, I think there are two issues with this:
1. If `tm.close` is closed, there will still be a CPU consuming loop because that select will always be true until something else is randomly chosen (the <-upstream case).
2. If `ctx.Done()` is closed, then that should have higher precedence than `tm.close`. It should indeed just return.

thallgren (commented):
I know this is merged already, but I don't think it's quite right.

PR #2107, opened 2021-09-22
Njayp/webhook ensure agent

njayp:
## Description

Deployments with the webhook annotation that were deployed to the cluster before telepresence would timeout on intercept because their agent was never installed. This PR finds the pod from the deploy name and, if the agent is missing, rolls the pod to install the agent.

## Checklist


pkg/client/connector/userd_k8s/k8s_cluster.go
```diff
@@ -170,6 +171,24 @@ func (kc *Cluster) FindAgain(c context.Context, obj kates.Object) (kates.Object,
...
+	regex, err := regexp.Compile(`^` + name + `-[a-z0-9]+-[a-z0-9]+$`)
+	if err != nil {
+		return nil, err
+	}
+	for _, pod := range pods {
+		if regex.MatchString(pod.Name) {
+			return pod, nil
+		}
+	}
+	return nil, errors.NewNotFound(corev1.Resource("pod"), name+"."+namespace)
```
josecv: what about statefulsets? if a pod is called `my-sts-name-0` and the statefulset is called `my-sts-name`, the regex will not match

Instead of doing this, I suggest doing a label selector -- that should ensure things work regardless of controller (i.e. statefulset or deployment).
The workload object (i.e. `obj` in `ensureAgent`) should have a `spec.selector.matchLabels` object that you can use as a label selector to get the pods.

josecv (changes requested):

pkg/client/connector/userd_k8s/k8s_cluster.go
njayp: oops

njayp: obj seems to only have prebuilt functions rather than field access--there is no spec or getSpec. svc, however, has Spec, so I used Spec.Selector["service"] as the label selector. I do need you to look at it again to make sure im doing this right

josecv (approved):

pkg/client/connector/userd_k8s/k8s_cluster.go
```diff
@@ -170,6 +171,26 @@ func (kc *Cluster) FindAgain(c context.Context, obj kates.Object) (kates.Object,
...
+		return nil, err
+	}
+
+	for _, pod := range pods {
+		podLabels := pod.GetLabels()
+		// check if selector is in labels
+		for key, val := range selector {
+			if podLabels[key] == val {
+				return pod, nil
+			}
```
josecv: make sure _all_ the labels in the selector match, not just one

josecv (changes requested):

pkg/client/connector/userd_k8s/k8s_cluster.go
```diff
@@ -170,6 +171,30 @@ func (kc *Cluster) FindAgain(c context.Context, obj kates.Object) (kates.Object,
...
+		return nil, err
+	}
+
+	for _, pod := range pods {
+		podLabels := pod.GetLabels()
+		match := true
+		// check if selector is in labels
+		for key, val := range selector {
+			if podLabels[key] != val {
+				match = false
```
josecv, line 187:
```suggestion
				match = false
				break
```

josecv (approved):
feel free to merge after adding the `break`, i don't think i need to review this again :)

PR #2111, opened 2021-09-24
Misc fixes to get the CLI integration test more reliable

thallgren:
This PR contains a number of minor bug fixes to Telepresence, discovered while running the
`pkg/client/cli/telepresence_test.go` tests, and also a cleanup of that test to get more consistent
handling of the config.

I recommend reviewing one commit at a time.

### DIFF
--- a/pkg/client/cli/telepresence_test.go
+++ b/pkg/client/cli/telepresence_test.go
@@ -397,7 +397,7 @@ func (ts *telepresenceSuite) TestC_Uninstall() {
 			require.NoError(ts.kubectl(ctx, "delete", "svc,deploy", jobname))
 		}()
 
-		require.NoError(ts.kubectl(ctx, "rollout", "status", "-w", deployname, "-n", ts.namespace))
+		require.NoError(ts.rolloutStatusWait(ctx, deployname, ts.namespace))
 		stdout, stderr := telepresence(ts.T(), "list", "--namespace", ts.namespace, "--agents")
 		require.Empty(stderr)
 		require.Contains(stdout, jobname+": ready to intercept (traffic-agent already installed)")
@@ -408,12 +408,15 @@ func (ts *telepresenceSuite) TestC_Uninstall() {
 		require.Contains(stdout, "Root Daemon quitting... done")
 
 		// Double check webhook agent is uninstalled
-		require.NoError(ts.kubectl(ctx, "rollout", "status", "-w", deployname, "-n", ts.namespace))
-		stdout, err = ts.kubectlOut(ctx, "get", "pods", "-n", ts.namespace)
-		require.NoError(err)
-		match, err := regexp.MatchString(jobname+`-[a-z0-9]+-[a-z0-9]+\s+1/1\s+Running`, stdout)
-		require.NoError(err)
-		require.True(match)
+		require.NoError(ts.rolloutStatusWait(ctx, deployname, ts.namespace))
+		ts.Eventually(func() bool {
+			stdout, err = ts.kubectlOut(ctx, "get", "pods", "-n", ts.namespace)
+			if err != nil {
+				return false
+			}
+			match, err := regexp.MatchString(jobname+`-[a-z0-9]+-[a-z0-9]+\s+1/1\s+Running`, stdout)
+			return err == nil && match
+		}, 10*time.Second, 2*time.Second)
 
 		require.Eventually(
 			func() bool {
@@ -961,7 +964,7 @@ func (is *interceptedSuite) SetupSuite() {
 }
 				`)
 	is.NoError(err)
-	is.NoError(run(ctx, "kubectl", "rollout", "status", "-w", "deploy/hello-3", "-n", is.tpSuite.namespace))
+	is.NoError(is.tpSuite.rolloutStatusWait(ctx, "deploy/hello-3", is.tpSuite.namespace))
 
 	is.Run("all intercepts ready", func() {
 		rxs := make([]*regexp.Regexp, serviceCount)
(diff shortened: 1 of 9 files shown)


pkg/client/cli/telepresence_test.go
```diff
@@ -1834,3 +1761,55 @@ func telepresenceContext(ctx context.Context, args ...string) (string, string) {
...
+		t.Fatal(err)
+	}
+	c = client.WithEnv(c, env)
+
+	config := client.GetDefaultConfig(c)
+	config.LogLevels.UserDaemon = logrus.DebugLevel
+	config.LogLevels.RootDaemon = logrus.DebugLevel
+
+	to := &config.Timeouts
+	to.PrivateAgentInstall = 240 * time.Second
```
donnyyung, line 1793: If we have to bump all these timeouts in the tests, that makes me think maybe our default timeouts are too short?  While we need to overwrite config to test certain things, and some we need to overwrite for all tests (e.g. the registry), I think we should try to use as many defaults as possible in our tests so we are testing, as best as possible, what people get out of the box when they install Telepresence.  So perhaps bump the timeouts in `config.go` and remove setting the defaults here

thallgren: Not sure I agree. Our situation, where we use a sometimes extremely slow CI and not very fast cluster is, IMO out of the ordinary and this is a better safe than sorry timeout tailored for that use-case. For instance, bumping this default to 240s will leave the user hanging for four full minutes until telepresence gives up. That's quite exceptional and it's very likely to make the user unhappy when the problems are elsewhere (such as failure to pull the image, etc).

I also think that our approach of telling the user what times out on very slow systems is sufficient. It gives the user a chance to trim timeouts for a slow system without having a negative impact on all others.

donnyyung: yeah fair enough. Maybe it's worth trying to use more performant clusters in our CI in the future. My main concern is wanting to make sure our timeouts are good on average because we don't want a user's first use of telepresence to result in a timeout since that's not the best UX imo. But that's not exactly a problem for this PR and that should probably be a metric's based decision and not one based on a "hunch" which is all I have at this point.

thallgren: Totally agree with that. We should collect metrics for how long various things take. Perhaps we do already?

pkg/client/daemon/dns/flush_darwin.go
```diff
@@ -2,15 +2,58 @@ package dns
...
 	// As of macOS 11 (Big Sur), how to flush the DNS cache hasn't changed since 10.10.4 (a Yosemite version release in mid 2015),
 	// other than that in 10.12 (Sierra) it became necessary to also kill mDNSResponderHelper. On older versions the call to kill
 	// mDNSResponderHelper is unnecessary but harmless, as the process doesn't exist.
-	_ = dexec.CommandContext(c, "killall", "-HUP", "mDNSResponder").Run()
-	_ = dexec.CommandContext(c, "killall", "mDNSResponderHelper").Run()
-	_ = dexec.CommandContext(c, "dscacheutil", "-flushcache").Run()
+	dlog.Debug(c, "Flushing DNS")
+	if majorProductVersion >= 11 {
+		// Big Sur, a killall -HUP mDNSResponder doesn't restart the mDNSResponderHelper but killall -TERM of both, does`
+		dlog.Debug(c, "Flushing DNS")
```
donnyyung, line 50: Either change to "Flushing DNS >= Big Sur" or delete this one since we have the same log at line 48.

pkg/client/daemon/dns/flush_darwin.go
```diff
@@ -2,15 +2,58 @@ package dns
...
+		devNullCmd(c, "killall", "-HUP", "mDNSResponder")
```
donnyyung, line 51: Is this supposed to be "killall", "-TERM" based on the comment? or am I misunderstanding

josecv: wait i'm confused, maybe i'm reading this wrong but it sounds like the comment doesn't match the code -- should we be doing `killall -TERM` instead of `-HUP` here?

thallgren: Oh, that comment is wrong. Need to fix.

thallgren: The comment is wrong. Will fix.

pkg/client/connector/userd_grpc/service.go
```diff
@@ -77,39 +77,54 @@ func (s *service) Version(_ context.Context, _ *empty.Empty) (*common.VersionInf
...
 func (s *service) CreateIntercept(c context.Context, ir *rpc.CreateInterceptRequest) (result *rpc.InterceptResult, err error) {
+	c = s.callCtx(c, "CreateIntercept")
+	dlog.Debug(c, "called")
 	result = s.callbacks.InterceptStatus()
 	if result != nil {
+		dlog.Debug(c, "returned")
 		return result, nil
 	}
-	c = s.callCtx(c, "CreateIntercept")
+	dlog.Debug(c, "called")
```
donnyyung: I think this is an extra "called" log?

pkg/client/cli/telepresence_test.go
donnyyung: I don't think we do, but it should be easy enough to sent a report every time we timeout w/ what the timeout was and what its value was.. so we can track that :) I'll make an issue just so we don't lose track of it.

donnyyung (approved):

PR #2118, opened 2021-09-25
Fix bug causing segfault when intercepts collide on port number.

thallgren:
## Description

The `LOCAL_TARGET_IN_USE` intercept error didn't provide the expectd
`InterceptInfo` struct which caused a segfault panic. Also, the
name reported was the name of the current intercept, which is kind of
useless, instead of the conflicting one.

## Checklist

### DIFF
--- a/pkg/client/daemon/dns/flush_darwin.go
+++ b/pkg/client/daemon/dns/flush_darwin.go
@@ -45,14 +45,15 @@ func Flush(c context.Context) {
 	// As of macOS 11 (Big Sur), how to flush the DNS cache hasn't changed since 10.10.4 (a Yosemite version release in mid 2015),
 	// other than that in 10.12 (Sierra) it became necessary to also kill mDNSResponderHelper. On older versions the call to kill
 	// mDNSResponderHelper is unnecessary but harmless, as the process doesn't exist.
+	sig := "-HUP"
 	if majorProductVersion >= 11 {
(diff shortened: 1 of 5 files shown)


pkg/client/daemon/dns/flush_darwin.go
```diff
@@ -45,14 +45,15 @@ func Flush(c context.Context) {
 	// As of macOS 11 (Big Sur), how to flush the DNS cache hasn't changed since 10.10.4 (a Yosemite version release in mid 2015),
 	// other than that in 10.12 (Sierra) it became necessary to also kill mDNSResponderHelper. On older versions the call to kill
 	// mDNSResponderHelper is unnecessary but harmless, as the process doesn't exist.
+	sig := "-HUP"
 	if majorProductVersion >= 11 {
```
donnyyung, line 49: re: the commit message, should we bump the macos version used in our CI tests to use an image that is using Big Sur?

donnyyung (approved):
LGTM

pkg/client/daemon/dns/flush_darwin.go
thallgren: I thought about that, but perhaps that would be a mistake. I think we all use Big Sur on our laptops so it does get a lot of test coverage. Older versions will not get any if we bump the CI version.

Ideally, we should test both in CI, but perhaps that can wait until we move to GH actions?

PR #2119, opened 2021-09-27
Gather-logs enhancements 

donnyyung:
## Description
Currently a WIP, only added the PR so I could test some circle changes

This adds a few features to gather-logs:
1. The ability to anonymize logs pod names + namespaces from the logs you submit (we'll work on anonymizing even more things in the future)
2. Adding a flag to gather-logs that will get the pod yaml for the traffic-manager + pods that contain a traffic-agent... this can be very helpful in debugging

## Checklist


pkg/client/cli/cmd_gatherlogs.go
```diff
@@ -119,6 +139,7 @@ func (gl *gatherLogsArgs) gatherLogs(ctx context.Context, cmd *cobra.Command, st
 	scout.SetMetadatum("daemon_logs", daemonLogs)
 	scout.SetMetadatum("traffic_manager_logs", gl.trafficManager)
 	scout.SetMetadatum("traffic_agent_logs", gl.trafficAgents)
+	scout.SetMetadatum("get_pod_yaml", gl.podYaml)
```
josecv, line 143: let's add a metadatum for the anonymize flag too, just for completeness' sake

pkg/client/cli/cmd_gatherlogs.go
```diff
@@ -203,6 +226,40 @@ func (gl *gatherLogsArgs) gatherLogs(ctx context.Context, cmd *cobra.Command, st
...
+// manager and writing its components into logical files.
+func writeResponseToFiles(lr *manager.LogsResponse, anonymizer *anonymizer, exportDir string, anonymize bool) error {
+	createFileWithContent := func(fileName, content string) error {
+		fd, err := os.Create(fileName)
+		if err != nil {
+			return err
+		}
+		defer fd.Close()
+		fdWriter := bufio.NewWriter(fd)
+		_, _ = fdWriter.WriteString(content)
```
josecv: should we be throwing out this error? might make sense to at least report it in some capacity (unless it can't happen?)

pkg/client/cli/cmd_gatherlogs.go
```diff
@@ -260,12 +317,122 @@ func zipFiles(files []string, zipFileName string) error {
...
+	// If this pod name has already been mapped, return that
+	if anonName, ok := anonymizer.podNames[podName]; ok {
+		return anonName
+	}
+
+	// the podName hasn't been anonymized yet so we split it up
+	// so we can anonymize the namespace
+	nameComponents := strings.SplitN(podName, ".", 2)
+	if len(nameComponents) != 2 {
+		unknownPodName := "anonPod.anonNamespace"
```
josecv: hmm could we give this an ordinal based on `len(anonymizer.podNames)` too? im worried that if for whatever  reason this happens to one pod, it will happen to  all of em and we won't be able to distinguish them

pkg/client/cli/cmd_gatherlogs.go
```diff
@@ -260,12 +317,122 @@ func zipFiles(files []string, zipFileName string) error {
...
+		// Strip the namespace off of the podName
+		podParts := strings.Split(fullPodName, ".")
+
+		// replace pod name + full hash with the anonymized name
+		podName := podParts[0]
+		stringContent = strings.ReplaceAll(stringContent, podName, anonPodName)
+
+		// Strip the hash off of the pod name + replace remainder with anonymized name
+		brokenPodName := strings.Split(podName, "-")
+		podName = strings.Join(brokenPodName[:len(brokenPodName)-2], "-")
```
josecv: watch out cause this is not going to work for statefulsets -- if a statefulset is called `my-sts` the pods will be `my-sts-0`, `my-sts-1`, etc -- you might need a regex of some sort to check what kind of pod it is

this will similarly have trouble for pods created from Jobs which also have their own format IIRC

pkg/client/cli/cmd_gatherlogs.go
```diff
@@ -260,12 +317,122 @@ func zipFiles(files []string, zipFileName string) error {
...
+		anonPodName := anonPodParts[0]
+
+		// Strip the namespace off of the podName
+		podParts := strings.Split(fullPodName, ".")
+
+		// replace pod name + full hash with the anonymized name
+		podName := podParts[0]
+		stringContent = strings.ReplaceAll(stringContent, podName, anonPodName)
+
+		// Strip the hash off of the pod name + replace remainder with anonymized name
```
josecv: wait im confused why do we do both of these things? it seems either/or would be sufficient?

josecv (changes requested):
v nice work overall just some concerns -- also i think this is missing the timestamp change (unless im dumb and i missed it which is always possible) so i can re-review when that is added 🙂

donnyyung:
I had gotten called into the other flurry of alerts before I could add the timestamp fixes :( will add that first and then go through your comments, thanks Jose!

donnyyung:
Fixes https://github.com/telepresenceio/telepresence/issues/2126

pkg/client/cli/cmd_gatherlogs.go
donnyyung: Oh whoops that was my bad. I error checked in the other place I used this, not sure why I didn't here. will fix

pkg/client/cli/cmd_gatherlogs.go
donnyyung: Ah great catch. Will do.

pkg/client/cli/cmd_gatherlogs.go
donnyyung: Ah I should better describe this in the logs.. So the problem was in the traffic-manager logs we might have something like this:
```
2021-10-01 18:45:36.8322 info    agent-injector : Injecting traffic-agent into pod echo-auto-inject-869ff95bc7-      .default
```
but in the traffic-agent we have logs like this:
```
2021-10-01 18:45:38.3415 info    {Name:echo-auto-inject Namespace:default PodIP:10.42.0.35 AgentPort:9900 AppMo      unts:/tel_app_mounts AppPort:8080 ManagerHost:traffic-manager.ambassador ManagerPort:8081}
```

So I could just ReplaceAll without the hash, but then the first log would look like:
```
2021-10-01 18:45:36.8322 info    agent-injector : Injecting traffic-agent into pod pod-1-869ff95bc7-.namespace-2:
```

Although I just noticed my code wouldn't anonymize this bc it doesn't include the 2nd part of the hash...

Hm I think I have an idea that might remedy this + the issue with statefulsets

pkg/client/cli/cmd_gatherlogs.go
donnyyung: ah yeah let me think about that... I don't think we need to worry about Jobs (yet) since we can't intercept those (afaik), but it won't hurt to be forward thinking if it's not too difficult.

pkg/client/cli/cmd_gatherlogs.go
```diff
@@ -239,7 +242,10 @@ func writeResponseToFiles(lr *manager.LogsResponse, anonymizer *anonymizer, expo
 		}
 		defer fd.Close()
 		fdWriter := bufio.NewWriter(fd)
-		_, _ = fdWriter.WriteString(content)
+		_, err = fdWriter.WriteString(content)
+		if err != nil {
+			return nil
```
josecv: should this return err?

k8s/echo-easy.yaml
```diff
@@ -2,7 +2,7 @@
 apiVersion: v1
 kind: Service
 metadata:
-  name: echo-easy
+  name: "a123f"
```
josecv: hmm is this an accident? it looks a little like you were testing the `getSignificantPodNames` locally; it would probably be a good idea to actually add this to the unit tests

pkg/client/cli/cmd_gatherlogs.go
```diff
@@ -455,3 +463,43 @@ func anonymizeLog(stdout io.Writer, logFile string, anonymizer *anonymizer) erro
...
+		// Indicates we found a Pod that comes from a StatefulSet
+		if _, err := strconv.Atoi(component); err == nil {
+			significantComponent = true
+		}
+
+		if significantComponent {
+			nameComponents = nameComponents[:i+1]
+			break
+		}
+	}
```
josecv: hmm im not sure im sold -- what if i have a deployment called `hello-0`? i almost wanna say we should loop in reverse; what about something like this:

```go
lastSignificant := false
subNames := []string{}
for i := len(nameComponents)-1; i >= 0; i-- {
                component := nameComponents[i]
                significant := false
                // Indicates we found a Pod that comes from either a
                // ReplicaSet and/or a Deployment
                if _, err := hex.DecodeString(component); err == nil {
                    significant = true
                }
                // Indicates we found a Pod that comes from a StatefulSet
                if _, err := strconv.Atoi(component); err == nil && !lastSignificant {
                    significant = true
                }
                
                if !significant && lastSignificant {
                     // my-sts or my-deploy
                     subNames = append(subNames, strings.Join(nameComponents[:i], "-"))
                     break
                } else if significant {
                     // my-sts-0 or my-deploy-asdf-asdf or my-deploy-asdf
                     subNames = append(subNames, strings.Join(nameComponents[:i], "-"))
                     lastSignificant = true
                }
}
```

although now that i look at that i think it won't work for a deployment called `a123f` 🤔 it might be easier to just do this with regexes 😅 should be easy to make one for statefulsets, one for deployments, and one for replicasets and see which is matched

josecv (changes requested):

pkg/client/cli/cmd_gatherlogs.go
donnyyung: 🤦‍♂️

pkg/client/cli/cmd_gatherlogs.go
donnyyung: yeah there wasn't really any way that I was super happy with. I will try to think of a regexes.

pkg/client/cli/cmd_gatherlogs.go
```diff
@@ -469,37 +468,42 @@ func anonymizeLog(stdout io.Writer, logFile string, anonymizer *anonymizer) erro
...
+	// if the pods ends in an ordinal we can be pretty sure it's
+	// coming from a StatefulSet.
+	statefulSetRegex := regexp.MustCompile("(?P<appName>.*)-([0-9]+)$")
+	// ReplicasSets, and therefore Deployments because they create
+	// ReplicaSets, have a hash followed by a 5 character identity
+	// string attached to the end.
+	replicaSetRegex := regexp.MustCompile("(?P<appName>.*)-(?P<hash>[0-9a-f]+)-([0-9a-z]{5})$")
+	sigNames := []string{}
+	switch {
+	case statefulSetRegex.MatchString(podName):
```
donnyyung, line 480: this seems ridiculously complicated to get matches but I couldn't find another way to do it?? If another way exists, I'd love to know.

pkg/client/cli/cmd_gatherlogs.go
```diff
@@ -469,37 +468,29 @@ func anonymizeLog(stdout io.Writer, logFile string, anonymizer *anonymizer) erro
...
+		match := statefulSetRegex.FindStringSubmatch(podName)
+		appName := match[1]
+		// Add the pod name with and without the ordinal
+		sigNames = append(sigNames, podName, appName)
+	case replicaSetRegex.MatchString(podName):
+		match := replicaSetRegex.FindStringSubmatch(podName)
+		appName := match[1]
+		rsName := fmt.Sprintf("%s-%s", appName, match[2])
+		// add the app name with and without generated ReplicaSet hash
+		sigNames = append(sigNames, rsName, appName)
```
josecv: should we also append podName here?

josecv (approved):
lgtm with just one minor issue

pkg/client/cli/cmd_gatherlogs.go
donnyyung: I hadn't seen the full pod name show up in any logs, but it can't hurt to be extra thorough, will do :)

PR #2123, opened 2021-09-29
Check if k8s version is supported

donnyyung:
## Description

We have known that we don't support k8s versions older than 1.17, so let's let users know.

## Checklist

### DIFF
--- a/cmd/traffic/cmd/manager/internal/cluster/info.go
+++ b/cmd/traffic/cmd/manager/internal/cluster/info.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/blang/semver"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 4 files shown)


cmd/traffic/cmd/manager/internal/cluster/info.go
```diff
@@ -53,8 +56,29 @@ func NewInfo(ctx context.Context) Info {
 	oi := info{}
 	oi.waiter.L = &oi.accLock
 	clientset := managerutil.GetK8sClientset(ctx)
-	client := clientset.CoreV1()
 
+	// Validate that the kubernetes server version is supported
+	dc := clientset.Discovery()
+	info, err := dc.ServerVersion()
+	if err != nil {
+		dlog.Errorf(ctx, "unable to get server information")
```
josecv: let's log `err` itself here

josecv (approved):
looks good, just one minor logging nit

PR #2125, opened 2021-10-01
Use one tunnel per connection instead of multiplexing them into one tunnel.

thallgren:
## Description

This PR is a refactoring of the Telepresence tunneling framework. It's main objective is to increase performance by deprecating the old solution which multiplexed all connections over the same tunnel and instead use one tunnel per connection.

I found several very minor issues during the work to switch this over, and rather than creating one PR for each little thing I found, I decided to create a commit here even if some of the them aren't directly related to the tunnels.

I recommend reviewing one commit at a time and to use the "ignore white space" mode when reviewing.

## Checklist


pkg/tunnel/pool.go
```diff
@@ -10,56 +10,59 @@ import (
...
-	lock sync.Mutex
+	lock sync.RWMutex
 }
 
 type Handler interface {
 	// Close closes the handle
 	Close(context.Context)
 
-	Start(ctx context.Context)
+	Start(ctx context.Context) error
```
donnyyung: Might just be where I'm at in the commit-by-commit review (abd2681c328f745f22df107eaaf791152b6e0b64), but you added error as an output and I don't currently see any `errors` returned in the start functions edited in this commit... so not sure of the purpose.

cmd/traffic/cmd/agent/client.go
```diff
@@ -157,7 +159,7 @@ func lookupHostWaitLoop(ctx context.Context, manager rpc.ManagerClient, session
 	for ctx.Err() == nil {
 		lr, err := lookupHostStream.Recv()
 		if err != nil {
-			if ctx.Err() == nil {
+			if ctx.Err() == nil && !errors.Is(err, io.EOF) {
 				dlog.Debugf(ctx, "lookup request stream recv: %+v", err) // May be io.EOF
```
josecv: we should probably update the `May be io.EOF` comment if it won't be io.EOF

pkg/tunnel/pool.go
```diff
@@ -69,25 +72,29 @@ func (p *Pool) GetOrCreate(ctx context.Context, id ConnID, createHandler Handler
...
 	if ok {
 		// Toss newly created handler. It's not started anyway.
 		return old, true, nil
 	}
-	handler.Start(handlerCtx)
-	dlog.Debugf(ctx, "++ POOL %s, count now is %d", id, len(p.handlers))
+	if err = handler.Start(handlerCtx); err != nil {
+		release()
+		return nil, false, err
+	}
```
josecv: should we remove it from `p.handlers` if it fails to start?

pkg/client/daemon/outbound.go
```diff
@@ -48,7 +48,7 @@ type outbound struct {
 
 	dnsConfig *rpc.DNSConfig
 
-	scout chan scout.ScoutReport
+	scout chan<- scout.ScoutReport
```
josecv, line 51: TIL about `<-chan` and `chan<-`

pkg/client/daemon/tunrouter.go
```diff
@@ -452,7 +447,30 @@ func (t *tunRouter) handlePacket(c context.Context, data *buffer.Data) {
...
+		dlog.Debugf(c, "Unhandled protocol %d", ipHdr.L4Protocol())
+		reply(icmp.DestinationUnreachablePacket(ipHdr, icmp.ProtocolUnreachable))
+	}
+}
+
+type vifWriter struct {
+	*vif.Device
+}
+
+func (w vifWriter) Write(ctx context.Context, pkt ip.Packet) (err error) {
```
josecv, line 485: does this need to softRelease the packet?

pkg/log/testlogger.go
```diff
@@ -0,0 +1,66 @@
...
+
+		for i, k := range parts {
+			if i > 0 {
+				sb.WriteString(" ")
+			}
+			fmt.Fprintf(&sb, "%s=%#v", k, w.fields[k])
+		}
+	}
+	w.TB.Log(sb.String())
+}
```
josecv, line 66: i don't really have an issue with this, but the weird thing is that the dlib test logger should add timestamps https://github.com/datawire/dlib/blob/master/dlog/logger_testing.go#L71-L73 -- maybe we should fix it there? there seems to be some bug where some log lines get timestamped and others do not

josecv (commented):
so im not yet done reviewing this, but i wanna post my comments so far (i'm up to `aeb2c353a6574f03b853e1681468fd181663c11e`) so that you can take a look at em since i  won't be here on monday

pkg/log/testlogger.go
thallgren: The dlib testlogger uses fields and escaped quoted strings (lots of `\"`) which makes the timestamps unaligned and the lines hard to read even if we fix the lack-of-timestamp problem. The `log/testlogger.go` is backported from the integration test refactoring that I'm working on and what I'm after is something more similar to other logs that Telepresence outputs. I'm all for suggesting that the dlog test-logger would do the same, but I also think it might be good to show what it would look like.

pkg/tunnel/pool.go
thallgren: We should, and that's exactly what the call to `release()` does.

pkg/client/daemon/tunrouter.go
thallgren: No, it's supposedly taken care of by the caller. But thanks for the reminder. The `Packet.SoftRelease()` should be removed completely since it's no longer needed.

pkg/tunnel/pool.go
thallgren: You're right. There's no need for an error return.

pkg/tunnel/pool.go
thallgren: Moot point now, since `Start()` no longer returns error.

donnyyung:
Mostly some tiny nits, want to do some manual testing tomorrow and may do a few more sanity checks, there's still some gaps in my knowledge (around networking concepts, not anything specific in your PR) that I'd like to fill in.

pkg/client/daemon/tunrouter.go
```diff
@@ -507,11 +527,35 @@ func (t *tunRouter) udp(c context.Context, dg udp.Datagram) {
...
 	if err != nil {
 		dlog.Error(c, err)
 		return
 	}
 	uh.(udp.DatagramHandler).HandleDatagram(c, dg)
 }
+
+func (t *tunRouter) maybeOpenStream(c context.Context, id tunnel.ConnID) (tunnel.Stream, error) {
+	if t.muxTunnel != nil {
+		// tunnelVersion <= 2, so use the multiplexing tunnel
```
josecv: should this  be `< 2`?

pkg/vif/tcp/handler.go
```diff
@@ -342,31 +331,28 @@ func (h *handler) processPayload(ctx context.Context, data []byte) {
 	n := len(data)
 	for n > start {
 		h.sendLock.Lock()
-		window := h.sendWindow
-		aqz := h.ackWaitQueueSize
-		for window == 0 || aqz > maxAckWaits {
-			if window == 0 {
+		window := int(h.peerWindow) - int(h.sequence()-h.seqAcked)
+		for window <= 0 {
+			if window <= 0 {
```
josecv: do we need the if? there's no more `||` in the loop invariant so it seems we could just directly log

josecv (approved):
woo i think this lgtm now

i really like the new tunnel semantics, i feel like the interfaces are a whole lot cleaner than the previous package -- what are your thoughts around getting rid of the deprecated muxTunnel? It seems there's a whole lot of places that check for it, so i think it might be good to have a plan for eventually getting rid of it, esp if we need to be adding new features to it

thallgren:
@josecv I *really want* to get rid of the deprecated muxTunnel a.s.a.p. But we must provide a window of backward compatibility for some reasonable time. The fact that an explicit Helm install has precedence, combined with users that act on the "there's a new version available" message, will probably make it common to have a 2.4.5 talking to an older manager.

I think we should inform the user about the deprecation and when we plan to end support for it for some, yet to be determined, period of time, before we remove it.

PR #2146, opened 2021-10-07
Directly mount volumes, obviating TELEPRESENCE_ROOT when using --docker-run

AaronFriel:
## Description

When using `--docker-run`, volumes are mounted directly to the correct paths, obviating the need to use `$TELEPRESENCE_ROOT` and enabling Telepresence to work with containers that use volume mounts without modifying application code.

Caveats:

* This should be tested, but I am unfamiliar with the test harness.
* Paths with a colon in them will pose a difficulty. Ideally rather than parsing the agent-provided env var, a list of paths would be provided as a list.

Example output from an intercepted container:

```
root@b05aba634a8b:/app# ls /var/run/secrets/
kubernetes.io
```

## Checklist

 - [X] I made sure to update `./CHANGELOG.md`.
 - [X] I made sure to either submit a docs PR, or tell Matt about the necessary documentation changes.
 - [X] I updated `DEVELOPING.md` with any any special dev tricks I had to use to work on this code efficiently.
 - [X] I updated `TELEMETRY.md` if I added, changed, or removed a metric name.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+### 2.4.6 (TBD)
+
+- Feature: When using `--docker-run`, volumes are mounted directly to the correct paths, obviating
+  the need to use `$TELEPRESENCE_ROOT` and enabling Telepresence to work with containers that use
+  volume mounts without modifying application code.
+
 ### 2.4.5 (October 15, 2021)
 
 - Feature: Intercepting headless services is now supported. It's now possible to request a headless service on whatever port it exposes and get a response from the intercept.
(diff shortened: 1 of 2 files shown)


thallgren (approved):
Thanks. This looks like simple and very useful addition.

AaronFriel:
@thallgren thanks! It's failing lint and probably needs some unit tests. Before looking at that, what do you think the best way to address the issue of `:` characters in paths is? I'm unfamiliar with the agent and it wasn't obvious to me how the `TELEPRESENCE_MOUNTS` var is set, is it computed by the agent and then sent to the CLI via RPC? If so, is there a way to add an array/list of volume mounts?

AaronFriel:
I've modified the MR to be ready for review, it would be pretty nifty to merge this if someone could validate that it works on the Mac CLI.

AaronFriel:
@thallgren I fixed the lint, rebased on the upstream, and so on, but the lint job appears to be stuck on installing shellcheck, which works locally. Would you be able to take this PR from here?

lloydchang (approved):

thallgren:
Superseded by #3130

PR #2148, opened 2021-10-08
Don't set the GID on the traffic agent unless it's necessary

josecv:
## Description

## Checklist


thallgren (commented):
Does this mean that we cannot ever use the iptables approach in OpenShift clusters? If so, is there some other way to than using a GID to identify the sender in the rules? Can we let the initContainer detect a randomly assigned GID?

thallgren (approved):

PR #2155, opened 2021-10-13
Fixes for some minor problems and instabilities discovered during testing.

thallgren:
Details in the commits. None of this is relevant to add to the changelog since they don't really affect any released behavior.


donnyyung (approved):
seems sane and LGTM!

PR #2159, opened 2021-10-15
Remove two bad "See logs for details" advices

thallgren:
Telepresence shouldn't print the "See logs for details" message for normal error conditions such as "user not logged in" or "no workload found with an installed  agent". All relevant info has already been conveyed to the user.


donnyyung (approved):
LGTM!

efunk:
Fixed and released in Telepresence 2.4.6 GA

PR #2164, opened 2021-10-20
Prevent requests to the API server from ever being proxied

josecv:
## Description

When the API server happens to be in the same subnet as the pods, we should not proxy traffic to it; this is because we need to be able to talk to the API server to proxy traffic -- attempting to proxy to it will lead to and infinite loop.

## Checklist


pkg/client/daemon/tunrouter.go
```diff
@@ -220,6 +269,18 @@ func (t *tunRouter) setOutboundInfo(ctx context.Context, mi *daemon.OutboundInfo
...
 			}
 		}
+		if len(mi.NeverProxySubnets) > 0 {
+			t.neverProxySubnets = make([]routing.Route, len(mi.NeverProxySubnets))
+			for i, np := range mi.NeverProxySubnets {
+				npSn := iputil.IPNetFromRPC(np)
+				dlog.Infof(ctx, "Adding never-proxy subnet %s", npSn)
+				route, err := routing.GetRoute(ctx, npSn)
+				if err != nil {
+					return fmt.Errorf("unable to get route for %s: %w", iputil.IPNetFromRPC(np), err)
```
josecv: question for the reviewers: do you think it's a good idea to return an error here? i sorta went back and forth between returning the error and just logging it and moving on.
I guess the risk is that if for whatever reason we fail to parse the routes for the API server we might fail `telepresence connect` --even if in the end it turns out there was no need to actually add a static route for the API server.

pkg/client/daemon/tunrouter.go
```diff
@@ -248,56 +250,68 @@ func (t *tunRouter) setOutboundInfo(ctx context.Context, mi *daemon.OutboundInfo
...
-			close(cfgComplete)
-			cfgComplete = nil
+		}
+		dtime.SleepWithContext(ctx, backoff)
+		backoff *= 2
+		if backoff > 3*time.Second {
+			backoff = 3 * time.Second
 		}
 	}
+	return nil
```
donnyyung: If we're no longer returning an error, does it make sense to have the function not have a return value at all?

pkg/client/connector/userd_trafficmgr/traffic_manager.go
```diff
@@ -484,9 +485,36 @@ func (tm *trafficManager) Uninstall(c context.Context, ur *rpc.UninstallRequest)
 }
 
 // getClusterCIDRs finds the service CIDR and the pod CIDRs of all nodes in the cluster
-func (tm *trafficManager) getOutboundInfo() *daemon.OutboundInfo {
+func (tm *trafficManager) getOutboundInfo(ctx context.Context) *daemon.OutboundInfo {
+	neverProxy := []*manager.IPNet{}
```
donnyyung, line 494: nit: I would add a comment here or below that notes that what we are doing is "preventing requests to the api server from being proxied". I know the code here is pretty self-explanatory but just for future readers of the code, why that could be mad isn't potentially obvious (especially since it took us so long to figure it out) so isn't something we want to get lost in a future refactor.

pkg/client/daemon/tunrouter.go
donnyyung: I think if there's a case where we don't get the routes for the API server for whatever reason and it still works, I think we should _just_ log the error then.  But make sure the error is verbose and states what went wrong ("couldn't get the route for the apiserver so if that ip falls within this subnet, things might not work" obviously better than that but you get my drift.

I do also wonder the way we have AlsoProxy that a user can specify manually, so should we add something like "NeverProxy" as well.. and you could throw that in the error message like, "if you know the IP of your API server, you can add this to your kubeconfig extension"... sigh It's getting wordy at this point but I guess what do you think about these two things?

pkg/client/cli/telepresence_test.go
```diff
@@ -333,6 +334,57 @@ func (ts *telepresenceSuite) TestA_WithNoDaemonRunning() {
...
+		}}
+
+		require.NoError(clientcmd.WriteToFile(*cfg, kubeconfigFileName), "unable to write modified kubeconfig")
+
+		ctx := testContext(t)
+		defer os.Setenv("KUBECONFIG", origKubeconfigFileName)
+		os.Setenv("KUBECONFIG", kubeconfigFileName)
+
+		_, stderr := telepresenceContext(ctx, "connect")
+		require.Empty(stderr)
```
donnyyung, line 381: This might be extra but should we add validation here that `telepresence status` contains what we expect in the `Also Proxy` line to ensure that things were configured / operating correctly?

pkg/client/daemon/tunrouter.go
josecv: > I think if there's a case where we don't get the routes for the API server for whatever reason and it still works, I think we should just log the error then. But make sure the error is verbose and states what went wrong ("couldn't get the route for the apiserver so if that ip falls within this subnet, things might not work" obviously better than that but you get my drift.

yea i think that makes sense, especially given if we've come to this point we're able to connect to the API server at least

> I do also wonder the way we have AlsoProxy that a user can specify manually, so should we add something like "NeverProxy" as well

i definitely think this makes sense, but i don't wanna add it on to this PR, especially because...

> and you could throw that in the error message like, "if you know the IP of your API server, you can add this to your kubeconfig extension"

i'm not sure how helpful this would be; let's say the user manually does an `nslookup` for their API server, and adds its IP address to the NeverProxy config -- wouldn't it just be the very IP address that the connector daemon reported, and that we were unable to discover a route for here? i feel like we'd just fail to find it a second time and that could be frustrating. Maybe the action item is as simple as "open a bug and let's investigate why we can't find a route to an API server that you're obviously able to connect to"
i do think exposing the NeverProxy config would be a good feature for different reasons, e.g. overlapping VPNs and podCIDRs

donnyyung: Yeah I like the idea about saying, "okay please open an issue cuz we need to figure out what is going wrong in your specific cluster" which is really waht we need in that scenario. And yeah agreed the `NeverProxy` shouldn't be bundled into this PR, just an idea for a future improvement I thought of :)

donnyyung (approved):
LGTM as long as tests pass!

josecv:
I've seen the windows tests pass a bunch of times on this branch, and the failure looks to be in install code, which this PR doesn't touch at all, so I think the failure is a flake; i'm gonna merge this 🤞

PR #2166, opened 2021-10-20
publish arm64 builds mac

donnyyung:

## Description

Build + publish arm64 builds for mac when releasing Telepresence.

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -342,6 +342,7 @@ jobs:
           command: |
             curl -k https://curl.se/ca/cacert.pem -o ~/.cacert.pem
             export CURL_CA_BUNDLE=~/.cacert.pem
+            export AWS_CA_BUNDLE=~/.cacert.pem
       - run: TELEPRESENCE_VERSION=$CIRCLE_TAG make push-executable
       - run: |
           [[ $CIRCLE_TAG == *-* ]] || TELEPRESENCE_VERSION=$CIRCLE_TAG make promote-to-stable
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -342,9 +342,23 @@ jobs:
           command: |
             curl -k https://curl.se/ca/cacert.pem -o ~/.cacert.pem
             export CURL_CA_BUNDLE=~/.cacert.pem
-      - run: TELEPRESENCE_VERSION=$CIRCLE_TAG make push-executable
-      - run: |
-          [[ $CIRCLE_TAG == *-* ]] || TELEPRESENCE_VERSION=$CIRCLE_TAG make promote-to-stable
+            export AWS_CA_BUNDLE=~/.cacert.pem
```
donnyyung, line 345: I think this _should_ work but if it doesn't I think we've exhausted everything we can try and will have to bump our mac executor :(

pkg/client/cli/telepresence_test.go
```diff
@@ -924,16 +924,22 @@ func (cs *connectedSuite) TestN_ToPodPortForwarding() {
...
 	b := make([]byte, 1024*1024*5)
 	b[0] = '!'
 	b[1] = '\n'
 	for i := range b[2:] {
 		b[i+2] = 'A'
 	}
-	req, err := http.NewRequest(http.MethodPut, fmt.Sprintf("http://hello-0.%s/put", cs.ns()), bytes.NewBuffer(b))
-	require.NoError(err)
 
+	// Ensure the service is accessible before we do the put request
```
donnyyung, line 936: For some reason this is where tests would fail frequently on my Apple silicon mac. The first request would fail but then future ones would succeed but by that time the test had failed, so this seems like an okay bandaid until Thomas gets back and we have the complete test refactor?

josecv (approved):

PR #2171, opened 2021-10-25
Prevent a panic when a route fails to be found.

josecv:

## Description

## Checklist

### DIFF
--- a/pkg/client/daemon/tunrouter.go
+++ b/pkg/client/daemon/tunrouter.go
@@ -270,18 +270,19 @@ func (t *tunRouter) setOutboundInfo(ctx context.Context, mi *daemon.OutboundInfo
 			}
 		}
 		if len(mi.NeverProxySubnets) > 0 {
-			t.neverProxySubnets = make([]routing.Route, len(mi.NeverProxySubnets))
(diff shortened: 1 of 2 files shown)


pkg/client/daemon/tunrouter.go
```diff
@@ -270,18 +270,19 @@ func (t *tunRouter) setOutboundInfo(ctx context.Context, mi *daemon.OutboundInfo
 			}
 		}
 		if len(mi.NeverProxySubnets) > 0 {
-			t.neverProxySubnets = make([]routing.Route, len(mi.NeverProxySubnets))
-			for i, np := range mi.NeverProxySubnets {
```
donnyyung, line 274: Was this panicking bc we'd make a slice of some known size but if we didn't get a route then we wouldn't put anything in said slice and then we'd panic upon accessing it?

donnyyung (approved):

pkg/client/daemon/tunrouter.go
josecv: yeah, basically -- the `neverProxySubnets` slice would be full of `Route{}`, and then when we'd access those `Route{}`, they'd contain nil pointers and we'd panic on trying to use them

PR #2172, opened 2021-10-25
Allow the traffic agent to be manually injected

josecv:
## Description

Allows manual agent injection. For this, the sidecar container must be generated:

```
telepresence genyaml --manifest-type container --container-name echo-easy --port 8080 --output - --input k8s/echo-easy.yaml
```

As well as the volume:

```
telepresence genyaml --manifest-type volume --container-name echo-easy --port 8080 --output - --input k8s/echo-easy.yaml
``` 

Then once both are injected into the application's deployment YAML, the pods will have to be annotated as `telepresence.getambassador.io/manually-injected: "true"`

## Checklist

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -32,7 +32,6 @@ require (
 	golang.zx2c4.com/wireguard/windows v0.3.11
 	google.golang.org/grpc v1.38.0
 	google.golang.org/protobuf v1.25.0
-	gopkg.in/yaml.v2 v2.4.0
 	gopkg.in/yaml.v3 v3.0.0-20200615113413-eeeca48fe776
 	helm.sh/helm/v3 v3.6.3
 	k8s.io/api v0.21.0
@@ -151,6 +150,7 @@ require (
 	google.golang.org/genproto v0.0.0-20201110150050-8816d57aaa9a // indirect
 	gopkg.in/gorp.v1 v1.7.2 // indirect
 	gopkg.in/inf.v0 v0.9.1 // indirect
+	gopkg.in/yaml.v2 v2.4.0 // indirect
 	k8s.io/apiextensions-apiserver v0.21.0 // indirect
 	k8s.io/apiserver v0.21.0 // indirect
 	k8s.io/cli-runtime v0.21.0 // indirect
--- a/go.mod
+++ b/go.mod
@@ -32,6 +32,7 @@ require (
 	golang.zx2c4.com/wireguard/windows v0.3.11
 	google.golang.org/grpc v1.38.0
 	google.golang.org/protobuf v1.25.0
+	gopkg.in/yaml.v2 v2.4.0
 	gopkg.in/yaml.v3 v3.0.0-20200615113413-eeeca48fe776
 	helm.sh/helm/v3 v3.6.3
 	k8s.io/api v0.21.0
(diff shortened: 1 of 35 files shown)


pkg/client/cli/cmds_genyaml.go
```diff
@@ -0,0 +1,185 @@
...
+	scheme.AddKnownTypes(schema.GroupVersion{Group: appsv1.GroupName, Version: "v1"}, &appsv1.StatefulSet{}, &appsv1.Deployment{}, &appsv1.ReplicaSet{})
+	codecFactory := serializer.NewCodecFactory(scheme)
+	deserializer := codecFactory.UniversalDeserializer()
+	obj, kind, err := deserializer.Decode(b, nil, nil)
+	if err != nil {
+		return fmt.Errorf("unable to parse yaml in %s: %w", i.inputFile, err)
+	}
+	var containers []corev1.Container
+	switch o := obj.(type) {
+	case *appsv1.Deployment:
```
donnyyung, line 170: Also need something here for replicasets no?

pkg/client/connector/userd_trafficmgr/install.go
```diff
@@ -214,39 +269,14 @@ func (ki *installer) ensureAgent(c context.Context, namespace, name, svcName, po
 	kind := obj.GetObjectKind().GroupVersionKind().Kind
 
 	var svc *kates.Service
-	if a := podTemplate.ObjectMeta.Annotations; a != nil && a[install.InjectAnnotation] == "enabled" {
-		// agent is injected using a mutating webhook. Get its service and skip the rest
-		svc, err = install.FindMatchingService(c, ki.Client(), portNameOrNumber, svcName, namespace, podTemplate.Labels)
+	a := podTemplate.ObjectMeta.Annotations
+	webhookInjected := a != nil && a[install.InjectAnnotation] == "enabled"
+	manuallyManaged := a != nil && a[install.ManualInjectAnnotation] == "true"
+	if webhookInjected || manuallyManaged {
```
donnyyung: Should we make sure this is an exclusive or, and raise an error if for some reason someone sets both annotations?

go.mod
```diff
@@ -32,6 +32,7 @@ require (
 	golang.zx2c4.com/wireguard/windows v0.3.11
 	google.golang.org/grpc v1.38.0
 	google.golang.org/protobuf v1.25.0
+	gopkg.in/yaml.v2 v2.4.0
```
donnyyung: Do we use something specific here that's not in the 3.0.0 version we already use?

pkg/client/cli/cmds_genyaml.go
```diff
@@ -0,0 +1,239 @@
...
+	_ = cmd.MarkPersistentFlagRequired("input")
+	cmd.AddCommand(
+		genContainerSubCommand(&info),
+		genVolumeSubCommand(&info),
+	)
+	return cmd
+}
+
+func (i *genYAMLInfo) getOutputWriter() (io.WriteCloser, error) {
+	if i.outputFile == "-" {
```
donnyyung, line 55: I might just not know this, but does "-" always mean stdout?  If so, it might be worth making that the default instead of "", or mentioning in the help text that `-` will output it to stdout.

go.mod
josecv: ah lemme see i didn't realize we already had a version

pkg/client/cli/cmds_genyaml.go
josecv: conventionally `-` means stdin/out yeah, but i'll add a note. I also agree it makes sense as the default for output

donnyyung (approved):
Assuming the tests pass, this looks good to me! Thanks for addressing my comments :)

PR #2184, opened 2021-10-28
Fix default image config blocking extensions

donnyyung:

## Description

We were setting this config value all the time which broke extensions' ability to get their own agent image.

## Checklist


smoke-tests/run_smoke_test.sh
```diff
@@ -353,6 +353,7 @@ fi
 
 VERYLARGEJAVASERVICE=verylargejavaservice.default:8080
 $TELEPRESENCE connect >"$output_location"
+$TELEPRESENCE loglevel debug --duration 5m >"$output_location"
```
donnyyung, line 356: This was unrelated to the fix but in general is probably a good idea for debugging purposes.

josecv (approved):

PR #2185, opened 2021-10-28
added rule for sourcePort

njayp:

## Description

Added a rule to break iptables loop

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@
 - Feature: Telepresence now supports manually injecting the traffic-agent YAML into workload manifests.
   Use the `genyaml` command to create the sidecar YAML, then add the `telepresence.getambassador.io/manually-injected: "true"` annotation to to your pods to allow Telepresence to intercept them.
 
+- Feature: Added an iptables rule to prevent looping requests
+
 ### 2.4.5 (October 15, 2021)
 
 - Feature: Intercepting headless services is now supported. It's now possible to request a headless service on whatever port it exposes and get a response from the intercept.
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,8 @@
 - Feature: Telepresence now supports manually injecting the traffic-agent YAML into workload manifests.
   Use the `genyaml` command to create the sidecar YAML, then add the `telepresence.getambassador.io/manually-injected: "true"` annotation to to your pods to allow Telepresence to intercept them.
 
+- Feature: Added an iptables rule to prevent looping requests
```
josecv, line 14: This should probably be a bugfix, also pls state what it does e.g. `prevent the fallback DNS resolver on linux from requesting itself`

pkg/client/daemon/outbound_linux.go
```diff
@@ -176,7 +177,7 @@ func (o *outbound) runOverridingServer(c context.Context) error {
 			// Give DNS server time to start before rerouting NAT
 			dtime.SleepWithContext(c, time.Millisecond)
 
-			err := routeDNS(c, o.dnsConfig.LocalIp, dnsResolverAddr.Port, conn.LocalAddr().(*net.UDPAddr))
+			err := routeDNS(c, o.dnsConfig.LocalIp, dnsResolverAddr.Port, conn.LocalAddr().(*net.UDPAddr), sourcePort)
```
josecv, line 180: wait hang on i'm confused now... isn't `conn.LocalAddr()` the same as sourcePort here?

pkg/client/daemon/outbound_linux.go
```diff
@@ -314,6 +315,14 @@ func routeDNS(c context.Context, dnsIP net.IP, toPort int, fallback *net.UDPAddr
 		return err
 	}
 
+	if err = runNatTableCmd(c, "-A", tpDNSChain,
```
josecv, line 318: hmm if sourcePort is the same as fallback then this seems like it'd be very much the same as the previous rule

josecv (changes requested):
suddenly i'm not so sure about this -- is there a reason the rule in line 308 isn't enough to achieve this? maybe it needs to not have a `--source`

PR #2189, opened 2021-11-01
Don't use ipv4 masks for ipv6 addresses.

josecv:
Also added some fixes to ensure we use the right kinds of IP addresses
in the routing package (iputil.Parse vs net.ParseIP).

## Description

## Checklist

 Not necessary since the bug hasn't been released

### DIFF
--- a/pkg/client/connector/userd_trafficmgr/traffic_manager.go
+++ b/pkg/client/connector/userd_trafficmgr/traffic_manager.go
@@ -518,6 +518,7 @@ func (tm *trafficManager) getOutboundInfo(ctx context.Context) *daemon.OutboundI
 			mask := net.CIDRMask(128, 128)
 			if ipv4 := ip.To4(); ipv4 != nil {
 				mask = net.CIDRMask(32, 32)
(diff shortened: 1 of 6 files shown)


pkg/client/connector/userd_trafficmgr/traffic_manager.go
```diff
@@ -518,6 +518,7 @@ func (tm *trafficManager) getOutboundInfo(ctx context.Context) *daemon.OutboundI
 			mask := net.CIDRMask(128, 128)
 			if ipv4 := ip.To4(); ipv4 != nil {
 				mask = net.CIDRMask(32, 32)
+				ip = ipv4
```
josecv, line 521: this line is the actual bugfix

njayp (approved):

thallgren (approved):
Good catch (and I still wonder why `net.ParseIP()` creates ipv6 out of ipv4 addresses, but I guess it's in the spirit of "use ipv6 everywhere". Doesn't play well with subnets and masks though).

PR #2195, opened 2021-11-04
Complete rewrite of the telepresence integration tests

thallgren:
## Description

All integration tests, with the exception of some tests for the installer, resided in one single file, the `pkg/client/cli/telepresence_test.go`. This PR moves those tests into a more fine grained file structure under the `integration_test` folder. The tests are also split into harnesses and suites. A harness is typically a cluster, cluster with namespaces, or cluster with namespaces and a connected client.

The installer integration tests are also moved into this new structure to prevent that the cluster is created and the image is published twice.

## Checklist


josecv (approved):
looks great i really like the structure of it

i think there'll be further opportunities to organize stuff and break it down; e.g. it's currently a little hard to know that a given `_test.go` file uses a given `Suite`, but for now it's okay

PR #2198, opened 2021-11-05
Check that the client version is the same as version of the daemons.

thallgren:
Adds a version check that triggers when Telepresence connects to an
already started daemon. If there's a version mismatch, then the connect
is refused and the user is asked to quit and reconnect.


josecv (approved):

efunk:
Released in Telepresence 2.4.7

PR #2199, opened 2021-11-05
Remove explicit DNS flush on darwin.

thallgren:
## Description

Updating files under `/etc/resolver` is enough to flush the DNS cache
on macOS. There's no need to flush it explicitly. Actually, doing so
makes the workstations DNS somewhat unstable since requests made during
the flush can be interrupted and return failures.

## Checklist


josecv (approved):
looks good and tracks with what we said on wednesday but i'm a little concerned about the failing tests; approving bc i assume it's just a flake though

thallgren:
Closing this in favor of #2210.

PR #2200, opened 2021-11-05
Enable global flags when using legacy flags.

thallgren:
## Description

All global flags are now accepted when using legacy flags such as
`--swap-deployment`.

This fix is primarily driven by the fact that we need to use the `--as`
flag when running our integration tests.

## Checklist


josecv (approved):

efunk:

PR #2202, opened 2021-11-08
Ensure that outbound connection is closed when closed by peer

thallgren:
## Description

An outbound TCP connection that got closed by the peer would cancel all
activity, but it didn't send the correct packages to negotiate the
EOF to the client. This commit ensures that the FIN,ACK is sent and
that the connection terminates.

## Checklist


josecv (approved):

efunk:

PR #2209, opened 2021-11-10
Some minor improvements to the integration test framework

thallgren:


josecv (approved):

integration_test/itest/cluster.go
```diff
@@ -42,7 +43,7 @@ const TestUser = "telepresence-test-developer"
 const TestUserAccount = "system:serviceaccount:default:" + TestUser
 
 type Cluster interface {
-	CapturePodLogs(ctx context.Context, app, ns string)
+	CapturePodLogs(ctx context.Context, app, container, ns string)
```
donnyyung, line 46: Nit: Perhaps change `app` to `labelSelector` or `selector` since that's what it was changed to.

donnyyung (approved):
This looks good to me, I will try to look into the ReplicaSet flakiness more but since we have passing tests I see no reason to block this PR on that.  I left one nit comment that you can change if you want, but approving this either way.

thallgren:
I'm merging this as-is since we now have other things rebase on top of it. @donnyyung I'll address your "app should be labelSelector" remark in a future PR.

PR #2210, opened 2021-11-10
Use short TTL for DNS lookups in combination with a local cache.

thallgren:
## Description

Unifies how DNS queries are cached by implementing a proper caching
mechanism with recursion detection. The recursion detection is needed
when running clusters DNS server locally, as it forwards queries that
it fails to resolve to the local DNS resolver, which in turn forwards
the query to Telepresence DNS resolver.

The number of queries that are forwarded to the cluster is minimized
when using a local cache which means that a very short TTL can be used
without increasing the load on the cluster's DNS resolver. A short TTL
prevents caching in the host's DNS system and hence, removes the need
for explicit host-dependent DNS flushing. Instead, the local cache
is flushed when needed.

## Checklist


pkg/client/daemon/dns/dns.go
```diff
@@ -22,23 +28,103 @@ type Server struct {
 	fallback     *dns.Conn
 	resolve      Resolver
 	requestCount int64
+	cache        *sync.Map
+	recursive    int32 // 0 = never tested, 1 = not recursive, 2 = recursive
```
josecv, line 32: could we turn these into constants with a custom int type?

pkg/client/daemon/dns/dns.go
```diff
@@ -22,23 +28,103 @@ type Server struct {
...
+func (s *Server) resolveThruCache(q *dns.Question) []dns.RR {
+	newDv := &dnsValue{wait: make(chan struct{}), created: time.Now()}
+	if v, loaded := s.cache.LoadOrStore(q.Name, newDv); loaded {
+		oldDv := v.(*dnsValue)
+		if atomic.LoadInt32(&s.recursive) == 2 && atomic.LoadInt32(&oldDv.recursion) == int32(q.Qtype) {
+			// We have to assume that this is a recursion from the cluster.
+			return nil
+		}
+		<-oldDv.wait
+		if time.Since(oldDv.created) < 60*time.Second {
```
josecv: is 60s the TTL? can that be made a constant?

pkg/tunnel/pool.go
```diff
@@ -63,14 +100,81 @@ func (p *Pool) GetOrCreate(ctx context.Context, id ConnID, createHandler Handler
...
+		return handler, true, nil
+	}
+
+	if blocker != nil {
+		<-blocker.InitDone()
+		if blocker.Proceed() {
+			return p.GetOrCreate(ctx, id, createHandler)
+		}
+		if err := blocker.Reset(ctx, initialPacket); err != nil {
+			dlog.Errorf(ctx, "failed to send RST after recursion block")
```
josecv: let's log `err` here

josecv (approved):
lgtm just some minor things

pkg/client/daemon/dns/dns.go
thallgren: I'd rather not do that because a) it's very internal, and b) then I'll also need to do type casting on all the atomic load and store operations which makes the code less readable.

efunk:

PR #2214, opened 2021-11-16
Njayp/ingress flag

njayp:
## Description

In order to create a preview URL with the docker-descktop ext UI, I need to be able to skip the dialogue. I did this by adding the option of setting the ingress fields using flags. 

Example command: ./build-output/bin/telepresence intercept dataprocessingservice -i --ingressHost dataprocessingservice.default --ingressPort 3000 --ingressL5 dataprocessingservice.default -p 3000:3000

## Checklist


pkg/client/cli/cmds_intercept.go
```diff
@@ -151,6 +159,13 @@ func interceptCommand(ctx context.Context) *cobra.Command {
 
 	flags.StringVarP(&args.namespace, "namespace", "n", "", "If present, the namespace scope for this CLI request")
 
+	flags.BoolVarP(&args.cmdLineIngress, "cmd-ingress", "i", false, "if set, uses cmd flags instead of opening a dialogue for ingress settings."+
```
thallgren: Why isn't this flag implied by the other ingressXXX flags?

pkg/client/cli/cmds_intercept.go
```diff
@@ -151,6 +159,13 @@ func interceptCommand(ctx context.Context) *cobra.Command {
...
+		"Uses --ingressHost, --ingressPort, --ingressTLS, and --ingressL5")
+	flags.StringVar(&args.ingressHost, "ingressHost", "", "if cmd-ingress is set, this value will be used as the ingress hostname")
```
thallgren: I suggest you use all lower-case and dash separated words. All current command flags follow those rules.

pkg/client/cli/cmds_intercept.go
```diff
@@ -522,6 +537,20 @@ func (is *interceptState) EnsureState(ctx context.Context) (acquired bool, err e
 		}
 	}()
 
+	if is.args.previewEnabled && is.args.cmdLineIngress && is.args.previewSpec.Ingress == nil {
+		ingress := &manager.IngressInfo{}
+		ingress.Host = is.args.ingressHost
+		ingress.Port = is.args.ingressPort
```
thallgren: Some sanity checks would be welcome here. An ingress host must be a valid hostname (see `var hostRx` declared further down). The port cannot be zero.

pkg/client/cli/cmds_intercept.go
```diff
@@ -522,6 +537,20 @@ func (is *interceptState) EnsureState(ctx context.Context) (acquired bool, err e
...
+		ingress.UseTls = is.args.ingressTLS
+		if ingress.L5Host == "" {
+			ingress.L5Host = is.args.ingressHost
+		} else {
+			ingress.L5Host = is.args.ingressL5
```
thallgren: Check that it's a valid host name.

thallgren (changes requested):
Should these flags also be added to the `preview` command?

pkg/client/cli/cmds_intercept.go
donnyyung: I agree with Thomas here. I think instead of having `cmdLineIngress` as a flag, just add the others, and if the user has already specified the flag, don't call the `askfor*` function inside of `selectIngress` function and use the value provided by the user.  (you could add some output like: "Using value for x" so the user knows why we aren't asking them to provide input.

njayp: Yes, I will get rid of the -i flag and check if any of the other flags are set.

I dont really want to make the flags answer individual questions in `selectIngress` because I don't think that is a use-case. I dont think anyone would ever want to only skip half the dialogue questions.

njayp:
> Should these flags also be added to the `preview` command?

Yes, they should. I took a quick look, and it doesnt seem trivial, as cmd_intercept reuses the preview flag struct. I can do this during my cool-down.

pkg/client/cli/cmds_intercept.go
```diff
@@ -505,6 +520,37 @@ func (is *interceptState) getMountPoint() (string, bool, error) {
...
+	if hostRx.MatchString(ingressHost) {
+		if ingressPort > 0 {
+			ingress.Host = ingressHost
+			ingress.Port = ingressPort
+			ingress.UseTls = ingressTLS
+
+			if ingress.L5Host == "" { // if L5Host is not present
+				ingress.L5Host = ingressHost
+				return ingress, nil
+			} else { // if L5Host is present
```
thallgren, line 534: Nitpick, but you don't need an `else` after an `if` that does a `return`.

pkg/client/cli/cmds_intercept.go
```diff
@@ -505,6 +520,37 @@ func (is *interceptState) getMountPoint() (string, bool, error) {
...
+				if hostRx.MatchString(ingressL5) {
+					ingress.L5Host = ingressL5
+					return ingress, nil
+				} else {
```
thallgren, line 538: Same thing here. No `else` needed.

pkg/client/cli/cmds_intercept.go
```diff
@@ -505,6 +520,37 @@ func (is *interceptState) getMountPoint() (string, bool, error) {
...
+					return nil, fmt.Errorf("the address provided by --ingress-l5, %s, must match the regex"+
+						" [a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)* (e.g. 'myingress.mynamespace')",
+						ingressL5)
+				}
+			}
+		} else {
```
thallgren, line 544: And once the above `else`s have been removed, it becomes very clear that the `if` for this `else` always returns, so this `else` is also redundant.

thallgren (approved):
Approved with a nit. Feel free to ignore.

efunk:

PR #2221, opened 2021-11-17
Login always uses credentials

donnyyung:
Currently, if there is cached user-info, we don't try a new login. This
creates issues when the token expires, so this will ensure that if the
token has expired and a user runs login, it will make them login again.

## Description

This is basically a bandaid until I can figure out why the refresh token logic doesn't seem to be working. 

Here is the problem users are seeing:
```
User Daemon: Running
  Version           : v2.4.7-52-geb199abe-1637180824 (api 3)
  Ambassador Cloud  : Login expired (or otherwise no-longer-operational)
  Status            : Connected
  Kubernetes server : https://35.224.222.254
  Kubernetes context: default
  Telepresence proxy: ON (networking to the cluster is enabled)
  Intercepts        : 0 total
```

But login currently doesn't do anything because the user-info is cached on their machine, but the token has expired... With this fix when you do a login, this is what happens:
```
(⎈ |default:default)➜  ~/go/telepresenceio git:(donnyyung/expired-login) ✗ telepresence login
Launching browser authentication flow...
Login successful.
(⎈ |default:default)➜  ~/go/telepresenceio git:(donnyyung/expired-login) ✗ telepresence status
Root Daemon: Running
  Version   : v2.4.7-52-geb199abe-1637180824 (api 3)
  DNS       :
    Remote IP       : 10.43.0.10
    Exclude suffixes: [.arpa .com .io .net .org .ru]
    Include suffixes: []
    Timeout         : 4s
  Also Proxy: (1 subnets)
    - 35.224.222.0/24
User Daemon: Running
  Version           : v2.4.7-52-geb199abe-1637180824 (api 3)
  Ambassador Cloud  : Logged in
  Status            : Connected
  ``` 

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 - Feature: The agent injector now supports a new annotation, `telepresence.getambassador.io/inject-service-name`, that can be used to set the name of the service to be intercepted.
   This will help disambiguate which service to intercept for when a workload is exposed by multiple services, such as can happen with Argo Rollouts
 
+- Feature: The kubeconfig extensions now support a `never-proxy` argument, analogous to `also-proxy`, that defines a set of subnets that will never be proxied via telepresence.
+
 - Bugfix: Legacy flags such as `--swap-deployment` can now be used together with global flags.
 
 - Change: Telepresence check the versions of the client and the daemons and ask the user to quit and restart if they don't match.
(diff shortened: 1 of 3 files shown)


thallgren (approved):
LGTM

efunk:

PR #2222, opened 2021-11-18
Ensure update of both version and app-version in the embedded Helm chart

thallgren:
Add a `--app-version=VERSION` to the embedded Helm chart packager so
that the `app-version` value of the embedded chart is the same as the
version of the binary.


donnyyung (approved):

donnyyung (approved):
Nice catch on this bug. LGTM!

PR #2224, opened 2021-11-18
Implement a configuration to never proxy certain subnets.

josecv:

## Description

## Checklist

 - [x] I made sure to either submit a docs PR, or tell Matt about the necessary documentation changes. https://github.com/telepresenceio/telepresence.io/pull/125

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@
 
 - Bugfix: The configured webhookRegistry is now propagated to the webhook installer even if no webhookAgentImage has been set.
 
+- Feature: The kubeconfig extensions now support a `never-proxy` argument, analogous to `also-proxy`, that defines a set of subnets that will never be proxied via telepresence.
 
 ### 2.4.6 (November 2, 2021)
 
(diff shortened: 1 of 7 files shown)


thallgren (commented):
What happens if you `never-proxy` a subset of a proxied subnet? And vice versa, what happens if you `always-proxy` a subset of a `never-proxy` subnet?

josecv:
Great question -- so the never proxies are implemented as a routing override to restore their original route, which means that effectively routing precedence rules apply. If I never proxy a subnet with a larger mask than an also-proxied subnet, the never proxied subnet wins. If the also-proxy subnet has a larger mask, then that one will win.

So e.g. if I have:
```
never-proxy: 10.0.0.0/16
also-proxy: 10.0.5.0/24
```
Then only `10.0.5.0/24` will be proxied through. If I have:

```
also-proxy: 10.0.0.0/16
never-proxy: 10.0.5.0/24
```

Then all of `10.0.0.0/16` will be proxied *except* `10.0.5.0/24`

I am going to go add a test to make sure I'm not lying :^)

thallgren (approved):

efunk:

PR #2226, opened 2021-11-22
Export baseLogger() function.

thallgren:
This enables unified logging in the OSS and proprietary agents.


donnyyung (approved):
Thank you for this!!

PR #2227, opened 2021-11-22
Add a RESTful service to help determine if messages should be consumed or not

thallgren:
## Description

This PR adds a RESTful API service to the Telepresence `traffic-agent`. The service has one end-point `consume-here` which can be queried using a set of http headers to determine if those headers belong to an active intercept in the `traffic-agent`. When that's the case, the `consume-here` returns `false` since the message presumably should be consumed by the interceptor running at the local workstation (in fact, the OSS `traffic-agent` will always return true when it is intercepting because it is unaware of personal intercepts and headers).

Another RESTful service, with the same API is added to the user daemon running on the local workstation so that the application can query the same end-point when running locally.  The reply from `consume-here` will be inverted so that matching headers yield `true`.

## Checklist


charts/telepresence/values.yaml
```diff
@@ -171,6 +171,15 @@ agentInjector:
...
 
+################################################################################
+## Telepresence API Server Configuration
+################################################################################
+telepresenceAPI:
+  # Default: false
+  create: false
+  # The port on agent's localhost where the API service can be found
+  # Default: 0
+  port: 0
```
josecv, line 180: should we have a default instead of 0? that way users can install with `--set telepresenceAPI.create=true` and get a reasonable port for the service

cmd/traffic/cmd/manager/internal/mutator/agent_injector.go
```diff
@@ -274,13 +280,55 @@ func addAgentContainer(
...
+	keys := make([]string, len(env))
+	i := 0
+	for k := range env {
+		keys[i] = k
+		i++
+	}
+	sort.Strings(keys)
+	if cn.Env == nil {
+		patches = append(patches, patchOperation{
+			Op:    "replace",
```
josecv, line 313: should this be add at this point, since they're not there already? similar to what we do to `initContainers`

pkg/header/matcher.go
```diff
@@ -0,0 +1,116 @@
...
+// string that has no regexp-meta characters or that cannot be compiled into a regexp will yield an
+// exact text matcher.
+func NewValueMatcher(v string) ValueMatcher {
+	if regexp.QuoteMeta(v) == v {
+		return textMatcher(v)
+	}
+	rx, err := regexp.Compile(v)
+	if err != nil {
+		return textMatcher(v)
+	}
```
josecv: hmmm i feel like the error should be logged or even returned; if i accidentally mess up my regex, there should be some kind of indicator

pkg/restapi/api.go
```diff
@@ -0,0 +1,92 @@
...
+	"net/http"
+	"strconv"
+
+	"github.com/datawire/dlib/dhttp"
+	"github.com/datawire/dlib/dlog"
+)
+
+const HeaderCallerInterceptID = "x-telepresence-caller-intercept-id"
+const HeaderInterceptID = "x-telepresence-intercept-id"
+const EndPontConsumeHere = "/consume-here"
```
josecv: should this be `EndpointConsumeHere`? instead of `EndPont`

josecv (changes requested):
looks good, just some questions/comments

charts/telepresence/values.yaml
thallgren: The way it's designed (and documented), the service will be enabled when the user chooses a port. I'm not a fan of default ports as they tend to collide with other vendors choice of defaults. There's no authority to register such defaults with so the whole port space is sort of random.

cmd/traffic/cmd/manager/internal/mutator/agent_injector.go
thallgren: In my mind, this is an assignment, and assigning is always "replace", whereas an "add" is adding to a collection. Consider this in go.
```go
foo = 1
```
here, we replaced the value in the variable `foo` with `1`. One could argue that:
```go
foo := 1
```
is an add, because here, `foo` is also declared. So the question is, do we treat the patch as something that declares the `Env` or assigns a value to something that already exists in the container? In my mind, the `Container` is a struct, and the `Env` is a field in that struct. Since we don't declare it, we don't add it. We replace the value of an existing field. Consequently, the `initiContainers` should probably change from "add" to "replace".

josecv (approved):

PR #2240, opened 2021-11-26
Let auto-installer default to using Telepresence API port from manager

thallgren:
The auto-installer must use the API port stipulated by the traffic
manager unless `telepresenceAPI.port` is explicitly set in the
`config.yml`. This means that it's enough to declare this port when
doing the Helm install. All clients will follow (unless they explicitly
override).


pkg/client/connector/userd_trafficmgr/intercept.go
```diff
@@ -304,10 +305,22 @@ func (tm *trafficManager) AddIntercept(c context.Context, ir *rpc.CreateIntercep
...
 	}
 
+	cfg := client.GetConfig(c)
+	apiPort := uint16(cfg.TelepresenceAPI.Port)
+	if apiPort == 0 {
+		// Default to the API port declared by the traffic-manager
+		apiInfo, err := tm.managerClient.GetTelepresenceAPI(c, &empty.Empty{})
+		if err != nil {
+			// Traffic manager is probably outdated. Not fatal, but deserves to be logged
+			dlog.Errorf(c, "failed to obtain Telepresence API info from traffic manager")
```
josecv: should we log `err`?

josecv (approved):

pkg/client/connector/userd_trafficmgr/intercept.go
thallgren: Definitely :-)

PR #2243, opened 2021-11-29
Update source of LoginClientID

donnyyung:
## Description

## Checklist


pkg/client/version.go
```diff
@@ -14,3 +20,39 @@ func Version() string {
...
+	// so we follow the symlink to get the actual binary path
+	mechanism := "undetermined"
+	binaryPath, err := filepath.EvalSymlinks(execPath)
+	if err != nil {
+		wrapErr := fmt.Errorf("error following executable symlink %s: %s", execPath, err)
+		return mechanism, wrapErr
+	}
+	switch {
+	case runtime.GOOS == "darwin" && strings.Contains(binaryPath, "Cellar"):
+		mechanism = "brew"
```
josecv, line 51: nice i like that we'll have numbers for that

pkg/client/version_test.go
```diff
@@ -0,0 +1,136 @@
...
+		testcases["brew-install-sym"] = testcase{
+			binaryPath:               "Cellar/telepresence",
+			symLinkPath:              "telepresence",
+			expectedInstallMechanism: "brew",
+			errFile:                  false,
+		}
+	} else {
+		// I'm not sure why anybody would have brew paths in a non-macos
+		// but brew is *only* for macos so we should report website in
+		// this case.
```
josecv: well there is brew for linux, but i suppose we don't support it, given that our brew formulae just download the binary

josecv (approved):

pkg/client/version_test.go
donnyyung: TIL, I will update that comment so it's accurate 😬

PR #2250, opened 2021-12-02
Fix access token not refreshing

donnyyung:
## Description
This fixes an error where if you do `telepresence login` after `telepresence connect` the access tokens never refresh because the login command overrides the newTokenSource with a context that is canceled as soon as the command finishes, which causes the refresh token function to fail since it uses the expired context.

I've fixed this by passing the token to the goroutine via a channel and the refresh goroutine will update the newTokenSource with that token and a context that isn't expired so the refresh token function no longer fails.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+### 2.4.9 (TBD)
+
+- Bugfix: Fixed an error where access tokens were not refreshed if you login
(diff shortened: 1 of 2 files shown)


josecv (approved):
i think this is pretty reasonable for now, but this seems like a good example of why we shouldn't store contexts in structs

maybe we should consider refactoring this, but that can wait

PR #2256, opened 2021-12-08
Thallgren/chart nil exception

thallgren:
## Description

The new `telepresenceAPI.port` value causes a "nil pointer" error when
upgrading from earlier chart versions using the `--reuse-values` flag
because the container for the `.port` is nil. This commit ensures that
the part of the template that is failing is ignored when the value is
missing.

## Checklist


donnyyung (approved):
LGTM!

efunk:
Released in Telepresence 2.4.9 GA

PR #2265, opened 2021-12-14
Add github actions for tests

donnyyung:
## Description
Picking up the work jose left off on. This just moves the unit + integration tests to github actions, we will move the release bits later, perhaps next cooldown.

Successful run of the tests can be seen on this PR:
https://github.com/telepresenceio/telepresence/pull/2266
## Checklist


.github/workflows/unit_tests.yaml
```diff
@@ -0,0 +1,178 @@
+name: "Unit tests"
```
josecv: is there any way that we can have like a parameterized workflow? would be nice not to repeat the stuff from the integration tests here

smoke-tests/run_smoke_test.sh
```diff
@@ -485,6 +487,26 @@ verify_output_empty "${output}" false
...
 
+###############################################
+#### Step 3b (temp) - Verify mount works   ####
+###############################################
+# Due to some issues with the github actions macOS executors, macfuse
+# doesn't work in our integration tests, so we ensure that mounts
+# in our smoke tests. The integration tests *do* test mounts on
+# Windows and Linux so this testing is really just to be extra cautious.
+# We can remove this whole step when/if the macfuse issue is cleared up
+# in the github actions executors
```
josecv: do we have a reference of some sort here? maybe we should open up an issue with github to track this

josecv (approved):
nice one Donny, thank you for taking over my half-finished branch 😄

smoke-tests/run_smoke_test.sh
donnyyung: So perhaps I should change this comment... This issue was happening in the newer circle ci executors as well (which is why we were pinned to an old version) so I think it's actually due to changes in macos since it was affecting both.  And so once it seemed like a macos problem, I decided not to dig too much further since adding a smoke test would be an easy workaround... but I'll try to add more information here so people are aware of the "state" I'm leaving it in.

build-aux/main.mk
```diff
@@ -229,13 +229,27 @@ format: $(tools/golangci-lint) $(tools/protolint) ## (QA) Automatically fix lint
 	$(tools/golangci-lint) run --fix --timeout 2m ./... || true
 	$(tools/protolint) lint --fix rpc || true
 
-.PHONY: check
+.PHONY: check-all
```
maxlegault, line 232: Looks like the name of the target below isn't renamed to `check-all`

donnyyung: oops good catch thank you!

.github/workflows/unit_tests.yaml
donnyyung: So I spent some time looking into this (and I did turn the cache logic into a reusable action) and tried to do the same with the tests, but that won't work for annoying reasons.

1. For the windows tests we use powershell and for mac/linux we use bash, which was nbd bc I could have a run step for the windows tests and one for mac/linux and then "if" them based on the platform
2. "if" isn't supported on composite actions so then I found a hack where I could do an "if" in the command logic and if we were doing the windows tests on a non-windows machine, then just exit 0.
3. But the retry action we are using doesn't let you use the powershell shell on non-windows executors, so that step would always fail.

So basically it was hacks on hacks so I decided to stop investigating. I think perhaps we could make the windows tests run on bash instead of using powershell but I've sorta already spent 1.5 days on this and since we'll be revisiting the actions to add the release steps, I can do more investigation then.

josecv: no worries, let's leave this as it is then and we'll figure something out :) this is a great start

PR #2268, opened 2021-12-17
Let dialer ignore DialOK controls

thallgren:
## Description

A `Dialer` is reused as a tunnel endpoint. The peer endpoint may also be
a `Dialer` and may respond with the `DialOK` control message. It's an
indication that everything is OK and can be safely ignored.

## Checklist


donnyyung (approved):
LGTM

PR #2269, opened 2021-12-21
Ensure that intercept can be made before login and asking for ingress

thallgren:
## Description

Split intercept creation into a CanIntercept and CreateIntercept
This split is made so that the user can be informed about any problems
with an intercept (typically that a workload doesn't exist or that an
intercept already exists) before logging in or engaging in a dialog
to provide ingress information.

## Checklist


rpc/connector/connector.proto
```diff
@@ -161,7 +164,7 @@ enum InterceptError {
   NO_ACCEPTABLE_WORKLOAD = 8;
   AMBIGUOUS_MATCH = 9;
   FAILED_TO_ESTABLISH = 10;
-  reserved 11;
```
donnyyung, line 164: Aren't we supposed to not use numbers that were reserved in the past? or is it not a big deal since these are all InterceptErrors?

donnyyung (approved):
One minor comment but this all seems sane and I did some local testing and didn't run into any issues!

rpc/connector/connector.proto
thallgren: We are supposed to do that where backward compatibility can be an issue. That's not the case with the connector protocol because telepresence of later versions will refuse to continue if there's a version mismatch between the cli and the connector.

Issue #2248, opened 2021-12-02
Daemons shutdown when intercepting without previously connecting or starting the local process


douglascamata:
**Describe the bug**

Daemons are shutting down and the intercept isn't being created whenever I try to run the `intercept` command without having previously started the daemons or the local process (that should receive intercepted traffic).

if I previously start the daemons with `telepresence connect` I see the CONN errors, but the intercept gets created and the daemon does not quit.

**To Reproduce**

0. I run `telepresence quit`
1. When I run `telepresence intercept ...`
2. I see `See logs for details (2 errors found): "/Users/douglascamata/Library/Logs/telepresence/connector.log"`
3. So I look at the logs
4. See error: 

```
2021-12-02 17:31:19.8655 debug   connector/background-manager/intercept-port-forward : search paths posted successfully
2021-12-02 17:31:20.0062 debug   connector/background-manager/dial-request-watcher :    CONN tcp 127.0.0.1:39292 -> 127.0.0.1:9090, dialing
2021-12-02 17:31:20.0066 error   connector/background-manager/dial-request-watcher : !! CONN tcp 127.0.0.1:39292 -> 127.0.0.1:9090, failed to establish connection: dial tcp4 127.0.0.1:9090: connect: connection refused
2021-12-02 17:31:20.0119 debug   connector/background-manager/dial-request-watcher :    CONN tcp 127.0.0.1:47932 -> 127.0.0.1:9090, dialing
2021-12-02 17:31:20.0121 error   connector/background-manager/dial-request-watcher : !! CONN tcp 127.0.0.1:47932 -> 127.0.0.1:9090, failed to establish connection: dial tcp4 127.0.0.1:9090: connect: connection refused
2021-12-02 17:31:24.6905 debug   connector/server-grpc/conn=1/CreateIntercept-5 : returned
2021-12-02 17:31:25.1457 debug   connector/server-grpc/conn=1/UserNotifications-1 : returned
2021-12-02 17:31:25.1469 info    connector/server-grpc/conn=2:shutdown_logger : shutting down (gracefully)...
2021-12-02 17:31:25.1469 debug   connector/server-grpc/conn=2/manager->daemon : goroutine "/connector/server-grpc/conn=2/manager->daemon" exited without error
2021-12-02 17:31:25.4016 debug   connector/server-grpc/conn=3/UserNotifications-6 : called
2021-12-02 17:31:25.4017 debug   connector/server-grpc/conn=3/Quit-7 : called
[...]

PR #2273, opened 2022-01-03
Add strategy for application protocol selection

thallgren:
## Description

This PR adds a new `intercept` setting to the `config.yml` file. The `intercept` has two values:

- appProtocolStrategy A strategy used when an agent for personal intercept determines what application protocol to use when talking to the intercepted container running in the cluster and/or the interceptor process running on the local workstation.
- defaultPort the port to use by default for the `telepresence intercept --port` flag. Defaults to 8080.

The appProtocolStrategy for agents injected by a webhook can be set in the Helm chart. During an auto-install, this value defaults to the setting of the `intercept.appProtocolStrategy` in the `config.yml` file.

The `appProtocolStrategy` is only useful when using personal intercepts and requires and updated "smart-agent".

## Checklist


thallgren:
Will likely fix #2248.

pkg/client/cli/extensions/builtin.go
```diff
@@ -48,6 +48,15 @@ func builtinExtensions(ctx context.Context) map[string]ExtensionInfo {
 								`If this flag is given multiple times, then it will only intercept traffic that matches *all* of the specifiers. ` +
 								`(default "auto" if you are logged in with 'telepresence login', default "all" otherwise)`,
 						},
+						"protocol": {
+							Type: "string",
+							Usage: `` +
+								`Supported protocols are ` +
+								`"http" (Plaintext HTTP/1.1), ` +
+								`"http2" (Plaintext HTTP/2), ` +
+								`"tcp" (Opaque TCP stream, implies --http_match=all), and ` +
```
josecv: it feels a little strange that `--http-protocol=tcp` is a valid value. Maybe `--net-protocol` would make more sense?
At the same time i understand that this makes it consistent with the other `--http-*` flags so it does work from that perspective.

pkg/install/util.go
```diff
@@ -37,6 +41,49 @@ func GetPort(cn *corev1.Container, portName string) (*corev1.ContainerPort, erro
 	return nil, fmt.Errorf("unable to locate port %q in container %q", portName, cn.Name)
 }
 
+// GetAppProto determines the application protocol of the given ServicePort. The given AppProtocolStrategy
+// used if the port's appProtocol attribute is unset.
+func GetAppProto(ctx context.Context, aps client.AppProtocolStrategy, p *corev1.ServicePort) string {
```
josecv, line 46: I'm just gonna dump the link to the KEP [here](https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/1507-app-protocol) because i did not know about `appProtocol`

pkg/install/util.go
```diff
@@ -37,6 +41,49 @@ func GetPort(cn *corev1.Container, portName string) (*corev1.ContainerPort, erro
...
+	if p.AppProtocol != nil {
+		appProto := *p.AppProtocol
+		if appProto != "" {
+			dlog.Debugf(ctx, "Using application protocol %q from service appProtocol field", appProto)
+			return appProto
+		}
+	}
```
josecv, line 53: should this function account for a possible `--http-protocol` flag? I'm specifically worried about `--http-protocol=tcp` when the `AppProtocol` is set to, say, `http`.

pkg/client/cli/extensions/builtin.go
thallgren: Yeah, "tcp" was more meant to signify "opaque", i.e. to tell Envoy to just dispatch without any attempt to do L7 routing. The name stems from Istio's port selection. I think I'll just remove it for now.

pkg/install/util.go
thallgren: I don't think our agent should override the protocol used when talking to the container app in the cluster it is explicitly stated in the `ports.appProtocol` field.

The `--http-protocol` flag only affects the protocol used during an intercept when forwarding to the interceptor (the process on the local workstation). It has no effect on the communication between the agent and the container app.

pkg/client/config.go
```diff
@@ -633,6 +637,76 @@ func (g *TelepresenceAPI) merge(o *TelepresenceAPI) {
...
+// in case the service.spec.ports.appProtocol is not set.
+type AppProtocolStrategy int
+
+var apsNames = [...]string{"http2Probe", "portName", "http", "http2"}
+
+const (
+	// Http2Probe means never guess. Choose HTTP/1.1 or HTTP/2 by probing (this is the default behavior)
+	Http2Probe AppProtocolStrategy = iota
+
+	// PortName means trust educated guess based on port name when appProtocol is missing and perform a http2 prob
```
donnyyung: nit: spelling prob -> probe

pkg/install/util.go
josecv: ah okay got it, makes sense

josecv (approved):

donnyyung:
We should update the docs page for https://www.telepresence.io/docs/latest/reference/config/ to show the new config options being added.

PR #2276, opened 2022-01-04
Long-running daemon

thallgren:
## Description

This PR refactors the code of the root daemon and changes its behavior so that it doesn't terminate when a connection ends.  Internally, the connection (and the TUN-device and DNS-resolver) is maintained by a session which has the same lifecycle as the connection.

Three significant changes deserves a mention:
1. The daemon.log is rotated daily instead of once per connection.
2. The config.yml is loaded automatically when it is modified.
3. The messages printed when starting and quitting a connection mentions "network" rather than "root daemon".

## Checklist


pkg/client/daemon/service.go
```diff
@@ -49,6 +50,9 @@ type service struct {
 	rpc.UnsafeDaemonServer
 	session       *session
 	cancel        context.CancelFunc
+	sessionsCtx   context.Context
+	sessionLock   sync.Mutex
+	session       *session
```
josecv, line 59: is this a dupe? there's another `session *session` on line 51

pkg/client/daemon/service.go
```diff
@@ -197,38 +233,15 @@ func run(c context.Context, loggingDir, configDir, dns string) error {
...
-	// it based on what is read from that work queue.
-	scoutUsers.Add(1)
-	g.Go("server-router", func(ctx context.Context) error {
-		defer scoutUsers.Done()
-		return d.session.run(ctx)
+	g.Go("session", func(c context.Context) (err error) {
+		c, d.cancel = context.WithCancel(c)
+		d.sessionsCtx = c
+		<-c.Done()
+		return nil
```
josecv: hmm i feel like the handling of the sessions' lifecycle is a little bit awkward in general. I don't really like the context being stored in the struct here, or that successive sessions share the same context.
It's also a little hard to tell by looking at `daemon/service.go` where exactly the session's goroutines are getting started.

IMO what we should consider doing is split `newSession` into a `newSession` that doesn't start any new goroutines and a `runSession` that actually kicks everything off.
We could then have this goroutine (`g.Go("session")`) receive the `newSession` on some channel and `runSession` on it. Here's sort of what I'm picturing:

```
type service struct {
	sessionCh      chan *session
        sessionErrCh chan error
	sessionLock   sync.Mutex
	session           *session
}

func (d *service) SetOutboundInfo(ctx context.Context, info *rpc.OutboundInfo) (*empty.Empty, error) {
	d.sessionLock.Lock()
	defer d.sessionLock.Unlock()
	if d.session != nil {
		return nil, status.Error(codes.AlreadyExists, "an active session exists")
	}
	s, err := newSession(ctx, d.scout, info)
	if err != nil {
		return nil, status.Error(codes.Internal, err.Error())
	}
	d.sessionCh <- s
        err = <-d.sessionErrCh
        // handle err
        d.session = s // session was started successfully, so store it
	return &empty.Empty{}, nil
}

g.Go("session", func(c context.Context) (err error) {
      c, d.cancel = context.WithCancel(c)
      for {
          sess := <-d.sessionCh
          sCtx, sCancel := context.WithCancel(c)
          doneCh, err := sess.runSession(sCtx)
          d.sessionErrCh <- err
          select {
                case <-c.Done():
                     sCancel()
                     return nil
                case <-doneCh: // assuming doneCh is already closed if runSession failed
                     sCancel()
          }
      }
})
```

This looks a bit bulkier than what you have but i think it gives us a few advantages:

* opens it up to refactor session cancellation so that just doing `sCancel()` is good enough to stop the session -- eventually we could just store `sCancel` and call it in `disconnect`, dropping the need for `session.stop()`. This would be good because right now it's not clear to me that `d.cancel` will clean everything up correctly
* makes it really clear that the sessions are running in the "session" goroutine
* prevents the sessions' context from being shared by multiple sessions, or leaked to the rest of this struct

josecv (changes requested):
this all looks good but it'd be good to discuss about session lifecycles as per my other comment

pkg/client/daemon/service.go
thallgren: Not sure what's going on here. Must be a glitch in how the diff is displayed somehow. There's only one session in the struct and that's the one below the `sessionsLock`.

pkg/client/daemon/service.go
thallgren: Not sure I understand your reasoning here. The correspondence to your `sCancel` is currently stored as `cancel` in the `session` struct. It's created by the `newSession` function from the parent `ctx` which is the `d.sessionsCtx`.  A `disconnect` calls `d.session.stop` to do necessary clean-up. That method in turn calls that `cancel` and terminates the session.

How would `disconnect` get hold of the `sCancel`? It needs to be stored somewhere. What place is better than the session that it cancels?

Please note that the `ctx` received by `Connect` is short lived. Using that context would terminate the connection as soon as the connect call returns.  Hence the need for the `d.sessionsCtx` which lives until the process terminates.

thallgren: `Connect` in my reasoning is `SetOutboundInfo` before it was renamed.

thallgren: Your design is cleaner so I decided to go ahead and implement something similar. The new design is based on your channel pair and moving the actual session execution into the "session" goroutine. It differs slightly in how the cancellation is handled but there's no explicit call to `stop` now. The session can only be cancelled by calling its `cancel` (which is the context cancellation method).

josecv (approved):

PR #2283, opened 2022-01-06
[wip] allow the connector command to take in an interface.

josecv:

## Description

## Checklist

### DIFF
--- a/pkg/client/connector/command.go
+++ b/pkg/client/connector/command.go
@@ -69,8 +69,12 @@ type service struct {
 	connectResponse chan *rpc.ConnectInfo     // connectWorker -> server-grpc.connect()
 }
 
+type Extension interface {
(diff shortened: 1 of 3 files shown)


pkg/client/connector/command.go
```diff
@@ -69,16 +69,20 @@ type service struct {
 	connectResponse chan *rpc.ConnectInfo     // connectWorker -> server-grpc.connect()
 }
 
+type Extension interface {
```
donnyyung, line 72: This name might be confusing in the short-term just bc we already have a notion of extensions (https://github.com/telepresenceio/telepresence/tree/release/v2/pkg/client/cli/extensions). But eventually I think we'll want to yank all//most of that out since the "v2" of extensions we are thinking about is having everything be in an extended daemon.

PR #2288, opened 2022-01-08
Ensure correct shutdown order when session is closed

thallgren:
The DNS service needs the TUN-device to be alive during shutdown, and
closing the TUN-device is what unblocks the routerWorker's read on that
device. This commit ensures that the DNS-worker dies first, followed by
a close down of the TUN-device (which implicitly ends the routerWorker).

### DIFF
--- a/pkg/client/daemon/session.go
+++ b/pkg/client/daemon/session.go
@@ -426,42 +426,47 @@ func (s *session) run(c context.Context) error {
 	case <-c.Done():
 		return nil
 	case <-cfgComplete:
-		g.Go("dns", func(ctx context.Context) error {
-			return s.dnsServer.Worker(ctx, s.dev, s.configureDNS)
-		})
-		g.Go("router", s.routerWorker)
-		g.Go("stop", s.stop)
 	}
+
+	// Start the router and the DNS service and wait for the context
+	// to be done. Then shut things down in order. The following happens:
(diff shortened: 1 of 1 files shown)


pkg/client/daemon/session.go
```diff
@@ -426,42 +426,47 @@ func (s *session) run(c context.Context) error {
...
-	}
-	if atomic.CompareAndSwapInt32(&s.closing, 1, 2) {
-		cc := dcontext.WithoutCancel(c)
-		for _, np := range s.curStaticRoutes {
-			err := s.dev.RemoveStaticRoute(cc, np)
-			if err != nil {
-				dlog.Warnf(c, "error removing route %s: %v", np, err)
-			}
+func (s *session) stop(c context.Context) {
+	atomic.StoreInt32(&s.closing, 1)
```
maxlegault, line 445: Is it still valuable to keep the closing state if it's not compared anymore?

donnyyung (approved):
LGTM!

pkg/client/daemon/session.go
thallgren: There's no point in comparing now. The `stop` function is the only one storing the value and it's guaranteed to only be called once when the session ends.

thallgren: And the value is used by the router.

maxlegault: Ah got it, I hadn't looked at the router.go file. Thanks

PR #2289, opened 2022-01-08
Improved Telemetry Reporting

thallgren:
## Description

This PR makes some improvements to the Telepresence Telemetry Reporting.

1. All reporting is buffered (CLI reporting wasn't, which caused delays in air-gapped environments)
2. Remove stutter (e.g. `scout.Scout` and `scout.ScoutMeta`)
3. Improve performance by not parsing OS-environment on each call.

## Checklist

### DIFF
--- a/pkg/client/connector/command.go
+++ b/pkg/client/connector/command.go
@@ -53,7 +53,7 @@ type parsedConnectRequest struct {
 
 // service represents the state of the Telepresence Connector
 type service struct {
-	scout *scout.Scout
+	scout *scout.Reporter
 
 	cancel func()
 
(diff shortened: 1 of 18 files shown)


pkg/client/connector/command.go
```diff
@@ -204,23 +199,18 @@ func (s *service) connectWorker(c context.Context, cr *rpc.ConnectRequest, k8sCo
...
-		report := ScoutReport{
-			Action: "connecting_traffic_manager",
-			PersistentMetadata: map[string]interface{}{
-				"cluster_id":        cluster.GetClusterId(c),
-				"mapped_namespaces": len(cr.MappedNamespaces),
-			},
-		}
-		return report
-	}()
+	s.scout.SetMetadatum(c, "cluster_id", cluster.GetClusterId(c))
```
donnyyung, line 202: nit: Should we also set the mapped_namespaces with `SetMetadatum` instead of adding it to the Report?

thallgren: The `cluster_id` used to be a metadatum, so I kept it that way because it makes sense that subsequent reports are identified with that ID. Didn't keep the `namespaces` count as metadatum though, since that felt simply wrong. Also, please note that the old way using `PersistentMetadata` also sent a `Report` (an empty one, unless it also included other `Metadata`).

donnyyung (approved):
LGTM! thanks for also adding the test fix in here!

PR #2298, opened 2022-01-11
Ensure that agent can accept the given mechanism flags

thallgren:
Check if an existing agent is present, and if it is, then ensure that
it can handle the given mechanism args.


donnyyung (approved):
LGTM!

PR #2299, opened 2022-01-12
Send DNS telemetry as counters per session instead of once per query

thallgren:
Sending one telemetry report to metriton for each DNS query that
telepresence performs with a success or failure status is quite heavy.
It's far more efficient to send the data as counters when the session
ends.

### DIFF
--- a/pkg/client/daemon/session.go
+++ b/pkg/client/daemon/session.go
@@ -121,6 +121,10 @@ type session struct {
 
 	// rndSource is the source for the random number generator in the TCP handlers
 	rndSource rand.Source
+
+	// Telemetry counters for DNS lookups
+	dnsLookups  int
+	dnsFailures int
 }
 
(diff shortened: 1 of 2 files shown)


maxlegault (approved):

pkg/client/daemon/session.go
```diff
@@ -442,6 +446,10 @@ func (s *session) run(c context.Context) error {
 }
 
 func (s *session) stop(c context.Context) {
+	s.scout.Report(c, "incluster_dns_queries",
```
donnyyung, line 449: This looks good. I think if we notice we have huge delays with getting these metrics bc sessions are long, we could always send updates on some cadence (every hour or so).. but I agree that that's more complicated and we should stick to the simpler solution unless we ~ need ~ the more involved one.

donnyyung (approved):
LGTM!

PR #2301, opened 2022-01-12
Never close the Report buffer

thallgren:
Send a zero entry instead, and let processing finish when it arrives.
This ensures that stray calls to Report after the context has been
cancelled are ignored rather than triggering a panic.


donnyyung (approved):
ah good catch!

PR #2303, opened 2022-01-13
Add make target for generating LICENSES.md

ghost:
## Description

In the current cycle, infra team is working on a story to make sure that our applications have information about dependencies (OPENSOURCE.md) and the licenses those dependencies use (LICENSES.md).

In the near future, these files will be verified on every build to make sure they are up to date and if they aren't, build will fail so developer updates them. (How to will be documented soon).

This PR is the first step, adding these two files to the repo.

## Checklist


thallgren:
@aosoriodw , I only see one file added in this PR, the `LICENSES.md`. Is it supposed to be auto-generated when we generate the `OPENSOURCE.md` file? If so, is @LukeShu aware of this? I think it would be a fairly simple addition to the `go-mkopensource` tool.

ghost:
> 

@thallgren OPENSOURCE.md was already up to date.
I'm working on the part to check on every build that the files are up to date and if they aren't then build will break, forcing developer to update them.
I'll also be working on the documentation explaining the steps to update these files.

go-mkopensource was updated to generate JSON output that can be used to generate LICENSES.md.

In the future we could automate this process (maybe sending a PR) but for now we're doing what Emissay Ingress does.

thallgren:
You can disregard the check-list for this PR. It's there to remind us and OSS-users and keep us honest.

We always auto-generate the `OPENSOURCE.md` file when we change the dependencies and our CI will detect if it isn't done. Given that `go-mkopensource` already can output JSON, I'd suggest you add the generation of the `LICENSES.md` to this PR at the same place as the `OPENSOURCE.md` is generated (see `build-aux/main.mk` around line 48). Having one of the files generated but not the other would be somewhat confusing and also make us prone to ambiguities.

ghost:
I'll look into ho OPENSOURCe.md is generated and do the same for LICENSES.md and update PR.

Thanks for the info!

ghost:
@thallgren I noticed the version for the Go std library is `v1.17` in [OPENSOURCES.md](https://github.com/telepresenceio/telepresence/blob/release/v2/OPENSOURCE.md), but when I generate the file it becomes `v1.17.6`

Do you modify this file manually after it's generated?
Should OPENSOURCES.md show the full version or just the major version?

thallgren (approved):
Looks good to me now. And yes, I usually strip the patch version from the Go standard library. It picks it up from the runtime that you're using and rather than running into constant merge conflicts, I just omit that change as long as it's the patch version.

ghost:
Makes sense. I'll make changes to go-mkopensource later to strip minor version from Go std library and submit a new PRs with that change once telepresence is updated.

PR #2309, opened 2022-01-15
Refactor part of the user-daemon to make it long-running

thallgren:
## Description

This PR contains the changes needed to make the user daemon (AKA connector) long running, i.e. the process does not
terminate when a connection terminates. The changes include

- Reloading the configuration using a file watcher
- Changing the long lived traffic-manager client in accordance with what's currently connected
- A new `Disconnect` method in the gRPC (the existing `Quit` will still terminate the process)
- Fixes for some obvious wasteful resource consumption related to ingressInfo and mapped namespaces
- More consistent handling of kubernetes flags and networking flags (fixes #2096)
- Misc clean up and simplifications around how a session is started

## Checklist


rpc/connector/connector.proto
```diff
@@ -141,7 +141,7 @@ message ConnectInfo {
 
   string cluster_server = 3;
   string cluster_context = 4;
-  bool bridge_ok = 5;
+  reserved 5;
```
josecv, line 149: food for thought (not necessarily something you need to change) should we even do this?
I guess usually a number becomes reserved bc old clients might still be supplying something in it, but in this case we make the CLI be in sync with the daemons, so we could just as well delete `5` entirely. Not that it makes a huge difference, but it makes it easier to read through the protos

pkg/client/userd/k8s/outbound.go
```diff
@@ -62,27 +63,72 @@ func (kc *Cluster) RunWatchers(c context.Context) (err error) {
...
+
+func (kc *Cluster) IngressInfos(c context.Context) ([]*manager.IngressInfo, error) {
 	kc.accLock.Lock()
-	kc.mappedNamespaces = namespaces
+	if kc.ingressInfo == nil {
+		kc.accLock.Unlock()
+		ingressInfo, err := kc.detectIngressBehavior(c)
+		if err != nil {
+			// Don't fetch again unless namespaces change
+			kc.ingressInfo = []*manager.IngressInfo{}
```
josecv: hmmm what if err is transient? on retry the ingress info will still be empty, but maybe we would've been able to detect it this time

pkg/client/userd/service.go
```diff
@@ -71,8 +71,19 @@ func (s *service) SetManagerClient(managerClient manager.ManagerClient, callOpti
...
+		return s.daemonClient, nil
+	}
+	// establish a connection to the root daemon gRPC grpcService
+	dlog.Info(c, "Connecting to root daemon...")
+	conn, err := client.DialSocket(c, client.DaemonSocketName)
+	if err != nil {
+		dlog.Errorf(c, "unable to connect to root daemon: %+v", err)
+		return nil, err
+	}
+	s.daemonClient = daemon.NewDaemonClient(conn)
```
josecv, line 85: should this be protected by a lock?

josecv (approved):

rpc/connector/connector.proto
thallgren: I agree. We should fix all of these. The important thing is that the client can detect the version of a running daemon. If there's a mismatch, it'll refuse to continue anyway so it really doesn't matter if everything but the Version call is incompatible.

pkg/client/userd/service.go
thallgren: That's not  necessary. It is only called when a session is created that always happens in sequence (only one session at a time).

josecv: ah makes sense ok

josecv (approved):

PR #2311, opened 2022-01-18
Remove Telepresence dependency on datawire/kates

thallgren:
## Description

Removes Telepresence dependency on the `datawire/kates` package in favor of a more OO-oriented approach using
bare metal  `kubernetes.Interface`.

I think the best approach for reviewing this is to make two iterations over the commits because there are methods that were added temporarily to make the commits build.

## Checklist


pkg/client/userd/trafficmgr/traffic_manager.go
```diff
@@ -252,7 +254,11 @@ func connectMgr(c context.Context, cluster *k8s.Cluster, installID string, svc S
 	}
 
 	dlog.Debug(c, "traffic-manager started, creating port-forward")
-	grpcDialer, err := dnet.NewK8sPortForwardDialer(c, cluster.ConfigFlags, cluster.Client())
+	restConfig, err := cluster.ConfigFlags.ToRESTConfig()
+	if err != nil {
+		return nil, errors.Wrap(err, "ToREstConfig")
```
josecv: dumb nit but this should probably be `ToRESTConfig`

josecv (approved):
lgtm! if only go had generics i suspect we'd have even less code but it's a v nice change regardless 🙃

PR #2319, opened 2022-01-22
Add a WorkloadsWatcher function

thallgren:
## Description

In order to provide a watcher for workloads with reasonable performance it needs to watch the services and workloads of all mapped namespaces. So that's what this PR is adding. Well, almost. There's a conceptual difference between a mapped namespace (a namespace that can be reached with outbound traffic) and an accessible mapped namespace (a namespace that where user is allowed to watch and get services and workloads). In this PR, watchers are added for the latter.

The implementation focus on a bare-metal `cache.Store` and `cache.Controller` to avoid unnecessary overhead and to achieve correct context propagation and logging. No additional caching is made outside the `Store` and each watcher uses a tailored `equals` method to decide whether or not a resources has changed to avoid that snapshots are sent for changes that Telepresence doesn't care about.

Since the PR starts with moving some code into places that are more suitable, I recommend reviewing one commit at a time.

## Checklist


josecv (approved):

pkg/client/userd/trafficmgr/traffic_manager.go
```diff
@@ -400,148 +410,130 @@ func (tm *TrafficManager) Run(c context.Context) error {
...
-				if err != nil {
-					continue
-				}
-				if len(matchingSvcs) == 0 {
-					reason = "No service with matching selector"
-				}
+		for _, workload := range wls {
+			name := workload.GetName()
+			dlog.Debugf(ctx, "Getting info for %s %s.%s, matching service %s.%s", workload.GetKind(), name, workload.GetNamespace(), svc.Name, svc.Namespace)
+			wlInfo := &rpc.WorkloadInfo{
```
josecv, line 439: it'd be good if the service, with its UID, name and namespace were stored (could be optional) in the WorkloadInfo so that we can work our way back from service -> workload

josecv (changes requested):

pkg/client/userd/trafficmgr/traffic_manager.go
thallgren: Agree. I'll add that in a future commit.

josecv: it's ok i can take care of it, i think i'll need to make some changes to this code anyway @thallgren

PR #2325, opened 2022-01-26
Telepresence pro 

donnyyung:
## Description

## Checklist


pkg/client/cli/cliutil/systema.go
```diff
@@ -127,3 +141,92 @@ func GetCloudLicense(ctx context.Context, outputFile, id string) (string, string
...
+		if err != nil {
+			return errcat.Unknown.Newf("error reading input: %s", err)
+		}
+
+		// If the user doesn't want to install it, then we we'll proceed
+		// with launching the daemon normally
+		reply = strings.TrimSpace(reply)
+		if reply == "n" {
+			return nil
+		}
```
josecv: i feel like we should rather return if `reply != "y"` otherwise we risk installing it by accident. This is in line with how eg terraform works

josecv (approved):
one minor thing but otherwise lgtm

pkg/client/cli/cliutil/systema.go
donnyyung: ah very smart! ty

PR #2328, opened 2022-01-27
Allow personal intercepts based on URL path

thallgren:
## Description

Adds the following new flags to the `http` mechanism:

- path-equal Only intercept if path is equal to the given value
- path-prefix Only intercept if path is prefixed by the given value
- path-regex Only intercept if the path is matched by the given regexp

Also updates the Telepresence API service to take an additional `path=xxx` URL parameter.
 
## Checklist


pkg/matcher/request.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+// NewRequest creates a new Request based on the values of the given map. Aside from http headers,
+// the map may contain one of three special keys.
+//
+//   :path-equal: path will match if equal to the value
+//   :path-prefix: path will match prefixed by the value
+//   :path-regex: path will match it matches the regexp value
+//
+func NewRequest(m map[string]string) (Request, error) {
```
donnyyung, line 35: This suggests that we can only have one of the `--path-*` flags at a time, do we need to add logic to prevent users from using these flags in combination with each other?  bc right now I think it'd use the `pm` value of the last one it found in the map which may lead to unexpected behavior, unless I'm missing where we only allow one of those keys to be in the dict.

thallgren: I thought about that, but it's tricky because the current extension mechanism doesn't really allow for adding a post-validator that checks the resulting flags and it seemed a bit of an overkill to implement that at this point.

I would like to remove the extension mechanism altogether at some point and move all things dealing with `--http-xxx` over to the proprietary daemon. Perhaps we could defer this improvement until we do that?

donnyyung: Yeah that seems reasonable.

donnyyung (approved):

PR #2335, opened 2022-01-28
Creating an intercept returns the service port number as part of interceptResult

njayp:

## Description

The goal of the PR is to have pro-intercepts created by the client use the cloud to determine the default ingress values. This requires two major parts, first to get the resolved service port number from the created intercept, then to use that service port number to call the cloud in the creation of the preview url.

## Checklist


pkg/client/userd/trafficmgr/install.go
```diff
@@ -291,14 +300,27 @@ func (ki *installer) EnsureAgent(c context.Context, obj k8sapi.Workload,
 	var svc k8sapi.Object
 	autoInstall, err := useAutoInstall(podTemplate)
 	if err != nil {
-		return "", "", err
+		return "", "", 0, err
+	}
+
+	matchingSvc, err := install.FindMatchingService(c, portNameOrNumber, svcName, namespace, podTemplate.Labels)
```
thallgren: The creation of a personal intercept was recently divided into two steps to get rid of some friction. One to ensure that an intercept is indeed possible (the `trafficmgr.CanIntercept()` which is executed before any questions are asked regarding ingress), and one that actually creates the intercept (the `trafficmgr.AddIntercept()`). The new actions performed here to find the matching service and its port, should probably be performed as part of the first step so that errors are caught early.

thallgren (changes requested):
The title of this PR is nondescriptive.

pkg/client/userd/trafficmgr/install.go
njayp: that makes sense

pkg/client/userd/trafficmgr/install.go
```diff
@@ -210,20 +215,13 @@ var agentNotFound = errors.New("no such agent")
 
 func (ki *installer) getSvcForInjectedPod(
```
thallgren: Perhaps I'm missing the point here. But what does this method actually do now? If anything, the name `getSvcForInjectedPod` is gravely misleading.

thallgren (changes requested):

pkg/client/userd/trafficmgr/install.go
njayp: I only changed the input to this function, as we have already found the matching service. Its logic remains (hopefully) unchanged

njayp: Now that this function no longer finds the matching service, but instead only rolls the pod if the agent is missing, perhaps it should be renamed. I can rename it "ensureinjectedagent" or something

pkg/client/userd/systema_proxy/service_proxy.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/telepresenceio/telepresence/v2/pkg/client/cli/cliutil"
+	"github.com/telepresenceio/telepresence/v2/pkg/client/userd/trafficmgr"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
+	"google.golang.org/protobuf/types/known/emptypb"
+)
+
+// This proxy is a temporary solution to allow the client to get the
+// preferred ingress info from the cloud
+type SystemAProxy struct {
```
thallgren: `systema_proxy.SystemAProxy` breaks the Golang "no stuttering" rule. How about naming the package `systema` and the type `Proxy`? You'd call `systema.NewProxy()` to get a `systema.Proxy`. Then use an import alias to resolve the rpc name collision:
```
rpc "github.com/telepresenceio/telepresence/rpc/v2/systema"
```

pkg/client/userd/systema_proxy/service_proxy.go
```diff
@@ -0,0 +1,88 @@
...
+	return &SystemAProxy{
+		systemacli: systema.NewConnSystemAProxyClient(conn),
+	}, nil
+}
+
+func (s *SystemAProxy) ResolveIngressInfo(ctx context.Context, req *systema.IngressInfoRequest) (*systema.IngressInfoResponse, error) {
+	return s.systemacli.ResolveIngressInfo(ctx, req)
+}
+
+type systemaCredentials string
```
thallgren: Just `credentials` in a `systema` package.

pkg/client/userd/systema_proxy/service_proxy.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+func (c systemaCredentials) RequireTransportSecurity() bool {
+	return true
+}
+
+// ConnectSystemA tries to create a connection to the given systemaURL
+// using the apiKey for authorization and returns the connection if it
+// was successful in creating one
+func ConnectSystemA(ctx context.Context, systemaURL, apiKey string) (*grpc.ClientConn, error) {
```
thallgren: and this would probably be `systema.Connect`

pkg/client/userd/systema_proxy/service_proxy.go
```diff
@@ -0,0 +1,88 @@
...
+		grpc.WithTransportCredentials(credentials.NewTLS(&tls.Config{ServerName: u.Hostname()})),
+		grpc.WithPerRPCCredentials(creds))
+	if err != nil {
+		dlog.Errorf(ctx, "error establishing gRPC connection: %s", err)
+		return nil, err
+	}
+	return conn, nil
+}
+
+func ConnectSessionToSystemA(ctx context.Context, session trafficmgr.Session) (*grpc.ClientConn, error) {
```
thallgren: together with `systema.ConnectSession`

rpc/systema/cli2systema.proto
```diff
@@ -23,3 +23,41 @@ message CommandMessageResponse {
   // intercept is the message to give when a user performs an intercept command.
   string intercept = 1;
 }
+
+// copied from systemA
```
thallgren: Why would we ever copy the declaration of an rpc service? Why not use the original?

pkg/client/userd/trafficmgr/install.go
thallgren: Yes. But with camel-case `ensureInjectedAgent`.

pkg/client/userd/trafficmgr/install.go
```diff
@@ -275,12 +273,59 @@ func useAutoInstall(podTpl *core.PodTemplateSpec) (bool, error) {
...
 
+type ServiceProps struct {
+	Service            *core.Service
+	ServicePort        *core.ServicePort
+	Container          *core.Container
+	ContainerPortIndex int
+}
+
+// ExploreSvc finds the matching service, its comtainters, and their ports
+func ExploreSvc(c context.Context, portNameOrNumber, svcName string, obj k8sapi.Workload) (*ServiceProps, error) {
```
thallgren: Unless there's a plan to later use this function from other packages, I suggest not exporting it. Use the name `exploreSvc` and return the type `*serviceProps`.

thallgren (dismissed):

pkg/client/userd/systema_proxy/service_proxy.go
thallgren: Why do we need this proxy in the first place? Why can't the intended caller instead call directly to SystemA?

pkg/client/userd/trafficmgr/install.go
```diff
@@ -210,20 +215,13 @@ var agentNotFound = errors.New("no such agent")
...
 	c context.Context,
-	name,
-	namespace,
-	svcName,
-	portNameOrNumber string,
+	svc *core.Service,
+	name, namespace string,
 	podTemplate *core.PodTemplateSpec,
 	obj k8sapi.Object,
 ) (*core.Service, error) {
```
thallgren: The method should only return `error` now. There's no point in always returning the service that was passed in because the caller already has it. Returning it here gives the impression that the method might somehow change it.

pkg/client/cli/cmds_intercept.go
```diff
@@ -594,9 +596,20 @@ func (is *interceptState) EnsureState(ctx context.Context) (acquired bool, err e
...
+			if err != nil {
+				return false, err
+			}
+			spec.Ingress = ingress
 		}
+		/*
+			if err = is.ensureIngress(ctx, args, ir, needLogin); err != nil {
+				return false, err
+			}
+		*/
```
josecv: let's just remove this instead of commenting it out pls

rpc/systema/cli2systema.proto
josecv: there's precedent for doing this @thallgren -- it's not trivial to import the service as it would require a private repo to be pulled in, which would prevent open source contributors from building telepresence

josecv: eventually the intercept command can be pushed into a daemon command and at that point the OSS functionality can be separated from the proprietary functionality and we can fix this

josecv (changes requested):
i find the use of a systema proxy a little concerning, i think i would prefer it if you just added an RPC to the connector that executes that call, instead of a whole proxy mechanism (i.e. if systema can't be called from the CLI, add a `ResolveIngressInfo` to the connector and have that call systema)

rpc/systema/cli2systema.proto
thallgren: Ok, that makes sense.

josecv (approved):

PR #2340, opened 2022-02-01
Fix proper session shutdown on process shutdown

thallgren:
The process shutdown didn't wait for the session to complete
its shutdown, leaving static routes behind.

### DIFF
--- a/pkg/client/rootd/session.go
+++ b/pkg/client/rootd/session.go
@@ -420,7 +420,7 @@ func (s *session) onClusterInfo(ctx context.Context, mgrInfo *manager.ClusterInf
 }
 
 func (s *session) run(c context.Context) error {
-	g := dgroup.NewGroup(c, dgroup.GroupConfig{ShutdownOnNonError: true})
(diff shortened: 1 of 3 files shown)


pkg/client/rootd/session.go
```diff
@@ -448,6 +448,8 @@ func (s *session) run(c context.Context) error {
 }
 
 func (s *session) stop(c context.Context) {
+	dlog.Debug(c, "Brining down TUN-device")
```
donnyyung, line 451: nit: Bringing*

pkg/client/userd/service.go
```diff
@@ -135,13 +135,15 @@ func (s *service) configReload(c context.Context) error {
 func (s *service) manageSessions(c context.Context, sessionServices []trafficmgr.SessionService) error {
 	// The d.quit is called when we receive a Quit. Since it
```
donnyyung, line 136: Is this "d.quit" the "s.quit" or is it something else? If the former, I think we should update the comment just for clarity

donnyyung (approved):
Just some comment-related points but otherwise seems sane to me!

PR #2343, opened 2022-02-01
Fix bogus proxy subnet status printout in daemon status

thallgren:
## Description

The `AlsoProxySubnets` was listed after the "Never Proxy" heading and
the `NeverProxySubnets` were never listed.


donnyyung (approved):
Nice catch!

PR #2346, opened 2022-02-01
Try preview provider using SNI relay.

rhs:
## Description

This PR enables the kubeception preview provider along with the SNI relay feature flag. Previously the preview provider would not work with the telepresence test suite due to intermittent "404 NR"s that were introduced by the proxies in front of the API server. The SNI relay feature flag enables a fix for this problem by routing connections directly to the API server using SNI.

I'd like to get broader telepresence CI usage of the preview provider (with the SNI relay feature flag enabled) because (a) it should be faster and more reliable than the legacy provider at this point and (b) once this has proven stable enough for the whole team to depend on, I can promote the preview provider to production and retire the legacy provider.

## Checklist

### DIFF
--- a/.github/actions/prepare-kluster/action.yaml
+++ b/.github/actions/prepare-kluster/action.yaml
@@ -40,7 +40,7 @@ runs:
       run: |
         export KUBECONFIG="$HOME/kubeconfig"
         kubectl apply -f build-aux/image-importer.yaml
+        kubectl rollout status -w deployment/image-importer
         POD_NAME=$(kubectl get pod -ojsonpath='{.items[0].metadata.name}' -l app=image-importer)
-        kubectl wait --for=condition=ready pod $POD_NAME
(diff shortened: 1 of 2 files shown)


donnyyung (approved):
I think Thomas had more context on the "SNI" bits, but this seems good to me!

rhs:
> I think Thomas had more context on the "SNI" bits, but this seems good to me!

I'll wait for @thallgren to have a quick look just in case.

FYI the "enableSNIRelay" parameter is a feature flag that controls an aspect of the kubeception implementation (the use of SNI to route to the API server). The reason it is required for the telepresence tests is that (as @thallgren unfortunately discovered) the telepresence tests happen to be sensitive to a bug in the initial routing strategy used by the preview provider.

Once it is sufficiently battle tested, I intend to make the preview provider with the SNI routing strategy the default (and eventually eliminate the alternative), but I didn't want to switch up the implementation behavior from underneath you. Hence this PR.

PR #2348, opened 2022-02-02
Create some new metrics

josecv:
## Description

## Checklist

### DIFF
--- a/pkg/client/userd/grpc.go
+++ b/pkg/client/userd/grpc.go
@@ -27,10 +27,12 @@ import (
 	"github.com/telepresenceio/telepresence/v2/pkg/client/cli/cliutil"
 	"github.com/telepresenceio/telepresence/v2/pkg/client/errcat"
 	"github.com/telepresenceio/telepresence/v2/pkg/client/logging"
+	"github.com/telepresenceio/telepresence/v2/pkg/client/scout"
 	"github.com/telepresenceio/telepresence/v2/pkg/client/userd/auth"
 	"github.com/telepresenceio/telepresence/v2/pkg/client/userd/commands"
 	"github.com/telepresenceio/telepresence/v2/pkg/client/userd/trafficmgr"
 	"github.com/telepresenceio/telepresence/v2/pkg/k8sapi"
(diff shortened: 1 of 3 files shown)


pkg/client/userd/grpc.go
```diff
@@ -52,6 +55,18 @@ func (s *service) logCall(c context.Context, callName string, f func(context.Con
 }
 
 func (s *service) withSession(c context.Context, callName string, f func(context.Context, trafficmgr.Session) error) (err error) {
+	defer func() {
```
thallgren: Not sure if this is that useful. It affects the subset of the grpc calls that *require* a session. It does not affect calls like `Connect`, `Disconnect`, `Login`, or `Quit`. It does however affect SystemA's call to `WatchWorkloads`. A side-effect is that whenever a new call is added, the implementor must remember to add an entry in the `TELEMETRY.md` (I guess the names implicitly added here should go in the too?).

In general, I think we should avoid generic approaches like this and instead focus on what we know is useful such as the outcome of intercept calls or connect attempts. The scout calls are not lightweight. They do affect performance on our side and as the number of users grow, they will consume a significant amount of resources from the metriton service.

thallgren (changes requested):

pkg/client/userd/grpc.go
```diff
@@ -52,6 +55,18 @@ func (s *service) logCall(c context.Context, callName string, f func(context.Con
...
+		name := strcase.ToSnake(callName)
+		traits := []scout.Entry{}
+		if err != nil {
+			name += "_fail"
+			traits = append(traits, scout.Entry{Key: "error", Value: err.Error()})
+		} else {
+			name += "_success"
```
thallgren: This assumption is not true because many (most) errors are reported in the returned message, not as an gRPC `error`. We should in fact clean this up further. Errors here should only be `grpc.status` errors with proper codes.

pkg/client/userd/grpc.go
josecv: ok makes sense, i'll instrument the specific RPCs that we're interested in which are `CanIntercept` and `CreateIntercept`

thallgren (approved):

PR #2357, opened 2022-02-07
Throw away repeated SYN packages during connect

thallgren:
A slow connect will cause repeated SYN-packages to arrive while the
connection is establishing. Those packages should be disregarded instead
of handled once the connection is established.


josecv (approved):

thallgren:
Rebased on top of release/v2 to get the OPENSOURCE.md license versions correct.

PR #2365, opened 2022-02-09
Prevent leaking of the subscribe goroutine

josecv:
## Description

## Checklist

### DIFF
--- a/pkg/client/userd/grpc.go
+++ b/pkg/client/userd/grpc.go
@@ -61,7 +61,8 @@ func (s *service) withSession(c context.Context, callName string, f func(context
 			return
 		}
 		defer func() { err = callRecovery(recover(), err) }()
-		err = f(s.sessionContext, s.session)
+		ctx := dgroup.WithGoroutineName(s.sessionContext, "-"+callName)
+		err = f(ctx, s.session)
 	})
 	return
 }
--- a/pkg/client/userd/grpc.go
+++ b/pkg/client/userd/grpc.go
@@ -41,8 +41,24 @@ func callRecovery(r interface{}, err error) error {
 	return err
 }
 
+type reqNumberKey struct{}
+
+func getReqNumber(ctx context.Context) int64 {
+	num := ctx.Value(reqNumberKey{})
+	if num == nil {
+		return 0
+	}
+	return num.(int64)
+}
+
+func withReqNumber(ctx context.Context, num int64) context.Context {
+	return context.WithValue(ctx, reqNumberKey{}, num)
+}
+
 func (s *service) callCtx(ctx context.Context, name string) context.Context {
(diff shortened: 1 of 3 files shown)


pkg/client/userd/grpc.go
```diff
@@ -61,7 +61,8 @@ func (s *service) withSession(c context.Context, callName string, f func(context
 			return
 		}
 		defer func() { err = callRecovery(recover(), err) }()
-		err = f(s.sessionContext, s.session)
+		ctx := dgroup.WithGoroutineName(s.sessionContext, "-"+callName)
```
thallgren: I think you must do what `callCtx` does here and use an integer suffix, to be able to differentiate between parallel calls.

pkg/client/userd/trafficmgr/traffic_manager.go
```diff
@@ -487,6 +487,10 @@ func (tm *TrafficManager) getInfosForWorkloads(
 }
 
 func (tm *TrafficManager) WatchWorkloads(c context.Context, wr *rpc.WatchWorkloadsRequest, stream WatchWorkloadsStream) error {
+	c, cancel := context.WithCancel(c)
+	// We need to make sure the subscription ends when we leave this method, since this is the one consuming the snapshotAvailable channel.
+	// Otherwise, the goroutine that writes to the channel will leak.
+	defer cancel()
 	snapshotAvailable := tm.wlWatcher.subscribe(c)
```
thallgren: This won't fly. We can't mix wlWatcher's using this context with wlWatcher's using the session context. This has to use the `sessionContext`. It will linger, for sure, and we might want to address that at some point.

thallgren (changes requested):

pkg/client/userd/grpc.go
```diff
@@ -61,7 +77,9 @@ func (s *service) withSession(c context.Context, callName string, f func(context
 			return
 		}
 		defer func() { err = callRecovery(recover(), err) }()
-		err = f(s.sessionContext, s.session)
+		num := getReqNumber(c)
```
thallgren, line 80: But of course. Good solution.

pkg/client/userd/trafficmgr/traffic_manager.go
```diff
@@ -487,7 +487,11 @@ func (tm *TrafficManager) getInfosForWorkloads(
 }
 
 func (tm *TrafficManager) WatchWorkloads(c context.Context, wr *rpc.WatchWorkloadsRequest, stream WatchWorkloadsStream) error {
-	snapshotAvailable := tm.wlWatcher.subscribe(c)
+	sCtx, sCancel := context.WithCancel(c)
+	// We need to make sure the subscription ends when we leave this method, since this is the one consuming the snapshotAvailable channel.
+	// Otherwise, the goroutine that writes to the channel will leak.
+	defer sCancel()
```
thallgren, line 493: I don't think this is needed actually. gRPC will cancel the context when the call ends. But it's harmless.

thallgren (approved):

PR #2368, opened 2022-02-09
Don't use a default for userDaemonBinary

donnyyung:

## Description

Setting a default in the config leads to some non-ideal behavior with certain IDEs, so let's not do that.

## Checklist

### DIFF
--- a/pkg/client/cli/cliutil/connector.go
+++ b/pkg/client/cli/cliutil/connector.go
@@ -52,6 +52,12 @@ func withConnector(ctx context.Context, maybeStart bool, withNotify bool, fn fun
 		return fn(ctx, connectorClient)
 	}
 
+	// If the UserDaemonBinary isn't set, use the same executable as the
+	// CLI binary
+	connectorDaemon := client.GetConfig(ctx).Daemons.UserDaemonBinary
+	if connectorDaemon == "" {
+		connectorDaemon = client.GetExe()
+	}
 	var conn *grpc.ClientConn
 	started := false
 	for {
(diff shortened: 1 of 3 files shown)


pkg/client/cli/cliutil/connector.go
```diff
@@ -84,9 +89,8 @@ func withConnector(ctx context.Context, maybeStart bool, withNotify bool, fn fun
 	ctx = context.WithValue(ctx, connectorConnCtxKey{}, conn)
 	connectorClient := connector.NewConnectorClient(conn)
 	if !started {
-		daemonBinary := client.GetConfig(ctx).Daemons.UserDaemonBinary
 		// Ensure that the already running daemon has the correct version
-		if err := versionCheck(ctx, "User", daemonBinary, connectorClient); err != nil {
+		if err := versionCheck(ctx, "User", client.GetConfig(ctx).Daemons.UserDaemonBinary, connectorClient); err != nil {
```
thallgren, line 93: Shouldn't this use the `connectorDaemon` initialized around line 59? The value from the config isn't guaranteed.

thallgren (changes requested):

pkg/client/cli/cliutil/connector.go
donnyyung: Yeah which means we won't do this check: https://github.com/telepresenceio/telepresence/blob/release/v2/pkg/client/cli/cliutil/version_check.go#L29-L32 which is what I thought you wanted when the config value wasn't set?

thallgren: Oh, I didn't realize that `versionCheck` compared the binaries too. Then this makes sense.

thallgren (approved):

PR #2369, opened 2022-02-09
Deprecate --http-match in favor of --http-header

thallgren:
## Description

The flag `--http-match` is renamed to `--http-header` to avoid confusion
as the new `--http-path-xxx` flags are introduced. A command line like:

    telepresence intercept --http-match=all --http-path-prefix=/api ...

didn't really make sense. Is it matching all requests, or just the
ones prefix with '/api'? Using `--http-header` makes it clearer:

    telepresence intercept --http-header=all --http-path-prefix=/api ...

The old flag can still be used but will render a deprecation warning
that suggests using the new flag. It no longer shows up in the help.

### DIFF
--- a/pkg/client/cli/extensions/extensions.go
+++ b/pkg/client/cli/extensions/extensions.go
@@ -202,8 +202,13 @@ func LoadExtensions(ctx context.Context, existingFlags *pflag.FlagSet) (es *Exte
 					flagname,
 					err)
 			}
-			es.flags.Var(val, mechname+"-"+flagname,
-				flagdata.Usage+fmt.Sprintf(" (implies %q)", "--mechanism="+mechname))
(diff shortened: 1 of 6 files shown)


pkg/client/cli/extensions/extensions.go
```diff
@@ -427,4 +432,10 @@ type FlagInfo struct {
 	// Default is the default value for this flag.  This field is optional; if it isn't
 	// specitified then the zero value is used.
 	Default json.RawMessage `json:"default,omitempty"`
+
+	//nolint:gocritic // this is not a deprecation comment
```
josecv, line 436: lol

josecv (approved):
nice one on the compatibility checks for the flags

PR #2371, opened 2022-02-10
Improved metriton reporting

thallgren:
## Description

- Adds more metriton reporting around gRPC calls
- Ensures that metriton reporter is flushed before process terminates
- Fixes problem with false successful login reports

### DIFF
--- a/pkg/client/scout/reporter.go
+++ b/pkg/client/scout/reporter.go
@@ -6,6 +6,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
+	"time"
 
 	"github.com/google/uuid"
 
@@ -29,6 +30,7 @@ type bufEntry struct {
 type Reporter struct {
 	index    int
 	buffer   chan bufEntry
+	done     chan struct{}
 	reporter *metriton.Reporter
 }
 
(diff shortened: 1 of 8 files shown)


pkg/client/scout/reporter.go
```diff
@@ -200,6 +203,22 @@ func (r *Reporter) Start(ctx context.Context) {
...
 
+func (r *Reporter) Close() {
+	// Send a zeroed bufEntry here instead of closing the buffer so that
+	// any stray Reports that arrive after the context is cancelled aren't
+	// sent on a closed channel
+	select {
+	case r.buffer <- bufEntry{}:
+	default:
+	}
+	// Wait for the done channel to close. Give up after 2 seconds (that
```
josecv, line 214: this should probably read 3 seconds as does the code

josecv (approved):

pkg/client/scout/reporter.go
thallgren: Indeed.

thallgren: I fix that in the next PR which unfortunately builds on this one.

PR #2402, opened 2022-02-16
DNS improvements and bug-fixes

thallgren:
## Description

This PR contains the following:
- A fix for an earlier regression that cause the DNS recursion check to fail (never released).
- Improved usage of iptables for the overriding resolver
- A fix to avoid the overridden resolver's `search` declared in the `/etc/resolv.conf` to affect the TP resolver
- Much improved DNS logging

### DIFF
--- a/pkg/client/rootd/dns/server_linux.go
+++ b/pkg/client/rootd/dns/server_linux.go
@@ -107,7 +107,16 @@ func (s *Server) runOverridingServer(c context.Context, dev *vif.Device) error {
 				fields := strings.Fields(line)
 				s.config.LocalIp = net.ParseIP(fields[1])
 				dlog.Infof(c, "Automatically set -dns=%s", net.IP(s.config.LocalIp))
-				break
+			}
+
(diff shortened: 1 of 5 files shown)


pkg/client/rootd/dns/server_linux.go
```diff
@@ -288,42 +288,45 @@ func runNatTableCmd(c context.Context, args ...string) error {
...
-	return runNatTableCmd(c, "-A", tpDNSChain,
+	if err = runNatTableCmd(c, "-A", tpDNSChain,
 		"-p", "udp",
 		"--dest", dnsIP.String()+"/32",
 		"--dport", "53",
-		"-j", "REDIRECT",
-		"--to-ports", strconv.Itoa(toPort),
-	)
+		"-j", "DNAT",
+		"--to-destination", toAddr.String(),
```
josecv, line 323: eh pretty sure a REDIRECT is just a dnat with an implicit host

josecv (approved):

PR #2407, opened 2022-02-18
Ensure that zero entries in the Config are omitted during marshal

thallgren:
- Replaces the approach of defining `client.Config` defaults using environment variables with getters.
- Adds an `IsZero()` method to `client.Config` entries that have a default value that is different from the go zero value.
- Save the config.yml as config.yml.bak when updating it


maxlegault (approved):

thallgren:
Merged

PR #2410, opened 2022-02-18
Only use ipv4 addresses for the resolv.conf override

josecv:

## Description

## Checklist

### DIFF
--- a/pkg/client/rootd/dns/server_linux.go
+++ b/pkg/client/rootd/dns/server_linux.go
@@ -105,8 +105,12 @@ func (s *Server) runOverridingServer(c context.Context, dev *vif.Device) error {
 		for _, line := range strings.Split(string(dat), "\n") {
 			if strings.HasPrefix(strings.TrimSpace(line), "nameserver") {
 				fields := strings.Fields(line)
-				s.config.LocalIp = net.ParseIP(fields[1])
-				dlog.Infof(c, "Automatically set -dns=%s", net.IP(s.config.LocalIp))
+				ip := net.ParseIP(fields[1])
+				if ip.To4() != nil {
+					s.config.LocalIp = ip.To4()
+					dlog.Infof(c, "Automatically set -dns=%s", net.IP(s.config.LocalIp))
+					break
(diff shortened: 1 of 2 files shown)


maxlegault (approved):

pkg/client/rootd/dns/server_linux.go
```diff
@@ -105,8 +105,12 @@ func (s *Server) runOverridingServer(c context.Context, dev *vif.Device) error {
 		for _, line := range strings.Split(string(dat), "\n") {
 			if strings.HasPrefix(strings.TrimSpace(line), "nameserver") {
 				fields := strings.Fields(line)
-				s.config.LocalIp = net.ParseIP(fields[1])
-				dlog.Infof(c, "Automatically set -dns=%s", net.IP(s.config.LocalIp))
+				ip := net.ParseIP(fields[1])
+				if ip.To4() != nil {
+					s.config.LocalIp = ip.To4()
+					dlog.Infof(c, "Automatically set -dns=%s", net.IP(s.config.LocalIp))
+					break
```
thallgren: This break is wrong. It will prevent parsing of "search" if that comes after "nameserver".

thallgren (changes requested):

pkg/client/rootd/dns/server_linux.go
thallgren: What would it take to actually support IPv6?

josecv: ah ok good catch thank you

josecv: > What would it take to actually support IPv6?

not sure really i think there might be a way to make iptables work with it (i believe it's a matter of using iptables6 but i've never done it), but rn it fails

thallgren (approved):

pkg/client/rootd/dns/server_linux.go
thallgren: Not for this release, for sure.

thallgren:
Merged

PR #2422, opened 2022-02-19
Improve flow when upgrading the pro-daemon

### LINKED ISSUE #2421: User Daemon Version Mismatch
**Describe the bug**
After upgrading from Telepresence 2.5.0 to 2.5.1, I get the following error:

```
❯ telepresence connect
telepresence: error: version mismatch. Client v2.5.1 != User Daemon v2.5.0, please run 'telepresence quit -u' and reconnect
```

After running `telepresence quit -u`, I get the same error again.

# To see all options, run the following command
telepresence gather-logs --help
```

**To Reproduce**
1. Install Telepresence 2.5.0
2. Run `telepresence login` to download the pro daemon
3. Install Telepresence 2.5.1
4. Run `telepresence connect`

**Expected behavior**
A compatible daemon version should automatically be downloaded

2.5.0

thallgren:
When a version mismatch is detected for the pro-daemon, the user will
now always be queried if it's ok to update. If it is, then a running
user daemon is stopped, a new one is downloaded, and then the new daemon
is started if it was previously running.


njayp (approved):
The way the context was used to store/update the conn pointer was interesting

PR #2423, opened 2022-02-19
Always pick the first nameserver in resolv.conf

thallgren:


pkg/client/rootd/dns/server_linux.go
```diff
@@ -103,7 +103,7 @@ func (s *Server) runOverridingServer(c context.Context, dev *vif.Device) error {
 			return err
 		}
 		for _, line := range strings.Split(string(dat), "\n") {
-			if strings.HasPrefix(strings.TrimSpace(line), "nameserver") {
+			if s.config.LocalIp == nil && strings.HasPrefix(strings.TrimSpace(line), "nameserver") {
```
njayp, line 106: I probably missed something, but I don't see how `s.config.LocalIp` could change between this line and line 100

njayp (changes requested):

pkg/client/rootd/dns/server_linux.go
josecv: @njayp this is happening in a loop -- line 110 will set `config.LocalIp` if it's an ipv4 address. The purpose of this guard is to prevent it from being overriden from that initial value

josecv (approved):

njayp (approved):

PR #2424, opened 2022-02-19
Send "--header" instead of "--match" to agents more recent than 1.11.8

thallgren:


njayp (approved):

pkg/client/userd/trafficmgr/intercept.go
```diff
@@ -386,27 +383,39 @@ func makeFlagsCompatible(agentVer *semver.Version, args []string) ([]string, err
...
+		m["header"] = ds
 	}
 	if agentVer != nil {
-		if agentVer.LE(semver.MustParse("1.11.8")) {
-			for ma := range m {
-				switch ma {
-				case "meta", "path-equal", "path-prefix", "path-regex":
-					return nil, errcat.User.New("--http-" + ma)
-				}
+		if agentVer.LE(semver.MustParse("1.11.9-rc.3")) {
```
njayp: Just curious why this isn't `1.11.9`

thallgren: Hmm, it should be. 1.11.9 wasn't released yet when I created this. I'll change it.

PR #2444, opened 2022-02-28
Enable nss_wrapper in images to enable use of SecurityContext

thallgren:
This commit replaces the `ko` based build with a three-step docker
build. All steps are based on Alpine.

1. `tel2-base` Builds the `nss_wrapper` from source and copies the password
   template and script to activate it.
2. `tel2-build` Builds the `traffic` image from source. The step uses
   `go mod download` to ensure that dependencies are cached by docker
   and then copies the sources needed to build the binary.
3. `tel2` starts with clean alpine image and copies the necessary built
   artifacts from `tel2-base` and `tel2-build`

Everything related to `ko` is removed from the repository.


base-image/startscript.sh
```diff
@@ -0,0 +1,11 @@
+#! /bin/sh
```
maxlegault: I think there's an extra space here between the shebang and /bin/sh

maxlegault: And to be safer, we should probably add a `set -eu` to avoid unexpected behaviours for missing variables or continued execution over a previous failure.

maxlegault (commented):
Minor comment in my review, it looks good to me in general, but I'd like a review from someone who's more used to this code

josecv (approved):
This looks good to me but have you tested it with istio? I'd just like to be sure that the firewall rules with os.Getuid() won't conflict (they shouldn't bc istio uses a custom UID)

PR #2445, opened 2022-02-28
Add install id, user id, account id and json flag to status

maxlegault:
## Description

This PR adds the install id, user id and account id as metadata
returned by the status command. It also adds support for a `--json` flag
so that its output can be easily consumed by other processes.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### 2.5.4 (TBD)
 
+- Change: The status command includes the install id, user id and account id in its result, and can print output as JSON
+
 - Bugfix: Removed a bad concatenation that corrupted the output path of `telepresence gather-logs`.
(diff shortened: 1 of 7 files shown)


thallgren (changes requested):
Looks good to me except that the fairly large structs passed by value (both as return and as method parameters). Doing so implies a full by value copy and I think that's uncalled for. So please use pointers throughout.

maxlegault:
> Looks good to me except that the fairly large structs passed by value (both as return and as method parameters). Doing so implies a full by value copy and I think that's uncalled for. So please use pointers throughout.

@thallgren I addressed this issue in addition to fixing the usage of `systeminfo` in Windows as discussed on Slack.

thallgren (approved):
Looks good!

PR #2449, opened 2022-03-03
Allow random UID at all times

thallgren:
## Description

The `sftp-server` didn't play well with randomly generated UID's so this PR replaces it with a pure go sftp server.

The PR also contains a set of build improvements, mainly targeted at enabling CGO in the traffic binary (to get DNS resolutions correct at all times) and also allowing further improvements in downstream projects (the check-in of the telepresence-chart.tgz)

## Checklist


cmd/traffic/cmd/agent/agent.go
```diff
@@ -179,11 +181,18 @@ func SftpServer(ctx context.Context, sftpPortCh chan<- int32) error {
 			return nil
 		}
 		go func() {
-			dlog.Debugf(ctx, "Serving sshfs connection from %s", conn.RemoteAddr())
-			err := dpipe.DPipe(ctx, conn, "/usr/lib/ssh/sftp-server")
+			s, err := sftp.NewServer(conn)
```
josecv, line 184: this is so much better lol

cmd/traffic/cmd/agentinit/agent_init.go
```diff
@@ -30,7 +29,7 @@ func configureIptables(ctx context.Context, iptables *iptables.IPTables, loopbac
 	// However, if there's a service mesh we want to make sure we don't bypass the mesh, so the traffic will flow request -> mesh -> agent -> app
 	appPort := strconv.Itoa(cfg.AppPort)
 	agentPort := strconv.Itoa(cfg.AgentPort)
-	agentUID := strconv.FormatInt(install.AgentUID, 10)
+	agentUID := strconv.Itoa(os.Getuid())
```
josecv, line 32: hmmmm the problem with this is that the application container will be running under the same UID as the sidecar, no? So i guess if the running application requests localhost it will actually get itself and not the intercept...

maybe that's okay but we should document it explicitly if so

josecv (approved):
works for me just one comment around UIDs that we should be conscious of

cmd/traffic/cmd/agentinit/agent_init.go
thallgren: You mean if the app requests the `localhost:<intercepted port>`? Then yeah, I think it must be rerouted to itself, not the intercept. That behavior is consistent with what happens with the alternative approach, where we actually patch the service to use the port of our sidecar.

PR #2468, opened 2022-03-14
+ an RPC for list for active intercepts in all ns

njayp:

## Description

## Checklist


pkg/client/cli/cmd_list.go
```diff
@@ -33,6 +35,7 @@ func listCommand() *cobra.Command {
 	}
 	flags := cmd.Flags()
 	flags.BoolVarP(&s.onlyIntercepts, "intercepts", "i", false, "intercepts only")
+	flags.BoolVarP(&s.allIntercepts, "all-intercepts", "A", false, "gets all live intercepts, overrides all other flags")
```
maxlegault, line 38: I think it would be better to name this flag `all-namespaces` like with `kubectl`, `helm` and most k8s tools

njayp: We never want to search through all workloads in all namespaces. I could do the above and also change the implementation so that if `-A` is set, `-i` or `-a` must also be set, or throw an error.

PR #2470, opened 2022-03-14
add flag and logic to search all ns

njayp:

## Description

Added a `--all-namespaces` flag to `telepresence list`. In order to not search all namespaces, this command will default to filter level agents, and will search all namespaces with agents. `--intercepts` can be used to raise the filter level to only search namespaces with intercepts.

## Checklist


pkg/client/userd/trafficmgr/traffic_manager.go
```diff
@@ -572,12 +585,25 @@ nextIs:
...
+
 	aMap := make(map[string]*manager.AgentInfo)
-	for _, ns := range nss {
-		for k, v := range tm.getCurrentAgentsInNamespace(ns) {
-			aMap[k] = v
+	if allNamespaces && filter >= connector.ListRequest_INSTALLED_AGENTS {
+		// add all agents to agent map and namespaces
+		for _, v := range tm.getCurrentAgents() {
+			aMap[v.Name] = v
+			if _, value := nsmap[v.Namespace]; !value {
```
josecv: pls rename value to `ok` here just to be consistent with go-standards

josecv (approved):

pkg/client/userd/trafficmgr/traffic_manager.go
```diff
@@ -544,24 +544,37 @@ func (tm *TrafficManager) WorkloadInfoSnapshot(
...
+		}
+	} else {
+		// validate user-provided namespaces
+		for _, ns := range namespaces {
+			ns = tm.ActualNamespace(ns)
+			if ns != "" {
+				nss = append(nss, ns)
+			}
+		}
+		if len(nss) == 0 {
```
thallgren, line 571: It strikes me as odd that this check was moved into the else-clause. Why isn't it valid in the `allNamspaces` case? Surely there will be cases when there are no intercepts? If the objective is to list all current intercepts regardless of namespace, then why continue here?

pkg/client/userd/trafficmgr/traffic_manager.go
```diff
@@ -572,12 +585,25 @@ nextIs:
...
-	for _, ns := range nss {
-		for k, v := range tm.getCurrentAgentsInNamespace(ns) {
-			aMap[k] = v
+	if allNamespaces && filter >= connector.ListRequest_INSTALLED_AGENTS {
+		// add all agents to agent map and namespaces
+		for _, v := range tm.getCurrentAgents() {
+			aMap[v.Name] = v
+			if _, ok := nsmap[v.Namespace]; !ok {
+				nsmap[v.Namespace] = true
+				nss = append(nss, v.Namespace)
```
thallgren, line 596: I really don't like this part. Wasn't the intention to capture all current intercepts? If so, then why are we scanning *all* namespaces where there's an agent? Please note that this can be agents that in no way are related to the current user.

thallgren (commented):
I know this is merged already, but I'd like some changes regardless.

PR #2473, opened 2022-03-16
Have traffic-manager use channels instead of conditions

thallgren:
Clients and agents can't wait on conditions because the
condition won't release the wait until the condition is
notified. So even if a client or agent departs, their calls
to wait will not end which causes an ever-growing demand
of resources.

### DIFF
--- a/cmd/traffic/cmd/manager/internal/cluster/subscriber.go
+++ b/cmd/traffic/cmd/manager/internal/cluster/subscriber.go
@@ -1,0 +1,107 @@
+package cluster
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"sync"
+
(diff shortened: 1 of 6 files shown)


cmd/traffic/cmd/manager/internal/cluster/subscriber.go
```diff
@@ -0,0 +1,107 @@
...
+		select {
+		case <-ctx.Done():
+			return nil
+		case ll := <-ch:
+			if ll == nil {
+				return nil
+			}
+			if err := rec.Send(ll); err != nil {
+				if ctx.Err() == nil {
+					return fmt.Errorf("WatchLogLevel.Send() failed: %w", err)
```
josecv: nit: this should probably not be WatchLogLevel

josecv (approved):

PR #2478, opened 2022-03-18
Don't route Cluster IPs if they're already reachable.

josecv:

## Description

## Checklist


pkg/client/rootd/session.go
```diff
@@ -419,10 +430,47 @@ func (s *session) onClusterInfo(ctx context.Context, mgrInfo *manager.ClusterInf
 	}
 }
 
+func (s *session) checkConnectivity(ctx context.Context) error {
```
thallgren: Strikes me as odd that a separate call is needed to retrieve the cluster's subnets. The root daemon will receive those before it configures the TUN-device anyway since it's listening on updates from `WatchClusterInfo`. If `cfgComplete` isn't `nil` [here](https://github.com/telepresenceio/telepresence/blob/release/v2/pkg/client/rootd/session.go#L382), then nothing has been started yet, so this ought to be a good point to determine connectivity using the provided `mgrInfo`.

thallgren (changes requested):
I understand the objective, but I don't think we need an extra traffic-manager call in order to accomplish it.

josecv:
@thallgren thank you for reviewing! pls take another look when you have a chance, i reworked basically the entire PR with your feedback 🙂

thallgren (approved):
LGTM. I think this will make a lot of users happy.

PR #2481, opened 2022-03-21
Use connection pooling for the DNS fallback

josecv:
This fixes a bug on Linux where the fallback resolver would sometimes
incorrectly report timeouts for DNS queries due to a concurrency bug in
miekg/dns' ExchangeWithConn method

## Description

## Checklist


pkg/client/rootd/dns/connpool.go
```diff
@@ -0,0 +1,111 @@
...
+		finished:   make(chan *dns.Conn),
+		cancel:     cCancel,
+		RemoteAddr: addr,
+	}
+	heap.Init(&pool.clients)
+	for i := 0; i < poolSize; i++ {
+		conn, err := dns.Dial("udp", net.JoinHostPort(addr, "53"))
+		if err != nil {
+			fmt.Printf("failed to create conn: %v", err)
+			os.Exit(1)
```
maxlegault: We should probably bubble up the error here instead of exiting and let the caller deal with logging.

pkg/client/rootd/dns/connpool.go
```diff
@@ -0,0 +1,111 @@
...
+		case client := <-cp.newArrival:
+			heap.Push(&cp.clients, client)
+		case conn := <-cp.finished:
+			cp.items[conn] = false
+		}
+		for conn, inUse := range cp.items {
+			if !inUse && len(cp.clients) > 0 {
+				cp.items[conn] = true
+				client := heap.Pop(&cp.clients).(*waitingClient)
+				client.returnCh <- conn
```
maxlegault: I think we can end up locking on this line if the caller's context times out on line 104 before we manage to send an available connection on the waiting client's return chan. Maybe buffering the channel when we create it on line 94 could mitigate that?

maxlegault: But even then, this conn would end up being unusable 🤔

pkg/client/rootd/dns/connpool.go
thallgren: I second this. We cannot just exit since that'll bypass all deferred functions everywhere. A `runtime.Goexit()` could be feasible in some cases but a simple `panic` with the error, accompanied by a recover that takes care of the error where the goroutine is started is almost always better in cases where the error cannot be propagated by just returning it.

pkg/client/rootd/dns/client_queue.go
```diff
@@ -0,0 +1,47 @@
+package dns
```
thallgren: Since this only applies to the overriding resolver, and only on Linux (no other resolver will ever use a fallback), it would be better if all of this was conditioned using file names with a `_linux.go` suffix.

thallgren (changes requested):
Good catch!

Question: Did you check what happens if you set the `SingleInflight = true` in the `dns.Client`? I can see that it relies on `context.Background` which is unfortunate, but it does some other tricks as well. like dealing with multiple requests for the same name.

pkg/client/rootd/dns/connpool.go
josecv: oops this is a copy paste error, will fix

pkg/client/rootd/dns/connpool.go
josecv: aah good catch i'll store the ctx.Done for the client and check it here

josecv:
re SingleInFlight, I tried setting it but it doesn't fix anything --it really only means "single in flight per domain" so multiple concurrent queries for different domains don't work

pkg/client/rootd/dns/client_queue.go
josecv: I'm not sure it's possible because server.go relies on storing the connection pool

thallgren: So, does that mean that all DNS implementations will create 10 connections even though they are never used?

josecv: oh no -- the pool is set to `nil` everywhere except the _linux implementations, i just meant that the type has to be there for the server to compile

thallgren: Ah, I see. But couldn't the type just be a small interface then? Seems a bit wasteful to drag the implementation along when it's never used.

thallgren:
@josecv  will you report that concurrency problem to miekg/dns ?

josecv:
@thallgren yep, i'm gonna switch over to fixing #2478 next then i'll report the issue

thallgren (approved):

maxlegault (approved):
LGTM! Good job on the connection pool implementation, I like how neat it is without needing any mutex 👌🏼

josecv:
Issue has been opened at https://github.com/miekg/dns/issues/1350

PR #2483, opened 2022-03-23
Enable multi-port intercepts

### LINKED ISSUE #1607: [v2] - Feature request - Develop an ability to intercept several ports of deployment
First of all, I'm a big fan of telepresence as it's making the development process more convenient and saves my time. While I'm still not able to migrate to Telepresence 2 because of two problems. And this issue is about one of them. 

**Problem**: In our infrastructure micro-services listen two ports: one for public and one for internal requests (inside kubernetes network). When I'm intercepting deployment - the traffic should be proxied for all ports.

### LINKED ISSUE #2479: telepresence uninstall --agent removes it from all agents
**Describe the bug**
According to documentation, telepresence uninstall can accept 
`--agent`
`--all-agents`
`--everything`

If single-agent name passed, it should remove traffic-agent from a specific workload. When you use uninstall --agent on cluster side it behave expected, however on local it removes all workload access and requires re-intercepting workloads.

**To Reproduce**
1. Start 2 interception process on different workloads
2. List intercepted workloads to verify 2 of them is running 
3. Uninstall one of them by using `telepresence uninstall --agent workload-name`
4. List again and you will see nothing intercepting

**Expected behavior**
It should not close connection to intercepted services

v2.5.3

### LINKED ISSUE #2489: multiple port deployment and no service deployment
Hi guys, I haven't found a solution for this cases:
1. I have a deployment, that creates container with multiple ports. And one service for both ports. I split service on two, but no luck. 
```
Deployment already being used for intercept with a different service
```

2. Is it possible to start interception on the deployment which has no services?

1. It would be nice to have an option to intercept all ports in deployment 
2. There are some services which is a part of cluster but did not have open ports, because they communicate with storage or by using queries. Maybe you need to support them too. 

  - Output of `telepresence version` (just in case the feature exists in future versions)
```
git:(master) ✗ telepresence version
Client: v2.5.3 (api v3)
Root Daemon: v2.5.3 (api v3)
User Daemon: v2.5.3 (api v3)
```

thallgren:
## Description

With this PR in place, the user can create intercepts on the same workload using different service ports.

The PR changes the way the telepresence traffic-agent is installed and configured, so that it uses a `ConfigMap` with one `agent.Config` per workload (replaces the traffic-agents environment variables), and it also changes the traffic-agent itself so that it is capable of handling multiple intercept configs simultaneously.

### Intercepting

The auto-install performed by the telepresence client, where the agent container was inserted into the workload's pod template, is no longer used. Instead, an intercept  configuration is added to the "telepresence-agents" `ConfigMap`, and all pod injection is made by the traffic-manager's mutating webhook injector. The new auto-inject workflow will typically perform the following actions.

1. The user issues a `telepresence intercept xyz` command
2. The user-daemon sends a `PrepareIntercept` request to the `traffic-manager` (roughly equivalent to `CanIntercept`)
3. The traffic-manager ensures that an entry for the intercepted workflow is present in the "telepresence-agents" `ConfigMap`:
    1. If not already present, an initial "create" entry is added. The traffic-manager then waits for the agent to arrive.
    2. The watcher for the "telepresence-agents" wakes up, discovers the "create" entry, and generates a proper intercept configuration for the agent.
    3. The watcher performs a rollout of the workload.
    4. The agent-injector receives a patch request for the workload's pod(s) and injects the traffic-agent based on the new config.
4. The traffic-manager matches the prepare intercept request against the agent configuration and returns an updated reply.
5. The user-daemon updates its intercept request and sends a `CreateIntercept` to the traffic-manager.
6. The traffic-manager creates the intercept, and returns additional info (container environment, mount point etc).
7. The user-daemon mounts the remote filesystem and starts forwarding extra ports.

### Uninstalling

The helm chart now contains a pre-delete hook that ensures that the webhook-injector stops injecting agents and then performs a rollout of all workloads that has an agent. This ensures the uninstall of all agents. In fact, a `telepresence uninstall --everything` now calls `helm uninstall` and that's it.

#### agents only

When only agents are uninstalled, this is done by dropping their respective entries in the "telepresence-agents" `ConfigMap`. This triggers the watcher which performs rollouts of the affected workloads, which in turn triggers the webhook. The webhook, not finding a config for the workload, will not inject a traffic-agent.

### Upgrading

The helm chart has a post-upgrade hook that will ensure that all workloads that were modified by earlier versions are restored to their original state (in essence, it performs actions very similar to the former `telepresence uninstall --all-agents`). It generates new agent configs for the affected workloads so that they are rolled out with a new agent installed.

### Backward compatiblity

The telepresence client will fall back to the old auto-inject behavior when an older traffic-manager is detected.
The traffic-manager will reject older clients as their auto-inject behavior is no longer permitted.

## Checklist


rpc/manager/manager.proto
```diff
@@ -203,6 +203,13 @@ message InterceptInfo {
   string pod_ip = 10;
   int32 sftp_port = 11;
 
+  // The directory where the client mounts the remote mount_point. Only
+  // set when obtaining InterceptInfo from the user daemon.
+  string client_mount_point = 2;
```
josecv, line 219: are we reusing 2 just bc it was there? if so could we move the declaration to be right under `spec` so the numbers are in order? and also what's the "procedure" for these reserved things? when do we reserve a number, when do we just drop the field from the RPC, and when do we re-use a reserved number?

rpc/manager/manager.proto
```diff
@@ -110,6 +109,8 @@ message InterceptSpec {
 
   // The dial timeout to use when a dial is made on the intercepting workstation.
   int64 dial_timeout = 17;
+
+  reserved 11;
```
josecv: similar to my comment about the 2 in the InterceptInfo, could we add a comment here stating what this used to be and when it went away? i'd like to keep track of these rpcs a bit better haha so that we can have more confidence around version (in)compatibilities

pkg/client/userd/k8s/k8s_cluster.go
```diff
@@ -95,31 +94,6 @@ func (kc *Cluster) check(c context.Context) error {
...
-		for key, val := range selector {
-			if podLabels[key] != val {
-				match = false
-				break
-			}
-		}
-		if match {
-			return pods[i], nil
-		}
-	}
```
josecv, line 118: omg thank you for replacing this haha for a big enough namespace this is probably a small but noticeable performance boost

pkg/client/sockets_unix.go
```diff
@@ -39,6 +40,12 @@ func dialSocket(ctx context.Context, socketName string, opts ...grpc.DialOption)
 			return conn, nil
 		}
 
+		// The google.golang.org/grpc/internal/transport.ConnectionError does not have an
+		// Unwrap method. It does have a Origin method though.
+		// See: https://github.com/grpc/grpc-go/pull/5148
+		if oe, ok := err.(interface{ Origin() error }); ok {
+			err = oe.Origin()
+		}
```
josecv: i see the PR's been merged so can we get rid of the workaround? disregard this comment if a future commit does that but i couldn't find one by just skimming

charts/telepresence/templates/agent-configmap.yaml
```diff
@@ -0,0 +1,18 @@
+{{- if not .Values.rbac.only }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: telepresence-agents
+  namespace:  {{ include "telepresence.namespace" . }}
+data:
+  agentInjector: |
```
josecv: is the `agentInjector` field going to be modified by telepresence? If so, I suggest we split it up into a mutable and immutable one:

```
data:
    config: |
        namespaced: ....
    agentInjector: |
        bits-that-change:
```

If we don't do that, then modifications made by the traffic manager are liable to get erased by helm and/or argocd because they're technically changes to a managed field (different systems will handle this differently and may or may not erase the changes based on different things, but having two separate fields usually fixes it)

cmd/traffic/cmd/manager/internal/mutator/agentconfig/watcher.go
```diff
@@ -0,0 +1,325 @@
...
+			switch event.Type {
+			case watch.Deleted:
+				if m, ok := event.Object.(*core.ConfigMap); ok {
+					dlog.Infof(ctx, "%s %s.%s", event.Type, m.Name, m.Namespace)
+					c.update(ctx, m.Namespace, nil)
+				}
+			case watch.Added, watch.Modified:
+				if m, ok := event.Object.(*core.ConfigMap); ok {
+					dlog.Infof(ctx, "%s %s.%s", event.Type, m.Name, m.Namespace)
+					if m.Name != agent.ConfigMap {
```
josecv: does this actually happen? didn't we ask the API to just watch our config map?

cmd/traffic/cmd/manager/internal/mutator/agentconfig/watcher.go
```diff
@@ -0,0 +1,325 @@
...
+		data:       make(map[string]map[string]string),
+	}
+}
+
+type configWatcher struct {
+	sync.RWMutex
+	cancel     context.CancelFunc
+	name       string
+	namespaces []string
+	data       map[string]map[string]string
```
josecv: could we rename this field pls? it looks too much like the Data in the configmap, but it's structured _juuuust_ differently enough that it makes my head spin --ideally the name should indicate that it maps from namespaces to agent names to agent configs

cmd/traffic/cmd/manager/internal/mutator/agentconfig/watcher.go
```diff
@@ -0,0 +1,325 @@
...
+		case <-ctx.Done():
+			return false
+		case event, ok := <-evCh:
+			if !ok {
+				return true // restart watcher
+			}
+			switch event.Type {
+			case watch.Deleted:
+				if m, ok := event.Object.(*core.ConfigMap); ok {
+					dlog.Infof(ctx, "%s %s.%s", event.Type, m.Name, m.Namespace)
```
josecv: corollary to my question a few lines down: do we need to check `m.Name` here?

pkg/tunnel/stream.go
```diff
@@ -79,7 +79,11 @@ func ReadLoop(ctx context.Context, s Stream) (<-chan Message, <-chan error) {
 			if err != nil {
 				close(msgCh) // Must close before posting the error to avoid potential deadlock
 				if ctx.Err() == nil && !(errors.Is(err, io.EOF) || errors.Is(err, net.ErrClosed)) {
-					errCh <- fmt.Errorf("!! %s %s, read from grpc.ClientStream failed", s.Tag(), s.ID())
+					err = fmt.Errorf("!! %s %s, read from grpc.ClientStream failed", s.Tag(), s.ID())
+					select {
+					case errCh <- err:
+					default:
```
josecv, line 85: why default instead of say ctx.Done()?

charts/telepresence/templates/pre-delete-hook.yaml
```diff
@@ -0,0 +1,50 @@
...
+        app.kubernetes.io/managed-by: {{ .Release.Service | quote }}
+        app.kubernetes.io/instance: {{ .Release.Name | quote }}
+        helm.sh/chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    spec:
+      restartPolicy: Never
+      containers:
+        - name: uninstall-agents
+          securityContext:
+            {{- toYaml .Values.securityContext | nindent 12 }}
+          image: "curlimages/curl"
```
josecv, line 32: could we pin this version pls? I wouldn't want the job to break bc they upgrade something silently

cmd/traffic/cmd/manager/internal/mutator/agent_injector.go
```diff
@@ -46,241 +45,278 @@ func agentInjector(ctx context.Context, req *admission.AdmissionRequest) ([]patc
...
-		if ports[i].ContainerPort == env.AgentPort {
-			err := fmt.Errorf("the %s pod container %s is exposing the same port (%d) as the %s sidecar", refPodName, appContainer.Name, env.AgentPort, install.AgentContainerName)
-			dlog.Info(ctx, err)
+	case "", "enabled":
+		config, err = a.findConfigMapValue(ctx, pod, nil)
+		if err != nil {
+			if strings.Contains(err.Error(), "unsupported workload kind") {
+				// This isn't something that we want to touch
+				err = nil
+			}
```
josecv: could we add a debugf statement here? i'd love it if every pod that comes thru is logged at some level so we can quickly answer questions of the `why dont it work` variety

cmd/traffic/cmd/manager/internal/mutator/agent_injector.go
```diff
@@ -364,3 +415,176 @@ func hidePorts(pod *core.Pod, cn *core.Container, portName string, patches []pat
...
+		if or := &refs[i]; or.Controller != nil && *or.Controller {
+			wl, err := k8sapi.GetWorkload(ctx, or.Name, pod.GetNamespace(), or.Kind)
+			if err != nil {
+				return nil, err
+			}
+			return a.findConfigMapValue(ctx, pod, wl)
+		}
+	}
+	return nil, nil
+}
```
josecv: I find this function mildly confusing --what do you think about changing its signature to `findConfigMapValue(ctx context.Context, pod *core.Pod)` and using a helper for the recursion? As it stands it seems like it's doing two things: getting the config value, yes, but also working its way back through a ref chain to get to the pod's "original" owner (I assume the intent of the recursion is to go pod -> replicaset -> deployment, is that right?)

cmd/traffic/cmd/manager/internal/mutator/agent_injector.go
```diff
@@ -364,3 +415,176 @@ func hidePorts(pod *core.Pod, cn *core.Container, portName string, patches []pat
...
+		if ok && (ag.WorkloadKind == "" || ag.WorkloadKind == wl.GetKind()) {
+			return &ag, nil
+		}
+		refs = wl.GetOwnerReferences()
+	} else {
+		refs = pod.GetOwnerReferences()
+	}
+	for i := range refs {
+		if or := &refs[i]; or.Controller != nil && *or.Controller {
+			wl, err := k8sapi.GetWorkload(ctx, or.Name, pod.GetNamespace(), or.Kind)
```
josecv, line 523: is it possible for err and wl to both be nil here? bc if that happens we'll recurse forever

cmd/traffic/cmd/manager/internal/mutator/agent_injector.go
```diff
@@ -364,3 +415,176 @@ func hidePorts(pod *core.Pod, cn *core.Container, portName string, patches []pat
...
+				ContainerPort: int32(ic.AgentPort),
+				Protocol:      core.Protocol(ic.Protocol),
+			})
+		}
+	}
+	if len(ports) == 0 {
+		return nil
+	}
+
+	evs := make([]core.EnvVar, 0, len(config.Containers)*5)
```
josecv: are the capacities just a guess? there's an avg of 5 variables in a container kinda thing?

cmd/traffic/cmd/manager/internal/mutator/agent_injector.go
```diff
@@ -86,11 +86,17 @@ func (a *agentInjector) inject(ctx context.Context, req *admission.AdmissionRequ
...
 			dlog.Debugf(ctx, `The %s.%s pod has not enabled %s container injection through %q configmap or %q annotation; skipping`,
 				pod.Name, pod.Namespace, agent.ContainerName, agent.ConfigMap, agent.InjectAnnotation)
 			return nil, nil
 		}
+		if config, err = agentconfig.GenerateForPod(ctx, pod); err != nil {
+			return nil, err
+		}
+		if err = a.agentConfigs.Store(ctx, config, true); err != nil {
+			return nil, err
+		}
```
josecv: should this be `if config == nil`? or is it idempotent?

cmd/traffic/cmd/manager/internal/mutator/agent_injector.go
josecv: Seems to me we could also use `FindOwnerWorkload`

cmd/traffic/cmd/manager/internal/state/intercept.go
```diff
@@ -0,0 +1,350 @@
...
+		case ev, ok := <-wi.ResultChan():
+			if !ok {
+				return nil, status.Error(codes.Canceled, fmt.Sprintf("watch of ConfigMap  %s[%s]: channel closed", agent.ConfigMap, agentName))
+			}
+			if !(ev.Type == watch.Added || ev.Type == watch.Modified) {
+				continue
+			}
+			if m, ok := ev.Object.(*core.ConfigMap); ok {
+				if y, ok := m.Data[agentName]; ok {
+					conf, ir := unmarshalConfigMapEntry(y, agentName, namespace)
```
josecv: any reason this isn't called `err` as usual? scope? if so, can we call it something more clear like `confErr`?

pkg/client/userd/trafficmgr/intercept.go
```diff
@@ -444,20 +563,28 @@ func (tm *TrafficManager) AddIntercept(c context.Context, ir *rpc.CreateIntercep
...
-	if result.Error != rpc.InterceptError_UNSPECIFIED {
-		return result, nil
+	var agentEnv map[string]string
+	if svcProps.preparedIntercept == nil {
+		// It's OK to just call addAgent every time; if the agent is already installed then it's a
+		// no-op.
+		agentEnv, result = tm.addAgent(c, svcProps, ir.AgentImage, apiPort)
+		if result.Error != rpc.InterceptError_UNSPECIFIED {
+			return result, nil
+		}
```
josecv, line 585: would we be here bc of an older traffic manager?

cmd/traffic/cmd/agentinit/agent_init.go
```diff
@@ -8,77 +8,129 @@ import (
...
+	agentUID := strconv.Itoa(os.Getuid())
+
+	outputInsertCount := 0
+	for _, proto := range []string{"tcp", "udp"} {
+		hasRule := false
+		for _, cn := range c.Containers {
+			for _, ic := range cn.Intercepts {
+				if strings.EqualFold(proto, ic.Protocol) {
+					hasRule = true
+					break
```
josecv: should this break the Containers loop instead of the Intercepts loop?

josecv (commented):
lgtm overall, and im very very excited to see how much this is gonna simplify everything. I've a few QoL comments, and some questions for you tomorrow, then we can merge this in

charts/telepresence/templates/post-upgrade-hook.yaml
```diff
@@ -0,0 +1,58 @@
...
+          args:
+            - '--retry'
+            - '5'
+            - '--retry-delay'
+            - '0'
+            - '--retry-max-time'
+            - '20'
+            - '--retry-all-errors'
+            - '--fail'
+            - '--silent'
```
josecv: could we remove this `--slient`? it makes it v hard to debug if something goes wrong

charts/telepresence/templates/agent-configmap.yaml
thallgren: This was an alternative solution to get the agent aware of the managed namespaces. It can be removed completely since we now pass them in the `TELEPRESENCE_MANAGED_NAMESPACES` env.

rpc/manager/manager.proto
thallgren: The only significance of the numbers is to send data in the correct order so that compatibility is maintained. The "procedure" is to ensure that we never introduce a compatibility issue. In this case, the number 2 was reserved a very long time ago when we first introduced connection tunnels between the client and the TM (around the release of 2.3.0) and we don't support that old clients anyway.

IMO, it's important to group attributes together contextually. The numbers are mostly irrelevant as long as we notice them when we do modifications. Contextually, I felt that the `client_mount_point` should be close to `pod_id`, `sftp_port` and `mount_point` since those three four are all about the remote mount, so I'd like to keep it that way even though the number sequence is 10, 11, 2, 16.

rpc/manager/manager.proto
thallgren: Yes, such a comment would be good practice. It took me a while to track the number 2 above :-)

thallgren: Ah, I see that I've already added a comment in a later commit, so I guess we're in totally agreement on this one.

pkg/client/sockets_unix.go
thallgren: Woot! It was released in 1.45.0 so I'll remove it.

cmd/traffic/cmd/manager/internal/mutator/agent_injector.go
thallgren: Yes, it's a guess. Thought it might limit the number of reallocs needed to build the final result.

cmd/traffic/cmd/manager/internal/mutator/agent_injector.go
thallgren: It's idempotent. If the config is the same, then there should be zero patches. I take special care when comparing our injected containers to accomplish this.

pkg/client/userd/trafficmgr/intercept.go
thallgren: Yes, the `preparedIntercept` will never be nil when using the new traffic-manager (and always be nil when using the old one). I'll add a comment to clarify.

cmd/traffic/cmd/agentinit/agent_init.go
thallgren: Yes, it should.

cmd/traffic/cmd/manager/managerutil/envconfig.go
```diff
@@ -37,22 +37,26 @@ type envKey struct{}
...
+		APIPort:             uint16(e.APIPort),
+		QualifiedAgentImage: e.QualifiedAgentImage(),
+		ManagerNamespace:    e.ManagerNamespace,
 	}
 }
 
+func (e *Env) QualifiedAgentImage() string {
+	img := e.AgentImage
+	if img == "" {
+		img = "tel2:" + strings.TrimPrefix(version.Version, "v")
```
josecv, line 50: we could also load up the traffic manager's image as an environment variable and return that here

i don't really care either way, but food for thought if in practice it's gonna be the same thing

josecv (approved):

PR #2487, opened 2022-03-23
Use objectSelector functionality of MutatingWebhookConfiguration to filter objects

### LINKED ISSUE #2486: Use `objectSelector` capability to limit what objects get sent to mutating webhook
MutatingWebhooks can be defined with an `objectSelector`:

```
› kubectl explain MutatingWebhookConfiguration.webhooks.objectSelector
KIND:     MutatingWebhookConfiguration
VERSION:  admissionregistration.k8s.io/v1

RESOURCE: objectSelector <Object>

DESCRIPTION:
     ObjectSelector decides whether to run the webhook based on if the object
     has matching labels. objectSelector is evaluated against both the oldObject
     and newObject that would be sent to the webhook, and is considered to match
     if either object matches the selector. A null object (oldObject in the case
     of create, or newObject in the case of delete) or an object that cannot
     have labels (like a DeploymentRollback or a PodProxyOptions object) is not
     considered to match. Use the object selector only if the webhook is opt-in,
     because end users may skip the admission webhook by setting the labels.
     Default to the empty LabelSelector, which matches everything.

     A label selector is a label query over a set of resources. The result of
     matchLabels and matchExpressions are ANDed. An empty label selector matches
     all objects. A null label selector matches no objects.

FIELDS:
   matchExpressions	<[]Object>
     matchExpressions is a list of label selector requirements. The requirements
     are ANDed.

   matchLabels	<map[string]string>
     matchLabels is a map of {key,value} pairs. A single {key,value} in the
     matchLabels map is equivalent to an element of matchExpressions, whose key
     field is "key", the operator is "In", and the values array contains only
     "value". The requirements are ANDed.
```

Currently the agent injector mutating code handles this exclusion itself by searching for a specific `annotation`. This code can be removed and use the capabilities provided by Kubernetes itself to handle this filter.

cyrus-mc:
Kubernetes MutatingWebhookConfiguration supports specifying an `objectSelector` which filters whether tor un the webhook based on if the object has matching labels.

Currently this field is left out of the configuration and the filtering is handled in the agent injector code itself. This present a problem if you decided to set the failurePolicy too fail, in that if the traffic agent restarts it will never restart because it will fail mutation.

PR to resolve #2486

### DIFF
--- a/charts/telepresence/templates/agentInjectorWebhook.yaml
+++ b/charts/telepresence/templates/agentInjectorWebhook.yaml
@@ -40,6 +40,9 @@ webhooks:
   name: agent-injector.getambassador.io
   sideEffects: {{ .Values.agentInjector.webhook.sideEffects }}
   timeoutSeconds: {{ .Values.agentInjector.webhook.timeoutSeconds }}
+  objectSelector:
+    matchLabels:
+      telepresence.getambassador.io/inject-traffic-agent: enabled
 {{- if .Values.managerRbac.namespaced }}
   namespaceSelector:
     matchExpressions:
--- a/cmd/traffic/cmd/manager/internal/mutator/agent_injector.go
+++ b/cmd/traffic/cmd/manager/internal/mutator/agent_injector.go
@@ -62,12 +62,6 @@ func agentInjector(ctx context.Context, req *admission.AdmissionRequest) ([]patc
 		return nil, nil
 	}
 
-	if pod.Annotations[install.InjectAnnotation] != "enabled" {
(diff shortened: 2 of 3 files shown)


thallgren:
Thanks for providing this, but I'm not sure we'd want it. The behavior of the mutating webhook is about to change to instead be controlled by `ConfigMap`. The annotation will still be valid, but only meaningful if it is set to disabled. See #2483 for more info.

thallgren:
Closing this because the mutating webhook is now the default way of injecting.

PR #2495, opened 2022-03-30
New cli flag to add request headers to preview-url requests

alexgervais:
## Description

This feature will allow setting extra headers to be injected in all preview-url requests with the `--preview-url-add-headers` CLI flag. 
There are changes in proto, which will need to be implemented in Ambassador Cloud to fully support the feature.

## Checklist


pkg/client/cli/cmd_preview.go
```diff
@@ -16,6 +16,8 @@ import (
 // fields in the given 'spec'.  If 'prefix' is given, long-flag names are prefixed with it.
 func addPreviewFlags(prefix string, flags *pflag.FlagSet, spec *manager.PreviewSpec) {
 	flags.BoolVarP(&spec.DisplayBanner, prefix+"banner", "b", true, "Display banner on preview page")
+	flags.StringToStringVarP(&spec.InjectHeaders, prefix+"add-headers", "", map[string]string{},
```
LukeShu: Why did you call it "add" externally, but "inject" internally?

Nit: IMO, whichever word you think is clearer should be the word used in both places.

rpc/manager/manager.proto
```diff
@@ -170,6 +170,7 @@ message IngressInfo {
 message PreviewSpec {
   IngressInfo ingress = 1;
   bool display_banner = 2;
+  map<string,string> inject_headers = 3;
```
LukeShu: Suggestion: Taking a lesson from Emissary's `add_request_headers`, it seems likely to me that at some point we will want to have extra instructions associated with the value; like "append vs overwrite". Even if we don't want to implement that right now, it seems wise to me to leave that possibility open; as

```protobuf
message InjectedHeader {
    string value = 1;
}
map<string,InjectedHeader> inject_headers = 3;
```

pkg/client/cli/cmd_preview.go
LukeShu: Change request: Taking a lesson from Emissary history, it's probably worth clarifying this as "add-request-headers"; so that no one (confused Ambassador Labs engineer or confused user alike) mistakes it as adding response headers.

LukeShu (changes requested):
One nitpick and one suggestion that you can choose to ignore; and one thing that I'm actually requesting be changed.

pkg/client/cli/cmd_preview.go
alexgervais: Yeah, 💯. I was trying different names: `inject`, `add`, `add-request`, etc. and settled on `add` simply to make the CLI arg shorter and not skew the `--help` output. I do agree `--preview-url-add-request-headers` makes the most sense.

alexgervais:
Will need rebasing after https://github.com/telepresenceio/telepresence/pull/2496

alexgervais:
Closing in favor of https://github.com/telepresenceio/telepresence/pull/2501

PR #2500, opened 2022-04-02
Move gathering of logs from the traffic-manager to the client

### LINKED ISSUE #2499: The telepresence gather-logs command must take mapped-namespaces into account
The `telepresence gather-logs` command might leak information from namespaces that the user doesn't have access to, and also scale very poorly for customers that have a high number of namespaces and users that are confined to one (or very few) of those namespaces.

The command must only gather logs from namespaces that the user is able to access and it should not ask the `traffic-manager` to provide them because the `traffic-manager` will not impersonate the user and may provide logs that the user doesn't have access to.

thallgren:
## Description

Having the traffic-manager do it had two problems:
1. There was no way of telling what namespaces to gather logs for.
2. The gathered logs should perhaps not be visible to the caller.

This commit moves the whole process to the client to ensure that the
client's set of accessible namespaces are taken into account and that
no logs can be retrieved unless the user has the permissions needed
to do so.

## Checklist

### DIFF
--- a/pkg/client/userd/trafficmgr/gather_logs.go
+++ b/pkg/client/userd/trafficmgr/gather_logs.go
@@ -1,0 +1,156 @@
+package trafficmgr
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"io"
+	"strings"
+	"sync"
+
+	core "k8s.io/api/core/v1"
+	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	typed "k8s.io/client-go/kubernetes/typed/core/v1"
+	"sigs.k8s.io/yaml"
+
+	"github.com/datawire/dlib/dlog"
+	"github.com/telepresenceio/telepresence/rpc/v2/connector"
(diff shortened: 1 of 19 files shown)


njayp (approved):

pkg/client/userd/trafficmgr/gather_logs.go
```diff
@@ -0,0 +1,156 @@
...
+					logs, yml := getPodLog(ctx, podsAPI, pod, container, request.GetPodYaml)
+					logWriteMutex.Lock()
+					resp.PodLogs[podAndNs] = logs
+					if request.GetPodYaml {
+						resp.PodYaml[podAndNs] = yml
+					}
+					logWriteMutex.Unlock()
+				}(pod)
+			}
+			wg.Wait()
```
njayp, line 116: could we do all ns in parallel instead of one ns at a time?

thallgren: This solution will do all pods in a namespace in parallel. Extended to do all namespaces in parallel too, means all pods with agents in parallel. I'm concerned that it may result in too many simultaneous connections. Given that the command isn't performance critical, I think I'd like to keep it as-is.

PR #2503, opened 2022-04-05
Watch agents only in namespaces of interest.

thallgren:
## Description

Clients will no longer watch agents from all namespaces. Instead it is
now confined to the namespaces that are accessible or mapped using the
`connect` command's --mapped-namespaces` flag.


cmd/traffic/cmd/manager/service.go
```diff
@@ -212,6 +212,103 @@ func (m *Manager) WatchAgents(session *rpc.SessionInfo, stream rpc.Manager_Watch
...
+		}
+		return false
+	}
+
+	for {
+		select {
+		case snapshot, ok := <-snapshotCh:
+			if !ok {
+				// The request has been canceled.
+				dlog.Debug(ctx, "WatchAgentsNS request cancelled")
```
josecv, line 282: could we add the session id to this log line?

josecv (approved):

cmd/traffic/cmd/manager/service.go
thallgren: I believe that's covered already by `ctx := managerutil.WithSessionInfo(stream.Context(), request.Session)` at the beginning of the function. It adds a "session_id" field to the log output.

PR #2505, opened 2022-04-05
Ensure that root daemon session terminates before another is started

thallgren:
## Description

Ensure that root daemon session terminates before another is started

The root daemon risked entering a bad state when a disconnect was
rapidly followed by a new connect. The change in this commit ensures
that the disconnect completes before the new session can start.

## Checklist

### DIFF
--- a/pkg/client/rootd/service.go
+++ b/pkg/client/rootd/service.go
@@ -58,9 +58,10 @@ type service struct {
 	quit           context.CancelFunc
 	connectCh      chan *rpc.OutboundInfo
 	connectReplyCh chan sessionReply
-	sessionLock    sync.Mutex
+	sessionLock    sync.RWMutex
 	sessionContext context.Context
 	session        *session
+	cancelCh       chan struct{}
 	timedLogLevel  log.TimedLevel
 
 	scout *scout.Reporter
@@ -88,8 +89,8 @@ func (d *service) Version(_ context.Context, _ *empty.Empty) (*common.VersionInf
 }
 
 func (d *service) Status(_ context.Context, _ *empty.Empty) (*rpc.DaemonStatus, error) {
-	d.sessionLock.Lock()
-	defer d.sessionLock.Unlock()
+	d.sessionLock.RLock()
+	defer d.sessionLock.RUnlock()
 	r := &rpc.DaemonStatus{}
 	if d.session != nil {
 		r.OutboundConfig = d.session.getInfo()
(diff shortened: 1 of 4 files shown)


pkg/client/rootd/service.go
```diff
@@ -222,9 +226,27 @@ nextSession:
 		var session *session
 		reply := sessionReply{}
 
+		d.sessionLock.Lock()
+		// If a cancelCh exists, we must wait for it to close and
+		// then check for it again. This ensures that the session
+		// cannot be accessed during shutdown and that a new connect
+		// must wait for an old to complete.
+		for {
```
njayp, line 234: Do we need this for loop? It seems like we are just waiting for `cancelCh` to be closed. Is the `if` statement by itself sufficient?

josecv: @njayp we're not just waiting for `cancelCh` to be closed, we're waiting for it to be `nil`. If we don't loop here we could have a race condition where the channel is set but we go and start a session anyway

josecv (approved):

pkg/client/rootd/service.go
josecv: Nick and I discussed offline and there's actually no risk of a race condition -- i didn't realize the close() and the `d.cancelCh = nil` happened when the lock was held.
All the same it's probably best to keep the for loop just for safety's sake and as a typical concurrency pattern to wait for a condition to hold

thallgren: @njayp @josecv But there is a risk for a race condition. Consider the following flow of events:
1. One goroutine (henceforth called "this") is at line 237, waiting for the `select` to end.
2. Another goroutine, holding the lock, now closes the `cancelCh` and sets it to nil.
3. This goroutine wakes up, perhaps one of many that are waiting on the same select, *and is beaten to the lock* so it is now waiting for the lock at line 241.
4. A goroutine with the lock creates the `cancelCh` again and releases the lock.
5. This goroutine now manages to take the lock.

Unless this goroutine checks for the `cancelCh` again once it owns the lock, it won't find it although it's there. The point I think you're both missing is waking up due to the channel being closed and grabbing the lock is not an atomic operation (step 3). Any number of things can happen between the `cancelCh` being closed and the time when this goroutine owns the lock again. In fact, this goroutine might not even reach the step where it calls the `Lock()` before the `cancelCh` has been recreated.

PR #2506, opened 2022-04-05
Make telepresence list command work without preceeding connect

thallgren:
## Description

The `telepresence list` command would produce an empty list unless it
was preceded by a `telepresence connect`. This commit ensures that the
needed synchronization always takes place so that the list output is
correct at all times.

## Checklist


njayp (approved):
lgtm

josecv (approved):

PR #2541, opened 2022-04-20
Make initial DNS recursion check more resilient to slow DNS update

thallgren:
## Description

When the TUN-device starts and is configured with a new DNS IP and
search paths, it takes several seconds before DNS quires gets routed
to that IP on a Windows box. This delay made the initial DNS recursion
check fail and in a state where recursion was falsely considered, which
in turn meant that queries to for the same name that arrived faster than
they could be resolved by the cluster, would yield a NXNAME response.

This commit changes the logic so that the recursion check is retried
until it actually hits the resolver.

## Checklist

### DIFF
--- a/pkg/client/rootd/dns/server.go
+++ b/pkg/client/rootd/dns/server.go
@@ -39,6 +39,13 @@ type FallbackPool interface {
 	LocalAddrs() []*net.UDPAddr
 }
 
+const (
+	recursionUnknown = int32(iota)
+	recursionNotDetected
+	recursionDetected
+	recursionTestInProgress
+)
+
 // Server is a DNS server which implements the github.com/miekg/dns Handler interface
 type Server struct {
 	ctx          context.Context // necessary to make logging work in ServeDNS function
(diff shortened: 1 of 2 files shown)


pkg/client/rootd/dns/server.go
```diff
@@ -45,7 +46,7 @@ type Server struct {
 	resolve      Resolver
 	requestCount int64
 	cache        sync.Map
-	recursive    int32 // 0 = never tested, 1 = not recursive, 2 = recursive
+	recursive    int32 // 0 = never tested, 1 = not recursive, 2 = recursive, 99 = test in progress
```
josecv: can we turn these into constants please

pkg/client/rootd/dns/server.go
```diff
@@ -370,15 +373,37 @@ func (d dfs) String() string {
...
+		}
+		// Check didn't hit our resolver. Try again after a second
+		dtime.SleepWithContext(c, time.Second)
+
+		// Check that the resolver didn't get hit during our wait. We don't want
+		// to retry if it did, because that will give the false impression that
+		// the resolver is recursive.
+		if atomic.LoadInt32(&s.recursive) != 99 {
+			return
+		}
```
josecv: i'm like okay with this the way it's implemented because it's realistic but i do feel the need to point out that there's a potential race condition here --say the resolver, for whatever reason, doesn't get hit until exactly when we hit line 403, we'll  end up producing that false impression anyway

i don't think there's a need to really address that though, because a DNS query to localhost should really not take longer than a second

josecv (approved):
lgtm let's just turn the `recursive` values into constants pls

josecv (approved):

PR #2543, opened 2022-04-21
Limit namespaced traffic-manager RBAC to managed namespaces

### LINKED ISSUE #2521: Namespace-scoped traffic manager install creates ClusterRole and ClusterRoleBinding
**Describe the bug**

Using the namespaced install of the traffic manager shouldn't require cluster-level permissions. However, when installing the traffic manager this way, it actually tries to create a `ClusterRole` and `ClusterRoleBinding`. This won't work for a user who doesn't have cluster-wide permissions.

**To Reproduce**

1. Create a Helm `values.yaml` file with the following values:

```
managerRbac:
  create: true
  namespaced: true
  namespaces:
  - default
```

I want to install the traffic manager in the `default` namespace, following the doc here: https://www.getambassador.io/docs/telepresence/latest/install/helm/#installing-a-namespace-scoped-traffic-manager

2. Install the traffic manager with:

```
helm install traffic-manager -n default datawire/telepresence -f values.yaml
```

3. Check the `ClusterRole`s and `ClusterRoleBinding`s:

```
kubectl get clusterrole, clusterrolebinding -A
```

4. See that a `ClusterRole` named `traffic-manager-default` and a `ClusterRoleBinding` named `traffic-manager-default` were created. (A `Role` named `traffic-manager` and a `RoleBinding` named `traffic-manager` were also created.)

**Expected behavior**

It is expected that no `ClusterRole`s or `ClusterRoleBinding`s should be created with the namespace-scoped traffic manager install. Only `Role`s and `RoleBinding`s should be created.

Telepresence version:

```
Client: v2.5.4 (api v3)
Root Daemon: v2.5.4 (api v3)
User Daemon: v2.5.4 (api v3)
```

MacOS, Kubernetes 1.20

thallgren:
## Description

The `ServiceAccount` used by a traffic-manager that is confined to a set
of namespaces must only be a `RoleBinding`-subject in those namespaces.
No `ClusterRoleBinding`, nor associated `ClusterRole`, is needed and
must not be created by the Helm chart.

Removing the cluster-wide access means that nodes become unavailable.
Consequently, the pod-CIDRs must be retrieved by watching pods, and
the watcher(s) must be limited to the managed namespaces.

Another consequence is that the cluster-id cannot be set to the UID of
the "default" namespace because the traffic-manager will typically not
have access to it. Therefore, the cluster-id will fall back to the UID
of the namespace where the traffic-manager is installed.

## Checklist


charts/telepresence/templates/trafficManagerRbac/namespace-scope.yaml
```diff
@@ -28,16 +28,25 @@ rules:
...
 {{- if eq . (include "telepresence.namespace" $) }}
+# Must be able to get the manager namespace in order to get the cluster-id
+- apiGroups:
+  - ""
+  resources:
+  - namespaces
+  resourceNames:
+  - {{ . }}
+  verbs:
+  - get
```
josecv, line 46: wow im very surprised that giving get namespaces permissions at the namespace scope works, but i guess it makes sense to get the namespace the rolebinding is in

cmd/traffic/cmd/manager/internal/cluster/info.go
```diff
@@ -77,14 +87,20 @@ func NewInfo(ctx context.Context) Info {
...
-	// Get the clusterID from the default namespaces
-	// We use a default clusterID because we don't want to fail if
-	// the traffic-manager doesn't have the ability to get the namespace
-	ns, err := client.Namespaces().Get(ctx, "default", metav1.GetOptions{})
+	// Get the clusterID from the default namespace, or from the manager's namespace if
+	// the traffic-manager doesn't have access to the default namespace.
+	nsForID := "default"
+	ns, err := client.Namespaces().Get(ctx, nsForID, metav1.GetOptions{})
+	if err != nil {
+		nsForID = env.ManagerNamespace
```
josecv, line 96: could we dlog.warn or info on the err? i can see it being useful if the user expects to be able to get the default ns

josecv (approved):

PR #2547, opened 2022-04-25
fix-unit-testing-issues

raphaelreyna:
## Description

Fixes 3 issues with unit tests

1. Skips OS specific tests if OS does not match `runtime.GOOS`.
2. Fixes panic while serializing `core.Secret` by updating `github.com/json-iterator/go` to latest version.
3. Removes dependency on the `TELEPRESENCE_VERSION` environment variable.

## Checklist

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -106,7 +106,7 @@ require (
 	github.com/imdario/mergo v0.3.11 // indirect
 	github.com/inconshreveable/mousetrap v1.0.0 // indirect
 	github.com/jmoiron/sqlx v1.3.1 // indirect
-	github.com/json-iterator/go v1.1.10 // indirect
+	github.com/json-iterator/go v1.1.12 // indirect
 	github.com/kr/fs v0.1.0 // indirect
 	github.com/lann/builder v0.0.0-20180802200727-47ae307949d0 // indirect
 	github.com/lann/ps v0.0.0-20150810152359-62de8c46ede0 // indirect
@@ -122,7 +122,7 @@ require (
 	github.com/mitchellh/reflectwalk v1.0.1 // indirect
 	github.com/moby/spdystream v0.2.0 // indirect
 	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
-	github.com/modern-go/reflect2 v1.0.1 // indirect
+	github.com/modern-go/reflect2 v1.0.2 // indirect
 	github.com/monochromegane/go-gitignore v0.0.0-20200626010858-205db1a8cc00 // indirect
 	github.com/morikuni/aec v1.0.0 // indirect
 	github.com/opencontainers/go-digest v1.0.0 // indirect
--- a/pkg/client/scout/reporter_test.go
+++ b/pkg/client/scout/reporter_test.go
@@ -21,6 +21,7 @@ import (
 	"github.com/datawire/dlib/dcontext"
 	"github.com/datawire/dlib/dlog"
 	"github.com/telepresenceio/telepresence/v2/pkg/filelocation"
+	"github.com/telepresenceio/telepresence/v2/pkg/version"
 )
 
 func TestInstallID(t *testing.T) {
@@ -256,9 +257,17 @@ func TestInstallID(t *testing.T) {
 		}
 	}
 	origEnv := os.Environ()
+	version.Version = "testing"
+	defer func() { version.Version = "" }()
+
 	for tcName, tcData := range testcases {
 		tcData := tcData
 		t.Run(tcName, func(t *testing.T) {
+			if tcData.InputGOOS != runtime.GOOS {
+				t.Skip()
+				return
+			}
+
 			ctx := dlog.NewTestContext(t, true)
 			homedir := t.TempDir()
 			defer func() {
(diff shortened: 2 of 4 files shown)


go.mod
```diff
@@ -122,7 +122,7 @@ require (
 	github.com/mitchellh/reflectwalk v1.0.1 // indirect
 	github.com/moby/spdystream v0.2.0 // indirect
 	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
-	github.com/modern-go/reflect2 v1.0.1 // indirect
+	github.com/modern-go/reflect2 v1.0.2 // indirect
```
thallgren, line 125: Please don't modify the `go.mod` file unless it's absolutely necessary at this point. My pending "multi-port-support" PR contains several changes and upgrades (including this one), so this might cause merge conflicts.

pkg/client/scout/reporter_test.go
```diff
@@ -256,9 +257,17 @@ func TestInstallID(t *testing.T) {
 		}
 	}
 	origEnv := os.Environ()
+	version.Version = "testing"
+	defer func() { version.Version = "" }()
```
thallgren: If the version changes, I think it should be restored to the value it had before the change, not simply blanked out.

pkg/client/scout/reporter_test.go
```diff
@@ -256,9 +257,17 @@ func TestInstallID(t *testing.T) {
...
+
 	for tcName, tcData := range testcases {
 		tcData := tcData
 		t.Run(tcName, func(t *testing.T) {
+			if tcData.InputGOOS != runtime.GOOS {
```
thallgren, line 269: Is this what makes the test fail when using Go 1.18? A short description of the fix in the commit description would be nice ;-)

thallgren (changes requested):

thallgren:
FYI, the license verification failure happens when the `go.mod` file is modified without a subsequent `make generate`.

go.mod
raphaelreyna: There is a panic while serializing a `core.Secret` struct that happens on `github.com/json-iterator/go v1.1.10` that is resolved in `v1.1.12`.

pkg/client/scout/reporter_test.go
raphaelreyna: The version was already an empty string which is why I set it back to an empty string.

thallgren: But that makes the test assume that the version is an empty string. What if the test runs under some other condition? Or, to word it differently, why would the test make that assumption?

pkg/client/scout/reporter_test.go
raphaelreyna: Hmm it wasn't my original intention to fix testing on 1.18 but I just checked and it's fixed now; a nice side-effect.

go.mod
thallgren: I had no idea that this PR would fix problems related to serializing a `core.Secret`. Is that a linux specific problem?

raphaelreyna: It is a bit weird that this would be an OS specific issue but I only encountered it while testing under linux.

pkg/client/scout/reporter_test.go
thallgren: Please note that this test runs just fine in CI without this change, on all platforms. I have never had problems with it either (and I'm on Linux).

thallgren: Just tested with Go 1.18.1. That fails miserably without this PR.

pkg/client/scout/reporter_test.go
raphaelreyna: The wording of the comment on `version.Version` made me confident that it was always assumed to be an empty string during testing. 

```
// Version is a "vSEMVER" string, and is either populated at build-time using `--ldflags -X`, or at
// init()-time by inspecting the binary's own debug info.
```

The `init` function in `version.go` implicitly leaves `version.Version` blank while testing if it's not set in the environment. My understanding is that it is not great practice to have the unit tests depend on the environment.

If it is not the intention that `init` in `version.go` leave the version blank for testing, I think the best solution here would be to make this a more general fix and set some non empty string at the end of that `init` function if the version string is still blank by that end of the function.

pkg/client/scout/reporter_test.go
thallgren: I installed Go 1.18.1 on my mac. It fails there too, so this isn't linux specific.

raphaelreyna: It's strange that it would pass for you and in CI since that would mean that non-linux tests are being run under Linux which one would expect to fail...

pkg/client/scout/reporter_test.go
thallgren: Version initialization should be improved for sure. I'd like to use an embedded file instead of `--ldflags` or an `init()`, now when that's an option. In any case, I think tests that need to modify that version should make no assumptions at all, e.g.:
```
	sv := version.Version
	version.Version = "v0.0.0-bogus"
	defer func() { version.Version = sv }()
```

pkg/client/scout/reporter_test.go
thallgren: I'm fairly certain that this is related to Go 1.18 (we use 1.17 in CI). It runs fine with 1.17 on all platforms. It breaks with 1.18 on all platforms.

thallgren (approved):

PR #2558, opened 2022-05-03
Single implementation of a systema conn pool

josecv:

## Description

This'll make it so there's a single implementation of the systema conn pool with uniform headers passed through on all requests. Part of some lifting so we can pass the install id to systema requests

## Checklist


cmd/traffic/cmd/manager/manager.go
```diff
@@ -136,9 +137,13 @@ func (m *Manager) runSystemAGCLoop(ctx context.Context) error {
 					}
 
 					// Release the connection we got to delete the domain + intercept
-					if err := systema.Done(); err != nil {
+					if err := systema.Done(ctx); err != nil {
 						dlog.Errorln(ctx, "systema: release management connection:", err)
 					}
+					// Double free needed to decrease refcount of connection acquired, but not released, in UpdateIntercept
```
thallgren: Where's the call to `UpdateIntercept`? And why isn't `UpdateIntercept` releasing the connection that it acquired?

pkg/a8rcloud/systema.go
```diff
@@ -0,0 +1,169 @@
...
+func GetSystemAPool[T Closeable](ctx context.Context, poolName string) SystemAPool[T] {
+	if p, ok := ctx.Value(systemaPoolKey(poolName)).(SystemAPool[T]); ok {
+		return p
+	}
+	return nil
+}
+
+type SystemAPool[T Closeable] interface {
+	Get(ctx context.Context) (T, error)
+	Done(ctx context.Context) error
```
thallgren, line 59: An alternative to a Get/Done pair could be:
```
   Do(context.Context, func(context.Context, T) error) error
```
that would use a private get/done pair internally.

cmd/traffic/cmd/manager/manager.go
josecv: So it's not a call that comes from here, but rather the original UpdateIntercept call that created the domain --it [acquires a connection](https://github.com/telepresenceio/telepresence/blob/release/v2/cmd/traffic/cmd/manager/service.go#L466) but doesn't drop it except [for](https://github.com/telepresenceio/telepresence/blob/release/v2/cmd/traffic/cmd/manager/service.go#L508-L510) [errors](https://github.com/telepresenceio/telepresence/blob/release/v2/cmd/traffic/cmd/manager/service.go#L541-L543). I dug through the history and this seems to be deliberate: the pool's .Get() sets up a reverse connection to systema, so that has to stay open as long as there's an intercept with a preview url live

pkg/a8rcloud/systema.go
josecv: I'd love to do that, the only issue is the way UpdateIntercept currently expects the connection to be left dangling --it'd have to be changed to have a long-running goroutine that mirrors the intercept lifecycle

cmd/traffic/cmd/manager/manager.go
thallgren: So the `update` that is read from the `snapshot.Updates` "carries" an open connection with it (the comment could be clearer about that)? Isn't there a risk for a leak here? What if the `systema.Get` fails on line 122 for instance? Shouldn't the second `system.Done` be called regardless?

josecv: > So the update that is read from the snapshot.Updates "carries" an open connection with it (the comment could be clearer about that)?

yes it basically does --for it to get here that connection must've been opened by an updateintercept rpc

> Isn't there a risk for a leak here? What if the systema.Get fails on line 122 for instance? Shouldn't the second system.Done be called regardless?

yes great point, i'll go change this

pkg/a8rcloud/systema.go
thallgren: Yeah, I understand that now. Never mind.

thallgren (approved):

PR #2561, opened 2022-05-05
Add --output global flag

raphaelreyna:
Signed-off-by: Raphael Reyna <raphaelreyna@protonmail.com>

## Description

Allows for structured output. 
If enabled, the output of the subcommand is wrapped in a JSON object along with the subcommand name and an error if it was returned.
Doesn't interfere with subcommands that already have a `--json` flag by treating their output as a `json.RawMessage` type rather than a string type.

## Checklist


raphaelreyna:
For subcommands that already have a `--json` flag, should enabling json output globally automatically set the subcommand's `--json` flag as well?

pkg/client/cli/util.go
```diff
@@ -105,3 +107,72 @@ func connect(ctx context.Context, connectorClient connector.ConnectorClient, std
...
+		default:
+			return f(cmd, args)
+		}
+
+		return nil
+	}
+}
+
+func (o *Output) writeStructured(w io.Writer) {
+	switch o.nativeJSON {
```
thallgren: Seems a bit overkill to use a switch on a boolean. I suggest you use if/else instead.

pkg/client/cli/util.go
```diff
@@ -105,3 +107,72 @@ func connect(ctx context.Context, connectorClient connector.ConnectorClient, std
...
+		}{
+			Cmd:    o.cmd,
+			Stdout: json.RawMessage(o.stdout.String()),
+		}
+
+		if o.err != nil {
+			x.Err = o.err.Error()
+		}
+
+		json.NewEncoder(w).Encode(x)
```
thallgren: You must explicitly ignore the return using `_ = ...`.

pkg/client/cli/util.go
```diff
@@ -105,3 +107,72 @@ func connect(ctx context.Context, connectorClient connector.ConnectorClient, std
 	}
 	return false, nil, cat.Newf("connector.Connect: %s", msg)
 }
+
+type Output struct {
```
thallgren: Does this type need to be exported?

thallgren (changes requested):
Looks good overall. Just a couple of nits.

pkg/client/cli/util.go
raphaelreyna: Ah ok, I was thinking of adding an annotation for the linter since this will never produce an error but this is cleaner.

pkg/client/cli/util.go
raphaelreyna: Not sure yet...
I might end up importing into the `cliutil` package since its writing straight to stdout and we want to intercept that.

pkg/client/cli/output/output.go
```diff
@@ -0,0 +1,104 @@
...
+
+		if next != nil {
+			next(cmd, args)
+		}
+	}
+
+	return context.WithValue(ctx, key{}, &o)
+}
+
+func Outputs(ctx context.Context) (stdout, stderr io.Writer) {
```
thallgren: The move into an `output` package introduces stuttering, `output.Outputs`. Consider changing the `Outputs` to something that would be more suitable when prefixed with `output`. Perhaps `output.Formatted` ?

pkg/client/cli/output/output_test.go
```diff
@@ -0,0 +1,247 @@
...
+		}
+	})
+
+	t.Run("json output with native json", func(t *testing.T) {
+		expectedNativeJSONMap := map[string]float64{
+			"a": 1,
+		}
+		cmd, outBuf, errBuf := newCmdWithBufs()
+		cmd.LocalFlags().Bool("json", false, "")
+		cmd.LocalFlags().Set("json", "true")
```
thallgren: It's often helpful to run `make lint` before you commit ;-)

thallgren (dismissed):

josecv (approved):

PR #2563, opened 2022-05-06
Remove processes parented by an intercept when the intercept is removed

### LINKED ISSUE #2178: telepresence leave or quit does not clean up the container that's brought up with --docker-run when intercept
**Describe the bug**
I usually run telepresence intercept with a docker container by attaching `--docker-run` option. The issue is when quit telepresence or leave an intercept, it doesn't cleanup the container that's running automatically. 

**To Reproduce**
1. Run telepresence intercept with `--docker-run` option.
2. Then leave the intercept or just do telepresence quit
3. Check your containers that's running with `docker ps`. You'll still see the intercept container running.

**Expected behavior**
Telepresence quit or leave should automatically cleanup the running container that's brought by telepresence intercept.

 - Root Daemon: v2.4.4 (api v3), User Daemon: v2.4.4 (api v3)
 - Macos Bigsur (intel)
 - EKS

thallgren:
## The problem
An intercept can be started with a `--` flag to spawn a command that
inherits the environment of the intercepted process. E.g.
```
$ telepresence intercept mysvc -- localsvc
```
The command would start the `localsvc` binary and not terminate until
that process terminates. This became a problem when a `telepresence
leave mysvc` was issued from another terminal. The intercept ceased to
exist but the `localsvc` binary kept on running, unaware that its
"parent" no longer was active.

## The solution
The `user-daemon` has a new gRPC call `AddInterceptor` that registers
the relationship between an intercept ID and a interceptor PID. A
corresponding `RemoveInterceptor` exists to remove the relationship.
When the CLI process (`telepresence`) creates a parented interceptor, it
uses the `AddInterceptor` method to register a relationship between the
current intercept ID and its own PID.

Whenever the `user-daemon` is about to remove an intercept (as the
result of a call to `leave` or `quit`), it now checks if there's a
relationship between the ID of that intercept and a PID. The process is
terminated if such a relationship is found.

## Checklist


josecv (approved):

josecv:
oops didn't mean for the other PR to be merged into this one -- i think I'll just cherry pick what i need

PR #2576, opened 2022-05-12
Fallback to attempting to get cluster id from license

raphaelreyna:
## Description

Tries to pull the cluster id from the license audience claim if it could not be read from the Kubernetes API.

## Checklist


cmd/traffic/cmd/manager/service.go
```diff
@@ -70,32 +76,19 @@ func (*Manager) Version(context.Context, *empty.Empty) (*rpc.VersionInfo2, error
...
-
-	// This is the host domain associated with a license
-	hostDomainData, err := os.ReadFile("/home/telepresence/hostDomain")
-	if err != nil {
-		errMsg += fmt.Sprintf("error reading hostDomain: %s\n", err)
-	} else {
-		resp.Host = string(hostDomainData)
+	if m.license == nil {
+		var err error
+		m.license, err = newLicenseBundleFromDisk("/home/telepresence")
```
josecv: do we have reason to expect that this'll work if m.license is nil by this point?

cmd/traffic/cmd/manager/service.go
```diff
@@ -58,6 +58,12 @@ func NewManager(ctx context.Context) (*Manager, context.Context) {
 	}
 	ctx = a8rcloud.WithSystemAPool[managerutil.SystemaCRUDClient](ctx, a8rcloud.TrafficManagerConnName, &ReverseConnProvider{ret})
 	ret.ctx = ctx
+
+	l, err := newLicenseBundleFromDisk("/home/telepresence")
+	if err == nil {
+		ret.license = l
+	}
```
josecv: pls log the error at an Info level or similar if it's not nil so users can see if/why their license didn't load

josecv: also this is probably where we should check if we need to get the ClusterID from the license

cmd/traffic/cmd/manager/service.go
```diff
@@ -129,6 +122,18 @@ func (m *Manager) GetTelepresenceAPI(ctx context.Context, e *empty.Empty) (*rpc.
...
+	if env := managerutil.GetEnv(ctx); 0 < len(env.GetManagedNamespaces()) {
+		li, err := m.license.extractLicenseClaims()
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, err.Error())
+		}
+
+		cid := m.clusterInfo.GetClusterID()
+		if err := li.IsValidForCluster(cid); err != nil {
+			return nil, status.Errorf(codes.Unauthenticated, err.Error())
+		}
```
josecv: so the issue is that we may not have the cluster id if we're in namespaced mode because [we may lack permissions to get it](https://github.com/telepresenceio/telepresence/blob/18f551004dbe34a74c3c6bd1142a1290fe5e279a/cmd/traffic/cmd/manager/internal/cluster/info.go#L84-L102) --the idea is to get that ClusterID from a claim in the license if it's not available

josecv (changes requested):
so the license loading code looks okay to me but i think maybe i didnt explain the objective to well --we need to load the manager's ClusterID from the License if [we weren't able to get it](https://github.com/telepresenceio/telepresence/blob/18f551004dbe34a74c3c6bd1142a1290fe5e279a/cmd/traffic/cmd/manager/internal/cluster/info.go#L84-L102) because of insufficient permissions.

So you may wanna make `NewManager` return an error in case the ClusterInfo couldn't get the ClusterID and there's no license to get it from
I don't think you need the `ArriveAsClient` check just yet, but that is definitely where the check will go in for the other part of the licensing story

cmd/traffic/cmd/manager/license/license.go
```diff
@@ -0,0 +1,160 @@
...
+func (lc *LicenseClaims) GetClusterID() (string, error) {
+	n := len(lc.Audience)
+	if n == 1 {
+		return lc.Audience[0], nil
+	}
+
+	return "", fmt.Errorf("found %d audience claims", n)
+}
+
+type key struct{}
```
josecv: nit: let's call this something slightly more explicit, maybe `licenseKey`

josecv (approved):
super stuff thank you

PR #2583, opened 2022-05-16
Fix panic when using preview URLs with traffic-managers < 2.6.0

### LINKED ISSUE #2580: Telepresence is not intercepting
When I try to intercept, I get and error:
`telepresence: error: PANIC: assignment to entry in nil map`

My intercept command is : 
`telepresence intercept dashboard-depl --service dashboard-svc --port 3000:3000 --http-header=name=shakil `

thallgren:
## Description

Older traffic-managers would return an InterceptInfo without an
initialized env on calls to `UpdateIntercept`. This caused a panic
message to be printed on the console after the intercept had
been set up.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 - Bugfix: Telepresence will now handle multiple path entries in the KUBECONFIG environment correctly.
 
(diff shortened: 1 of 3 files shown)


njayp (approved):
makes sense to me

PR #2587, opened 2022-05-17
Fix agent-injector misbehaviour when owner is an unsupported workload.

thallgren:
## Description

Some workloads have owners of unsupported type. One example is Argo's
`Rollout`. The agent-injector doesn't handle this correctly. Instead
of generating a `ConfigMap` entry for the controlled workload
(typically a`ReplicaSet`), it attempts to use the `Rollout` and fails.
The correct action must be to actually use the `ReplicaSet` and don't
look further for an owner.

This PR ensures that objets controlled by a `Rollout` are
considered unowned with respect to Telepresence notion of an owner.

The PR also contains a fix for the use-case when multiple services
appoint the same container and containerPort.

## Checklist

### DIFF
--- a/pkg/agentmap/generator.go
+++ b/pkg/agentmap/generator.go
@@ -59,10 +59,16 @@ func Generate(ctx context.Context, wl k8sapi.Workload, cfg *GeneratorConfig) (*a
 	}
 
 	var ccs []*agentconfig.Container
-	agentPort := cfg.AgentPort
+	pns := make(map[int32]struct{})
+	portNumber := func(cnPort int32) uint16 {
+		// Increase the agent's port number only for unique container ports
+		pns[cnPort] = struct{}{}
+		return cfg.AgentPort + uint16(len(pns)-1)
+	}
+
 	for _, svc := range svcs {
 		svcImpl, _ := k8sapi.ServiceImpl(svc)
-		if ccs, err = appendAgentContainerConfigs(svcImpl, pod, &agentPort, ccs); err != nil {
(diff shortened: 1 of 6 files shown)


pkg/agentmap/generator.go
```diff
@@ -59,10 +59,16 @@ func Generate(ctx context.Context, wl k8sapi.Workload, cfg *GeneratorConfig) (*a
 	}
 
 	var ccs []*agentconfig.Container
-	agentPort := cfg.AgentPort
+	pns := make(map[int32]struct{})
+	portNumber := func(cnPort int32) uint16 {
+		// Increase the agent's port number only for unique container ports
+		pns[cnPort] = struct{}{}
+		return cfg.AgentPort + uint16(len(pns)-1)
```
njayp: If we are reusing ports, do we always want to reuse the last port?

thallgren: That's not what's happening. The port starts at some given value. Then, for each unique container port that is added, the port increases its value. So if the initial agent port is 9000 and the container ports are 80, 8080, 8080, 443, the the corresponding mappings will be:
```
80 -> 9000
8080 -> 9001
8080 -> 9001 // Mapping is unchanged because len(pns) is unchanged
443 -> 9002
```

njayp: right. could the order be 80, 8080, 443, 8080? in which case

80 -> 9000
8080 -> 9001
443 -> 9002
8080 -> 9002

thallgren: Doh, you're right! I need to fix that.

njayp (approved):

PR #2589, opened 2022-05-17
Thread the install and cluster id with every intercept request.

josecv:
Done by passing it in the SessionInfo struct which makes it available
everywhere and makes everything easy as now there's everything needed to
re-connect "as" the user in the SessionInfo.

## Checklist


rpc/manager/manager.proto
```diff
@@ -237,6 +237,11 @@ message InterceptInfo {
 
 message SessionInfo {
   string session_id = 1;
+  string cluster_id = 2;
+  oneof session {
+    ClientInfo client = 3;
+    AgentInfo agent = 4;
```
thallgren: Why is the `ClientInfo` / `AgentInfo` needed here? Isn't that a lot of unnecessary data?

josecv: So the idea is that we need the install id to be available, so it'd be included through the ClientInfo.
I guess we could just add an optional installid field instead of ClientInfo but my thinking is this is much more future-proof because agents will not have install ids anyway and this lets us see what kind of session it is

thallgren: I'd prefer an optional `install-id` here. The traffic-manager can always see what kind of session it is by just checking its type. E.g.
```
f _, ok := ss.(*agentSessionState); ok {
   ...
}
```
The `agentSessionState` already contains the `AgentInfo` (the same is not true for the `clientSessionState` though).

josecv: Fair enough 👍

thallgren (approved):

PR #2593, opened 2022-05-18
Fix issue related to intercept ports

thallgren:
## Description

This PR fixes two problems:
1. An intercepted workload with multiple matching services appointing the same target port wasn't handled correctly under some circumstances.
2. Services that appoint the container port by number (instead of symbolic, like "http") must always have an initContainer.

## Checklist


joseldmc (commented):
The code changes overall look good to me, but I would like someone more familiar with the code to give the final approval.

josecv (approved):

PR #2594, opened 2022-05-18
Refactor intercept cleanup logic to be done on removeIntercept.

josecv:
## Description

It really made no sense to do it in a separate GC goroutine when the
user's session may already be terminated and it may be no longer be
possible to use the user's credentials to contact systema for cleanup.

## Checklist

### DIFF
--- a/cmd/traffic/cmd/manager/internal/state/intercept.go
+++ b/cmd/traffic/cmd/manager/internal/state/intercept.go
@@ -410,3 +410,62 @@ func findIntercept(ac *agentconfig.Sidecar, spec *managerrpc.InterceptSpec) (fou
 	}
 	return nil, nil, errcat.User.Newf("%s %s.%s has no interceptable port%s", ac.WorkloadKind, ac.WorkloadName, ac.Namespace, ss)
 }
+
+type InterceptFinalizer func(ctx context.Context, interceptInfo *managerrpc.InterceptInfo) error
+type interceptRefresher func() (*managerrpc.InterceptInfo, bool)
+
+type interceptState struct {
+	sync.Mutex
+	lastInfoCh       chan *managerrpc.InterceptInfo
+	finalizers       []InterceptFinalizer
+	refreshIntercept interceptRefresher
+	interceptID      string
+}
+
+func newInterceptState(clientCtx context.Context, tmCtx context.Context, interceptID string, refreshIntercept interceptRefresher) *interceptState {
+	is := &interceptState{
+		lastInfoCh:       make(chan *managerrpc.InterceptInfo),
+		interceptID:      interceptID,
+		refreshIntercept: refreshIntercept,
+	}
+	go is.monitor(clientCtx, tmCtx)
+	return is
+}
+
+func (is *interceptState) addFinalizer(finalizer InterceptFinalizer) {
+	is.Lock()
+	defer is.Unlock()
+	is.finalizers = append(is.finalizers, finalizer)
+}
+
+func (is *interceptState) terminate(info *managerrpc.InterceptInfo) {
+	is.lastInfoCh <- info
+}
+
+func (is *interceptState) monitor(clientCtx context.Context, tmCtx context.Context) {
+	for {
+		select {
+		case <-tmCtx.Done():
+			dlog.Warnf(tmCtx, "Traffic manager is closing but intercept %s is still open", is.interceptID)
+			return
+		case interceptInfo := <-is.lastInfoCh:
+			parentCtx := clientCtx
+			if parentCtx.Err() != nil {
+				// Currently this has no implications, but in future it may mean that the operations below happen
+				// with the API key of a user other than the intercept's owner
+				dlog.Warnf(tmCtx, "Intercept %s will be closed with global context", is.interceptID)
+				parentCtx = tmCtx
+			}
+			is.Lock()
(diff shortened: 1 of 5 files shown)


cmd/traffic/cmd/manager/internal/state/intercept.go
```diff
@@ -410,3 +410,62 @@ func findIntercept(ac *agentconfig.Sidecar, spec *managerrpc.InterceptSpec) (fou
...
+	defer is.Unlock()
+	is.finalizers = append(is.finalizers, finalizer)
+}
+
+func (is *interceptState) terminate(info *managerrpc.InterceptInfo) {
+	is.lastInfoCh <- info
+}
+
+func (is *interceptState) monitor(clientCtx context.Context, tmCtx context.Context) {
+	for {
```
thallgren: Why does this happen in a loop? Seems all cases in the `select` will return.

cmd/traffic/cmd/manager/internal/state/state.go
```diff
@@ -399,7 +402,7 @@ func (s *State) addClient(sessionID string, client *rpc.ClientInfo, now time.Tim
 	ctx, cancel := context.WithCancel(s.ctx)
 	s.sessions[sessionID] = &clientSessionState{
 		sessionState: sessionState{
-			done:       ctx.Done(),
+			ctx:        ctx,
 			cancel:     cancel,
 			lastMarked: now,
 			dials:      make(chan *rpc.DialRequest),
```
thallgren: Perhaps create a `newSessionState(ctx, lastMarked, dials)` that performs the `ctx, cancel := context.WithCancel(ctx)` internally? Can be reused on line 454.

cmd/traffic/cmd/manager/internal/state/state.go
```diff
@@ -548,9 +551,26 @@ func (s *State) AddIntercept(sessionID, clusterID, apiKey string, client *rpc.Cl
 		return nil, status.Errorf(codes.AlreadyExists, "Intercept named %q already exists", spec.Name)
 	}
 
+	sess := s.sessions[sessionID].(*clientSessionState)
```
thallgren: This should be checked, even if it's not likely to fail. The `sess` can be obtained right at the beginning of the function and return `status.Errorf(codes.NotFound, "session %q not found", id)` if the session cannot be found and casted.

cmd/traffic/cmd/manager/internal/state/state.go
```diff
@@ -548,9 +551,26 @@ func (s *State) AddIntercept(sessionID, clusterID, apiKey string, client *rpc.Cl
...
+
 	return cept, nil
 }
 
+func (s *State) AddInterceptFinalizer(interceptID string, finalizer InterceptFinalizer) error {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+	state, ok := s.interceptStates[interceptID]
+	if !ok {
+		return fmt.Errorf("no such intercept %s", interceptID)
```
thallgren:
```suggestion
		return status.Errorf(codes.NotFound, "no such intercept %s", interceptID)
```

cmd/traffic/cmd/manager/internal/state/state.go
```diff
@@ -608,7 +628,18 @@ func (s *State) UpdateIntercept(interceptID string, apply func(*rpc.InterceptInf
 }
 
 func (s *State) RemoveIntercept(interceptID string) bool {
-	_, didDelete := s.intercepts.LoadAndDelete(interceptID)
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	return s.unlockedRemoveIntercept(interceptID)
+}
+
+func (s *State) unlockedRemoveIntercept(interceptID string) bool {
```
thallgren, line 638: Why is this method is prefixed with `unlocked` when it's called with the `s.mu.Lock()` in place?

thallgren (changes requested):
Some nits and some questions, but overall I like this very much.

cmd/traffic/cmd/manager/internal/state/intercept.go
josecv: great catch that's what I get for iterating so much over the code

cmd/traffic/cmd/manager/internal/state/state.go
josecv: that seems to be the naming convention for the class. If the method is supposed to be called with the caller holding the mutex then it's `unlocked`. See `unlockedRemoveSession` or `unlockedCheckAgentsForIntercept` for other examples. Doesn't mean I agree mind you but I did wanna keep it consistent 🙃

cmd/traffic/cmd/manager/internal/state/intercept.go
josecv: This code is actually gone now, pls check [this commit](https://github.com/telepresenceio/telepresence/pull/2594/commits/76d695f08a82c2008f09b3d4f470d6597878d0c0) because I replaced it with a simpler solution

cmd/traffic/cmd/manager/internal/state/state.go
thallgren: That is ... weird. But OK :-)

thallgren (approved):

PR #2597, opened 2022-05-19
Make traffic-agents configuration follow suit when things change

thallgren:
## Description

A service that is removed or added will often affect the set of possible
routes that can be intercepted. This commit adds a watcher for services
and logic to discover what existing entries that may be affected by
the service addition/modification/deletion.

In addition, the mutating webhook now also listens for pod deletions, so
that the corresponding configmap entry can be removed when the owner
workload cease to exist.

## Checklist


cmd/traffic/cmd/manager/internal/mutator/watcher.go
```diff
@@ -205,6 +245,15 @@ func (c *configWatcher) Store(ctx context.Context, ac *agentconfig.Sidecar, upda
 			return fmt.Errorf("unable to get ConfigMap %s: %w", agentconfig.ConfigMap, err)
 		}
 		create = true
+	} else {
+		// Ensure that we're not about to overwrite a manually added config entry
+		if currentYml, ok := cm.Data[ac.AgentName]; ok {
+			var currAc agentconfig.Sidecar
+			if err = decode(currentYml, &currAc); err == nil && currAc.Manual {
+				dlog.Warn(ctx, "avoided an attempt to overwrite manually added config entry for %s.%s", ac.AgentName, ns)
```
josecv: I think this should be warnf since you have a formatting directive at the end there

josecv (approved):

PR #2598, opened 2022-05-19
enforce license client limit

raphaelreyna:
## Description

Enforces the maximum client limit set in the license if users are using the `datawire.io/telepresence-traffic-agent` traffic-agent image.

## Checklist


cmd/traffic/cmd/manager/managerutil/systema.go
```diff
@@ -15,6 +16,10 @@ type SystemaCRUDClient interface {
 	a8rcloud.Closeable
 }
 
+func AgentIsProprietary(ctx context.Context) bool {
+	return strings.Contains(GetAgentImage(ctx), "datawire.io/telepresence-traffic-agent")
```
thallgren, line 20: I don't think this is correct. The image name is `ambassador-telepresence-agent`.

cmd/traffic/cmd/manager/managerutil/systema.go
```diff
@@ -15,6 +16,10 @@ type SystemaCRUDClient interface {
 	a8rcloud.Closeable
 }
 
+func AgentIsProprietary(ctx context.Context) bool {
```
thallgren, line 19: The term "proprietary" is about ownership. Could this method be named "AgentIsLicensed"?

cmd/traffic/cmd/manager/managerutil/systema.go
thallgren, line 20: Another comment on this. It's extremely easy for a skilled user to circumvent this and simply return `false` here, recompile the traffic-manager, and suddenly have a system that isn't really restricted.

thallgren (changes requested):
I find it odd that we should set license limits in our OSS-code. It's very easy to circumvent. Just change some lines and recompile, and wham! you have a fully fledged, limitless, freemium solution.

thallgren:
@raphaelreyna I guess this PR should be closed?

PR #2600, opened 2022-05-19
add --watch option to `tele list`

njayp:

## Description

A `WatchWorkloads` grpc already exists, i add a flag in `tele list` that uses it.

Reproducer: 
Run `telepresence list -w`
Hit `ctrl c`
Run a telepresence command. All commands will hang.

### DIFF
--- a/pkg/client/cli/cmd_list.go
+++ b/pkg/client/cli/cmd_list.go
@@ -181,7 +181,6 @@ func (s *listInfo) printList(workloads []*connector.WorkloadInfo, stdout io.Writ
 			}
 		}
 	}
-
 }
 
 func DescribeIntercepts(iis []*manager.InterceptInfo, volumeMountsPrevented error, debug bool) string {
--- a/pkg/client/cli/cmd_list.go
+++ b/pkg/client/cli/cmd_list.go
@@ -86,7 +86,7 @@ func (s *listInfo) list(cmd *cobra.Command, _ []string) error {
 		}
 
 		ch := make(chan *connector.WorkloadInfoSnapshot)
-		go func(ch chan *connector.WorkloadInfoSnapshot) {
+		go func() {
 			for {
 				r, err := stream.Recv()
 				if err != nil {
@@ -94,7 +94,7 @@ func (s *listInfo) list(cmd *cobra.Command, _ []string) error {
 				}
 				ch <- r
 			}
-		}(ch)
+		}()
 
 	looper:
 		for {
--- a/pkg/client/cli/cmd_list.go
+++ b/pkg/client/cli/cmd_list.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"io"
 	"net"
 	"strings"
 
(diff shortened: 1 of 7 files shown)


raphaelreyna (commented):
This might cause some issues with the global `--output` flag since it buffers prints to stdout when outputing json. 
Perhaps check if that global flag is set to `json` and if so return an error or just set the local `--json` flag?

njayp:
so i understand better, what issues would it cause?

raphaelreyna:
> so i understand better, what issues would it cause?

When the `--output=json` flag is set, the `io.Writer` returned by `cmd.OutOrStdout()` is not backed by `os.Stdout` but by a `strings.Builder` which is then used to form the json response that is flushed to `os.Stdout` _after_ the command exits.
Basically, it prevents any sort of "streaming" output.

pkg/client/cli/cmd_list.go
```diff
@@ -100,8 +118,17 @@ func (s *listInfo) list(cmd *cobra.Command, _ []string) error {
 		}
 	}
 
+	if len(workloads) == 0 {
```
thallgren: Nit, but shouldn't this test be the very first thing that happens in the function?

pkg/client/cli/cmd_list.go
```diff
@@ -69,22 +71,38 @@ func (s *listInfo) list(cmd *cobra.Command, _ []string) error {
...
+		stream, err := cs.userD.WatchWorkloads(ctx, &connector.WatchWorkloadsRequest{Namespaces: []string{s.namespace}}, grpc.MaxCallRecvMsgSize(int(maxRecSize)))
+		if err != nil {
+			return err
+		}
+		/*
+			go func() {
+				<-ctx.Done()
+				stream.CloseSend()
+			}()
+		*/
```
thallgren: Please remove this comment, or explain why it's there.

thallgren (changes requested):
Looks good overall. Perhaps add an integration test to verify the functionality?

njayp:
> I can see how the term "snapshot" makes sense for a watcher that delivers snapshots, but it's a bit far fetched as an option for a list command.
> 
> * What's the difference when you list with or without a `--snapshot`?
> * When you use `--snapshot` you get JSON output.
> * Oh? I though I'd get that with the global `--json` flag?
> * Yeah, you will, but that's different, because <fill in explanation here>.

My thinking was `--output json` gets you what you usually get, but in a json obj. `--snapshot` gets you the full snapshot object. Both gets you the snapshot object in your json object. I can explain this in the changelog as a change

thallgren:
@njayp "Both gets you the snapshot object in your json object" sounds like the `--snapshot` option is redundant, which I think it somehow isn't?

njayp:
> @njayp "Both gets you the snapshot object in your json object" sounds like the `--snapshot` option is redundant, which I think it somehow isn't?

its not redundant because errors. without the `--output json` flag set, errors are not in json format

thallgren:
Ok? And that's an distinction that is important enough to introduce this somewhat confusing flag?

njayp:
> Ok? And that's an distinction that is important enough to introduce this somewhat confusing flag?

i need json errors for docker. what we could do is detect if `--output json` is set, and if it is, print the snapshot object in the json object. This would remove the ability to output the human-readable list in a json object, but maybe that isn't a common use-case. Is this what you would prefer?

thallgren:
Ok, so why don't we change the `--json` flag to give us JSON errors then? @raphaelreyna would that be hard to do?

thallgren:
What I don't want is different ways of telling a command to produce JSON output, motivated needs that our users rarely know anything about.

njayp:
> Ok, so why don't we change the `--json` flag to give us JSON errors then? @raphaelreyna would that be hard to do?

I am against having multiple json flags floating around, imo having `--json` and `--output json` have different behavior is more confusing

thallgren:
I totally agree. I think we should have one single way of asking for JSON output. The only viable use-case for it is when the CLI is called from another process, and when that happens, surely both output and errors must be returned as JSON.

thallgren:
Something weird happened in PR #2602. It contains a whole slew of commits that doesn't belong to it and that affects the change-set of this PR (surely it doesn't span 32 files?). Some rebasing is needed in order to see what's included here.

thallgren (approved):

PR #2621, opened 2022-05-30
Allow the traffic manager to take in a proxy's CA config.

josecv:
## Description

Allows the traffic manager to accept a secret containing a root ca, meant to be a proxy's, that will be trusted when communicating to cloud endpoints.

## Checklist

 - [ ] I made sure to add any docs changes required for my change (including release notes). <- I'll post these tomorrow afternoon, should be quick


cmd/traffic/cmd/manager/service.go
```diff
@@ -49,19 +54,47 @@ func (wall) Now() time.Time {
...
+	env := managerutil.GetEnv(ctx)
+	ret := &rpc.AmbassadorCloudConfig{Host: env.SystemAHost, Port: env.SystemAPort}
+	if _, err := os.Stat(proxyCertsPath); err != nil {
+		if os.IsNotExist(err) {
+			return ret, nil
+		}
+		return nil, fmt.Errorf("could not stat %s: %w", proxyCertsPath, err)
+	}
+	f, err := os.Open(path.Join(proxyCertsPath, "ca.crt"))
+	if err != nil {
```
thallgren, line 69: Why is it OK if the directory doesn't exist but not OK if the directory exists and the file is missing?

cmd/traffic/cmd/manager/service.go
```diff
@@ -49,19 +54,47 @@ func (wall) Now() time.Time {
...
+		return nil, fmt.Errorf("unable to to open %s/ca.crt: %w", proxyCertsPath, err)
+	}
+	b, err := io.ReadAll(f)
```
thallgren, line 73: `f` is never closed.

pkg/a8rcloud/config.go
```diff
@@ -0,0 +1,19 @@
+package a8rcloud
+
+import (
+	"crypto/x509"
+	"fmt"
+
+	"github.com/telepresenceio/telepresence/rpc/v2/manager"
+)
+
+func certsFromConfig(cfg *manager.AmbassadorCloudConfig) (*x509.CertPool, error) {
```
thallgren, line 10: Shouldn't this function always return at least what's in `x509.SystemCertPool()` and then append the poxyCA if it is present?

thallgren (changes requested):

cmd/traffic/cmd/manager/service.go
josecv: if the directory isn't there that means the secret just isn't mounted. On the other hand if the directory is present but the file is missing, then the secret is misconfigured/lacks the correct keys. IMO this is the correct behavior

pkg/a8rcloud/config.go
josecv: I'm not super sure it makes a difference insofar as if we're behind a proxy presumably everything will be proxied, but i can go add it it's not hard

pkg/a8rcloud/config.go
```diff
@@ -0,0 +1,22 @@
...
+	if cfg.GetProxyCa() == nil {
+		return nil, nil
+	}
+	pool, err := x509.SystemCertPool()
```
thallgren, line 14: I guess this should be first in the function so that the `SystemCertPool` is returned as-is if no `proxyCA` is configured? Isn't there a fair chance that the proxy might actually also use a CA that's already in the system pool? If so, then always providing those CAs will make such proxies work without providing an explicit `proxyCA`.

cmd/traffic/cmd/manager/service.go
thallgren: Ok, that makes sense.

pkg/a8rcloud/config.go
josecv: so if no pool is returned, the TLS package will just use the system's by default (see [here](https://pkg.go.dev/crypto/tls#Config) under RootCAs) --returning it as is here is basically the same as returning nil.

thallgren: Aha.  Sorry for the noise then ;-)

thallgren (approved):

PR #2622, opened 2022-05-31
Deprecate config setting for images.webhookAgentImage

thallgren:
## Description

There is only one type of agent image since 2.6, and that is the one
injected by the webhook, so it no longer serves a purpose to have a
configuration setting for both `agentImage` and `webhookAgentImage`.
This commit deprecates the latter. If set, a warning is logged and the
value ends up in the `agentImage`.

This commit also changes the `images.webhookRegistry` so that it
defaults the setting of `images.registry` which in turn defaults to
the environment variable `TELEPRESENCE_REGISTRY`. Prior to this change,
the `images.webhookRegistry` would default directly to the environment
variable.

## Checklist

### DIFF
--- a/pkg/client/config.go
+++ b/pkg/client/config.go
@@ -520,10 +520,9 @@ func (ll *LogLevels) merge(o *LogLevels) {
 }
 
 type Images struct {
-	PrivateRegistry          string `json:"registry,omitempty" yaml:"registry,omitempty"`
-	PrivateAgentImage        string `json:"agentImage,omitempty" yaml:"agentImage,omitempty"`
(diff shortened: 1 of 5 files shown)


pkg/client/config.go
```diff
@@ -548,7 +547,8 @@ func (img *Images) UnmarshalYAML(node *yaml.Node) (err error) {
 		case "webhookRegistry":
 			img.PrivateWebhookRegistry = v.Value
```
njayp, line 548: should this also be depreciated?

njayp (approved):

pkg/client/config.go
thallgren: I didn't want to do that at this stage, because there might be situations when you want to get the traffic-manager from one place but the agents from another. But perhaps that's over ambitious.

PR #2640, opened 2022-06-09
add local address for intercept forward

te-simonren:
## Description

When intercepting remote service to local, we have an option to set local port to forward to, as:
```
-p, --port string    <local port>:<svcPortIdentifier>
```
But there is no option for local address that forward to, and that value is set in source code as:
https://github.com/telepresenceio/telepresence/blob/26c6f1b901f40692240d04bad4b756fbcd7b1a88/pkg/client/cli/cmds_intercept.go#L481

It will be nice if wen can choose which local interface the forward goes to, because not all local services bind on `127.0.0.1`, we may bind on a real ip address on local network, or just an interface alias ip for separating sub network. 

This PR added an option to set target local address of port forward.
```
      --address string            Local address to forward to, Only accepts IP address as a value. e.g. '--address 10.0.0.2' (default "127.0.0.1")
```

## Checklist

### DIFF
--- a/pkg/client/cli/cmds_intercept.go
+++ b/pkg/client/cli/cmds_intercept.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net"
 	"os"
 	"reflect"
 	"regexp"
@@ -38,6 +39,7 @@ type interceptArgs struct {
 	name        string // Args[0] || `${Args[0]}-${--namespace}` // which depends on a combinationof --workload and --namespace
 	agentName   string // --workload || Args[0] // only valid if !localOnly
 	namespace   string // --namespace
+	address     string // --address // only valid if !localOnly
 	port        string // --port // only valid if !localOnly
 	serviceName string // --service // only valid if !localOnly
 	localOnly   bool   // --local-only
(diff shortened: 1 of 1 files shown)


thallgren:
This looks like a good addition. Would it be possible for you to add an integration test? We need that to ensure that everything works as expected and to prevent that future additions break the functionality.
I would imagine a test somewhat similar to the tests in [docker_run_test.go](/telepresenceio/telepresence/blob/release/v2/integration_test/docker_run_test.go).

thallgren:
Also, please ensure that your commits are signed. The DCO test won't pass otherwise.

te-simonren:
Thank you for your fast reply.

Sure, I will try to add an integration test case later, and sign my commit.

thallgren:
@te-simonren that's great. A tip if you write integration tests. Running them can take some time (typically 10-45 minutes depending on CPU power and network). To only run one of them (or a few, it's actually a regexp pattern), you can use:
```
go test ./integration_test/... -testify.m=Test_DockerRun
```
(change "Test_DockerRun" for you test or a pattern)

thallgren:
Thanks for your contribution. It was integrated in #3110.

PR #2647, opened 2022-06-17
Let the telepresence client cache the session-info

### LINKED ISSUE #2645: Conflicts with the currently-served intercept
**Describe the bug**
Getting ```elepresence: error: Failed to establish intercept: intercept in error state AGENT_ERROR: Conflicts with the currently-served intercept "44db34b9-2267-4663-80f8-7b3b3b396759:deployment-name"``` often.

*logs were provided and examined by @thallgren already

**To Reproduce**
1. Running an intercept
2. losing connectivity (shutting down wifi, commuting from office to home, etc)
3. trying to intercept same service again, but the local client creates a new id for the intercept
4. See the conflict error above (the traffic-manager is locked on the intercept that the local client already lost)

**Expected behavior**
The local client should persist the id and then try to intercept again with the same id.
Another option is that there will be a way to manually close the currently served intercept (uninstalling the agent doesnt do the trick, killing the traffic-manager's pod, does.)

 - Output of `telepresence version` - Client: v2.6.6 (api v3)
 - Operating system of workstation running `telepresence` commands - MacOS, both 11 and 12
 - Kubernetes environment and Version - happens both GKE and AKS, 1.21 and 1.22

**Additional context**
Everything was discussed in this thread in slack
https://datawire-oss.slack.com/archives/CAUBBJSQZ/p1654771736018089

thallgren:
## Description

Prior to this change, the client would forget its session-id when it
was forcefully disconnected and reconnecting to the traffic-manager
would create a new session. As a result, the old session was lost and
all intercepts belonging to it was unavailable. The traffic-manager
would however keep the session alive for 24 hours, and during that time
the intercepts would remain active. An active tcp intercept effectively
blocks all other intercepts to the same workload.

This commit stores the session-id in the client's cache and removes it
only after a successful `Depart`. If the session-id is found in the
cache when the client connects, then the client will use that in an
attempted call to `Remain`. If the call succeeds, the client will reuse the
session and not call `ArriveAsClient` to create a new session.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+### 2.6.7 (TBD)
+
+- Bugfix: The Telepresence client will remember and reuse the traffic-manager session after a network failure
+  or other reason that caused an unclean disconnect.
+
 ### 2.6.6 (June 9, 2022)
 
 - Bugfix: The propagation of the `TELEPRESENCE_API_PORT` environment variable now works correctly.
--- a/pkg/client/userd/trafficmgr/session_cache.go
+++ b/pkg/client/userd/trafficmgr/session_cache.go
@@ -1,0 +1,45 @@
+package trafficmgr
+
+import (
+	"context"
+	"os"
+
+	"github.com/telepresenceio/telepresence/rpc/v2/manager"
+	"github.com/telepresenceio/telepresence/v2/pkg/client/cache"
+)
+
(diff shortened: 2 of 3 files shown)


sambrex:
@thallgren thanks for looking into this! Is there a workaround in the meantime?

raphaelreyna (approved):
Very nice.

Issue #538, opened 2018-03-28
Feature request: bash completion


cv:
Please consider using https://argcomplete.readthedocs.io/en/latest/ to generate an autocomplete script for bash. It's an easy win for teh fat-fingered among us :)

ark3:
See also [how pip 10 does it](https://github.com/pypa/pip/blob/master/src/pip/_internal/__init__.py), though using argcomplete is probably a lot less work.

donnyyung:
This is still a feature we should add to Telepresence 2.

jef:
Since we're using cobra, could we leverage https://github.com/spf13/cobra/blob/master/shell_completions.md?

sgreene570:
This feature would be greatly appreciated, and has already been implemented in `kubectl` and `oc`. IMO telepresence should have full feature parity with `kubectl` and `oc`'s bash completions (e.g. tab complete service names).


Is there anything I could hack on or look into to help this effort?

thallgren:
@sgreene570 if you'd like to contribute, I believe you should start looking at our cobra command structure: https://github.com/telepresenceio/telepresence/tree/release/v2/pkg/client/cli

thallgren:
Implemented

PR #2649, opened 2022-06-17
add completion subcommand

raphaelreyna:
Signed-off-by: Raphael Reyna <raphaelreyna@protonmail.com>

## Description

Adds a `completion` subcommand to generate a completion script for the given shell (addresses #538).
Also adds the line `see 'telepresence <subcommand_name> -h' for help` on `errcat.User` errors.

## Checklist


pkg/client/cli/cmd.go
```diff
@@ -32,6 +32,30 @@ Unless the daemons are already started, an attempt will be made to start them.
...
+
+func GetCalledSubcommandName(ctx context.Context) string {
+	n, _ := ctx.Value(calledSubcommandNameKey{}).(*string)
+	if n == nil {
+		return ""
+	}
+
+	return *n
+}
+
```
thallgren, line 58: While this will surely work, it bugs me a little. A cleaner approach would be to move the error handling from main into a `PersistentPostRunE`. That way, I believe the sub command will be passed as an argument and the `PersistentPreRun` and context variable that you add wouldn't be necessary.

pkg/client/cli/cmd_completion.go
```diff
@@ -0,0 +1,48 @@
...
+			switch shell {
+			case "zsh":
+				err = rootCmd.GenZshCompletionNoDesc(os.Stdout)
+			case "bash":
+				err = rootCmd.GenBashCompletionV2(os.Stdout, false)
+			case "fish":
+				err = rootCmd.GenFishCompletion(os.Stdout, false)
+			case "ps":
+				fallthrough
+			case "powershell":
```
thallgren:
```suggestion
			case "ps", "powershell":
```

jef:
out of curiosity, what does the default completion generation look like?

pkg/client/cli/cmd.go
raphaelreyna: Yeah, you're right; it bugged me to write it.

raphaelreyna:
> out of curiosity, what does the default completion generation look like?

@jef, do you mean if no shell is passed as an argument (just running `telepresence completion`)?
if so then the user gets the message: 
```
telepresence: error: shell not specified
see 'telepresence completion -h' for help
```

jef:
sorry for being vague. what i mean is: isn't there an implementation built-in without your `addCompletionCommand(rootCmd)` function? i ask only because i've written some (much smaller) cobra cli tools that didn't need much to generate. that said, perhaps this extends it much more! thanks 😌

raphaelreyna:
> sorry for being vague. what i mean is: isn't there an implementation built-in without your `addCompletionCommand(rootCmd)` function? i ask only because i've written some (much smaller) cobra cli tools that didn't need much to generate. that said, perhaps this extends it much more! thanks relieved

ah, yeah this was a feature of the cobra cli tool that I believe is no longer supported.

pkg/client/cli/cmd.go
raphaelreyna: > While this will surely work, it bugs me a little. A cleaner approach would be to move the error handling from main into a `PersistentPostRunE`. That way, I believe the sub command will be passed as an argument and the `PersistentPreRun` and context variable that you add wouldn't be necessary.

After looking into this, there's no way to get the error from `RunE` into `PersistentPostRunE` for processing through the mechanisms provided by cobra. We'd have to pass the error through the context, through a root command struct, etc.

jef:
> I believe is no longer supported.

makes a lot more sense! 😅 thanks for your feedback 😊

pkg/client/cli/cmd.go
thallgren: Could we call it directly from each of the subcommands then? So a CLI command like `cmd_list` would do this?
```
		RunE: func(cmd *cobra.Command, args []string) error {
			return HandleError(cmd, s.list(cmd, args))
		},
```
Then we could also relieve the daemons from this error handling. It doesn't make much sense there anyway.

It is kind of weird that there's no way to handle sub-command errors generically. Are we overlooking something?

PR #2650, opened 2022-06-18
Use PreRunE instead of PostRun to deprecate k8s flags

thallgren:
The error telling the user that the k8s flags are deprecated was never
printed. This is because the code used `PostRun` which never runs
when `RunE` is used (which it is). This commit changes that to instead
use `PreRunE`.


raphaelreyna (approved):

raphaelreyna:
good catch

Issue #2596, opened 2022-05-19
v2.6.x crashes on tel-agent-init container start


ventsislav-georgiev:
Using telepresence in k3s cluster which is setup using [colima](https://github.com/abiosoft/colima) on Mac with arm processor. 

Colima uses [alpine-lima](https://github.com/abiosoft/alpine-lima) for VM and k3s for kubernetes cluster.
```
telepresence intercept ...
```
Adds the `traffic-manager` which starts normally.
The target deployment is also updated to contain `tel-agent-init` and `traffic-agent`. 

Then `tel-agent-init` fails to start with the following log:
```
info    Traffic Agent Init v2.6.2
error   failed to clear chain TEL_INBOUND_TCP: running [/sbin/iptables -t nat -N TEL_INBOUND_TCP --wait]: exit status 3: iptables v1.8.7 (legacy): can't initialize iptables table `nat': iptables who? (do you need to insmod?) Perhaps iptables or your kernel needs to be upgraded.
error   quit: failed to clear chain TEL_INBOUND_TCP: running [/sbin/iptables -t nat -N TEL_INBOUND_TCP --wait]: exit status 3: iptables v1.8.7 (legacy): can't initialize iptables table `nat': iptables who? (do you need to insmod?) Perhaps iptables or your kernel needs to be upgraded.
```

I am not sure if it is related to update to the underlying VM or k3s or telepresence.
I've tried running the following commands in the VM:
```
sudo modprobe ip_tables
sudo echo 'ip_tables' >> /etc/modules
```
But it didn't help.

Any clue what is the missing piece here for the above error?

ventsislav-georgiev:
FYI v2.5.8 seems to work properly without issues:
```
info    Traffic Agent v2.5.8 [pid:1]
info    {Name:account-data-service Namespace:default PodIP:172.17.0.34 AgentPort:9900 AppMounts:/tel_app_mounts AppPort:8080 ManagerHost:traffic-manager.ambassador ManagerPort:8081 APIPort:0}
info    new agent secrets mount path: /var/run/secrets/kubernetes.io
info    client : Connected to Manager 2.5.8
info    client : Setting intercept "3e1be60b-5ad8-4351-8955-7ec448933832:account-data-service" as ACTIVE
```

thallgren:
[...]

PR #2665, opened 2022-06-23
feat: allow users to use an annotation to ignore specified volume mounts during injection

kolorful:
## Description

Allow users to set annotation `telepresence.getambassador.io/inject-ignore-volume-mounts: "foo,bar"` to instruct the mutating webhook to ignore certain volume mounts to "be copied" during injection. 

This WIP PR is currently more of a proposal or PoC to address an issue discovered in https://github.com/telepresenceio/telepresence/issues/2596#issuecomment-1164617536

Our application unfortunately has a host path mount on `/var/xyz/spam` (xyz owned by root and spam owned by 1001). When it gets mounted by the agent sidecar and all the symlinks in place, `/tel_app_mounts/app/var` becomes owned by root, and hence cause error why the sidecar tries to `mkdir /tel_app_mounts/app/var/run`
```
ls -la /tel_app_exports/app
lrwxrwxrwx    1 1001     1001            24 Jun 23 16:00 var -> /tel_app_mounts/app/var

ls -la /tel_app_mounts/app/var
drwxr-xr-x    3 root     root          4096 Jun 23 16:15 .
drwxr-xr-x    3 root     root          4096 Jun 23 16:15 xyz
``` 
I wonder if this is actually a proper fix or there is other alternatives (we don't really want to manually generate the yamls).

Thanks!

## Checklist

### DIFF
--- a/pkg/agentconfig/container.go
+++ b/pkg/agentconfig/container.go
@@ -60,7 +60,7 @@ func AgentContainer(
 
 	mounts := make([]core.VolumeMount, 0, len(config.Containers)*3)
 	EachContainer(pod, config, func(app *core.Container, cc *Container) {
-		mounts = appendAppContainerVolumeMounts(app, cc, mounts)
+		mounts = appendAppContainerVolumeMounts(app, cc, mounts, pod.ObjectMeta.Annotations)
 	})
 
 	mounts = append(mounts,
@@ -173,7 +173,7 @@ func EachContainer(pod *core.Pod, config *Sidecar, f func(*core.Container, *Cont
 	}
 }
 
-func appendAppContainerVolumeMounts(app *core.Container, cc *Container, mounts []core.VolumeMount) []core.VolumeMount {
+func appendAppContainerVolumeMounts(app *core.Container, cc *Container, mounts []core.VolumeMount, annotations map[string]string) []core.VolumeMount {
 	isVrs := func(s string) bool {
 		return strings.HasPrefix(s, "/var/run/secrets/")
 	}
@@ -188,7 +188,12 @@ func appendAppContainerVolumeMounts(app *core.Container, cc *Container, mounts [
 		return false
 	}
 
+	ignoredVolumeMounts := getIgnoredVolumeMounts(annotations)
+
 	for _, m := range app.VolumeMounts {
+		if _, ok := ignoredVolumeMounts[m.Name]; ok {
+			continue
+		}
 		if isVrs(m.MountPath) {
 			if vrsAdded() {
 				continue // Only add /var/run/secrets once, not once per container
@@ -215,4 +220,13 @@ func appendAppContainerEnvFrom(app *core.Container, cc *Container, es []core.Env
 		es = append(es, e)
 	}
 	return es
+}
+
+func getIgnoredVolumeMounts(annotations map[string]string) map[string]struct{} {
+	vmSlice := strings.Split(annotations["telepresence.getambassador.io/inject-ignore-volume-mounts"], ",")
+	ivms := make(map[string]struct{}, len(vmSlice))
+	for _, vm := range vmSlice {
+		ivms[strings.TrimSpace(vm)] = struct{}{}
+	}
+	return ivms
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+### 2.6.9 (TBD)
+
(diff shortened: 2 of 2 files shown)


thallgren:
Thanks for providing this PR. It highlights a generic problem in how the traffic-agent propagates the mounts. It doesn't solve that problem, but there might be other reasons for preventing a remote volume mount selectively, so your addition has value regardless.

A proper fix must detect that there are several things to mount under `/var` and instead create that as a directory, and then symlink from there, so you'd get:
```
ls -la /tel_app_exports/app
drwxr-xr-x    3 1001     1001          4096 Jun 23 16:15 .
drwxr-xr-x    3 1001     1001          4096 Jun 23 16:15 var

ls -la /tel_app_exports/app/var
drwxr-xr-x    3 1001     1001          4096 Jun 23 16:15 .
lrwxrwxrwx    1 1001     1001            24 Jun 23 16:00 xyz -> /tel_app_mounts/app/var/xyz
lrwxrwxrwx    1 1001     1001            24 Jun 23 16:00 run -> /tel_app_mounts/app/var/run

ls -la /tel_app_mounts/app/var/xyz
drwxr-xr-x    3 root     root          4096 Jun 23 16:15 .
drwxr-xr-x    3 root     root          4096 Jun 23 16:15 spam
```
If you want to proceed with your PR as it stands now, I'd approve that. If you want to give the proper fix a try, that'd be great too. In any case, you must make sure that your commits are signed (DCO tests currently fail), and add an entry to the CHANGELOG.md.

kolorful:
@thallgren thank you for the encouraging feedback! I have tested and verified the functionality in our testing clusters.

I've renamed the annotation to `telepresence.getambassador.io/inject-ignore-volume-mounts`, so it's consistent with [other injector related annotations](https://www.getambassador.io/docs/telepresence/latest/reference/cluster-config/#service-name-and-port-annotations).

Doc is updated here as well: https://github.com/telepresenceio/docs/pull/86

kolorful:
@thallgren would you mind take a look this PR when you have time, I can't post in datawire-oss slack cause I can't register with my email. Thanks!

pkg/agentconfig/container.go
```diff
@@ -216,3 +221,12 @@ func appendAppContainerEnvFrom(app *core.Container, cc *Container, es []core.Env
 	}
 	return es
 }
+
+func getIgnoredVolumeMounts(annotations map[string]string) map[string]bool {
+	vmSlice := strings.Split(annotations["telepresence.getambassador.io/inject-ignore-volume-mounts"], ",")
+	ivms := make(map[string]bool, len(vmSlice))
+	for _, vm := range vmSlice {
+		ivms[vm] = true
```
thallgren:
```suggestion
		ivms[strings.TrimSpace(vm)] = true
```
because it's not unlikely that a user will use comma and space to separate the entries.

pkg/agentconfig/container.go
```diff
@@ -216,3 +221,12 @@ func appendAppContainerEnvFrom(app *core.Container, cc *Container, es []core.Env
 	}
 	return es
 }
+
+func getIgnoredVolumeMounts(annotations map[string]string) map[string]bool {
+	vmSlice := strings.Split(annotations["telepresence.getambassador.io/inject-ignore-volume-mounts"], ",")
+	ivms := make(map[string]bool, len(vmSlice))
```
thallgren: A bit of a nit-pick, and I'm fine if you leave it as is, but the usage for `ivms` is a unique set of strings. This gives the impression that it's a map of `true` or `false` values. I usually use `map[string]struct{}` (and `if _, ok = ignoredVolumeMounts[m_name]; ok {` on line 194) to make sure it's understood that the value is insignificant.

thallgren (changes requested):
Just some minor stuff. Looks good overall.

thallgren:
@kolorful What link are you using when trying to register with Slack? It should be https://datawire-oss.slack.com/

kolorful:
> @kolorful What link are you using when trying to register with Slack? It should be https://datawire-oss.slack.com/

I used the link in this page:  https://datawire-oss.slack.com/signup#/domain-signup, but my email is not in the allowed domain list.

kolorful:
> Just some minor stuff. Looks good overall.

Thanks for reviewing, all updated.

thallgren (approved):

kolorful:
One of the test failed but it doesn't seem to be related to my change?

Edit: I've tested this change with our own clusters.

thallgren:
I restarted that test. Regarding slack, try this link instead: http://a8r.io/slack.

PR #2670, opened 2022-06-28
improve: allow resolving <service-name>.<namespace>.svc by default

kolorful:
## Description

By default telepresence only resolves `<service-name>.<namespace>` and FQDN `<service-name>.<namespace>.svc.cluster.local`. Currently, in order to resolve address like `<service-name>.<namespace>.svc`, I have to add to all clusters in my kube config file.
```
    extensions:
    - name: telepresence.io
      extension:
        dns:
          include-suffixes:
           - .svc
```

If there isn't a specific reason to not resolve this by default, this PR will make the resolution work.

## Checklist

### DIFF
--- a/pkg/client/rootd/dns/server.go
+++ b/pkg/client/rootd/dns/server.go
@@ -155,6 +155,9 @@ func (s *Server) shouldDoClusterLookup(query string) bool {
 		case n == 2 && strings.HasPrefix(query, wpadDot):
 			// Reject "wpad.<namespace>."
 			return false
+		case n == 3 && strings.HasSuffix(query, ".svc"):
+			// Accept "<service-name>.<namespace>.svc"
+			return true
 		}
 	}
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 - Feature: The agent injector now supports a new annotation, `telepresence.getambassador.io/inject-ignore-volume-mounts`, that can be used to make the injector ignore specified volume mounts denoted by a comma-separated string.
 
+- Change: Allow resolving `<service-name>.<namespace>.svc` by default.
+
 ### 2.6.8 (June 23, 2022)
 
 - Feature: The name and namespace for the DNS Service that the traffic-manager uses in DNS auto-detection can now be specified.


pkg/client/rootd/dns/server.go
```diff
@@ -155,6 +155,9 @@ func (s *Server) shouldDoClusterLookup(query string) bool {
 		case n == 2 && strings.HasPrefix(query, wpadDot):
 			// Reject "wpad.<namespace>."
 			return false
+		case n == 3 && strings.HasSuffix(query, ".svc"):
+			// Accept "<service-name>.<namespace>.svc"
+			return true
```
thallgren, line 160: I'm afraid this won't work. For two reasons really.
1. At this point, the `query` will always end with a dot.
2. On most platforms, the query will never end up here because the suffix isn't known to be a top-level domain handled by the Telepresence's DNS resolver. Consequently, the DNS system will never delegate the query to us.

A fix that will work is to add "svc" as a mandatory entry to the `includeSuffixes` list. That way the macOS specific [updateResolverFiles](/telepresenceio/telepresence/blob/release/v2/pkg/client/rootd/dns/server_darwin.go#L187) and systemd-resolved specific [updateLinkDomains](/telepresenceio/telepresence/blob/release/v2/pkg/client/rootd/dns/resolved_linux.go#L102) will tell the OS resolver that it's handled by our resolver.

thallgren (changes requested):
Thanks for providing this improvement. The reason "svc" hasn't been enabled so far, is that it was our impression that such queries were extremely rare. We don't want to add mandatory TLD entries to the OS DNS system that we never use, and if there's a need for it, it can be added to the `includeSuffixes` list. If you can make a convincing argument that our reasoning is incorrect, then we will of course consider accepting a PR like this.

Can you please add a test that proves that your addition works? It could be as simple as adding a line here: https://github.com/telepresenceio/telepresence/blob/release/v2/integration_test/subdomain_test.go#L30

If you want to run individual integration tests (the full package takes a long time), then you can use `go test ./integration_test/... -testify.m Test_PodWithSubdomain`.

kolorful:
> If you can make a convincing argument that our reasoning is incorrect, then we will of course consider accepting a PR like this

I found this issue when trying to run `istio-proxy` locally and let it talk to `istiod` in QA, but its certificate is only valid for `istiod.istio-system.svc`. I found `includeSuffixes` as a workaround, but it's quite annoying that I have to add multiple lines to each cluster in my kube config. I wish it could just be set at the global level, and maybe that's a better fix?

thallgren:
We do plan to make all the settings in our kubeconfig extension available globally.

kolorful:
Oh, that's nice, is there an issue to track the effort?

Closing this PR out and gonna wait for the change.

thallgren:
I created this ticket #2675.

PR #2671, opened 2022-06-29
Add intercept definition rpcs

maxlegault:
## Description

This adds 2 new rpcs to communicate with System-A so that we can send it intercept execution history and retrieve intercept definitions based on its name.

Since I also wanted to expose the InterceptError that resulted in the execution so that we can add some business logic eventually on top of it (for the time being we’ll just make sure that it didn’t have any error for the history entry to be considered as an intercept definition if the proper flag was passed), I needed to extract the rpc/connector/connector.proto’s InterceptError to rpc/common/errors.proto to avoid circular dependencies between between rpc/userdaemon/userdaemon.proto and rpc/connector/connector.go (the latter already depending on the former). This is why I needed to bring some code changes for references to the `InterceptError` enum's values.

## Checklist


josecv (approved):

rpc/userdaemon/userdaemon.proto
```diff
@@ -127,24 +127,62 @@ message DaemonInfo {
...
-  string intercept_name = 4;
-  repeated string arguments = 5;
-  string command_to_execute = 6;
-  telepresence.common.InterceptError intercept_error = 7;
-  string intercept_error_text = 8;
+  // Telepresence CLI version
+  string telepresence_cli_version = 1;
+
+  // Telepresence User Daemon version
+  string telepresence_user_daemon_version = 2;
```
thallgren: This should use `= 10` at the end to prevent the renumbering of the otherwise unchanged fields that follow. Then use `cluster_id = 11` and `cluster_namespace = 12`. That way, the message will remain backward compatible (name changes are OK, because the names are not passed as data).

rpc/userdaemon/userdaemon.proto
```diff
@@ -127,24 +127,62 @@ message DaemonInfo {
...
 message InterceptDefinition {
-  string telepresence_version = 1;
-  string intercept_name = 2;
-  string arguments = 3;
-  string command_to_execute = 4;
+  // Telepresence CLI version
+  string telepresence_cli_version = 1;
+
+  // Telepresence User Daemon version
+  string telepresence_user_daemon_version = 2;
```
thallgren, line 189: Use `telepresence_user_daemon_version = 5` here to avoid breakage.

thallgren (changes requested):
The commit changes a public RPC so we must be careful to not break backward compatibility.

I'm a bit curious why we need both the CLI version and the version of the user daemon. Is it because we want the enhanced user daemon version?

Also, why is the version propagated in the `InterceptCreationResult`? Shouldn't that be known before an intercept is attempted?

rpc/userdaemon/userdaemon.proto
maxlegault: The `InterceptCreationResult` message is a new one, so the field order here is the initial one. Is your concern around the order more to keep a standard with other messages?

Also, to answer this question
> I'm a bit curious why we need both the CLI version and the version of the user daemon. Is it because we want the enhanced user daemon version?
> Also, why is the version propagated in the InterceptCreationResult? Shouldn't that be known before an intercept is attempted?

I guess the rpc or message are badly named if I need to clarify, so if you have any suggestions, it's more than welcome :)

The goal of the `ReportInterceptCreation` rpc is so that the cloud knows what command was executed with which result so that we can track a history of these and implement the intercept definition based on this information. Ambassador Cloud will also have a way to visualize and edit intercept definitions based on the intercept creation history.

Since some arguments can be added over time, we'll need to keep the CLI version when reporting the intercept creation result so that we can send it back when getting an intercept definition. This way we could implement some verification before instantiating an intercept based on its definition if some arguments aren't recognized. As I'm writing this though, I'm thinking that the intercept command will simply fail executing if we pass unrecognized arguments, so maybe that's not a valid use case.

Then the reason why I wanted the user daemon version to be passed was the same. Since we're moving the `intercept` command from the CLI to the user daemon so that we can add flags specific to the pro daemon (`--save-intercept-definition-as` in this case). Since the pro daemon can evolve without requiring a new CLI version, that's mainly why I was thinking of tracking that piece of information.

raphaelreyna:
> As I'm writing this though, I'm thinking that the intercept command will simply fail executing if we pass unrecognized arguments, so maybe that's not a valid use case.

Simply having it fail because of unrecognized commands due to a version mismatch without that being explained could be a frustrating UX.
I think its worth it to keep the version info so in the case of a mismatch, we can present the user with a clear error message e.g. `Version Mismatch Error: The saved intercept command you attempted to run is for Telepresence vX.X.X but Telepresence vX.Y.Z is currently installed.`

raphaelreyna (approved):

rpc/userdaemon/userdaemon.proto
thallgren: Aha, so the `InterceptCreationResult` has never been included in a release? Then you can disregard my comment about order.

Regarding the versions:
I'm just thinking that there should be some kind of version negotiation when the connection from the user-daemon to the cloud is initially established, and that those versions won't change between intercepts.

maxlegault: Hmm the thing we have that looks like that is the Authorization header that contains a JWT describing the user, but that doesn't include any information about the Telepresence version because Keycloak doesn't have knowledge about that. Would you suggest an approach where all calls from Telepresence to Ambassador Cloud include a set of headers with this kind of metadata?

Since these versions are pieces of information that we need in order to build business logic on top, my reflex would be to include the fields in the messages.

maxlegault: @thallgren any thoughts on this?

thallgren: No, the opposite. Since this is gRPC, the connection has an inherent state. What I suggest is that the cloud side of the connection should obtain the versions when it first connects to the daemon, and then remember those versions so that they don't need to be sent in every call in that sesssion, similar to how the daemon does when it connects to the traffic-manager.

thallgren (approved):

PR #2673, opened 2022-07-01
G vega cl/alicewasko lukeshu podd

g-vega-cl:
## Description

## Checklist


rpc/connector/connector.proto
```diff
@@ -35,6 +35,7 @@ service Connector {
   // Adds an intercept to a workload.  Requires having already called
   // Connect.
   rpc CreateIntercept(CreateInterceptRequest) returns (InterceptResult);
+  rpc CreatePoddIntercept(CreateInterceptRequest) returns (InterceptResult);
```
thallgren, line 39: Why is this method called `CreatePoddIntercept` and what differs it from a `CreateIntercept`. Is `Podd` really a term that we want to expose in our APIs?

CHANGELOG.md
```diff
@@ -170,7 +172,7 @@
 
 - Bugfix: The root daemon will no longer get into a bad state when a disconnect is rapidly followed by a new connect.
 
-- Bugfix: The client will now only watch agents from accessible namespaces, and is also constrained to namespaces explicitly mapped
+- Bugfix: The client will now only watch agents from accessible namespaces, and is also constrained to namespaces  explicitly mapped
```
thallgren: Please remove this change.

cmd/traffic/cmd/manager/service.go
```diff
@@ -868,3 +870,19 @@ func (m *Manager) expire(ctx context.Context) {
...
+    Security
+    Status
+    Docs
+    Contact GitHub
+    Pricing
+    API
+    Training
+    Blog
+    About
+
```
thallgren: What is this?

pkg/client/userd/grpc.go
```diff
@@ -56,20 +56,20 @@ func withReqNumber(ctx context.Context, num int64) context.Context {
 	return context.WithValue(ctx, reqNumberKey{}, num)
 }
 
-func (s *service) callCtx(ctx context.Context, name string) context.Context {
+func (s *Service) callCtx(ctx context.Context, name string) context.Context {
```
thallgren, line 60: Why does the grpc service need to be exported? The  methods here should normally be called from grpc.

pkg/client/userd/k8s/k8s_config.go
```diff
@@ -157,6 +157,47 @@ func NewConfig(c context.Context, flagMap map[string]string) (*Config, error) {
 	return k, nil
 }
 
+// This represents an inClusterConfig
+func NewConfigPodd(c context.Context, flagMap map[string]string) (*Config, error) {
```
thallgren: Can we find a better name than "podd" to represent "in cluster"?

pkg/client/userd/trafficmgr/intercept.go
```diff
@@ -406,12 +408,41 @@ func (tm *TrafficManager) CanIntercept(c context.Context, ir *rpc.CreateIntercep
 		return nil, interceptError(rpc.InterceptError_TRAFFIC_MANAGER_ERROR, errcat.Category(pi.ErrorCategory).Newf(pi.Error))
 	}
 
+	wl, err := k8sapi.GetWorkload(c, spec.Agent, spec.Namespace, spec.WorkloadKind)
```
thallgren, line 404: Any information needed to do the intercept must be available at this point. The client should not retrieve the workload here and reason about the `agentImage`. All of that must have been made by the traffic-manager already. If there are things missing from what's returned by `PrepareIntercept`, then it should be added to the `PreparedIntercept` gRPC message (it does have an `agentImage` field, and it will always be set).

pkg/client/userd/trafficmgr/intercept.go
```diff
@@ -712,6 +678,144 @@ func (tm *TrafficManager) AddIntercept(c context.Context, ir *rpc.CreateIntercep
 	}
 }
 
+// AddPoddIntercept adds one intercept
+// Used for the pod daemon deploy previews so it performs the same job as AddIntercept() only without the mount info
+func (tm *TrafficManager) AddPoddIntercept(c context.Context, ir *rpc.CreateInterceptRequest) (result *rpc.InterceptResult, err error) {
```
thallgren, line 676: Can this be controlled by a flag instead of duplicating all code?

build-aux/main.mk
```diff
@@ -121,8 +122,14 @@ endif
 
 # Build: artifacts that don't get checked in to Git
 # =================================================
-build: build-version ## (Build)  Generate a telepresence-chart.tgz, build all the source code, then git restore telepresence-chart.tgz
```
thallgren, line 124: Removing the distinction between `build` and `build-version` will cause problems down the line. The `telepresence-pro` component requires the chart to be checked in to Git, yet, we don't want a new version to be checked in everytime we create a PR in this project (unless of  course, the chart indeed was modified).

pkg/client/userd/grpc.go
g-vega-cl: I believe is because we are taking the `Service` struct from `userd/service.go`

base-image/Dockerfile
```diff
@@ -12,32 +14,22 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# The tel2-base target is reused by all telepresence cluster-side components. It's
-# never published though, as it's quite large and all builds will access to this Dockerfile
-FROM golang:alpine3.15 as tel2-base
+FROM golang:alpine3.15 as tel2-build
```
maxlegault, line 17: Cool stuff going on here with the cache! I think it breaks the smart agent image generation process though. As you can see in here https://github.com/datawire/telepresence2-proprietary/blob/master/Makefile#L43 the smart agent will first run `make tel2-base`, which will build the docker image specifying `tel2-base` as a target. I'd expect a PR in `telepresence2-proprietary` to fix that dependency.

I started looking into building multi-arch images and that dependency to tel2-base from the smart agent build was a problem for me, so I'm looking forward to this new solution!

LukeShu:
Closing in favor of https://github.com/telepresenceio/telepresence/pull/2692

PR #2674, opened 2022-07-02
enhance: allow override of misc images default registry

omer-azmon:
## Description
add a parameter to allow override of the default registry to pull misc images

## Checklist


charts/telepresence/values.yaml
```diff
@@ -52,6 +52,10 @@ affinity: {}
 ##
 priorityClassName: ""
 
+## The default registry to pull misc images
+tools:
+   registry: docker.io
```
josecv: could you just go all out on making these configurable please? similar to the way other images are configured

```yaml
hooks:
  busybox:
    registry: "..."
    image: "..."
    tag: "..."
  curl:
    registry: "..."
    image: "..."
    tag: "..."
```

josecv (changes requested):

josecv:
Thank you for the PR @omer-azmon! I just have one small ask before merging :) and could you also fix the DCO please? all you need to do is add a `-s` to your `git commit` command

charts/telepresence/values.yaml
josecv: If you could also put the config [under `hooks`](https://github.com/telepresenceio/telepresence/blob/release/v2/charts/telepresence/values.yaml#L258) instead of tools (it's been newly created) that would be super please

josecv (approved):

PR #2684, opened 2022-07-12
prometheus support

njayp:

## Description

This PR adds prometheus support to the traffic manager. It does so by adding a http server that serves prometheus metrics. It also adds a custom metric--how many clients are currently connected to the traffic manager.

https://prometheus.io/docs/guides/go-application/

## Checklist


charts/telepresence/templates/deployment.yaml
```diff
@@ -94,6 +94,8 @@ spec:
             containerPort: 8081
           - name: https
             containerPort: 8443
+          - name: prometheus
```
thallgren, line 102: I would like this to be configurable and have Prometheus disabled by default. We don't want to add extra services unless there's a need for them. It may conflict with Istio rules etc. in the cluster that must be configured.

charts/telepresence/templates/deployment.yaml
```diff
@@ -94,6 +94,8 @@ spec:
...
+            containerPort: 8082
```
thallgren: Can we please avoid having opinions about default ports? If users want prometheus, they should decide the port themselves. Perhaps just require this port to be configured in order to enable prometheus? That's consistent with how we enable the API server.

charts/telepresence/templates/service.yaml
```diff
@@ -31,4 +31,20 @@ spec:
...
+  labels:
+    {{- include "telepresence.labels" . | nindent 4 }}
+spec:
+  type: {{ .Values.service.type }}
+  ports:
+  - name: prometheus
+    port: 80
+    targetPort: prometheus
+  selector:
+    {{- include "telepresence.selectorLabels" . | nindent 4 }}
```
thallgren: Include only if prometheus is enabled.

thallgren (changes requested):

charts/telepresence/templates/deployment.yaml
josecv: altho i agree that prometheus should be disabled by default and that the port should be configurable we should still provide a default port IMO; something like:


```
prometheus:
  enabled: false
  port: 8082
```

Just so a sysadmin can go `helm install ... --set prometheus.enabled=true` and have it magically spin everything up without worrying about ports. I say this bc if I put on my sysadmin hat the truth is I don't necessarily know or care which ports the app I'm installing uses already (i.e. I don't know if my choice will cause a conflict) and I'm just trying to enable prometheus and move on with my day

thallgren: @josecv why should we provide a default port? I just see problems on the horizon. People doing `prometheus.enabled=true` and running into port conflicts, not knowing what's causing it. If they instead must to `prometheus.port=8082`, then they cause the conflict themselves.

What is the motivation for using port `8082`? Is that normally used for prometheus traffic?

thallgren (approved):

PR #2690, opened 2022-07-18
(pt1) Build and Helm-build improvements

LukeShu:
## Description

> This is split out from https://github.com/telepresenceio/telepresence/pull/2673

 - (first 2 commits) Do Docker buildkit caching to speed up builds
 - (middle 2 commits) Improve transparency around what `pkg/install/helm/` is doing. And don't consider a pending state to be corrupt; this could cause it to be difficult o get things to a working state on slow clusters; first apply creates the thing, it's pending, times out, user tries again, still pending so it deletes and re-creates. Rinse and repeat.
 - (last 2 commits) Don't build a Helm chart tarball that gets checked in to Git; don't burn-in the version numbers in `Chart.yaml`.  Instead of `helm `package` generating a tarball that gets included in to the binary, have the binary include the chart sources and turn those in to a chart tarball on-demand. When working on pod-daemon service previews, we had lots of trouble with skew between the Helm chart and the telepresence binary,

## Checklist

 - [ ] I made sure to update `./CHANGELOG.md`. - I don't think these changes merit a changelog update
 - [ ] I made sure to add any docs changes required for my change (including release notes). - I don't think these changes merit a release note
 - [x] I updated `DEVELOPING.md` with any any special dev tricks I had to use to work on this code efficiently. - yes
 - [ ] I updated `TELEMETRY.md` if I added, changed, or removed a metric name. - I didn't change metrics


LukeShu:
Note: I waffled on whether I thought the `pkg/install/helm/` changes merit a changelog entry or not. If yous think they do, I'm happy to add one.

josecv (approved):

PR #2700, opened 2022-07-26
Add a gather_traces command to collect distributed tracing from components

josecv:
## Description

Adds distributed tracing to the various components and a gather-traces command that will collect the traces.
Also adds an upload-traces command capable of uploading them to an OTLP collector

## Checklist


charts/telepresence/values.yaml
```diff
@@ -204,6 +204,15 @@ telepresenceAPI:
   # Default: 0
   port: 0
 
+
+################################################################################
+## Tracing configuration
+################################################################################
+tracing:
+  # The port on which a grpc server will be exposed for traces
+  # To diable, set tracing to an empty object ("tracing: {}")
```
raphaelreyna: typo on `disable`

cmd/traffic/cmd/agent/client.go
```diff
@@ -65,7 +66,12 @@ func TalkToManager(ctx context.Context, address string, info *rpc.AgentInfo, sta
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
-	conn, err := grpc.DialContext(ctx, address, grpc.WithTransportCredentials(insecure.NewCredentials()), grpc.WithBlock())
+	conn, err := grpc.DialContext(ctx, address,
```
raphaelreyna, line 69: does this not need to be wrapped in with `if config.AgentConfig().TracingPort != 0 { ... }`?

pkg/client/rootd/service.go
```diff
@@ -352,6 +358,14 @@ func run(c context.Context, loggingDir, configDir string) error {
 		return err
 	}
 
+	tracer, err := tracing.NewTraceServer(c, tracing.TraceConfig{
+		ProcessID:   4,
```
raphaelreyna: is this not supposed to be the OS PID?

pkg/tracing/tracer.go
```diff
@@ -0,0 +1,106 @@
...
+
+func (ts *TraceServer) ServeGrpc(ctx context.Context, port uint16) error {
+	opts := []grpc.ServerOption{}
+	grpcHandler := grpc.NewServer(opts...)
+	sc := &dhttp.ServerConfig{
+		Handler: http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			if r.ProtoMajor == 2 && strings.HasPrefix(r.Header.Get("Content-Type"), "application/grpc") {
+				grpcHandler.ServeHTTP(w, r)
+			} else {
+				w.WriteHeader(404)
```
raphaelreyna: using `http.StatusNotFound` would be more readable here.

raphaelreyna (changes requested):
lgtm; just minor nitpicks regarding readability

pkg/client/rootd/service.go
josecv: you're totally right, I didn't think this through; there's also a whole bunch of other resource stuff we can pass through, I'm gonna refactor this bit

cmd/traffic/cmd/agent/client.go
josecv: nah it's actually alright; if the config disables tracing the headers will get added but they're totally no-op, traces will just be discarded, i'm pretty sure there's no impact

cmd/traffic/cmd/agent/config.go
```diff
@@ -76,6 +77,18 @@ func (c *config) PodIP() string {
...
+func OtelResources(ctx context.Context, c Config) []attribute.KeyValue {
+	return []attribute.KeyValue{
+		attribute.Bool("tel2.has-mounts", c.HasMounts(ctx)),
+		attribute.String("tel2.workload-name", c.AgentConfig().WorkloadName),
+		attribute.String("tel2.workload-kind", c.AgentConfig().WorkloadKind),
+		attribute.String("tel2.manager-host", fmt.Sprintf("%s:%v", c.AgentConfig().ManagerHost, c.AgentConfig().ManagerPort)),
+		attribute.Bool("tel2.manual", c.AgentConfig().Manual),
+		attribute.String("k8s.namespace", c.AgentConfig().Namespace),
+		attribute.String("k8s.pod-ip", c.PodIP()),
+	}
```
raphaelreyna, line 89: nice

raphaelreyna (approved):

PR #2705, opened 2022-07-29
add helm command and install subcommand

njayp:

## Description

This PR removes the auto-install from telepresence, instead the user must run `telepresence helm install`. This command has `--upgrade` and `--values` flags. Likewise, `telepresence uninstall` has been moved to `telepresence helm uninstall`.

`alsoProxy` and `neverProxy` options have been added to the values file. These subnets are then stored in the traffic manager and passed to all connecting root daemons.

## Checklist


cmd/traffic/cmd/manager/internal/cluster/info.go
```diff
@@ -213,6 +213,23 @@ func NewInfo(ctx context.Context) Info {
 		dlog.Warnf(ctx, "Unable to get manager pod ip; env var says %s", env.PodIP)
 	}
 
+	oi.DnsConfig = &rpc.DNSConfig{}
+	oi.DnsConfig.AlsoProxySubnets, err = env.GetAlsoProxySubnets()
+	if err != nil {
+		dlog.Warnf(ctx, "AlsoProxySubnets not parsed: %v", err)
+	}
+	for _, subnet := range oi.DnsConfig.AlsoProxySubnets {
+		dlog.Infof(ctx, "Storing AlsoProxySubnet: %+v", subnet)
```
josecv: I suggest not saying "Storing" but just something like "Configured with AlsoProxySubnet" or similar, just to make it real easy for someone looking through the logs to feel confident they set it up right

pkg/client/cli/cmd_helm.go
```diff
@@ -0,0 +1,117 @@
...
+	if ia.upgrade {
+		err := cliutil.Disconnect(cmd.Context(), false, false)
+		if err != nil {
+			dlog.Debugf(cmd.Context(), "Dry run quit error: %v", err)
+		}
+	}
+
+	return cliutil.WithNetwork(cmd.Context(), func(ctx context.Context, daemonClient daemon.DaemonClient) error {
+		return cliutil.WithConnector(ctx, func(ctx context.Context, connectorClient connector.ConnectorClient) error {
+			resp, err := connectorClient.Install(ctx, request)
```
josecv, line 107: So now I'm actually seeing this it occurs to me that we should do things a little different; instead of passing in a ConnectRequest to Install, why don't we create an install request that encapsulates Upgrade, ValuePaths, and a ConnectRequest:

```
	request := &connector.InstallRequest{
	      Upgrade:    ia.upgrade,
	      ValuePaths: ia.values,
              ConnectRequest: &connector.ConnectRequest{ ... }
	} 

```
I think that'll be cleaner than adding parameters to ConnectRequest that will only sometimes be used, and it should only require some minor massaging of  the Install method (you can also make InstallRequest encapsulate the InstallInfo you already have if that's easier)

josecv (changes requested):

josecv (approved):

pkg/client/userd/trafficmgr/traffic_manager.go
```diff
@@ -351,6 +351,29 @@ func connectCluster(c context.Context, cr *rpc.ConnectRequest) (*k8s.Cluster, er
 	return cluster, nil
 }
 
+func EnsureManager(ctx context.Context, req *rpc.InstallRequest) error {
+	// seg guard
+	cr := req.GetConnectRequest()
+	if cr == nil {
+		dlog.Warn(ctx, "Connect_request in install_request was nil, using defaults")
```
thallgren, line 358: What is the user intended to do when seeing this warning in the log? I.e. what's the warning actually about?

pkg/install/helm/install.go
```diff
@@ -208,26 +231,19 @@ restart:
...
-
-		dlog.Infof(ctx, "EnsureTrafficManager(namespace=%q): upgrading Traffic Manager from %q to %q...",
-			namespace, releaseVer(existing), strings.TrimPrefix(client.Version(), "v"))
-		if err := upgradeExisting(ctx, releaseVer(existing), chrt, helmConfig, namespace); err != nil {
-			return err
+			if err := upgradeExisting(ctx, releaseVer(existing), chrt, helmConfig, namespace, values); err != nil {
+				return err
+			}
+		} else {
+			return fmt.Errorf("traffic manager version %q is already installed, use the '--upgrade' flag to replace it", releaseVer(existing))
```
thallgren, line 246: Use `errcat.User.Newf` instead.

pkg/install/helm/install.go
```diff
@@ -187,17 +184,43 @@ restart:
...
+		return fmt.Errorf("unable to load built-in helm chart: %w", err)
+	}
 
 	// OK, now install things.
+	values := getValues(ctx)
+
+	for _, path := range req.ValuePaths {
+		vals, err := chartutil.ReadValuesFile(path)
+		if err != nil {
+			return fmt.Errorf("--values path %q not readable: %v", vals, err)
```
thallgren, line 218: Use `errcat.User.Newf` instead of `fmt.Errorf` to prevent that this error prompts the user to seek more info in the logs.

rpc/connector/connector.proto
```diff
@@ -193,6 +196,16 @@ message IngressInfos {
...
 }
 
+message InstallRequest {
+  ConnectRequest connect_request = 1;
+  bool upgrade = 2;
+  repeated string value_paths = 3;
+}
+
+message InstallResult {
+  string error_text = 1;
```
thallgren, line 206: This should include an `error_category` and be handled similar to the current `UninstallResult` to avoid users being prompted to seek further information in the logs when entering flags etc. incorrectly.

cmd/traffic/cmd/manager/internal/cluster/info.go
```diff
@@ -213,6 +213,23 @@ func NewInfo(ctx context.Context) Info {
...
+	if err != nil {
+		dlog.Warnf(ctx, "AlsoProxySubnets not parsed: %v", err)
+	}
+	for _, subnet := range oi.DnsConfig.AlsoProxySubnets {
+		dlog.Infof(ctx, "Using AlsoProxySubnet: %+v", subnet)
+	}
+
+	oi.DnsConfig.NeverProxySubnets, err = env.GetNeverProxySubnets()
+	if err != nil {
+		dlog.Warnf(ctx, "NeverProxySubnets not parsed: %v", err)
```
thallgren, line 227: Why is this a warning? Looks like an error to me.

pkg/client/cli/cmd_helm.go
```diff
@@ -0,0 +1,118 @@
...
+	cmd := &cobra.Command{
+		Use:  "install",
+		Args: cobra.NoArgs,
+
+		Short: "Install telepresence traffic manager",
+		RunE:  ia.runInstall,
+	}
+
+	flags := cmd.Flags()
+	flags.BoolVarP(&ia.upgrade, "upgrade", "u", false, "replace the traffic mangaer if it already exists")
```
thallgren, line 46: Typo "mangaer" should be "manager"

charts/telepresence/values.yaml
```diff
@@ -99,6 +99,12 @@ dnsServiceNamespace: "kube-system"
 # dnsServiceIP specifies a fallback clusterIP for Telepresence to use should auto DNS discovery fail, this value is ignored if Telepresence finds a valid clusterIP
 dnsServiceIP: ""
 
+dnsConfig:
```
thallgren, line 102: This new struct must be mentioned in the chart's `README.md`.

thallgren (commented):
This PR changes the way telepresence installs the traffic-manager (good bye auto-install, hello helm command), yet, there's no entry in the CHANGELOG.md. The new `dnsConfig` feature must also be mentioned.

pkg/client/userd/trafficmgr/traffic_manager.go
njayp: A user should not see this, as it should never happen when using the `helm install` command. I thought it would be good practice to prevent panic.

cmd/traffic/cmd/manager/internal/cluster/info.go
njayp: yeah i agree

pkg/client/userd/trafficmgr/traffic_manager.go
thallgren: They way I see it, it's either acceptable that the cr is `nil` at this point (no warning), or it isn't, in which case the function should return an error. Logging a warning and assigning a zero value covers a use-case that I don't think we want to have, so it just adds confusion when looking at the code.

njayp: > They way I see it, it's either acceptable that the cr is nil at this point (no warning), or it isn't

Seems safer to have it be acceptable. I can change the `warn` to `info`.

PR #2709, opened 2022-08-04
Replace Telepresence network stack with the stack from gVisor

thallgren:
## Description

This PR replaces the Telepresence network stack with the stack from [gVisor](https://github.com/google/gvisor). It also contains several stability and performance improvements to the gRPC tunnel mechanism and the network boot process.

I strongly recommend reviewing it one commit at a time.

## Checklist


pkg/vif/endpoint.go
```diff
@@ -0,0 +1,117 @@
...
+	d.wg.Add(2)
+	dlog.Info(d.ctx, "Starting Endpoint")
+	ctx, cancel := context.WithCancel(d.ctx)
+	go d.tunToDispatch(cancel)
+	go d.dispatchToTun(ctx)
+}
+
+func (d *endpoint) Wait() {
+	d.wg.Wait()
+	dlog.Info(d.ctx, "Endpoint done")
```
josecv: should we/can we do Wait on d.Endpoint as well?

josecv (approved):
one question but this looks great overall

pkg/vif/endpoint.go
thallgren: There's no point. It looks like this:
```
func (*Endpoint) Wait() {}
```
The intention is that other endpoints can implement their own wait, which is what I've done.

PR #2719, opened 2022-08-10
Deprecate uninstall --everything instead of removing it

thallgren:
## Description

The `telepresence uninstall` command was completely removed in 2.7.0
which may create some unnecessary churn for our users. The functionality
was moved in unmodified under `telepresence helm uninstall`. That was
a bit unfortunate, because `helm` isn't involved when uninstalling
agents. Also, it requires a `--everything` flag to actually remove
the traffic-manager.

This commit reinstates the `telepresence uninstall` command but
deprecates the `--everything` flag. It can still be used, but now it
delegates to `telepresence helm uninstall` and informs the user that
it's deprecated.

This commit also changes `telepresence helm uninstall` into a command
that just performs a `helm uninstall` of the traffic-manager.


pkg/install/helm/install.go
```diff
@@ -233,22 +234,16 @@ func EnsureTrafficManager(ctx context.Context, configFlags *genericclioptions.Co
...
-			dlog.Infof(ctx, "EnsureTrafficManager(namespace=%q): replacing Traffic Manager from %q to %q...",
-				namespace, releaseVer(existing), strings.TrimPrefix(client.Version(), "v"))
-			if err := upgradeExisting(ctx, releaseVer(existing), chrt, helmConfig, namespace, values); err != nil {
-				return err
-			}
-		} else {
-			return errcat.User.Newf("traffic manager version %q is already installed, use the '--upgrade' flag to replace it", releaseVer(existing))
-		}
+	case releaseVer(existing) == strings.TrimPrefix(client.Version(), "v"):
+		dlog.Infof(ctx, "traffic manager version %q is up to date", releaseVer(existing))
```
njayp: what if the user wants to update the helm chart with a different --values file?

pkg/client/userd/trafficmgr/traffic_manager.go
```diff
@@ -351,8 +351,7 @@ func connectCluster(c context.Context, cr *rpc.ConnectRequest) (*k8s.Cluster, er
 	return cluster, nil
 }
 
-func EnsureManager(ctx context.Context, req *rpc.InstallRequest) error {
-	// seg guard
+func DeleteManager(ctx context.Context, req *rpc.HelmRequest) error {
 	cr := req.GetConnectRequest()
 	if cr == nil {
 		dlog.Info(ctx, "Connect_request in install_request was nil, using defaults")
```
njayp: nit log update

pkg/client/userd/trafficmgr/traffic_manager.go
```diff
@@ -364,14 +363,25 @@ func EnsureManager(ctx context.Context, req *rpc.InstallRequest) error {
...
 
-	ti, err := NewTrafficManagerInstaller(cluster)
+	return helm.DeleteTrafficManager(ctx, cluster.ConfigFlags, cluster.GetManagerNamespace(), false)
+}
+
+func EnsureManager(ctx context.Context, req *rpc.HelmRequest) error {
+	// seg guard
+	cr := req.GetConnectRequest()
+	if cr == nil {
+		dlog.Info(ctx, "Connect_request in install_request was nil, using defaults")
```
njayp: nit log update

pkg/client/userd/trafficmgr/traffic_manager.go
```diff
@@ -396,15 +406,14 @@ func connectMgr(c context.Context, cluster *k8s.Cluster, installID string, svc S
...
 	}
 
-	ti, err := NewTrafficManagerInstaller(cluster)
-	if err != nil {
-		return nil, stacktrace.Wrap(err, "new installer")
+	dlog.Debug(c, "checking that traffic-manager exists")
+	existing, _, isManagerErr := helm.IsTrafficManager(c, cluster.ConfigFlags, cluster.GetManagerNamespace())
+	if isManagerErr == nil && existing == nil {
+		return nil, errcat.User.New("traffic manager not found, if it is not installed, please run 'telepresence helm install'")
 	}
```
njayp, line 413: this was my first approach, but a test failed, i think it was because on one test we have some missing permissions, but we want the install to succeed, that's why I still attempted the install even if I got nil, nil. If I missed additional changes, pls disregard

rpc/connector/connector.proto
```diff
@@ -196,13 +196,19 @@ message IngressInfos {
...
+    UNINSTALL = 3;
+  }
   ConnectRequest connect_request = 1;
-  bool upgrade = 2;
+  Type type = 2;
   repeated string value_paths = 3;
 }
 
-message InstallResult {
+message Result {
```
njayp, line 211: nit can this be `HelmResult`

njayp (changes requested):
initial comments, I will look at it again in a bit

pkg/client/userd/trafficmgr/traffic_manager.go
thallgren: Yeah, I also changed the `IsTrafficManager` so that it actually returns an error if it runs into permission problems.

rpc/connector/connector.proto
thallgren: Not really, because it's also used by the `Uninstall` method.

njayp (approved):

PR #2726, opened 2022-08-11
Add tracing to gather trace command itself

josecv:
## Description

This way if we get traces that are missing data we can hopefully reconstruct why

## Checklist

### DIFF
--- a/pkg/client/userd/commands/gather_traces.go
+++ b/pkg/client/userd/commands/gather_traces.go
@@ -67,7 +67,7 @@ func (c *traceCommand) cobraCommand(ctx context.Context) *cobra.Command {
 
 func (*traceCommand) init(_ context.Context) {}
 
-func (*traceCommand) tracesFor(ctx context.Context, conn *grpc.ClientConn, ch chan []byte, component string) error {
+func (*traceCommand) tracesFor(ctx context.Context, conn *grpc.ClientConn, ch chan<- []byte, component string) error {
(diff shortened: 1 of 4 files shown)


pkg/client/userd/commands/gather_traces.go
```diff
@@ -63,10 +67,14 @@ func (c *traceCommand) cobraCommand(ctx context.Context) *cobra.Command {
 
 func (*traceCommand) init(_ context.Context) {}
 
-func (*traceCommand) tracesFor(ctx context.Context, conn *grpc.ClientConn, ch chan []byte) error {
-	daemonCli := common.NewTracingClient(conn)
-	result, err := daemonCli.DumpTraces(ctx, &emptypb.Empty{})
+func (*traceCommand) tracesFor(ctx context.Context, conn *grpc.ClientConn, ch chan []byte, component string) error {
```
thallgren: I'm a big fan of passing a dedicated write-only `ch chan<- []byte` instead of a read/write `ch chan []byte`. It brings clarity and AFAICS, this method never reads from the channel. The same is true for all callers.

thallgren: I have a similar comment about the `launchTraceWriter` return values. It really should return `(<-chan byte[], <-chan error, error)` unless it's intended that the caller also can write to those channels (which doesn't make much sense).

thallgren (changes requested):
Looks good overall, apart from my nits about channel directions.

thallgren (approved):

PR #2731, opened 2022-08-12
Raphaelreyna/shell completions

raphaelreyna:
## Description

Adds support for command and flag autocompletion as well as autcompletion for the `--namespace` flag and interceptable workload names on the `intercept` command,  autocompletion on the `--namespace` flag on the `list` command,  and autocompletion on the intercept name on the `leave` command.
Also adds an rpc definition for the pro-daemon to autocomplete on saved intercept names.

## Checklist


pkg/client/userd/commands/intercept.go
```diff
@@ -191,6 +189,58 @@ func (c *interceptCommand) init(ctx context.Context) {
...
+			return nil, cobra.ShellCompDirectiveError
+		}
+
+		var (
+			namespaceArgsIdx int
+			namespaceArg     string
+			namespace        = "default"
+		)
+		for i := 0; i < len(args); i++ {
+			if strings.HasPrefix(args[i], "--namespace") {
```
thallgren: Should this check for the short form `-n` too?

thallgren (approved):

pkg/client/userd/commands/intercept.go
raphaelreyna: ah yeah good catch; thanks

PR #2734, opened 2022-08-15
add global flags to remote commands

njayp:

## Description

I ran into a problem updating the docker extension where the remote commands would not use the global flags. This small PR should add the global flags to the userD commands.

Tested with `--output json`

TODO i didnt figure out how to get the global flags to show up in the help text for remote commands


raphaelreyna (approved):
lgtm :+1:
Regarding the TODO, I can take a look into this if you want to hold off on merging.

PR #2741, opened 2022-08-18
Ensure that tunnel streams are closed when connection attempts fail

thallgren:
## Description

If a traffic-agent makes an attempt to connect to an interceptor process
on the workstation, and no such process exists, that results in the
dial being rejected, which is correctly propagated, but unfortunately,
the stream between the traffic-manager and the workstation is not being
closed. This commit ensures that the close happens and also adds some
additional security around canceling of streams.

## Checklist

### DIFF
--- a/pkg/client/userd/commands/gather_traces.go
+++ b/pkg/client/userd/commands/gather_traces.go
@@ -71,7 +71,16 @@ func (*traceCommand) tracesFor(ctx context.Context, conn *grpc.ClientConn, ch ch
 	ctx, span := otel.GetTracerProvider().Tracer("").Start(ctx, "tracesFor", trace.WithAttributes(attribute.String("component", component)))
 	defer span.End()
 	cli := common.NewTracingClient(conn)
-	result, err := cli.DumpTraces(ctx, &emptypb.Empty{})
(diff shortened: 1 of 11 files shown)


pkg/client/userd/commands/gather_traces.go
```diff
@@ -71,7 +71,16 @@ func (*traceCommand) tracesFor(ctx context.Context, conn *grpc.ClientConn, ch ch
 	ctx, span := otel.GetTracerProvider().Tracer("").Start(ctx, "tracesFor", trace.WithAttributes(attribute.String("component", component)))
 	defer span.End()
 	cli := common.NewTracingClient(conn)
-	result, err := cli.DumpTraces(ctx, &emptypb.Empty{})
+	cfg := client.GetConfig(ctx)
+	maxRecSize := int64(1024 * 1024 * 20) // Default to 20 Mb here. There might be a lot of traces.
+	if !cfg.Grpc.MaxReceiveSize.IsZero() {
```
josecv, line 76: whoops, i realized there was a MaxReceive on grpc servers, but i had no idea it impacted clients, good catch

josecv (approved):

PR #2765, opened 2022-08-30
Let user daemon run CLI commands using a PTY device

thallgren:
## Description

This commit adds a pseudo TTY (PTY) when running CLI commands from the
user daemon, so that the commands have the same behaviour as they do
when running in a terminal.

## Checklist


pkg/client/userd/grpc.go
```diff
@@ -498,22 +501,56 @@ func (s *Service) ListCommands(ctx context.Context, _ *empty.Empty) (groups *rpc
...
+	isPTY := true
+	if runtime.GOOS == "windows" {
+		pi := ioutils.NewBytesPipe()
+		wr = pi
+		rd = pi
+		isPTY = false
+	} else {
+		// Use a PTY on unix systems.
+		var err error
+		if wr, rd, err = pty.Open(); err != nil {
```
raphaelreyna, line 565: should we log this error?
another option could be to open the pty and doing an error check before launching this func as goroutine, this way we also avoid the external process exiting because this silently failed.

josecv (approved):

raphaelreyna (changes requested):
lgtm just a minor change request regarding an error

pkg/client/userd/grpc.go
thallgren: Good catch! I meant for this error to be returned and propagated to the user.

pkg/client/userd/grpc.go
```diff
@@ -498,22 +502,91 @@ func (s *Service) ListCommands(ctx context.Context, _ *empty.Empty) (groups *rpc
...
+	for _, arg := range args {
+		if len(arg) >= 2 && arg[0] == '-' && arg[1] != '-' {
+			arg = arg[1:]
+			if !hasI {
+				hasI = strings.ContainsRune(arg, 'i')
+			}
+			if !hasT {
+				hasT = strings.ContainsRune(arg, 't')
+			}
+			if hasI && hasT {
```
raphaelreyna, line 551: So I think we only need `if hasT {}` here since `-t` / `--tty` alone is used to request a pty for the container, `-i`/`--interactive` is specific to stdin but we should still support streaming commands.

pkg/client/userd/grpc.go
```diff
@@ -617,6 +685,13 @@ func (s *Service) RunCommand(cmdStream rpc.Connector_RunCommandServer) (err erro
 			return
 		}
 
+		var rd io.Reader
+		if ctx, rd, cmdErr = stdinPump(ctx, cmdStream, needsPTY(cmd)); cmdErr != nil {
```
raphaelreyna, line 706: Do we want to log this error or make it known to the user somehow?

raphaelreyna (changes requested):

pkg/client/userd/grpc.go
thallgren: It gets known to the user. It's assigned to `cmdErr` which is handled in the defer function that sends the final error.

pkg/client/userd/grpc.go
thallgren: We actually don't need a PTY unless `-i` is used. Using only `-t` works fine without a PTY. The long forms `--tty` and `--interactive` must be supported though. I'll fix that.

thallgren: There's a good text regarding `-i` and `-t` here: https://stackoverflow.com/a/54254380/328043

raphaelreyna (approved):

PR #2767, opened 2022-08-31
Add support for mounting remote volumes using FTP

thallgren:
## Description

Adds an FTP server to the traffic-agent and a FUSE implementation based on an FTP-client to the user daemon's intercept mount logic. The feature is experimental at this stage and must be explicitly enabled using
a new boolean `intercept.useFtp` value in config.yml.

## Checklist


pkg/client/userd/trafficmgr/intercept.go
```diff
@@ -333,6 +450,34 @@ func (s *serviceProps) portIdentifier() (agentconfig.PortIdentifier, error) {
...
+			}
+		}
+		if ir.MountPoint != "" && iCept.ClientMountPoint == ir.MountPoint {
+			return &rpc.InterceptResult{
+				Error:         common.InterceptError_MOUNT_POINT_BUSY,
+				ErrorText:     spec.Name,
+				ErrorCategory: int32(errcat.User),
+				InterceptInfo: iCept.InterceptInfo,
+			}
+		}
```
josecv: can we make this a select please?

pkg/client/userd/trafficmgr/mount.go
```diff
@@ -34,6 +52,41 @@ func (ic *intercept) startMount(ctx context.Context, podWG *sync.WaitGroup) {
...
+		host := fs.NewHost(m.client, ic.ClientMountPoint)
+		host.Start(ctx)
+		go func() {
+			// Stop when the intercept dies (but not when the pod changes)
+			<-ctx.Done()
+			defer dlog.Infof(ctx, "Unmounting FTP file system for intercept %q (address %s) at %q", ic.Id, addr, ic.ClientMountPoint)
+			host.Stop()
+		}()
+		return nil
+	}
```
josecv, line 83: should we lock the client here?

josecv (changes requested):
lgtm overall but some nits

pkg/client/userd/trafficmgr/intercept.go
thallgren: No, but I can make it a switch ;-)

pkg/client/userd/trafficmgr/mount.go
thallgren: No, I don't think that's necessary. The `handleInterceptShapshot` method is dealing with intercepts sequentially, so there's no risk that the start method is called concurrently.

thallgren:
Rebased on top of 2.7.3 and updated CHANGELOG.md to anticipate release in 2.7.4

josecv (approved):

Issue #1806, opened 2021-06-19
How to add resources for traffic-agent to allow running in namespace with resource quotas


hmoravec:
Traffic agent container which is injected by telepresence to deployment for intercepted service does not specify resources so it is not possible to use it in namespace with resource quotas set where specifying resources is mandatory. 

Is it possible to set the resources for traffic agent container?

LukeShu:
Something coming soon is the ability to install via a Helm chart in order to have more control over the install.  This would be a very good thing for us to include in that.

aldycool:
Hi, is this feature already configurable in the Helm chart?

cindymullins-dw:
If your cluster requires you to set resource limits, this is now available by setting `[agent.resources](https://www.getambassador.io/docs/telepresence/latest/reference/cluster-config#resources)`

PR #2773, opened 2022-09-07
Detect, report, and enable fix for resource quota violations

thallgren:
## Description

If a namespace has a resource quota, all containers of all pods in that namespace must specify their resources. Failing to do so, means that the pod cannot be created. This PR contains what's necessary to specify such resources for the injected traffic-agent and init-container.

The PR also contains logic to detect when an intended injection of the traffic-agent fails. This is done by watching cluster events, and filtering out the ones that is relevant for the injection. When errors occur, the user is informed about them and given an advice on how to move forward (typically adding resource definitions to the Helm chart).

In order to make the use-case "a) update helm chart with resources, b) run a helm upgrade", this PR also adds the logic necessary to regenerate the telepresence-agent config map. The regeneration takes place when the injector service starts,
and when SystemA changes its response to what traffic-agent image that should be used.

There's also some clean-up in the area of obtaining the traffic-agent image from SystemA that hopefully will offload that service quite a bit.

## Checklist


thallgren:
Closes #1806

cmd/traffic/cmd/agent/agent_test.go
```diff
@@ -10,8 +10,8 @@ import (
 
 	"github.com/spf13/afero"
 	"github.com/stretchr/testify/require"
-	"gopkg.in/yaml.v3"
 	core "k8s.io/api/core/v1"
+	"sigs.k8s.io/yaml"
```
raphaelreyna, line 14: Why the swap in yaml dep?

cmd/traffic/cmd/manager/internal/state/intercept.go
```diff
@@ -355,10 +341,68 @@ func waitForConfigMapUpdate(ctx context.Context, cmAPI typed.ConfigMapInterface,
...
-func (s *State) waitForAgent(ctx context.Context, name, namespace string) error {
+func watchFailedInjectionEvents(ctx context.Context, name, namespace string) (<-chan *events.Event, error) {
+	// A timestamp with second granularity is needed here, because that's what the event creation time uses.
+	// Finer granularity will result in relevant events seemingly being created before this timestamp because
+	// they have the fraction of seconds trimmed off (which is odd, given that the type used is a MicroTime).
+	start := time.Unix(time.Now().Unix(), 0)
+
+	ei := k8sapi.GetK8sInterface(ctx).EventsV1().Events(namespace)
+	w, err := ei.Watch(ctx, meta.ListOptions{
+		FieldSelector: fields.OneTermNotEqualSelector("type", "Normal").String()})
```
raphaelreyna, line 352: could this be further filtered down to only get events related to the agent?

cmd/traffic/cmd/manager/internal/state/intercept.go
```diff
@@ -355,10 +341,68 @@ func waitForConfigMapUpdate(ctx context.Context, cmAPI typed.ConfigMapInterface,
...
+				msg := fe.Note
+				switch fe.Reason {
+				case "Unhealthy":
+					// Let readiness probe continue, this isn't fatal
+					continue
+				case "BackOff":
+					// The traffic-agent container was injected, but it fails to start
+					msg = fmt.Sprintf("%s\nThe logs of %s %s might provide more details", msg, fe.Regarding.Kind, fe.Regarding.Name)
+				case "FailedCreate", "FailedScheduling":
+					// The injection of the traffic-agent failed for some reason, most likely due to resource quota restrictions.
```
raphaelreyna, line 399: I feel like the "most likely" part of this comment should be reflected in the message sent to the user unless we are actually certain that a resource quota restriction was the cause of the issue.

raphaelreyna (approved):
lgtm; just left a suggestion regarding user message and a couple of questions

cmd/traffic/cmd/manager/internal/state/intercept.go
thallgren: The filtering mechanism is very limited. It only allows for exact matches and only AND conditions can be used. That why I decided to just match abnormal events here and then do more advanced filtering further down in this method. It shouldn't matter much unless the cluster is in a very bad shape and sends a very large number of abnormal events.

cmd/traffic/cmd/manager/internal/state/intercept.go
thallgren: It should. I made the hint a bit more descriptive.

cmd/traffic/cmd/agent/agent_test.go
thallgren: It's well explained in the commit comment.

raphaelreyna (approved):
nice

PR #2779, opened 2022-09-12
Run go-fusefpt via gRPC so that we don't require a fuse installation

thallgren:
If we link the telepresence binary with a fuse library, then that
library must be installed in order for telepresence to run. This is a
problem on platforms like macOS and Windows where the fuse installation
is either proprietary or GPL3, because we cannot bundle it with our
installation.

Running a linked binary via gRPC will not stop the telepresence binary
from functioning in case the gRPC server doesn't start.


pkg/client/userd/fuseftp.go
```diff
@@ -0,0 +1,123 @@
...
+		}
+	}()
+
+	suffix := ""
+	if runtime.GOOS == "windows" {
+		suffix = ".exe"
+	}
+
+	downloadURL := fmt.Sprintf(latestFuseFTPRelease, runtime.GOOS, runtime.GOARCH, suffix)
+	dlog.Debugf(ctx, "About to download telepresence-pro from %s", downloadURL)
```
josecv: this should say `go-fuseftp`

josecv (approved):
nit re logging

PR #2787, opened 2022-09-19
DNS query types and other improvements

thallgren:
## Description

This PR contains a series of both new features and some bug fixes, mainly targeted at DNS. The PR should be reviewed one commit at a time, or it will be hard to make sense of it all.

Details in the comments.

## Checklist


pkg/dnsproxy/lookup.go
```diff
@@ -0,0 +1,324 @@
...
+	r := &net.Resolver{StrictErrors: true}
+	switch qType {
+	case dns.TypeA, dns.TypeAAAA:
+		ips, err := lookupIP(ctx, "ip", qName, r)
+		if err != nil {
+			return makeError(err)
+		}
+		for _, ip := range ips {
+			if ip4 := ip.To4(); ip4 != nil {
+				if qType == dns.TypeA {
```
raphaelreyna, line 191: just a style nitpick but why not turn this nested if statement into a single one with an `&&` like in the following `else if` case (or vice-versa)?

pkg/dnsproxy/lookup_linux.go
```diff
@@ -0,0 +1,40 @@
...
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/telepresenceio/telepresence/v2/pkg/iputil"
+	"github.com/telepresenceio/telepresence/v2/pkg/proc"
+)
+
+func externalLookup(ctx context.Context, host string, timeout time.Duration) iputil.IPs {
+	cmd := proc.CommandContext(ctx, "nslookup", fmt.Sprintf("-timeout=%.2g", timeout.Seconds()), host)
```
raphaelreyna, line 16: perhaps we should consider falling back to using `dig` if `nslookup` is not present

pkg/client/userd/trafficmgr/intercept.go
```diff
@@ -268,7 +268,11 @@ func (tm *TrafficManager) watchInterceptsLoop(ctx context.Context) error {
 func (tm *TrafficManager) handleInterceptSnapshot(ctx context.Context, podIcepts *podIntercepts, intercepts []*manager.InterceptInfo) {
 	tm.setCurrentIntercepts(ctx, intercepts)
 	podIcepts.initSnapshot()
-	namespaces := make(map[string]struct{})
+	tm.currentInterceptsLock.Lock()
+	active := len(tm.localIntercepts)
+	ins := tm.interceptedNamespace
+	tm.currentInterceptsLock.Unlock()
```
raphaelreyna, line 274: should this lock be acquired for the entire duration of this function?
I'm concerned that if two requests for intercepts in different namespaces come in at roughly the same time, the lock will bounce between the two in the following for loop.

pkg/client/cli/cliutil/cliutil.go
```diff
@@ -1,5 +1,21 @@
...
+type quitting struct{}
+
+func IsQuitting(ctx context.Context) bool {
+	if q, ok := ctx.Value(quitting{}).(bool); ok {
+		return q
+	}
+	return false
+}
+
+func QuittingContext(ctx context.Context) context.Context {
```
raphaelreyna: it looks like this is not used?

pkg/client/cli/cliutil/systema.go
```diff
@@ -62,10 +47,7 @@ func ClientEnsureLoggedIn(ctx context.Context, apikey string, connectorClient co
 
 // Logout logs out of Ambassador Cloud.  Returns an error if not logged in.
 func Logout(ctx context.Context) error {
-	err := WithConnector(ctx, func(ctx context.Context, connectorClient connector.ConnectorClient) error {
-		_, err := connectorClient.Logout(ctx, &empty.Empty{})
-		return err
-	})
+	_, err := GetUserDaemon(ctx).Logout(ctx, &empty.Empty{})
```
raphaelreyna, line 50: this is so much cleaner! :+1:

cmd/traffic/cmd/manager/internal/state/intercept.go
```diff
@@ -392,6 +392,11 @@ func (s *State) waitForAgent(ctx context.Context, name, namespace string, failed
 				case "Unhealthy":
 					// Let readiness probe continue, this isn't fatal
 					continue
+				case "FailedMount":
```
raphaelreyna: this might be worth logging at the debug level

raphaelreyna (approved):
very nice!
I reviewed this one commit at a time as you suggested so I think there are 2 comments/suggestions that were addressed/invalidated in a later commit.

pkg/client/userd/trafficmgr/intercept.go
thallgren: This particular function is never called concurrently. It's part of the `watchInterceptsLoop`, and there must only be one watcher. The locks protect variables that may be accessed concurrently from other places in the code. The critical sections that they lock are deliberately kept as short as possible, to allow for maximum concurrency (with other functions).

cmd/traffic/cmd/manager/internal/state/intercept.go
thallgren: It is logged by the producer of the event.

pkg/dnsproxy/lookup.go
thallgren: Can't do that, because the `else if` further down must never be executed if `ip4 != nil`, regardless if the `qType == dns.TypeA` or not.

pkg/dnsproxy/lookup_linux.go
thallgren: Perhaps. But I've never encountered a linux system with `dig`, but not with `nslookup`. They are both bundled in the same package (bind-utils/dnsutils).

pkg/client/cli/cliutil/cliutil.go
thallgren: True. I'll remove it.

josecv (approved):
lgtm i think i looked at all the commits and i can't really spot anything odd

Issue #2756, opened 2022-08-25
CreateContainerConfigError


EugeneAikashev:
**Describe the bug**
I am trying to intercept using the following command:
```
telepresence intercept -n api aikashev -p 5000
```
But receive the following error:
`telepresence: error: rpc error: code = DeadlineExceeded desc = request timed out while waiting for agent aikashev.api to arrive`

When I check the pods, I see:
```
$ kubectl -n api get pods | grep aikashev
aikashev-7c89b866cf-57d62                                         1/2     CreateContainerConfigError   0          23s
```

And there is a warning when I do a describe pod:
```
$ kubectl -n api describe pods aikashev-7c89b866cf-57d62

Events:
  Type     Reason     Age               From               Message
  ----     ------     ----              ----               -------
  Normal   Scheduled  37s               default-scheduler  Successfully assigned api/aikashev-7c89b866cf-57d62 to ip-10-66-148-213.ec2.internal
  Normal   Pulled     36s               kubelet            Container image "nginx:alpine" already present on machine
  Normal   Created    36s               kubelet            Created container app
  Normal   Started    36s               kubelet            Started container app
  Normal   Pulled     9s (x5 over 36s)  kubelet            Container image "docker.io/datawire/ambassador-telepresence-agent:1.12.9" already present on machine
  Warning  Failed     9s (x5 over 36s)  kubelet            Error: missing value for NAMESPACE_NAME, POD_NAME
```

**To Reproduce**
Here is my simple deployment and service manifests to reproduce 
```
---
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app: aikashev
  name: aikashev
  namespace: api
spec:
  replicas: 1
  selector:
    matchLabels:
      app: aikashev
  template:
    metadata:
      labels:
        app: aikashev
    spec:
      containers:
      - image: nginx:alpine
        name: app
        ports:
        - containerPort: 80
          name: http-main
          protocol: TCP
        env:
        - name: POD_NAME
          valueFrom:
            fieldRef:
              apiVersion: v1
              fieldPath: metadata.name
        - name: NAMESPACE_NAME
          valueFrom:
            fieldRef:
              apiVersion: v1
              fieldPath: metadata.namespace
        volumeMounts:
        - name: logs-app
          mountPath: /var/log/app
          subPathExpr: $(NAMESPACE_NAME)_$(POD_NAME)
        resources: {}
      volumes:
      - hostPath:
          path: /var/log
          type: ""
[...]

PR #2797, opened 2022-09-29
Fix environment variable interpolation in generated traffic-agent

thallgren:
## Description

Environment variables that are defined in a pod's container can be interpolated in other values using a `$(<env name>)` syntax. This commit ensures that such interpolation works with the env-vars that are copied, and prefixed into the traffic agent by also prefixing the actual interpolation. E.g.
```
volumeMounts:
  - name: $(TOKEN_VOLUME)
```
becomes:
```
volumeMounts:
  - name: $(_TEL_APP_A_TOKEN_VOLUME)
```
when copied into the agent container definition.

## Checklist

### DIFF
--- a/pkg/agentconfig/container.go
+++ b/pkg/agentconfig/container.go
@@ -1,14 +1,20 @@
 package agentconfig
 
 import (
+	"bytes"
+	"context"
+	"encoding/json"
 	"strconv"
 	"strings"
 
 	core "k8s.io/api/core/v1"
+
+	"github.com/datawire/dlib/dlog"
 )
 
 // AgentContainer will return a configured traffic-agent
 func AgentContainer(
+	ctx context.Context,
 	pod *core.Pod,
 	config *Sidecar,
 ) *core.Container {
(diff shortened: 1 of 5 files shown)


raphaelreyna (approved):
Nice! This addresses issue #2756 correct?

thallgren:
Yes, this will close #2756.

pkg/agentconfig/container.go
```diff
@@ -219,22 +241,27 @@ func appendAppContainerVolumeMounts(app *core.Container, cc *Container, mounts [
...
 		mounts = append(mounts, m)
 	}
 	return mounts
 }
 
-func appendAppContainerEnv(app *core.Container, cc *Container, es []core.EnvVar) []core.EnvVar {
+func appendAppContainerEnv(app *core.Container, cc *Container, es []core.EnvVar, subst map[string]string) []core.EnvVar {
 	for _, e := range app.Env {
-		e.Name = EnvPrefixApp + cc.EnvPrefix + e.Name
+		pn := EnvPrefixApp + cc.EnvPrefix + e.Name
```
serhii-nochevnyi, line 252: Maybe I'm wrong, but it seems there is a logical error here, which leads to the message "missing value for _TEL_APP_A__TEL_APP_A_NAMESPACE_NAME, _TEL_APP_A__TEL_APP_A_POD_NAME" 

due to the conversion

```
volumeMounts:
        - mountPath: .....
          name: ....
          subPathExpr: $(NAMESPACE_NAME)_$(POD_NAME) 
```

to 

```
subPathExpr: >-
        $(_TEL_APP_A__TEL_APP_A_NAMESPACE_NAME)_$(_TEL_APP_A__TEL_APP_A_POD_NAME)
```

in my opinion, it would not hurt to check for the already added prefix

```
if stings.HasPrefix(e.Name, EnvPrefixApp + cc.EnvPrefix) {
    continue
}
```

thallgren: @serhii-nochevnyi Why would the prefix already have been added to `$(NAMESPACE)` and `$(POD_NAME)`? That's IMO very unexpected and an indication that something else is broken.

thallgren: Investigating this further. The `e`  above stems from the `app` container. The for-loop uses values, not pointers, so each `e` is a distinct copy. The `app` container is the original container from the deployment template. It should definitely never be modified and hence never have `e.Name` prefixed with `_TEL_APP_A_`. So something else is going on here. The only thing I can think of is if the `app` container passed in here actually is the `traffic-agent` container itself.

PR #2805, opened 2022-10-06
Ensure that stdout/stderr are synchronized at the end of remote command

thallgren:
A remote command would sometimes never print the last output sent to stdout/stderr because the channel that they were sent to wasn't drained before the command ended. This commit ensures that the drain is made.

### DIFF
--- a/pkg/client/stdio.go
+++ b/pkg/client/stdio.go
@@ -1,10 +1,10 @@
 package client
 
 import (
-	"context"
 	"io"
 
 	"github.com/telepresenceio/telepresence/rpc/v2/connector"
+	"github.com/telepresenceio/telepresence/v2/pkg/client/errcat"
 )
 
 // StdOutput sends everything written to its Stdout and Stderr to its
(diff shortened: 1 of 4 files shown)


pkg/client/stdio.go
```diff
@@ -50,6 +51,18 @@ func (d dispatchToCh) Write(p []byte) (n int, err error) {
 
 type stdioHandler chan *connector.StreamResult
 
+func (h stdioHandler) Finish(err error) {
+	r := connector.StreamResult{Final: true}
```
raphaelreyna, line 55: shouldn't `r.Final` be `false` if the err is nil?

raphaelreyna (approved):
lgtm just one question regarding error signaling

pkg/client/stdio.go
thallgren: No, `final` is always `true` in the final message. This maps to this change on the CLI side of things (command_remote.go):
```
		if sr.Final {
			if r != nil {
				err = errcat.FromResult(r)
			}
			return err
		}
```
which means that the sender will return nil here, and trigger the deferred `CloseSend()` on the stream, which concludes the finalization of the stream.

PR #2806, opened 2022-10-06
Improve error message on failure to parse JSON files.

### LINKED ISSUE #2804: Unhelpful error message on JSON parse errors
**Describe the bug**

If the file `~/.cache/telepresence/update-checks.json` is corrupted, telepresence fails with an obscure error:

```
$ telepresence intercept foo 
telepresence: error: invalid character '}' after top-level value
```

**To Reproduce**
1. Edit the file `~/.cache/telepresence/update-checks.json`
2. Add a `}` at the end.
3. Run `telepresence intercept foo` (it doesn't matter what `foo` is - it doesn't even need to be the name of any deployment)
4. Telepresence fails with an error

**Expected behavior**
If telepresence fails to parse a local config file, the error message should contain the name/path of the file so that the user can fix the problem.

Also, the name `update-checks` suggests that this file is only used to decide when to check for new versions. This isn't essential functionality, so failing to parse the file probably shouldn't be a fatal error.

 - Output of `telepresence version`
```
$ telepresence version
Client: v2.7.6 (api v3)
Root Daemon: v2.7.6 (api v3)
User Daemon: v2.7.6 (api v3)
```
 - Operating system of workstation running `telepresence` commands: Ubuntu 22.04
 - Kubernetes environment and Version: EKS

**Additional context**
It's unclear what caused the file to become corrupted in the first place, but it definitely wasn't a result of deliberately editing it to introduce a syntax error.

petur:
## Description

Include the path to the file in the error message when JSON parse errors occur. This should make it easier to diagnose these errors.

Fixes: #2804

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### 2.8.0 (TBD)
 
 - Bugfix: The Helm chart value `appProtocolStrategy` is now correctly named (used to be `appPortStategy`)
+- Bugfix: Include file name in error message when failing to parse JSON file.
 
 ### 2.7.6 (September 16, 2022)
 
--- a/pkg/client/cache/cache.go
+++ b/pkg/client/cache/cache.go
@@ -3,6 +3,7 @@ package cache
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"os"
 	"path/filepath"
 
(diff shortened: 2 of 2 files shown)


thallgren (approved):

thallgren:
Looks good to me. Thanks for contributing to Telepresence!

PR #2807, opened 2022-10-07
Remove the concept of "remote command" and all non-oss related code

thallgren:
This PR strips almost all code related to Ambassador Cloud out of this repository. That code will instead be maintained in a private repository. The PR also contains many changes to make the code more extendable, so that our enhanced client can reuse code instead of copying int.

The corresponing PR for the pro daemon is datawire/telepresence-pro#62 (private repo).


josecv (approved):
This is all super; I don't really have any comments because it seems like a lot of it is just moving logic from place to place, but I really like it overall. I feel like this gives us a good place to extend stuff without having to butcher the OSS code, which was the goal.

Issue #2819, opened 2022-10-19
Add ability to specify port for the webhook


actualrockstar:
**Please describe your use case / problem.**
The prospect uses a private GKE cluster and needs to create a firewall rule to allow the webhook to work, but if they could have the webhook use port 443 instead of port 8443 they wouldn't need to wait for their dev/ops team to create the firewall rule for them. 

**Describe the solution you'd like**
Make the port for the webhook use port 443 instead of 8443 and have this configurable through the helm chart.

**Describe alternatives you've considered**
The alternative is to add the firewall rule for their private cluster. 

**Versions (please complete the following information)**
  - TP v2.7.6
  - GKE 1.22

**Additional context**
This is related to prospect Commscope

thallgren:
Fixed by #2822.

PR #2822, opened 2022-10-20
Change mutating webhook port to 443 and make it configurable.

thallgren:
## Description

Prior to this commit, the mutating webhook was listening to port `8443`. The port was selected so that it was above the privileged port range, and hence would cause no conflicts when running the container as an unprivileged user. This caused problems with firewalls that enable port `443` but blocks `8443` by default.

The remedy for this problem was to add:

    RUN setcap 'cap_net_bind_service+ep' /usr/local/bin/traffic

to the Dockerfile that builds the tel2 image. This enables the binary to listen to privileged ports without being root.

This commit also fixes a bug in the Helm chart, causing the existing `agentInjector.webhook.port` to be completely ignored. The default setting for that value was incidentally already `443`, and now that value is propagated correctly.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+### 2.8.3 (TBD)
+
+- Feature: The port used for the mutating webhook can be configured using the Helm chart setting
+  `agentInjector.webhook.port`.
+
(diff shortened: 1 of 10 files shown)


thallgren:
Closes #2819

josecv (approved):

PR #2823, opened 2022-10-20
Allow connecting to traffic managers not installed via helm

josecv:
There's lots of ways to set up a helm chart, and not all set up the helm metadata the same way the helm command does. With this change, as long as the traffic-manager service is present, all will be good.

## Description

## Checklist

### DIFF
--- a/pkg/client/userd/trafficmgr/traffic_manager.go
+++ b/pkg/client/userd/trafficmgr/traffic_manager.go
@@ -385,17 +385,9 @@ func connectMgr(
 		dlog.Errorf(ctx, "unable to get APIKey: %v", err)
 	}
 
-	dlog.Debug(ctx, "checking that traffic-manager exists")
-	coreV1 := k8sapi.GetK8sInterface(ctx).CoreV1()
-	if _, err := coreV1.Services(cluster.GetManagerNamespace()).Get(ctx, "traffic-manager", meta.GetOptions{}); err != nil {
-		msg := fmt.Sprintf("unable to get service traffic-manager in %s: %v", cluster.GetManagerNamespace(), err)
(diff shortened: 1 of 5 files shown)


pkg/client/userd/trafficmgr/traffic_manager.go
```diff
@@ -385,13 +386,16 @@ func connectMgr(
...
 
 	dlog.Debug(ctx, "checking that traffic-manager exists")
-	existing, _, isManagerErr := helm.IsTrafficManager(ctx, cluster.ConfigFlags, cluster.GetManagerNamespace())
-	if isManagerErr == nil && existing == nil {
-		return nil, errcat.User.New("traffic manager not found, if it is not installed, please run 'telepresence helm install'")
-	}
-
-	if isManagerErr != nil {
-		dlog.Infof(ctx, "unable to look for existing helm release: %v. Assuming it's there and continuing...", isManagerErr)
+	coreV1 := k8sapi.GetK8sInterface(ctx).CoreV1()
```
thallgren: So this leaves us with two ways of detecting the traffic-manager install? One in `helm.IsTrafficManager` and one right here. I guess what I wonder is, why isn't this code part of `helm.IsTrafficManager`?

josecv: My logic was that `helm.IsTrafficManager` checks the helm release metadata, not the presence of a kubernetes service. I think what we should actually do is make it private, since now it's only used within the helm package. I can move this logic into some `trafficmgr.ServiceInstalled` or similar

thallgren: Yes,that sounds like a good idea.

thallgren (approved):

PR #2836, opened 2022-10-25
Add repeated --get flag to telepresence helm install

thallgren:
## Description

This adds the ability to specify values on the command line in addition to passing them in a yaml file by adding a repeated `--get` option. The flag value must be in the form `key=value` where the key can be dot separated.

## Checklist


josecv (changes requested):
lgtm but i think it should be called `--set` bc that's what it's called by the `helm install` CLI

thallgren:
Doh, of course it should be called `--set`.

josecv (approved):

PR #2838, opened 2022-10-26
Update integration tests to run on Mac M1

ghost:
## Description

This PR refactors the integration workflow as follows:
  - Removed action rerun-cache. This is not required since GitHub already supports re-running only failed jobs.
  - Updated actions to the latest version running on NodeJS 16. Required to run CLI tests on ARM64 platforms.
  - Simplified integration test job by using bash on Windows.
  - Consolidated installation of dependencies under a composite action to improve readability of the integration workflow.
  - Implemented test matrix.

[This](https://github.com/telepresenceio/telepresence/actions/runs/3371576352) run shows how will the integration test look like once this PR is merged.

As suggested by @rhs, I forced a test to fail to ensure that matrix behaves as expected. See job failure [here](https://github.com/telepresenceio/telepresence/actions/runs/3380068139/jobs/5612476046)

## Checklist


.github/workflows/dev.yaml
```diff
@@ -228,7 +228,7 @@ jobs:
       - uses: actions/setup-go@v2
         with:
           go-version: '1.19'
-      - uses: azure/setup-kubectl@v1
+      - uses: azure/setup-kubectl@v3.0
```
thallgren: Why use `v3.0` here and not just `v3`? I would assume that the latter would auto-upgrade the minor version.

.github/actions/install-dependencies/action.yaml
```diff
@@ -10,9 +10,9 @@ runs:
     - uses: actions/setup-go@v3
       with:
         go-version: '1.19'
-    - uses: azure/setup-kubectl@v1
+    - uses: azure/setup-kubectl@v3.0
```
thallgren: Same comment as before. Why `v3.0` and not just `v3`?

.github/workflows/dev.yaml
```diff
@@ -29,6 +29,7 @@ jobs:
           path: build-output/tel2-image.tar
   run_tests:
     strategy:
+      fail-fast: false
```
thallgren, line 32: Does this mean that we have to wait for all jobs to complete just in order to view the logs of one that failed?

thallgren (approved):
Looks great overall. Just a couple of questions.

.github/actions/install-dependencies/action.yaml
ghost: Good catch! I have updated it.

.github/workflows/dev.yaml
ghost: This has been moved to the file [.github/actions/install-dependencies/action.yaml](https://github.com/telepresenceio/telepresence/pull/2838/files/e58897d228406e995040de443ab24feaba2f9ebf..81bd18ef24f405e2c209258d4b87f54d73088d83#diff-342d59190b4737ee45e2062eb625ada477bcea5b4a843b25900ad55d7982f200). I updated that instance.

.github/workflows/dev.yaml
ghost: No, what this does is that if one job fails, it won't abort all the others.
But as soon as a job finishes, the logs will be visible, same as before.

josecv (approved):

PR #2843, opened 2022-10-28
Include traffic-manager version in version command output

thallgren:
## Description

Include the version of the traffic-manager in the output from `telepresence version` when connected.

## Checklist

### DIFF
--- a/pkg/client/cli/cmd_version.go
+++ b/pkg/client/cli/cmd_version.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/telepresenceio/telepresence/rpc/v2/common"
 	"github.com/telepresenceio/telepresence/rpc/v2/daemon"
+	"github.com/telepresenceio/telepresence/rpc/v2/manager"
 	"github.com/telepresenceio/telepresence/v2/pkg/client"
 	"github.com/telepresenceio/telepresence/v2/pkg/client/cli/ann"
 	"github.com/telepresenceio/telepresence/v2/pkg/client/cli/cliutil"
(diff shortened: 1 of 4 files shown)


pkg/client/cli/cmd_version.go
```diff
@@ -52,6 +53,14 @@ func printVersion(cmd *cobra.Command, _ []string) error {
...
 	case err == nil:
 		fmt.Fprintf(cmd.OutOrStdout(), "User Daemon: %s (api v%d)\n", version.Version, version.ApiVersion)
+		var mgrVer *manager.VersionInfo2
+		mgrVer, err = managerVersion(ctx)
+		switch {
+		case err == nil:
+			fmt.Fprintf(cmd.OutOrStdout(), "Traffic Manager: %s\n", mgrVer.Version)
+		default:
+			fmt.Fprintf(cmd.OutOrStdout(), "Traffic Manager: error: %v\n", err)
+		}
```
njayp, line 67: If the daemons are running but not connected, I think this will print an rpc error. I think we should print something like "Tele is not connected to a traffic manager" if we are not connected. We should add a test for this too.

njayp (changes requested):

pkg/client/cli/cmd_version.go
thallgren: Good catch. I'll fix that.

thallgren: I added a "Traffic Manager: not connected" state (wording similar to "XXX Daemon: not running").

njayp (approved):

PR #2846, opened 2022-11-01
Workflow for testing server/client compatibility.

rhs:
## Description

This PR includes a workflow for testing nightly builds against older client/server versions. The matrix of versions can reference nightly builds (e.g. "nightly"), absolution versions (e.g. "2.4.1"), and relative versions (e.g. "-1" is the current release, "-2" is the most recent Z release of one minor version prior to the current release).

## Checklist

### DIFF
--- a/.github/workflows/compatibility.yaml
+++ b/.github/workflows/compatibility.yaml
@@ -40,6 +40,8 @@ jobs:
           ./telepresence-server version
 
           ./telepresence-server helm install
+          # Quit the user and root daemons since they get started by the installation.
+          ./telepresence-server quit
           ./telepresence-client connect
           ./telepresence-client status
           ./telepresence-client list
--- a/.github/workflows/compatibility.yaml
+++ b/.github/workflows/compatibility.yaml
@@ -86,9 +86,8 @@ jobs:
           ./telepresence-server helm install
 
           # Quit the user and root daemons since they get started by the installation. Use multiple
-          # different ways since the arguments have changed.
-          ./telepresence-server quit -u || true
-          ./telepresence-server quit -r || true
+          # different ways and ignore failure since the arguments have changed.
+          ./telepresence-server quit -ur || true
           ./telepresence-server quit -s || true
 
           ./telepresence-client connect
--- a/.github/workflows/compatibility.yaml
+++ b/.github/workflows/compatibility.yaml
@@ -1,0 +1,46 @@
+# This github workflow tests client and server compatibility for telepresence using nightly builds.
+#
+# See the "matrix" field below for which versions are tested.
+
+name: "Telepresence Test Matrix"
+on:
+  # Run daily at 9am UTC.
+  schedule:
+    - cron: '0 9 * * *'
+  # Run whenever this file is modified.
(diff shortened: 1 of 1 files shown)


.github/workflows/compatibility.yaml
```diff
@@ -0,0 +1,54 @@
...
+  push:
+    paths:
+      - '.github/workflows/compatibility.yaml'
+
+jobs:
+  telepresence_matrix:
+    strategy:
+      fail-fast: false
+      matrix:
+        client_version: ["nightly"]
```
AidanWenzel, line 27: Do we need to test server -> old client compatibility also? Or is one direction enough?

I guess that makes the most sense as a separate matrix 🤔

rhs: If we want to test old client versions we can just extend the matrix to do that. My goal with the PR is mostly to get the mechanics in place.

.github/workflows/compatibility.yaml
```diff
@@ -40,6 +40,8 @@ jobs:
           ./telepresence-server version
 
           ./telepresence-server helm install
+          # Quit the user and root daemons since they get started by the installation.
+          ./telepresence-server quit
```
thallgren: A `telepresence quit` doesn't quit the daemons. It just quits the current session.  Use `-s` to ensure that the daemons really die.

.github/workflows/compatibility.yaml
```diff
@@ -18,7 +18,7 @@ jobs:
       fail-fast: false
       matrix:
         client_version: ["nightly"]
-        server_version:  ["nightly", "2.8.3", "2.7.6", "2.6.8"]
+        server_version:  ["nightly", "2.8.3", "2.7.6"]
```
thallgren: "2.8.5" is the latest one.

.github/workflows/compatibility.yaml
```diff
@@ -41,8 +41,13 @@ jobs:
...
 
           ./telepresence-server helm install
-          # Quit the user and root daemons since they get started by the installation.
-          ./telepresence-server quit
+
+          # Quit the user and root daemons since they get started by the installation. Use multiple
+          # different ways since the arguments have changed.
+          ./telepresence-server quit -u || true
+          ./telepresence-server quit -r || true
+          ./telepresence-server quit -s || true
```
thallgren, line 91: Two alternatives are enough. The old `quit -ur` and the new `quit -s`.

thallgren (approved):
A few nits, but looks good overall.

josecv (approved):

.github/workflows/compatibility.yaml
rhs: @thallgren You probably figured this out, but the more recent version does what you suggest. I should have either rebased or mentioned that the history is a bit dirty due to the usual CI debug dance and it's better to review the full diff.

.github/workflows/compatibility.yaml
rhs: @thallgren Just pushed a commit combining the -u and -r.

PR #2848, opened 2022-11-02
add shared login

njayp:

## Description
This PR addresses shared login. It adds a configmap to the helm chart. It watches for a configmap or secret with the correct data entry. If one does not exist, it patches that configmap with the apikey if the user connects while logged on or logs in while connected. Either way, once a apikey is in the cluster, the a-agent can login, so the watchers shut down. I also added the RBAC necessary for these changes.

I have to do a couple more things (like changelog) before merging, but it is ready for review

## Checklist

### DIFF
--- a/DEPENDENCIES.md
+++ b/DEPENDENCIES.md
@@ -34,6 +34,7 @@ following Free and Open Source software:
     github.com/datawire/envconfig                                                v0.0.0-20221012222025-09524dc7d59b        Apache License 2.0
     github.com/datawire/go-ftpserver                                             v0.1.1                                    Apache License 2.0
     github.com/datawire/go-fuseftp/rpc                                           v0.2.0                                    Apache License 2.0
+    github.com/datawire/k8sapi                                                   v0.1.1-0.20221102000402-bab67787b913      Apache License 2.0
     github.com/datawire/metriton-go-client                                       v0.1.1                                    Apache License 2.0
     github.com/davecgh/go-spew                                                   v1.1.1                                    ISC license
     github.com/docker/cli                                                        v20.10.18+incompatible                    Apache License 2.0
@@ -136,13 +137,13 @@ following Free and Open Source software:
     github.com/xlab/treeprint                                                    v1.1.0                                    MIT license
     go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc  v0.36.0                                   Apache License 2.0
     go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp                v0.36.0                                   Apache License 2.0
-    go.opentelemetry.io/otel                                                     v1.10.0                                   Apache License 2.0
+    go.opentelemetry.io/otel                                                     v1.11.1                                   Apache License 2.0
     go.opentelemetry.io/otel/exporters/otlp/internal/retry                       v1.10.0                                   Apache License 2.0
     go.opentelemetry.io/otel/exporters/otlp/otlptrace                            v1.10.0                                   Apache License 2.0
     go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc              v1.10.0                                   Apache License 2.0
     go.opentelemetry.io/otel/metric                                              v0.32.1                                   Apache License 2.0
     go.opentelemetry.io/otel/sdk                                                 v1.10.0                                   Apache License 2.0
-    go.opentelemetry.io/otel/trace                                               v1.10.0                                   Apache License 2.0
+    go.opentelemetry.io/otel/trace                                               v1.11.1                                   Apache License 2.0
     go.opentelemetry.io/proto/otlp                                               v0.19.0                                   Apache License 2.0
     go.starlark.net                                                              v0.0.0-20220817180228-f738f5508c12        3-clause BSD license
     golang.org/x/crypto                                                          v0.0.0-20220924013350-4ba4fb4dd9e7        3-clause BSD license
--- a/DEPENDENCIES.md
+++ b/DEPENDENCIES.md
@@ -34,7 +34,7 @@ following Free and Open Source software:
     github.com/datawire/envconfig                                                v0.0.0-20221012222025-09524dc7d59b        Apache License 2.0
     github.com/datawire/go-ftpserver                                             v0.1.1                                    Apache License 2.0
     github.com/datawire/go-fuseftp/rpc                                           v0.2.0                                    Apache License 2.0
-    github.com/datawire/k8sapi                                                   v0.1.1-0.20221102000402-bab67787b913      Apache License 2.0
+    github.com/datawire/k8sapi                                                   v0.1.1-0.20221102212418-ff24519a45be      Apache License 2.0
     github.com/datawire/metriton-go-client                                       v0.1.1                                    Apache License 2.0
     github.com/davecgh/go-spew                                                   v1.1.1                                    ISC license
     github.com/docker/cli                                                        v20.10.18+incompatible                    Apache License 2.0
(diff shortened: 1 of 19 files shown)


DEPENDENCIES.md
```diff
@@ -34,6 +34,7 @@ following Free and Open Source software:
     github.com/datawire/envconfig                                                v0.0.0-20221012222025-09524dc7d59b        Apache License 2.0
     github.com/datawire/go-ftpserver                                             v0.1.1                                    Apache License 2.0
     github.com/datawire/go-fuseftp/rpc                                           v0.2.0                                    Apache License 2.0
+    github.com/datawire/k8sapi                                                   v0.1.1-0.20221102000402-bab67787b913      Apache License 2.0
```
thallgren: I like this, but shouldn't it also mean that we remove the `pkg/k8sutil` directory altogether? I mean, what's the point in having two incarnations?

Also, it seems a lot of the `watcher` functionality has generalized and moved into that package now, which means that we have multiple copies of things like `Subscribe` (copied from `pkg/client/userd/trafficmgr/workloads.go`. I think when such things are generalized, the original should be removed.

That said, I don't think this is related to the shared login, so perhaps you could create a separate PR that deals with all the "swap k8sapi" stuff?

cmd/traffic/cmd/manager/internal/ambassadoragent/cloudtoken/service.go
```diff
@@ -0,0 +1,121 @@
...
+
+var (
+	suffix               = "agent-cloud-token"
+	cloudConnectTokenKey = "CLOUD_CONNECT_TOKEN"
+)
+
+// NewCreateConfigmapIfNotPresent will start up watchers to watch for a configmap or secret
+// with the apikey. If one is found, the watchers shut down. If MaybeAddToken is called
+// and successfully creates a configmap, the watchers are shut down.
+func NewCreateConfigmapIfNotPresent(ctx context.Context) *createConfigmapIfNotPresent {
```
thallgren: This function contains three fairly well contained and separate things. There's creation of the watcher, the search function, and the watcherloop. Keeping those three together make this function complex and harder to read. You end up with names like `cancelCtx`and `subCtx` because they are cancelled for different reasons.

cmd/traffic/cmd/manager/internal/ambassadoragent/cloudtoken/service.go
```diff
@@ -0,0 +1,121 @@
...
+	apiv1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+type Service interface {
+	MaybeAddToken(ctx context.Context, apikey string) error
+}
+
+type createConfigmapIfNotPresent struct {
+	createToken func(ctx context.Context, apikey string) error
```
thallgren: What motivates this field? Can't it just be a function?

cmd/traffic/cmd/manager/internal/ambassadoragent/cloudtoken/service.go
```diff
@@ -0,0 +1,121 @@
...
+	done        <-chan struct{}
+}
+
+var (
+	suffix               = "agent-cloud-token"
+	cloudConnectTokenKey = "CLOUD_CONNECT_TOKEN"
```
thallgren: Those should be `const`, not `var`.

cmd/traffic/cmd/manager/internal/ambassadoragent/cloudtoken/service.go
```diff
@@ -0,0 +1,121 @@
...
+	clientset := k8sapi.GetK8sInterface(ctx)
+	watchers := newTokenWatchers(clientset, managerns)
+
+	// ctx for watchers, cancelled if token is found
+	cancelCtx, cancel := context.WithCancel(ctx)
+
+	go func() {
+		// search checks if apikey token exists, if it does, shut down the watchers
+		search := func() {
+			configmaps, err := watchers.mapsWatcher.List(cancelCtx)
```
thallgren: There are two watchers, and one cond (one subscribe), which means that when something changes, you need to search both resources. Wouldn't it be simpler and more efficient to have two separate watchers, subscribers, and listers?

thallgren (changes requested):

cmd/traffic/cmd/manager/internal/ambassadoragent/cloudtoken/service.go
```diff
@@ -0,0 +1,121 @@
...
+			return err
+		},
+		done: cancelCtx.Done(),
+	}
+}
+
+// MaybeAddToken will add a token if one does not already exist
+func (c *createConfigmapIfNotPresent) MaybeAddToken(ctx context.Context, apikey string) error {
+	select {
+	case <-c.done: // apikey is already present or watcher err, do nothing
```
thallgren: What happens if the apiKey is removed from the config map after an API key has been found (or if the secret is removed )? Seems to me like the `createToken` will never be called unless the traffic-manager is restarted.

DEPENDENCIES.md
njayp: Yeah, I can do that in another PR in the near future

cmd/traffic/cmd/manager/internal/ambassadoragent/cloudtoken/service.go
njayp: Meh. If we need to manage the token, we have to keep both watchers alive indefinitely, which is ok but i dont like it. Imo, if someone deletes the apikey, they can add it again too. If they really want the traffic manager to do it, they can roll the traffic manager

I will document this ofc

cmd/traffic/cmd/manager/internal/ambassadoragent/cloudtoken/service.go
njayp: probably yes

cmd/traffic/cmd/manager/internal/ambassadoragent/cloudtoken/service.go
njayp: I feel like in the future we could present options, like `createConfigmapIfNotPresent` or `createSecretIfNotPresent`, which could both impl cloudToken.Service

cmd/traffic/cmd/manager/internal/ambassadoragent/cloudtoken/service.go
thallgren: I don't understand why the watcher is a one-shot thing. If we can assume that the traffic manager can be rolled out when the key is re-added, why can't we assume that the key has been added prior to it being installed and skip the watcher altogether?

In my view, we either assume that it's there, or we assume that it it's controlled by someone else (it can be created, updated, or deleted). Just catering for the fact that it may exist already or arrive after the traffic-manager boots up is, IMO, a somewhat strange middleground. What happens if the token expires and is updated?

njayp: I dont think the a-agent cares about expired tokens, it opens the connection and never closes it. I can ask the run team if expired tokens have been a problem.

I dont want to step on emissary apikeys because emissary requires a user to manually set the apikey, meaning it is likely an org key, and the incoming login might be a personal key, which would reduce functionality.

DEPENDENCIES.md
```diff
@@ -34,6 +34,7 @@ following Free and Open Source software:
     github.com/datawire/envconfig                                                v0.0.0-20221012222025-09524dc7d59b        Apache License 2.0
     github.com/datawire/go-ftpserver                                             v0.1.1                                    Apache License 2.0
     github.com/datawire/go-fuseftp/rpc                                           v0.2.0                                    Apache License 2.0
+    github.com/datawire/k8sapi                                                   v0.1.1-0.20221102212418-ff24519a45be      Apache License 2.0
```
thallgren: I'd like to avoid introducing this package until we also remove the existing one.

PR #2857, opened 2022-11-03
rm secret watcher

njayp:

## Description
Sometimes you write code, and then you write more code, and you dont realize your old code is kinda worthless. So it gets merged. And then you realize afterwards that you dont need that old code anymore. So you have to make a PR of shame to delete the unneeded stuff.

We don't need the secret watcher :sadpandas:

### DIFF
--- a/charts/telepresence/templates/trafficManagerRbac/cloud-token.yaml
+++ b/charts/telepresence/templates/trafficManagerRbac/cloud-token.yaml
@@ -13,13 +13,6 @@ rules:
       - list
       - watch
       - patch
-  - apiGroups:
-      - ""
-    resources:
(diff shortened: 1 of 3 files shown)


thallgren (approved):
Looks good to me, except for the fact that plural "watchers" is now singular "watcher", so the file, type, and local variable can be renamed. I'll leave that up to you.

njayp:
> Looks good to me, except for the fact that plural "watchers" is now singular "watcher", so the file, type, and local variable can be renamed. I'll leave that up to you.

I think i will spend some cooldown time replacing pkg/k8sapi with datawire/k8sapi, ill do the renaming then

PR #2866, opened 2022-11-06
Add Helm chart setting to disable personal intercepts.

MaxMuzh:
Signed-off-by: Max Muzh <mmuzh@augury.com>

## Description

Adds a new intercept structure to the Helm chart with one single `disablePersonal` boolean setting which defaults to false. When set to true, the traffic-manager will respond with an error if an attempt is made to create an intercept that doesn't use the `tcp` mechanism.

## Checklist


thallgren:
Thanks for providing this, but I'm afraid this feature was already released in Telepresence 2.8.3. See Helm chart setting `intercept.disableGlobal`.

thallgren:
@MaxMuzh Can you please provide a motivation in the commit comment explaining why this is needed?

MaxMuzh:
Hi @thallgren,
Thank you for your reply.
In our company we use only Global intercepts because our system doesn't entirely supports passing of http headers.
In addition, as users of the open source version of telepresence we sometime violate by mistake your EULA by using too many personal intercepts (which doesn't work for us as explained above)

thallgren:
@MaxMuzh You can already accomplish what you want by passing the following values to the Helm chart:
```yaml
agent:
  image:
    name: tel2
    tag: 2.8.5
```
This will cause the traffic manager to always inject the OSS agent. It's not capable of doing http intercepts and will not accept the http flags.

thallgren:
I'm closing this for two reasons:
1. Another solution for the problem already exists.
2. Similar to the traffic-agent, the telepresence client will be also available in two shapes in the near future. One pure OSS that doesn't accept --http options, and one enhanced "pro" version that does.

GuyCarmy:
Hi @thallgren, is there already a pure OSS client like described here?

PR #2867, opened 2022-11-07
Fix inconsistencies when --output=json is used.

### LINKED ISSUE #2854: intercept creation json output is proceeded by update message
**Describe the bug**
When creating an intercept with the `--output json` flag, the upgrade message appears before the json object in stdout. An example of stdout is pasted below.

`An update of telepresence from version 2.8.2 to 2.8.5 is available. Please visit https://www.getambassador.io/docs/telepresence/latest/install/upgrade/ for more info.
{"cmd":"intercept","stdout":"Using Deployment dataprocessingservice\nintercepted\n    Intercept name         : dataprocessingservice-default\n    State                  : ACTIVE\n    Workload kind          : Deployment\n    Destination            : 127.0.0.1:3000\n    Service Port Identifier: 3000\n    Volume Mount Error     : sshfs is not installed on your local machine\n    Intercepting           : HTTP requests with headers\n        'x-telepresence-intercept-id: 20c7da00-028e-4358-b9d8-5cb9d18ec786:dataprocessingservice-default'\n    Preview URL            : https://wonderful-napier-149.preview.edgestack.me/\n    Layer 5 Hostname       : verylargejavaservice.default.svc.cluster.local\n"}`

**To Reproduce**
1. Use an older version of telepresence
2. Run `telepresence intercept --output json ...`
3. The output has the update message before the json

**Expected behavior**
The output is in json, the update message is not printed

 - >=2.6.6 or something like that

**Additional context**
It is a problem for the docker extension

thallgren:
## Description

Some output, like "An update of telepresence ...", "Launching Telepresence Root Daemon", and "Connected to context ..." was printed verbatim as unformatted text instead of being included in the JSON output when the `--output=json` flag was used.

This PR ensures that information like that is discarded so that the JSON output is always valid, and the "stdout" is kept consistent.

The PR also adds `yaml` to the set of valid values for the `--output` flag and ensures that invalid values render a proper error message, and that attempts to combine string output with objects on formatted stdout renders a panic (since it's a coding error).

Closes #2854.

## Checklist


njayp (approved):

pkg/client/cli/cmd_status.go
```diff
@@ -69,18 +69,35 @@ func statusCommand() *cobra.Command {
...
+		Short:             "Show connectivity status",
+		RunE:              s.status,
+		PersistentPreRunE: fixFlag,
 		Annotations: map[string]string{
 			ann.RootDaemon: ann.Optional,
 			ann.UserDaemon: ann.Optional,
 		},
 	}
 	flags := cmd.Flags()
 	flags.BoolVarP(&s.json, "json", "j", false, "output as json object")
```
njayp, line 81: Can we remove this flag? For neatness, I would prefer that we have one json flag that also influences the behavior of certain subcommands like list

pkg/client/cli/cmd.go
```diff
@@ -268,7 +271,7 @@ func GlobalFlags() *pflag.FlagSet {
 	)
 	flags.String(
 		"output", "default",
-		"set the output format, supported values are 'json' and 'default'",
+		"set the output format, supported values are 'json', 'yaml', and 'default'",
 	)
```
njayp: can we have `-o yaml` like kubectl?

pkg/client/cli/output/output.go
```diff
@@ -1,162 +1,212 @@
...
-			return f(cmd, args)
+// Object sets the object to be marshalled and printed on stdout when formatted output
+// is requested using the `--output=<fmt>` flag. Otherwise, this function does nothing.
+//
+// If override is set to true, then the formatted output will consist solely of the given
+// object. There will be no "cmd", "stdout", or "stderr" tags.
+//
+// The function will panic if data already has been written to the stdout of the command
+// or if an Object already has been called.
+func Object(ctx context.Context, obj any, override bool) {
```
raphaelreyna, line 62: hmm the name `override` might be a bit confusing here, perhaps `raw` is more descriptive?

pkg/client/cli/output/output.go
```diff
@@ -1,162 +1,212 @@
...
+	if cmd, ok := ctx.Value(key{}).(*cobra.Command); ok {
+		if o, ok := cmd.OutOrStdout().(*output); ok {
+			if o.Len() > 0 {
+				panic("output.Object cannot be used together with output.Out")
+			}
+			if o.obj != nil {
+				panic("output.Object can only be used once")
```
raphaelreyna, line 69: in the case of a panic, we might still want to bubble up the object if its a `fmt.Stringer`, especially if we're choosing to not use json output on a panic.

raphaelreyna (approved):
nice :+1:; just 2 minor nitpicks

pkg/client/cli/cmd_status.go
thallgren: I would love to remove it, but removing it means breaking backward compatibility. So I hide it instead, and let it have the exact same meaning as `--output=json`.

pkg/client/cli/output/output.go
thallgren: I'm not sure I see the point in doing that. A panic here means that a developer made a coding error. It's not intended for the user.

pkg/client/cli/output/output.go
thallgren: Is it? I think "raw" can be confused with "unformatted" or "binary". I wanted to make it clear that the normal wrapping is overridden and I try to describe that in the function docs.

PR #2873, opened 2022-11-08
Helm chart client config improvments.

thallgren:
## Description

The PR contains some improvements for the Helm chart based client config.

## Checklist


cmd/traffic/cmd/manager/internal/cliconfig/cliconfig.go
```diff
@@ -3,98 +3,95 @@ package cliconfig
...
+	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/watch"
 
 	"github.com/datawire/dlib/dlog"
+	"github.com/telepresenceio/telepresence/v2/pkg/k8sapi"
 )
 
-const cfgFileName = "client.json"
+const (
+	cfgFileName      = "client.yaml"
```
raphaelreyna, line 17: should we accept both json and yaml going forward if json is something we used to support?

raphaelreyna (approved):
nice, I'm just a bit unfamiliar with how the json config used to work so there's a qq on that.

cmd/traffic/cmd/manager/internal/cliconfig/cliconfig.go
thallgren: Backward compatibility is not an issue in this case, because this whole thing is brand new. The configmap doesn't exist in 2.8.5.

PR #2877, opened 2022-11-10
feat(chart): support livenessProbe and readinessProbe

maximee-leroyy:
## Description

Adds a livenessProbe and readinessProbe so that it restarts the pod if it hangs up for whatever reason

## Checklist


charts/telepresence/CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 # Change Log
 
-### 2.6.8 (TBD)
+### 2.9.0
+
+- Feature: add `livenessProbe`, `readinessProbe` value.
+
```
thallgren: This must be moved to the CHANGELOG.md at the repository root. Also, please make the entry more verbose. Where is it added? For what purpose? What pod is it for?

The CHANGELOG.md here is no longer maintained (and yes, we should remove it).

thallgren (changes requested):
Thanks for providing this. It looks besides my comment about the changelog.

You'll need to sign off on the commit to pass our DCO check.

charts/telepresence/CHANGELOG.md
maximee-leroyy: okay thanks for the review, i have done the changes!

maximee-leroyy:
> Thanks for providing this. It looks besides my comment about the changelog.
> 
> You'll need to sign off on the commit to pass our DCO check.

thanks for the review @thallgren, it should be good now!

thallgren (approved):

PR #2889, opened 2022-11-17
Restore the ability to set loglevel in the root daemon.

thallgren:
## Description

The ability set the loglevel in the root daemon using the command `telepresence loglevel <level>` got lost 2.8.0 rewrite, where the CLI no longer communicates directly with the root daemon. The call was removed from the CLI but never added to the user daemon.

This PR adds the missing `SetLogLevel` call from the user daemon to the root daemon.

## Checklist

### DIFF
--- a/pkg/client/userd/daemon/grpc.go
+++ b/pkg/client/userd/daemon/grpc.go
@@ -392,6 +392,10 @@ func (s *Service) SetLogLevel(ctx context.Context, request *manager.LogLevelRequ
 		} else {
 			result = &empty.Empty{}
 		}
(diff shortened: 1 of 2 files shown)


pkg/client/userd/daemon/grpc.go
```diff
@@ -392,6 +392,10 @@ func (s *Service) SetLogLevel(ctx context.Context, request *manager.LogLevelRequ
 		} else {
 			result = &empty.Empty{}
 		}
+		_ = s.withRootDaemon(ctx, func(ctx context.Context, rd daemon.DaemonClient) error {
+			_, _ = rd.SetLogLevel(ctx, request)
```
njayp: why aren't we returning err here?

njayp (approved):

PR #2891, opened 2022-11-17
Improve tunnel performance

thallgren:
## Description

Improve performance of outbound tunnel traffic by eliminating two unnecessary I/O pumps and one extra process jump.

The PR also contains a fix for a race condition when quitting and reconnecting the root daemon.

## Checklist


pkg/client/userd/daemon/mgrproxy.go
```diff
@@ -218,91 +216,8 @@ func (p *mgrProxy) AgentTunnel(managerrpc.Manager_AgentTunnelServer) error {
...
-
-	wg := sync.WaitGroup{}
-	wg.Add(4)
-	go recvLoop(ctx, "manager", fhManager, mgrToClient, &wg)
-	go sendLoop(ctx, "manager", fhManager, clientToMgr, &wg)
-	go recvLoop(ctx, "client", fhClient, clientToMgr, &wg)
-	go sendLoop(ctx, "client", fhClient, mgrToClient, &wg)
-	wg.Wait()
-	return nil
+func (p *mgrProxy) Tunnel(managerrpc.Manager_TunnelServer) error {
```
raphaelreyna, line 219: Can this be deleted if it's no longer being used?

pkg/dnet/kpfconn.go
```diff
@@ -66,20 +68,19 @@ func NewK8sPortForwardDialer(logCtx context.Context, kubeConfig *rest.Config, k8
...
 		dlog.Errorf(pf.logCtx, "Error with k8sPortForwardDialer dial: %s", err)
 		return nil, err
 	}
-	inner, err := pf.dial(pod, podPortNumber)
+	conn, err := pf.dial(pod, podPortNumber)
 	if err != nil {
 		dlog.Errorf(pf.logCtx, "Error with k8sPortForwardDialer dial: %s", err)
-		return nil, err
 	}
-	return wrapUnbufferedConn(inner), nil
```
raphaelreyna, line 82: There's a comment on `dnet.bufferedConn` explaining why a buffered connection was used here; is that no longer applicable?

raphaelreyna (approved):
lgtm; a cool and clean approach. Just added 2 questions.

pkg/client/userd/daemon/mgrproxy.go
thallgren: Deleting it means that the proxied interface is no longer implemented fully, so no.

pkg/dnet/kpfconn.go
thallgren: No it's not. It might be applicable in other implementations of `unbufferedConn`, but the kpfConn no longer implements that interface. It contains a `spdystream.Stream` that implements the `net.Conn` interface, so there's no need for any `Recv` or `Send`. Instead, we use that `net.Conn` directly. The gvisor network stack performs the necessary TCP buffering, so adding more buffers here is pointless.

pkg/client/userd/daemon/mgrproxy.go
raphaelreyna: Ah yeah, I figured it had been left behind to satisfy some interface but hadn't realized it was for manager server rpcs.
:+1:

pkg/dnet/kpfconn.go
raphaelreyna: > The gvisor network stack performs the necessary TCP buffering, so adding more buffers here is pointless.

Makes sense, thanks.

PR #2894, opened 2022-11-18
Don't use `systemd.resolved` based DNS resolver unless cluster is proxied

### LINKED ISSUE #2811: systemd-resolveD DNS sanity check fails with local MicroK8s
**Describe the bug**
 When connecting telepresence to a local microk8s, the sanity check fails for systemd-resolved based DNS, hence falling back to the legacy DNS server

**To Reproduce**
1. When I run `telepresence connect` having a local microk8s as k8s context
2. I see _resolver did not receive requests from systemd-resolved_ within daemon logs
3. This happens after telepresence has configured the _tel0_ TUN device with a a _tel2-search_ domain and failed checking for 2 secs if query works properly
4. It logs _Unable to use systemd-resolved, falling back to local server_
5. Then the legacy server works properly, but we loose resolved-related features such as per-link config and so on

**Expected behavior**
When using a local k8s cluster we should be able to leverage the resolved config related to the tel0 device when resolving names in the form of <service>.<namespace>.

 - Output of `telepresence version`

> Client: v2.7.6 (api v3)
> Root Daemon: v2.7.6 (api v3)
> User Daemon: v2.7.6 (api v3)

 - Operating system of workstation running `telepresence` commands: Ubuntu 20.04 LTS
 - Kubernetes environment and Version: MicroK8s v1.25.2

**Additional context**
Issue filled after a [Slack message](https://datawire-oss.slack.com/archives/CAUBBJSQZ/p1665418007349829) discussed at the weekly help session.

Download [this Vagrantfile](https://gist.github.com/davidecavestro/bee90919f6ac3c43c2122088d3186393) to easily spin-up an Ubuntu vm and check the issue on a live box, After the `vagrant up` command completes, you get a fresh telepresence_logs.zip and a running vm you can play with using `vagrant ssh`

See also related issue #2809 
 #
[telepresence_logs.zip](https://github.com/telepresenceio/telepresence/files/9763678/telepresence_logs.zip)

thallgren:
## Description

There is no CIDR to place the DNS in when the cluster isn't proxied by the TUN-device and as a result, the `systemd.resolved` DNS resolver's call to `SetLinkDNS` is pointless.

This commit changes the Linux DNS behavior so that it always falls back to the overriding resolver when the cluster CIDR's are already available on the host.

Closes #2811

## Checklist

### DIFF
--- a/pkg/client/rootd/dns/server.go
+++ b/pkg/client/rootd/dns/server.go
@@ -83,6 +83,9 @@ type Server struct {
 
 	// ready is closed when the DNS server is fully configured
 	ready chan error
+
+	// Whether the TUN-device is proxying cluster CIDRs
+	proxyCluster bool
 }
 
 type cacheEntry struct {
@@ -633,10 +636,11 @@ func (s *Server) resolveQuery(q *dns.Question, dv *cacheEntry) (dnsproxy.RRs, in
 }
 
 // Run starts the DNS server(s) and waits for them to end.
-func (s *Server) Run(c context.Context, initDone chan<- struct{}, listeners []net.PacketConn, fallbackPool FallbackPool, resolve Resolver) error {
+func (s *Server) Run(c context.Context, initDone chan<- struct{}, listeners []net.PacketConn, fallbackPool FallbackPool, resolve Resolver, proxyCluster bool) error {
 	s.ctx = c
 	s.fallbackPool = fallbackPool
 	s.resolve = resolve
+	s.proxyCluster = proxyCluster
 
 	g := dgroup.NewGroup(c, dgroup.GroupConfig{})
 	for _, listener := range listeners {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,8 +12,10 @@
 
 - Bugfix: Multi valued kubernetes flags such as `--as-group` are now propagated correctly.
 
-- Bugfix: The root daemon would sometimes hang indefinetily when quit and connect were called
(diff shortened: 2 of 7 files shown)


pkg/client/rootd/dns/server.go
```diff
@@ -633,10 +636,11 @@ func (s *Server) resolveQuery(q *dns.Question, dv *cacheEntry) (dnsproxy.RRs, in
 }
 
 // Run starts the DNS server(s) and waits for them to end.
-func (s *Server) Run(c context.Context, initDone chan<- struct{}, listeners []net.PacketConn, fallbackPool FallbackPool, resolve Resolver) error {
+func (s *Server) Run(c context.Context, initDone chan<- struct{}, listeners []net.PacketConn, fallbackPool FallbackPool, resolve Resolver, proxyCluster bool) error {
 	s.ctx = c
 	s.fallbackPool = fallbackPool
 	s.resolve = resolve
+	s.proxyCluster = proxyCluster
```
raphaelreyna, line 643: According to my editor and grep it looks like this value is never used, am I missing something?

raphaelreyna (approved):
lgtm; just a question about a possibly unused variable.

pkg/client/rootd/dns/server.go
thallgren: Your editor and grep are correct.I added it there and then redesigned some stuff that rendered it unused. I left it in place anyway because I think we have more work to do in this area. Linux will simply select the `overridingResolver` when the cluster isn't proxied, but on Windows I think we might come up with a different solution.

PR #2903, opened 2022-11-25
wip: RunAsUser / RunAsGroup / RunAsNonRoot

### LINKED ISSUE #2730: runAsUser setting for traffic-agent
Hi 
I'm working at my DEV cluster with PodSecurityPolicies enforcing pods to run as nonRoot.
In order to do that I've created a custom image for traffic-agent using user 1000 (since the original image will run as a root).  
I could not use the name of the user since otherwise I've got an error: 
`Error: container has runAsNonRoot and image has non-numeric user (tp_user), cannot verify user is non-root`

Also, my image registry is bound this way: namespace <-> repo in the registry. 
For this reason, to use a custom traffic-agent image in each namespace I have to push it to each repository corresponding with namespace. 
Currently, in sidecar part of the manifest I can see only `runAsNonRoot: true`. 

I tried with values.yaml fields but to no avail: 
securityContext
hooks.podSecurityContext

Being able to set specific non-root user (through values.yml)  for traffic-agent would probably solve the problem. 
Also being able to set imagePullPolicy for the sidecar would be nice (for what i can tell we cannot set this now). 

Please, If there is already such a setting, please let me know where or which. 

I'm working with images: 
traffic-agent 1.12.9
tel2:2.7.0

denysvitali:
## Description

This PR (still WIP, review is appreciated) is going to fix #2730 and finally allow running telepresence in environments where it's not possible to run privileged containers / some restrictions are in place.

This effectively adds the following options:

- `AGENT_RUN_AS_USER`
- `AGENT_RUN_AS_GROUP`
- `AGENT_RUN_AS_NON_ROOT`
- `AGENT_CAPABILITIES_ADD`
- `AGENT_CAPABILITIES_DROP`

### Helm Chart

This change will require a slight modification of the Helm chart too, here is my proposal:

`deployment.yaml`
```diff
*** 181,186 ****
--- 181,206 ----
              value: "{{ join " " . }}"
            {{- end }}
            {{- end }}
+           {{- if .agent.securityContext.runAsUser }}
+           - name: AGENT_RUN_AS_USER
+             value: {{ .agent.securityContext.runAsUser | quote }}
+           {{- end }}
+           {{- if .agent.securityContext.runAsGroup }}
+           - name: AGENT_RUN_AS_GROUP
+             value: {{ .agent.securityContext.runAsGroup | quote }}
+           {{- end }}
+           {{- if .agent.securityContext.runAsNonRoot }}
+           - name: AGENT_RUN_AS_NON_ROOT
+             value: {{ .agent.securityContext.runAsNonRoot | quote }}
+           {{- end }}
+           {{- if gt (len .agent.securityContext.capAdd) 0 }}
+           - name: AGENT_CAPABILITIES_ADD
+             value: {{ join ", " .agent.securityContext.capAdd }}
+           {{- end }}
+           {{- if gt (len .agent.securityContext.capDrop) 0 }}
+           - name: AGENT_CAPABILITIES_DROP
+             value: {{ join ", " .agent.securityContext.capDrop }}
+           {{- end }}
          {{- /*
          Client configuration
          */}}
```

and the related `values.yaml`:

```diff
*** 209,214 ****
--- 209,220 ----
      logLevel: warning
      serverPort: 18000
      adminPort: 19000
+   securityContext:
+     runAsGroup: {}
+     runAsUser: {}
+     runAsNonRoot: {}
+     capAdd: []
+     capDrop: []
```

## Checklist

### DIFF
--- a/base-image/Dockerfile
+++ b/base-image/Dockerfile
@@ -1,5 +1,3 @@
-# syntax = docker/dockerfile:1.3
-
 # Copyright 2020-2022 Datawire. All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -19,7 +17,7 @@ FROM golang:alpine3.15 as tel2-build
 RUN apk add --no-cache gcc musl-dev fuse-dev libcap
 
 WORKDIR telepresence
-COPY go.mod go.sum .
+COPY go.mod go.sum ./
 COPY cmd/ cmd/
 COPY pkg/ pkg/
 COPY rpc/ rpc/
--- a/cmd/traffic/cmd/manager/managerutil/envconfig.go
+++ b/cmd/traffic/cmd/manager/managerutil/envconfig.go
@@ -61,6 +61,8 @@ type Env struct {
 	AgentRunAsUser           *int64                     `env:"AGENT_RUN_AS_USER,        parser=maybe-int64,    default="`
 	AgentRunAsGroup          *int64                     `env:"AGENT_RUN_AS_GROUP,       parser=maybe-int64,    default="`
 	AgentRunAsNonRoot        *bool                      `env:"AGENT_RUN_AS_NON_ROOT,    parser=maybe-bool,     default="`
+	AgentCapabilitiesAdd     []core.Capability          `env:"AGENT_CAPABILITIES_ADD,   parser=capability,     default="`
+	AgentCapabilitiesDrop    []core.Capability          `env:"AGENT_CAPABILITIES_DROP,  parser=capability,     default="`
 
 	ClientRoutingAlsoProxySubnets  []*net.IPNet  `env:"CLIENT_ROUTING_ALSO_PROXY_SUBNETS,  parser=split-ipnet, default="`
 	ClientRoutingNeverProxySubnets []*net.IPNet  `env:"CLIENT_ROUTING_NEVER_PROXY_SUBNETS, parser=split-ipnet, default="`
@@ -86,6 +88,10 @@ func (e *Env) GeneratorConfig(qualifiedAgentImage string) (*agentmap.GeneratorCo
 		EnvoyAdminPort:      e.AgentEnvoyAdminPort,
 		EnvoyLogLevel:       e.AgentEnvoyLogLevel,
 		RunAsUser:           e.AgentRunAsUser,
+		RunAsGroup:          e.AgentRunAsGroup,
+		RunAsNonRoot:        e.AgentRunAsNonRoot,
+		CapabilitiesAdd:     e.AgentCapabilitiesAdd,
+		CapabilitiesDrop:    e.AgentCapabilitiesDrop,
 	}, nil
 }
 
@@ -231,6 +237,28 @@ func fieldTypeHandlers() map[reflect.Type]envconfig.FieldTypeHandler {
 			},
 		},
 		Setter: func(dst reflect.Value, src interface{}) { dst.Set(reflect.ValueOf(src.(*core.ResourceRequirements))) },
+	}
+	fhs[reflect.TypeOf([]core.Capability{})] = envconfig.FieldTypeHandler{
+		Parsers: map[string]func(string) (any, error){
+			"capability": func(capabilityEnv string) (any, error) {
+				if capabilityEnv == "" {
+					return nil, nil
+				}
+				var caps []core.Capability
+				capsString := strings.Split(capabilityEnv, ",")
+				for _, v := range capsString {
+					caps = append(
+						caps,
+						core.Capability(
+							strings.TrimSpace(v),
+						),
+					)
+				}
+
+				return caps, nil
+			},
+		},
+		Setter: func(dst reflect.Value, src interface{}) { dst.Set(reflect.ValueOf(src.([]core.Capability))) },
 	}
 	return fhs
 }
--- a/cmd/traffic/cmd/manager/managerutil/envconfig.go
+++ b/cmd/traffic/cmd/manager/managerutil/envconfig.go
@@ -3,21 +3,19 @@ package managerutil
 import (
 	"context"
 	"encoding/json"
-	"net"
-	"reflect"
-	"strconv"
-	"strings"
-	"time"
-
-	core "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/resource"
-
 	"github.com/datawire/dlib/derror"
 	"github.com/datawire/envconfig"
 	"github.com/datawire/k8sapi/pkg/k8sapi"
 	"github.com/telepresenceio/telepresence/v2/pkg/agentconfig"
 	"github.com/telepresenceio/telepresence/v2/pkg/agentmap"
 	"github.com/telepresenceio/telepresence/v2/pkg/iputil"
+	core "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	"net"
+	"reflect"
+	"strconv"
+	"strings"
+	"time"
 )
 
 // Env is the traffic-manager's environment. It does not define any defaults because all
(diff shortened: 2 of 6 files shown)


base-image/Dockerfile
```diff
@@ -19,7 +17,7 @@ FROM golang:alpine3.15 as tel2-build
 RUN apk add --no-cache gcc musl-dev fuse-dev libcap
 
 WORKDIR telepresence
-COPY go.mod go.sum .
+COPY go.mod go.sum ./
```
thallgren, line 20: I don't think this change is needed. Nor is it relevant to this PR.

cmd/traffic/cmd/manager/managerutil/envconfig.go
```diff
@@ -3,21 +3,19 @@ package managerutil
...
-	core "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/resource"
-
 	"github.com/datawire/dlib/derror"
 	"github.com/datawire/envconfig"
 	"github.com/datawire/k8sapi/pkg/k8sapi"
 	"github.com/telepresenceio/telepresence/v2/pkg/agentconfig"
 	"github.com/telepresenceio/telepresence/v2/pkg/agentmap"
 	"github.com/telepresenceio/telepresence/v2/pkg/iputil"
+	core "k8s.io/api/core/v1"
```
thallgren, line 12: We're picky about import order. So please stick to the rules we have using three groups.

cmd/traffic/cmd/manager/managerutil/envconfig.go
```diff
@@ -60,6 +58,9 @@ type Env struct {
 	AgentEnvoyLogLevel       string                     `env:"AGENT_ENVOY_LOG_LEVEL,    parser=logLevel,       defaultFrom=AgentLogLevel"`
 	AgentEnvoyServerPort     uint16                     `env:"AGENT_ENVOY_SERVER_PORT,  parser=port-number"`
 	AgentEnvoyAdminPort      uint16                     `env:"AGENT_ENVOY_ADMIN_PORT,   parser=port-number"`
+	AgentRunAsUser           *int64                     `env:"AGENT_RUN_AS_USER,        parser=maybe-int64,    default="`
+	AgentRunAsGroup          *int64                     `env:"AGENT_RUN_AS_GROUP,       parser=maybe-int64,    default="`
+	AgentRunAsNonRoot        *bool                      `env:"AGENT_RUN_AS_NON_ROOT,    parser=maybe-bool,     default="`
```
thallgren, line 63: Do we really need pointer values here? I would prefer if just left the values out of the generated manifest if zero or false. Or, if you think it's essential to be able to specify `runAsUser: 0` or `runAsGroup: 0`, then use a `-1` as the default instead of an empty string.

cmd/traffic/cmd/manager/managerutil/envconfig.go
```diff
@@ -84,6 +85,7 @@ func (e *Env) GeneratorConfig(qualifiedAgentImage string) (*agentmap.GeneratorCo
 		EnvoyServerPort:     e.AgentEnvoyServerPort,
 		EnvoyAdminPort:      e.AgentEnvoyAdminPort,
 		EnvoyLogLevel:       e.AgentEnvoyLogLevel,
+		RunAsUser:           e.AgentRunAsUser,
```
thallgren, line 90: Where's the rest?

cmd/traffic/cmd/manager/managerutil/envconfig.go
```diff
@@ -111,6 +113,44 @@ func fieldTypeHandlers() map[reflect.Type]envconfig.FieldTypeHandler {
...
+				return &b, err
+			},
+		},
+		Setter: func(dst reflect.Value, src interface{}) {
+			if src != nil {
+				dst.Set(reflect.ValueOf(src))
+			}
+		},
+	}
+
```
thallgren, line 159: Wouldn't be necessary if we use zero values (or -1)

cmd/traffic/cmd/manager/managerutil/envconfig_test.go
```diff
@@ -77,6 +77,34 @@ func TestEnvconfig(t *testing.T) {
...
+				"AGENT_RUN_AS_NON_ROOT": "false",
+			},
+			Output: func(e *managerutil.Env) {
+				var user1000 = int64(1000)
+				var group1000 = int64(1000)
+				var falseVal = false
+
+				e.AgentRunAsUser = &user1000
+				e.AgentRunAsGroup = &group1000
+				e.AgentRunAsNonRoot = &falseVal
```
thallgren, line 106: Pointers really complicate things :-)

cmd/traffic/cmd/manager/managerutil/envconfig_test.go
```diff
@@ -97,7 +125,7 @@ func TestEnvconfig(t *testing.T) {
 			ctx, err := managerutil.LoadEnv(context.Background(), lookup)
 			require.NoError(t, err)
 			actual := managerutil.GetEnv(ctx)
-			assert.Equal(t, &expected, actual)
+			assert.EqualValues(t, &expected, actual) // Pointers are different here
```
thallgren, line 144: Another point complication.

cmd/traffic/cmd/manager/managerutil/envconfig.go
```diff
@@ -232,6 +238,28 @@ func fieldTypeHandlers() map[reflect.Type]envconfig.FieldTypeHandler {
...
 		Setter: func(dst reflect.Value, src interface{}) { dst.Set(reflect.ValueOf(src.(*core.ResourceRequirements))) },
 	}
+	fhs[reflect.TypeOf([]core.Capability{})] = envconfig.FieldTypeHandler{
+		Parsers: map[string]func(string) (any, error){
+			"capability": func(capabilityEnv string) (any, error) {
+				if capabilityEnv == "" {
+					return nil, nil
+				}
+				var caps []core.Capability
+				capsString := strings.Split(capabilityEnv, ",")
```
thallgren, line 248: We're using space delimited strings for other multi-valued environment variables, or if need be, a JSON string. This introduces comma separated values. I don't think we want that.

thallgren (changes requested):
Thanks for providing this. I have reviewed it, and I don't think this is the way to go. Instead, we should add a `securityContext` to the `agent` structure and then provide quoted JSON of whatever struct the user decides to put in there, just like we do today with `agent.resources` (env `AGENT_RESOURCES` using parser "json-resources").

This would also be more consistent with the `resources` element that we already provide at top-level for the traffic-manager.

cmd/traffic/cmd/manager/managerutil/envconfig_test.go
denysvitali: I know :(

cmd/traffic/cmd/manager/managerutil/envconfig.go
denysvitali: In the latest commit it should be there. I forgot to add them :D

cmd/traffic/cmd/manager/managerutil/envconfig.go
denysvitali: Sorry, missed that convention

cmd/traffic/cmd/manager/managerutil/envconfig.go
denysvitali: Although UID and GID seems to be defined as `uint` in Linux, Kubernetes uses pointers too.
I think this is because an user might decide to not set an ID, and `-1` might have different effects:
https://man7.org/linux/man-pages/man2/setreuid.2.html

Obviously leaving it as `0` when the user decides not to specify that is also an "error", as this will run the container as root

thallgren:
Closing this, as no further input was provided after review.

PR #2904, opened 2022-11-29
Defer initialization of fuseftp until it is really needed.

thallgren:
## Description

Setting `intercept.fuseFtp=true` in the Helm chart didn't work, because the fuseftp server was initialized when the session booted up. This commit ensures that the initialization is deferred until the first intercept is made.

## Checklist

### DIFF
--- a/pkg/client/userd/daemon/fuseftp.go
+++ b/pkg/client/userd/daemon/fuseftp.go
@@ -16,12 +16,51 @@ import (
 	"github.com/datawire/dlib/dlog"
 	"github.com/datawire/dlib/dtime"
 	"github.com/datawire/go-fuseftp/rpc"
+	"github.com/telepresenceio/telepresence/v2/pkg/client"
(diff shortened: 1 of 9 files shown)


pkg/client/userd/daemon/fuseftp.go
```diff
@@ -16,13 +16,52 @@ import (
...
+	}
+
+	select {
+	case <-ctx.Done():
+		return nil
+	case c, ok := <-s.fuseFtpCh:
+		if ok {
+			// Put the client back onto the queue for the next caller to read
+			s.fuseFtpCh <- c
+		}
```
njayp, line 60: interesting use of chan as storage

njayp (approved):

pkg/client/userd/daemon/fuseftp.go
thallgren: I'm quite happy about it actually. It's a storage that requires the first reader to wait, and then synchronizes all subsequent reads without using a mutex.

PR #2909, opened 2022-11-29
Close all spdy streams when session ends.

thallgren:
## Description

Prior to this change, all spdy streams (used for port forwards to the traffic-manager) were left open after a `telepresence quit`. This meant that both the client and the traffic-manager retained connections that were no longer used, resulting in ever-increasing resource consumption until the client was killed with an explicit `telepresence quit -s`.

## Checklist


raphaelreyna (approved):
nice

PR #2910, opened 2022-11-29
Fix session instabilities

thallgren:
## Description

This PR fixes three issues related to session management and stability. See commit comments for details.

## Checklist

### DIFF
--- a/pkg/client/userd/daemon/service.go
+++ b/pkg/client/userd/daemon/service.go
@@ -131,7 +131,10 @@ func (s *Service) configReload(c context.Context) error {
 	return client.Watch(c, func(ctx context.Context) error {
 		s.sessionLock.RLock()
 		defer s.sessionLock.RUnlock()
-		return s.session.ApplyConfig(ctx)
+		if s.session == nil {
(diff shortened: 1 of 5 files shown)


pkg/client/userd/daemon/service.go
```diff
@@ -131,106 +131,102 @@ func (s *Service) configReload(c context.Context) error {
...
-					}
-				} else {
-					sCancel()
-					s.sessionCancel = nil
-				}
+			return nil
+		case cr := <-s.connectRequest:
+			rsp := startSession(c, si, cr, &wg)
+			select {
+			case <-c.Done():
```
raphaelreyna, line 159: do we need this line since there is already a check on the context on line 152?

raphaelreyna (approved):
lgtm; just on question regarding a possible duplicate loc.

pkg/client/userd/daemon/service.go
thallgren: Yes, it's needed. Starting the session might take a while and the context might have been cancelled during that time (calling `quit -s` would do that, for instance).

PR #2912, opened 2022-11-30
Raphaelreyna/modes

raphaelreyna:
## Description
Adds support for setting single/team mode in the traffic manager via its configmap, which has been renamed to `traffic-manager`.
The `traffic-manager` configmap also has a new data field, `traffic-manager.yaml`, where the mode is set as `mode:  single` or `mode: team`.

## Checklist


cmd/traffic/cmd/manager/service.go
```diff
@@ -173,6 +175,17 @@ func (m *Manager) GetTelepresenceAPI(ctx context.Context, e *empty.Empty) (*rpc.
...
+	var (
+		mode               config.Mode
+		currentClientCount = m.state.CountAllClients()
+	)
+	m.config.RLock()
+	mode = m.config.Mode
+	m.config.RUnlock()
+	if 0 < currentClientCount && mode == config.ModeSingle {
+		return nil, status.Errorf(codes.FailedPrecondition, "additional client connections require the traffic-manager to be set to team mode")
+	}
```
njayp: this seems like it would block user2 from connecting, which is not what we want. We should instead just nag the user

pkg/client/errcat/errors.go
```diff
@@ -17,11 +17,13 @@ type categorized struct {
...
-	Config       // Errors in config.yml, extensions, or kubeconfig
-	NoDaemonLogs // Other error generated in the CLI process, so no use pointing the user to logs
-	Unknown      // Something else. Consult the logs
+	OK             = Category(iota)
+	User           // User made an error
+	Config         // Errors in config.yml, extensions, or kubeconfig
+	NoDaemonLogs   // Other error generated in the CLI process, so no use pointing the user to logs
+	APIKeyRequired // An API key is required
+	ModeWarning    // Just a warning message regarding the need to switch modes, not an actual error
+	Unknown        // Something else. Consult the logs
```
thallgren: The enums here must comply with the enums in the `connector.Result` gRPC message, and existing enums must not change, because that breaks backward compatibility between the traffic-manager and the client.

pkg/client/userd/trafficmgr/session.go
```diff
@@ -221,6 +227,12 @@ func NewSession(
 		SessionInfo:    tmgr.SessionInfo(),
 		Intercepts:     &manager.InterceptInfoSnapshot{Intercepts: tmgr.getCurrentInterceptInfos()},
 	}
+
+	if warning != nil {
+		ret.ErrorCategory = int32(errcat.GetCategory(err))
```
thallgren: The `err` variable is `nil` at this point, so retrieving its error category doesn't make sense.

cmd/traffic/cmd/manager/internal/state/state.go
```diff
@@ -630,3 +632,24 @@ func (s *State) InitialTempLogLevel() *rpc.LogLevelRequest {
...
+// modeCheck checks the current mode and returns either a warning or error based on number of current users
+func (s *State) ModeCheck() (string, error) {
+	currentClientCount := s.CountAllClients()
+	s.mu.RLock()
+	mode := s.config.Mode
+	s.mu.RUnlock()
+
+	if 0 < currentClientCount && mode == config.ModeSingle {
+		// TODO(raphaelreyna): come up with a better message here
+		return "additional client connections require the traffic-manager to be set to team mode", nil
```
thallgren: This should be allowed but result in some nagging, but I don't think the text should be produced here.

cmd/traffic/cmd/manager/internal/state/state.go
```diff
@@ -630,3 +632,24 @@ func (s *State) InitialTempLogLevel() *rpc.LogLevelRequest {
...
 	return s.llSubs.subscriberLoop(stream.Context(), stream)
 }
+
+func (s *State) SetConfig(c config.TrafficManager) {
+	s.mu.Lock()
+	s.config = c
+	s.mu.Unlock()
+}
+
+// modeCheck checks the current mode and returns either a warning or error based on number of current users
```
thallgren: Doesn't seem like this function ever returns an error.

cmd/traffic/cmd/manager/internal/state/state.go
```diff
@@ -630,3 +632,24 @@ func (s *State) InitialTempLogLevel() *rpc.LogLevelRequest {
...
+	s.mu.RLock()
+	mode := s.config.Mode
+	s.mu.RUnlock()
+
+	if 0 < currentClientCount && mode == config.ModeSingle {
+		// TODO(raphaelreyna): come up with a better message here
+		return "additional client connections require the traffic-manager to be set to team mode", nil
+	}
+
+	return "", nil
```
thallgren: Where's the block preventing an unauthenticated user from connecting to a team-mode cluster?

cmd/traffic/cmd/manager/service.go
```diff
@@ -185,7 +208,7 @@ func (m *Manager) ArriveAsClient(ctx context.Context, client *rpc.ClientInfo) (*
 		SessionId: sessionID,
 		ClusterId: m.clusterInfo.GetClusterID(),
 		InstallId: &installId,
-	}, nil
+	}, warning
```
thallgren: The second argument here is a error, not a warning. Please don't use the error type to pass warnings around. That's very confusing.

cmd/traffic/cmd/manager/service.go
```diff
@@ -177,6 +181,25 @@ func (m *Manager) ArriveAsClient(ctx context.Context, client *rpc.ClientInfo) (*
...
+		return nil, status.Error(codes.FailedPrecondition, err.Error())
+	}
+
+	var warning error
+	if warnMsg != "" {
+		// TODO(raphaelreyna): should probably use a custom proto here
+		msg := errdetails.LocalizedMessage{
+			Message: warnMsg,
+		}
+		stat, err := status.New(codes.FailedPrecondition, warnMsg).WithDetails(&msg)
```
thallgren: A warning can never be a `FailedPrecondition`. A failure is an error.

pkg/client/errcat/errors.go
```diff
@@ -17,11 +17,13 @@ type categorized struct {
...
-	OK           = Category(iota)
-	User         // User made an error
-	Config       // Errors in config.yml, extensions, or kubeconfig
-	NoDaemonLogs // Other error generated in the CLI process, so no use pointing the user to logs
-	Unknown      // Something else. Consult the logs
+	OK             = Category(iota)
+	User           // User made an error
+	Config         // Errors in config.yml, extensions, or kubeconfig
+	NoDaemonLogs   // Other error generated in the CLI process, so no use pointing the user to logs
+	APIKeyRequired // An API key is required
```
thallgren: This is an error in category `User`.

pkg/client/errcat/errors.go
```diff
@@ -17,11 +17,13 @@ type categorized struct {
...
+	ModeWarning    // Just a warning message regarding the need to switch modes, not an actual error
```
thallgren: This is definitely not an error category.

pkg/client/userd/trafficmgr/session.go
```diff
@@ -170,10 +170,16 @@ func NewSession(
 	connectStart := time.Now()
 
 	dlog.Info(ctx, "Connecting to traffic manager...")
+	var warning error
```
thallgren: This is an oxymoron. A warning is not an error.

thallgren (changes requested):
I react to the use of errors as warnings. That just doesn't feel right.
Even though we now do changes to the traffic-manager, I don't think the OSS-client needs any changes. In essence, I'd shy away from using `ArriveAsClient` and instead base the nagging on an extra call, that only the pro-client will make, on connect.

The pure OSS-client will not be aware of any of this until someone tries to use it to connect to a team-mode cluster. An authenticated connection attempt to a team-mode cluster must always be rejected. That hard rejection is fine to add to `ArriveAsClient` and it will not require any modifications to the OSS-client.

cmd/traffic/cmd/manager/internal/config/trafficManager.go
```diff
@@ -0,0 +1,44 @@
...
+	"gopkg.in/yaml.v3"
+)
+
+type Mode int32
+
+const (
+	ModeUnspecified Mode = iota
+	ModeSingle
+	ModeTeam
+)
```
njayp: Since we have an enum in the rpc, can we extend that enum here? Like `type Mode manager.Mode`? That way we dont have to sync two different enums

cmd/traffic/cmd/manager/internal/state/intercept.go
```diff
@@ -60,11 +61,26 @@ func (s *State) PrepareIntercept(ctx context.Context, cr *managerrpc.CreateInter
...
+			// block
 			return interceptError(errcat.User.New("Global intercepts are not allowed. Please log in and use http intercepts"))
 		}
+
+		// or if not in team mode another intercept is being created
+		if !teamMode && 1 < s.CountAllClients() {
+			// suggest switching to team mode
+			// TODO(raphaelreyna): swap this out for a better err message
+			dlog.Info(ctx, "Ambassador Labs recommends switching to Team mode when using multiple accounts")
+		}
```
njayp: delete this. This is covered by a PR in the pro repo

cmd/traffic/cmd/manager/service.go
```diff
@@ -459,9 +466,16 @@ func (m *Manager) WatchIntercepts(session *rpc.SessionInfo, stream rpc.Manager_W
...
 	ctx = managerutil.WithSessionInfo(ctx, request.Session)
 	dlog.Debugf(ctx, "PrepareIntercept called")
+
+	pi, err := m.state.PrepareIntercept(ctx, request)
+	if err != nil {
+		return nil, err
+	}
+
 	span := trace.SpanFromContext(ctx)
 	tracing.RecordInterceptSpec(span, request.InterceptSpec)
```
njayp: I think the trace recording has to start before prepareintercept is called? Why was this change made?

rpc/connector/connector.proto
```diff
@@ -182,6 +182,9 @@ message ConnectInfo {
 
   daemon.DaemonStatus daemon_status = 13;
 
+  // traffic manager details
+  telepresence.manager.StatusInfo status = 14;
```
njayp: can you rename status to manager_status? my bad :sad_panda:

njayp (changes requested):

cmd/traffic/cmd/manager/service.go
raphaelreyna: yeah; its left over from when I had this block the intercept based on mode.
My thinking was that it didn't make much sense to start tracing before we knew if the intercept would be blocked.

charts/telepresence/values.yaml
```diff
@@ -341,3 +341,6 @@ client:
 
     # Tell client's DNS resolver to always send names with these suffixes to the cluster side resolver
     includeSuffixes: []
+
+trafficManager:
+  mode: single
```
thallgren, line 346: What happens if the user does the following:
```
telepresence helm install --team-mode
```
and then, at a later time, does:
```
telepresence helm install --upgrade
```
From the looks of it, that'd make this default kick in, and thus revert the earlier setting. I would expect that unless the mode is explicitly given, the current mode would survive an upgrade.

cmd/traffic/cmd/manager/internal/config/trafficManager.go
```diff
@@ -0,0 +1,44 @@
...
+	"strings"
+
+	"gopkg.in/yaml.v3"
+
+	"github.com/telepresenceio/telepresence/rpc/v2/manager"
+)
+
+type Mode manager.Mode
+
+func (m *Mode) UnmarshalYAML(value *yaml.Node) error {
```
thallgren, line 15: What happens if you marshal a mode that is "unspecified"? Will it be left out?

cmd/traffic/cmd/manager/internal/state/intercept.go
```diff
@@ -60,9 +60,13 @@ func (s *State) PrepareIntercept(ctx context.Context, cr *managerrpc.CreateInter
...
 
+	if s.config.Mode.IsTeam() && cr.ApiKey == "" {
+		return interceptError(errcat.User.New("A login is required to intercept while in Team mode."))
+	}
+
 	env := managerutil.GetEnv(ctx)
-	if env.InterceptDisableGlobal {
-		if cr.InterceptSpec.Mechanism != "http" {
+	if cr.InterceptSpec.Mechanism != "http" {
+		if env.InterceptDisableGlobal {
```
thallgren: I don't understand why this change is needed. If anything, it's more efficient to test the env first (it's just one indirection).

njayp (approved):

cmd/traffic/cmd/manager/service.go
```diff
@@ -459,9 +466,16 @@ func (m *Manager) WatchIntercepts(session *rpc.SessionInfo, stream rpc.Manager_W
...
-	return m.state.PrepareIntercept(ctx, request)
+
+	pi, err := m.state.PrepareIntercept(ctx, request)
+	if err != nil {
+		return nil, err
+	}
+
+	return pi, nil
```
thallgren: I don't understand this change. It doesn't add anything.

pkg/client/cli/cmd_status.go
```diff
@@ -250,3 +257,14 @@ func (cs *userDaemonStatus) WriteTo(out io.Writer) (int64, error) {
...
+func modeToString(mode *manager.Mode) string {
+	switch mode {
+	case manager.Mode_MODE_SINGLE.Enum():
+		return "single-user"
+	case manager.Mode_MODE_TEAM.Enum():
+		return "team"
+	default:
+		return "unknown"
+	}
+}
```
thallgren: This is the second function that creates a string from a `manager.Mode` (the first one is in cmd/traffic/cmd/manager/internal/config/trafficManager.go), and it uses different strings.

rpc/manager/manager.proto
```diff
@@ -544,10 +544,25 @@ message CLIConfig {
...
+  VersionInfo2 version = 1;
+  Mode mode = 2;
+  int32 client_count = 3;
+}
+
 service Manager {
   // Version returns the version information of the Manager.
   rpc Version(google.protobuf.Empty) returns (VersionInfo2);
 
+  // Status returns traffic manager state information.
```
thallgren: This is misleading as it returns a very small fraction of the manager's state.

thallgren (changes requested):

charts/telepresence/values.yaml
njayp: I think the configmap would preserve the mode

rpc/manager/manager.proto
njayp: That begs the question, should it return more? Version, number of intercepts, etc?

charts/telepresence/values.yaml
raphaelreyna: From what I could find: helm defaults to using the `reuse-values` flag if the `--set` flag is _not_ used so in the case of a plain `telepresence helm install --upgrade` we should be fine. However i think this means that other values will be reset when using the `--X-mode` (or when any other value is changed, not just the mode) unless we add the appropriate flag.

https://medium.com/@kcatstack/understand-helm-upgrade-flags-reset-values-reuse-values-6e58ac8f127e

cmd/traffic/cmd/manager/service.go
raphaelreyna: ah yeah this is left over from when i was doing mode checks at the very end and needed to check this error. will remove

cmd/traffic/cmd/manager/internal/state/intercept.go
raphaelreyna: ah yeah i must have moved this (don't remember why) by deleting it and rewriting it rather than copy/paste and just kept the logic the same. changing this back

pkg/client/cli/util/connector.go
```diff
@@ -58,7 +56,7 @@ func UserDaemonDisconnect(ctx context.Context, quitDaemons bool) (err error) {
 		fmt.Fprint(stdout, "quitting...")
 	} else {
 		fmt.Fprint(stdout, "disconnecting...")
-		if _, err = ud.Disconnect(ctx, &empty.Empty{}); status.Code(err) != codes.Unimplemented {
+		if _, err = ud.Disconnect(ctx, &empty.Empty{}); grpcStatus.Code(err) != grpcCodes.Unimplemented {
```
thallgren, line 59: This is harmless, but what motivates this change?

pkg/client/userd/daemon/grpc.go
```diff
@@ -155,6 +155,14 @@ func (s *Service) Status(c context.Context, ex *empty.Empty) (result *rpc.Connec
 			})
 		} else {
 			result = s.session.Status(s.sessionContext)
+			err = s.WithSession(c, "status", func(ctx context.Context, session userd.Session) error {
+				status, err := session.ManagerClient().Status(c, &empty.Empty{})
+				if err != nil {
```
thallgren: If a new client connects to an older traffic-manager, then the error here will be `codes.Unimplemented`. This must be OK. It must not propagate as an error.

pkg/client/cli/cmd_status.go
```diff
@@ -236,6 +241,8 @@ func (cs *userDaemonStatus) WriteTo(out io.Writer) (int64, error) {
 		n += ioutil.Printf(out, "  Executable        : %s\n", cs.Executable)
 		n += ioutil.Printf(out, "  Install ID        : %s\n", cs.InstallID)
 		n += ioutil.Printf(out, "  Status            : %s\n", cs.Status)
+		n += ioutil.Printf(out, "    Mode            : %s\n", cs.Mode)
+		n += ioutil.Printf(out, "    Client Count    : %d\n", cs.ClientCount)
```
thallgren: These are not user daemon statuses. They are traffic manager statuses. If we want to print status info about the traffic-manager, then we should put it under a new "Traffic Manager:" heading (similar to the version command).

pkg/client/userd/daemon/grpc.go
```diff
@@ -155,6 +155,14 @@ func (s *Service) Status(c context.Context, ex *empty.Empty) (result *rpc.Connec
 			})
 		} else {
 			result = s.session.Status(s.sessionContext)
+			err = s.WithSession(c, "status", func(ctx context.Context, session userd.Session) error {
```
thallgren: It is redundant to call `WithSession` here, given that the sessionLock is in place already.

thallgren (changes requested):

thallgren:
Slept on it, woke up with a "what about backward compatibility?" thought. So I reviewed again. Sorry for not catching this earlier.

pkg/client/cli/util/connector.go
njayp: I did this one, neatness, a lib was imported twice

charts/telepresence/values.yaml
thallgren: Perhaps we can do it like this then.
- If `--X-mode` is combined with a `--upgrade`, and no `--values` or `--set` is provided, then we retain the default (i.e., we add the `--reuse-values` flag).
- If `--X-mode` is not set, but the user provides `--upgrade` together with a `--values` or `--set` flag, then we must somehow ensure that the current `--X-mode` is preserved.

cmd/traffic/cmd/manager/internal/config/trafficManager.go
```diff
@@ -0,0 +1,49 @@
+package config
+
+import (
+	"fmt"
+	"strings"
+
+	"gopkg.in/yaml.v3"
+
+	"github.com/telepresenceio/telepresence/rpc/v2/manager"
+	"github.com/telepresenceio/telepresence/v2/pkg/client/cli"
```
thallgren: The traffic manager should never use anything beneath the `pkg/client` package. Common functions must be placed in a relevant package directly under `pkg`.

charts/telepresence/values.yaml
thallgren: We should probably add the `--reuse-values` flag too, at some point. Not necessary in this PR though, unless it makes the implementation easier.

thallgren:
The PR is in need of a better title.

pkg/client/cli/cmd_status.go
```diff
@@ -240,9 +248,12 @@ func (cs *userDaemonStatus) WriteTo(out io.Writer) (int64, error) {
...
 		n += ioutil.Printf(out, "  Install ID        : %s\n", cs.InstallID)
-		n += ioutil.Printf(out, "  Status            : %s\n", cs.Status)
-		n += ioutil.Printf(out, "    Mode            : %s\n", cs.Mode)
-		n += ioutil.Printf(out, "    Client Count    : %d\n", cs.ClientCount)
+		if tms := cs.TrafficManagerStatus; tms != nil {
+			n += ioutil.Println(out, "Traffic Manager: Running")
+			n += ioutil.Printf(out, "  Status            : %s\n", cs.Status)
+			n += ioutil.Printf(out, "    Mode            : %s\n", tms.Mode)
+			n += ioutil.Printf(out, "    Client Count    : %d\n", tms.Mode)
+		}
```
thallgren: It seems a bit overkill to both report that the traffic-manager is running and that it is connected. Connected implies running. How about this?
```suggestion
		if tms := cs.TrafficManagerStatus; tms != nil {
			n += ioutil.Println(out, "Traffic Manager: Connected")
			n += ioutil.Printf(out, "  Mode            : %s\n", tms.Mode)
			n += ioutil.Printf(out, "  Client Count    : %d\n", tms.Mode)
		} else {
			n += ioutil.Println(out, "Traffic Manager: Not connected")
		}
```

pkg/client/userd/daemon/grpc.go
```diff
@@ -155,14 +155,20 @@ func (s *Service) Status(ctx context.Context, ex *empty.Empty) (result *rpc.Conn
 			})
 		} else {
 			result = s.session.Status(s.sessionContext)
-			err = s.WithSession(c, "status", func(ctx context.Context, session userd.Session) error {
-				status, err := session.ManagerClient().Status(c, &empty.Empty{})
-				if err != nil {
-					return err
+			stts, err := s.session.ManagerClient().Status(c, &empty.Empty{})
```
thallgren: Using `:=` here creates a new `err` variable that shadows the one in the outer scope. I don't think that's what you want.

pkg/client/userd/daemon/grpc.go
```diff
@@ -155,14 +155,20 @@ func (s *Service) Status(ctx context.Context, ex *empty.Empty) (result *rpc.Conn
...
+					return
 				}
-				result.ManagerStatus = status
-				return nil
-			})
+				if stat.Code() == codes.Unimplemented {
+					err = nil
+				} else {
+					return
+				}
```
thallgren: Lines 160 to 163 can be removed, and this can be replaced with:
```suggestion
				if status.Code(err) == codes.Unimplemented {
					err = nil
				}
				return
```

pkg/common/common.go
```diff
@@ -0,0 +1,14 @@
+package common
```
thallgren: From the Go Blog on [Package Names](https://go.dev/blog/package-names): "Avoid meaningless package names. Packages named util, common, or misc provide clients with no sense of what the package contains".

I know we have a `util` package under `cli`. That should change too at some point in time, but we should avoid creating new ones.

thallgren (changes requested):

pkg/client/cli/cmd_status.go
```diff
@@ -148,6 +150,12 @@ func BasicGetStatusInfo(ctx context.Context) (ioutil.WriterTos, error) {
 				Client: icept.Spec.Client,
 			})
 		}
+		if ms := status.ManagerStatus; ms != nil {
```
thallgren: Will `ManagerStatus` be set even when connecting to an older traffic-manager?

pkg/client/cli/cmd_status.go
raphaelreyna: yeah that looks better :+1:

pkg/client/userd/daemon/grpc.go
raphaelreyna: those named returns :face_exhaling:
good catch, thanks

pkg/client/cli/cmd_status.go
raphaelreyna: Yeah this was coming in as nil if the tm was old or not connected; no way to tell the difference between both cases.
Ended up propagating the tm version to the cli where it checks the version before printing the new status fields.

charts/telepresence/values.yaml
raphaelreyna: > If --X-mode is not set, but the user provides --upgrade together with a --values or --set flag, then we must somehow ensure that the current --X-mode is preserved.

Since the mode is in the `values.yaml` I don't think we should make it special by having it be immune to being reset

pkg/client/cli/cmd_status.go
```diff
@@ -250,9 +248,13 @@ func (cs *userDaemonStatus) WriteTo(out io.Writer) (int64, error) {
 		n += ioutil.Printf(out, "  Install ID        : %s\n", cs.InstallID)
 		if tms := cs.TrafficManagerStatus; tms != nil {
 			n += ioutil.Println(out, "Traffic Manager: Running")
```
thallgren: "Running/Not connected" doesn't add up. It should be "Connected/Not connected" or "Running/Not running". Since we cannot determine if it isn't running, unless we're connected, we must use the former.

pkg/client/userd/daemon/grpc.go
```diff
@@ -155,20 +155,23 @@ func (s *Service) Status(ctx context.Context, ex *empty.Empty) (result *rpc.Conn
...
 			result = s.session.Status(s.sessionContext)
-			stts, err := s.session.ManagerClient().Status(c, &empty.Empty{})
+			var stts *manager.StatusInfo
+			stts, err = s.session.ManagerClient().Status(c, &empty.Empty{})
 			if err != nil {
 				stat, ok := status.FromError(err)
 				if !ok {
 					return
 				}
 				if stat.Code() == codes.Unimplemented {
```
thallgren: As I pointed out earlier:
```suggestion
				if status.Code(err) == codes.Unimplemented {
```

pkg/strings/strings.go
```diff
@@ -1,12 +1,12 @@
-package common
+package strings
 
 import "github.com/telepresenceio/telepresence/rpc/v2/manager"
 
-func ModeToString(mode *manager.Mode) string {
+func FromMode[M ~int32](mode M) string {
```
thallgren: Using generics here seem like overkill to me. What's wrong with using `manager.Mode` (not a pointer)?

pkg/client/cli/cmd_status.go
```diff
@@ -5,17 +5,19 @@ import (
 	"io"
 
 	"github.com/spf13/cobra"
+	"golang.org/x/mod/semver"
```
thallgren: Don't think you need version parsing (see comment further down), and if you do, you should use `github.com/blang/semver` because that's what the rest of the code base is using.

pkg/client/cli/cmd_status.go
```diff
@@ -45,9 +47,10 @@ type userDaemonStatus struct {
...
 }
 
-type trafficManagerStatus struct {
-	Mode        string `json:"mode,omitempty" yaml:"mode,omitempty"`
-	ClientCount int32  `json:"client_count,omitempty" yaml:"client_count,omitempty"`
+type trafficManagerStatus manager.StatusInfo
+
+func (tms *trafficManagerStatus) validVersion() bool {
+	return 0 < semver.Compare(tms.Version.GetVersion(), "v2.9.5")
 }
```
thallgren: This isn't needed (see comment further down).

pkg/client/cli/cmd_status.go
```diff
@@ -250,9 +248,13 @@ func (cs *userDaemonStatus) WriteTo(out io.Writer) (int64, error) {
 		n += ioutil.Printf(out, "  Install ID        : %s\n", cs.InstallID)
 		if tms := cs.TrafficManagerStatus; tms != nil {
 			n += ioutil.Println(out, "Traffic Manager: Running")
-			n += ioutil.Printf(out, "  Status            : %s\n", cs.Status)
-			n += ioutil.Printf(out, "    Mode            : %s\n", tms.Mode)
-			n += ioutil.Printf(out, "    Client Count    : %d\n", tms.Mode)
+			n += ioutil.Printf(out, "  Version: %s\n", tms.Version.Version)
+			if tms.validVersion() {
```
thallgren:
```suggestion
			if tms.Mode != manager.Mode_MODE_UNDEFINED {
```

thallgren (changes requested):

pkg/client/userd/daemon/grpc.go
raphaelreyna: missed that

thallgren (approved):

PR #2917, opened 2022-12-05
Njayp/add mode to status

njayp:
## Description

Adds a status rpc to the traffic manager. Calls this status rpc as part of `telepresence status`


raphaelreyna (approved):
nice :+1:

PR #2918, opened 2022-12-05
Ensure that google credentials are picked up correctly by root daemon.

thallgren:
## Description

A user using `gcloud auth` to authenticate to a GKE cluster would not be able to login unless the kubeconfig also contained a gce config element. This was becaue the root daemon was looking in the wrong place for the generated credentials file.

This commit ensures that setting the `GOOGLE_APPLICATION_CREDENTIALS` environment variable has the desired effect, and also that the default credentials are found when it is not set.

## Checklist

### DIFF
--- a/pkg/client/k8s_config.go
+++ b/pkg/client/k8s_config.go
@@ -77,20 +77,28 @@ func NewKubeconfig(c context.Context, flagMap map[string]string) (*Kubeconfig, e
 	// respect to this option.
 	delete(flagMap, "namespace")
 
-	// The KUBECONFIG entry is a copy of the KUBECONFIG environment variable sent to us from the CLI to give
-	// this long-running daemon a chance to update it. Using the --kubeconfig flag to send the info isn't
-	// sufficient because that flag doesn't allow for multiple path entries like the KUBECONFIG does.
-	if kcEnv, ok := flagMap["KUBECONFIG"]; ok {
-		delete(flagMap, "KUBECONFIG")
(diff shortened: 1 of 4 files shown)


pkg/client/k8s_config.go
```diff
@@ -77,20 +77,28 @@ func NewKubeconfig(c context.Context, flagMap map[string]string) (*Kubeconfig, e
...
+	// The GOOGLE_APPLICATION_CREDENTIALS and KUBECONFIG entries are copies of the environment variables
+	// sent to us from the CLI to give this long-running daemon a chance to update them. Here we set/unset
+	// our them in our environment accordingly and remove them from the flagMap
+	transferEnvFlag := func(key string) error {
+		if env, ok := flagMap[key]; ok {
+			delete(flagMap, key)
+			if err := os.Setenv(key, env); err != nil {
+				return err
+			}
+			dlog.Debugf(c, "Using %s %s", key, env)
```
raphaelreyna, line 89: this might be nice as an info message

raphaelreyna (approved):
lgtm; just a comment on a loglevel

pkg/client/k8s_config.go
thallgren: I'm going to keep this at debug level. For two reasons:
1. It's been that way before.
2. The root daemon uses these settings as a trick to get around the fact that it runs as root. I'd like that to be revealed when debugging, but not during normal use.

PR #2928, opened 2022-12-07
Remove the kubernetes port-forward from the root daemon.

thallgren:
## Description

The introduction of a port-forward that was created directly in the root daemon to increase outbound traffic performance, turned out to be a bad idea. Both the gce and AWS authentication mechanisms stopped working, and possibly other 3rd-party auth mechanisms too.

This commit reintroduces the bridge maintained by the user daemon. This bridge is however tailored for the root daemon and doesn't expose the full set of methods in the traffic-manager gRPC.

## Checklist


njayp (approved):
seems good.

PR #2936, opened 2022-12-09
Use kubernetes.io/metadata.name when mutating webhook matches namespaces

### LINKED ISSUE #2913: Replacing `app.kubernetes.io/name` with `kubernetes.io/metadata.name` in mutatingwebhook
Interception not working with namespace-scoped traffic manager installation with default labels of namespace even with k8s version > 1.21 because the mutating webhook use `app.kubernetes.io/name` label for namespaces scoping which is not added by default to namespaces by k8s controller

I would recommend replacing `app.kubernetes.io/name` with `kubernetes.io/metadata.name` because according to the K8s documentation  `app.kubernetes.io/name` is used for workload resources and namespace is not part of workload resources, and `kubernetes.io/metadata.name` is set by the K8s API on all namespaces and can't be changed and can be used for targeting namespace

thallgren:
## Description

The `kubernetes.io/metadata.name` is always set in Kubernetes versions
1.21 and higher. This commit ensures that it is used in favor of
`app.kubernetes.io/name` on such versions.

Closes #2913

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Changelog
 
+### 2.9.6 (TBD)
+
+- Change: If the cluster is Kubernetes 1.21 or later, the mutating webhook will find the correct namespace
+  using the label `kubernetes.io/metadata.name` rather than `app.kuberenetes.io/name`.
+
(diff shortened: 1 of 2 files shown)


thallgren:
This was approved before I split the PR into smaller ones.

PR #2941, opened 2022-12-13
Make the traffic-manager extendable from other modules

thallgren:
## Description

This PR makes it possible to extend the traffic-manager by adding:
1. An extension point to change the traffic-manager name.
2. An extension point to add a filesystem overlay to the charts directory.

The PR also moves the RPC for errcat from pkg/client to pkg because we're using it from the traffic-manger.

## Checklist


raphaelreyna (approved):
nice; lgtm :+1:

PR #2942, opened 2022-12-13
Add a make rule for building the helm chart

ghost:
## Description

New make rule `make build-chart` that will be used by the Telepresence pro system tests

## Checklist

### DIFF
--- a/build-aux/main.mk
+++ b/build-aux/main.mk
@@ -24,6 +24,8 @@ BINDIR=$(BUILDDIR)/bin
 
 RELEASEDIR=$(BUILDDIR)/release
 
+CHARTDIR=$(BUILDDIR)/chart
+
 bindir ?= $(or $(shell go env GOBIN),$(shell go env GOPATH|cut -d: -f1)/bin)
 
 # DOCKER_BUILDKIT is _required_ by our Dockerfile, since we use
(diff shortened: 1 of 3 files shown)


thallgren (commented):
Why is this needed? I'm asking because I'm about to move the `packaging/*chart*` files to telepresence-pro, where the chart will also include the ambassador-agent. The act of publishing a chart that isn't embedded in the `telepresence` binary will be performed there.

I'm also contemplating a `telepresence helm pull` command that would extract the embedded chart from the binary. Would that help your use case?

ghost:
> Why is this needed? I'm asking because I'm about to move the `packaging/*chart*` files to telepresence-pro, where the chart will also include the ambassador-agent. The act of publishing a chart that isn't embedded in the `telepresence` binary will be performed there.
> 
> I'm also contemplating a `telepresence helm pull` command that would extract the embedded chart from the binary. Would that help your use case?

@thallgren I didn't want to duplicate code in the Telepresence Pro repo and instead use this make target to generate the chart and pass it to the system tests.

If you want to move these scripts to the Telepresence Pro repo I can rebase that once it's ready.
Or I can also use the helm pull command.

Let me know how do you want to proceed, and I can update the TP Pro PR.

Would you be able to do this tomorrow? If not, I can temporarily call `packaging/build_chart.sh` from the Pro repo and once you're done the system tests can be updated.

thallgren:
@aosoriodw do we still need this?

thallgren:
Ping @aosoriodw

Issue #2911, opened 2022-11-30
split proxyCluster to proxyPod and proxyService


j2gg0s:
We set proxyCluster to false when manager pod is accessible.
However, this can't represent service is accessible and don't need proxy.
Under most circumstance, access to service depends on kube-proxy and kube-proxy does not exists locally.

So let we split proxyCluster into proxyPod and proxyService?

Besides this, alsoProxySubnets should be omit when we don't need proxy pod.

cindymullins-dw:
Hi @j2gg0s , sorry to ask but is proxyCluster a kubernetes spec or specifically related to your deployment?  I don't find this in the Helm chart for Telepresence.  Can you give us some more context?

j2gg0s:
@cindymullins-dw 

Set here
https://github.com/telepresenceio/telepresence/blob/2dd5b5b431888348c5287234fa2331df5fd97541/pkg/client/rootd/session.go#L649

Used here
https://github.com/telepresenceio/telepresence/blob/2dd5b5b431888348c5287234fa2331df5fd97541/pkg/client/rootd/session.go#L600

thallgren:
@j2gg0s you write "Besides this, alsoProxySubnets should be omit when we don't need proxy pod". I'm not so sure of that. The fact that the client has access to the pod subnets doesn't necessarily mean that it has access to all networks that the pod has access to.

I agree though, that a more fine grained selection could be made here. Right now, you need to control this with `neverProxySubnets` and `alsoProxySubnets`.

j2gg0s:
@thallgren 
According to my understanding, 
if traffic-manager pod is accessible, and proxyCluster will be set to false,  and we will ignore alsoProxySubnets.
So alsoProxySubnets cant let telepresence proxy service subnet

j2gg0s:
@thallgren will we split this config? If the plan is determined, can i provide PR?

thallgren:
@j2gg0s I haven't had time to make a deeper assessment about what's needed here yet. If you have modifications that solve your use case, then I'd be happy to review them.

j2gg0s:
> @j2gg0s I haven't had time to make a deeper assessment about what's needed here yet. If you have modifications that solve your use case, then I'd be happy to review them.

@thallgren I offer a simple solution, https://github.com/telepresenceio/telepresence/pull/2946

cindymullins-dw:
[...]

PR #2946, opened 2022-12-16
feat: Still enable alsoProxySubnets even though we can access pod.

j2gg0s:
When we can access the pod but no the service, user can use --also-proxy enabled proxy service subnets.

Related https://github.com/telepresenceio/telepresence/issues/2911

### DIFF
--- a/pkg/client/rootd/session.go
+++ b/pkg/client/rootd/session.go
@@ -591,6 +591,11 @@ func (s *Session) onClusterInfo(ctx context.Context, mgrInfo *manager.ClusterInf
 		}
 
 		s.clusterSubnets = subnet.Unique(subnets)
+	}
+	// NOTE: When we can access traffic manager by pod ip, we set proxyCluster to false.
+	// Howevery, if you need and can't access cluster by service,
+	// you can use alsProxySubnets to enable proxy service subnets.
+	if s.proxyCluster || len(s.alsoProxySubnets) > 0 {
 		if err := s.refreshSubnets(ctx); err != nil {
 			dlog.Error(ctx, err)
 		}
@@ -626,6 +631,7 @@ func (s *Session) checkConnectivity(ctx context.Context, info *manager.ClusterIn
 		return
 	}
 	conn.Close()
+	// NOTE: we can access pod, but can we access service?
 	s.proxyCluster = false
 	dlog.Info(ctx, "Already connected to cluster, will not map cluster subnets")
 }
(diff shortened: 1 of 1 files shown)


pkg/client/rootd/session.go
```diff
@@ -591,6 +591,11 @@ func (s *Session) onClusterInfo(ctx context.Context, mgrInfo *manager.ClusterInf
 		}
 
 		s.clusterSubnets = subnet.Unique(subnets)
+	}
+	// NOTE: When we can access traffic manager by pod ip, we set proxyCluster to false.
+	// Howevery, if you need and can't access cluster by service,
+	// you can use alsProxySubnets to enable proxy service subnets.
```
thallgren:
```suggestion
	// However, if you need and can't access cluster by service,
	// you can use alsoProxySubnets to enable proxy service subnets.
```

pkg/client/rootd/session.go
```diff
@@ -591,6 +591,11 @@ func (s *Session) onClusterInfo(ctx context.Context, mgrInfo *manager.ClusterInf
...
+	if s.proxyCluster || len(s.alsoProxySubnets) > 0 {
```
thallgren, line 598: I consider this a quick-fix. The real fix is to discover what subnets that are accessible and what subnets that aren't, and then only map the latter. The `alsoProxySubnets` shouldn't be needed in order to accomplish that.

pkg/client/rootd/session.go
```diff
@@ -626,6 +631,7 @@ func (s *Session) checkConnectivity(ctx context.Context, info *manager.ClusterIn
 		return
 	}
 	conn.Close()
+	// NOTE: we can access pod, but can we access service?
```
thallgren: Please remove this note. It isn't helpful. A `TODO: <a more elaborated action>` would be different.

thallgren (changes requested):

pkg/client/rootd/session.go
j2gg0s: Agree.

We should test pod subnet and service subnet for accessibility separately.
However, i am not sure if we should save the manager service ip to test accessible of service subnet.

The bug will completely block the user's use of telepresence, a quick fix might be necessary.
At the same time, from the standpoint of my own habits:
if i set `alsoProxySubnet`, it should take affect, regardless of whether the cluster is accessible or not.

thallgren:
Superseded by #3074

PR #2955, opened 2022-12-28
Don't panic when --docker-run is combined with --name

### LINKED ISSUE #2953: Panic if --docker-run --name uses space not '=' to separate option name from value
**Describe the bug**

Just migrating from legacy telepresence and dropping in telepresence 2  with existing use of
legacy telepresence panicked. 

Stack trace led me to the 

`startInDocker` function in the 

https://github.com/telepresenceio/telepresence/blob/release/v2/pkg/client/cli/intercept/state.go

file.

Specifically the attempt to extract the value of the '--name' option:

`name, hasName := getArg("--name")`

fails and panics in the `getArg` function:

 ```go
if i+1 < len(args) {
					return parts[i+1], true
```
As it is assumed the `parts` array has enough elements, which is not the case if the `--name` option is of the form

`--name the-name`

rather than:

`--name=the-name`

as the string is only split on '=':

```go
   parts := strings.Split(arg, "=")
```

**To Reproduce**

run telepresence v2 using legacy options and in the --docker-run arguments specify `--name the-name` not `--name=the-name`

```sh
telepresence --swap-deployment deployment-name --namespace blah --docker-run --name the-name ...
```

**Expected behavior**
telepresence to not panic and to run the docker container as specified

 -  all v2.9.5
 - Linux Alma 8
 - Rancher k3s

thallgren:
## Description

The code that parses arguments after `--` would panic on flags that
used `--flag value` instead of `--flag=value`. This commit ensures that
both variants work.

## Checklist

### DIFF
--- a/pkg/client/cli/intercept/state.go
+++ b/pkg/client/cli/intercept/state.go
@@ -4,7 +4,6 @@ import (
 	"bufio"
 	"context"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"os"
 	"runtime"
@@ -329,7 +328,11 @@ func (s *state) CreateRequest(ctx context.Context) (*connector.CreateInterceptRe
 	return ir, nil
 }
 
-func getArg(args []string, flag string) (string, error) {
+// getUnparsedFlagValue returns the value of a flag that has been provided after a "--" on the command
+// line, and hence hasn't been parsed as a normal flag. Typical use case is:
+//
+//	telepresence intercept --docker-run ... -- --name <name>
+func getUnparsedFlagValue(args []string, flag string) (string, error) {
 	feq := flag + "="
 	for i, arg := range args {
 		var v string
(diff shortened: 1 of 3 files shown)


pkg/client/cli/intercept/state.go
```diff
@@ -329,34 +329,43 @@ func (s *state) CreateRequest(ctx context.Context) (*connector.CreateInterceptRe
 	return ir, nil
 }
 
+func getArg(args []string, flag string) (string, error) {
```
josecv: maybe this should be called `getArgForDocker`?

pkg/client/cli/intercept/state.go
```diff
@@ -329,34 +329,43 @@ func (s *state) CreateRequest(ctx context.Context) (*connector.CreateInterceptRe
...
+				}
+			}
+		case strings.HasPrefix(arg, feq):
+			v = arg[len(feq):]
+		default:
+			continue
+		}
+		if v == "" {
+			return "", errors.New("docker flag `--name` requires a value")
+		}
```
josecv: This shouldn't be `--name`, but should rather include the provided `flag`

josecv (changes requested):

pkg/client/cli/intercept/state.go
thallgren: Haha, yeah :-)

josecv (approved):

PR #2966, opened 2023-01-06
add team mode to smoke test

njayp:

## Description

Add team mode testing to smoke test. We test

- default behavior in single-user mode while logged in
- that upgrading to team mode does not roll the tm and destroy intercepts
- default behavior in team-mode while logged out


josecv (approved):

thallgren (approved):
Given that the smoke tests require the proprietary versions of Telepresence and the Traffic Manager, I suggest that once this PR has been merged, we move the smoke-tests directory to the proprietary repo (with retained commit-history).

PR #2982, opened 2023-01-19
add the ability to stream formatted objects

njayp:

## Description
Fixes panic caused by in `telepresence list --output json --watch`. Adds a function that will immediately print an object. **This PR is not well tested.**

## Checklist

### DIFF
--- a/pkg/client/cli/cmd_list.go
+++ b/pkg/client/cli/cmd_list.go
@@ -126,10 +126,10 @@ func (s *listInfo) list(cmd *cobra.Command, _ []string) error {
 looper:
 	for {
 		select {
-		case r := <-ch:
-			s.printList(ctx, r.Workloads, stdout, formattedOutput)
 		case <-ctx.Done():
 			break looper
+		case r := <-ch:
+			s.printList(ctx, r.Workloads, stdout, formattedOutput)
 		}
 	}
 	return nil
@@ -138,7 +138,7 @@ looper:
 func (s *listInfo) printList(ctx context.Context, workloads []*connector.WorkloadInfo, stdout io.Writer, formattedOut bool) {
 	if len(workloads) == 0 {
 		if formattedOut {
-			output.Object(ctx, []struct{}{}, false)
+			output.StreamObject(ctx, []struct{}{}, false)
 		} else {
 			fmt.Fprintln(stdout, "No Workloads (Deployments, StatefulSets, or ReplicaSets)")
 		}
@@ -161,7 +161,7 @@ func (s *listInfo) printList(ctx context.Context, workloads []*connector.Workloa
 	}
 
 	if formattedOut {
-		output.Object(ctx, workloads, false)
+		output.StreamObject(ctx, workloads, false)
 	} else {
 		includeNs := false
 		ns := s.namespace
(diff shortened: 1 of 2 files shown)


pkg/client/cli/cmd_list.go
```diff
@@ -126,10 +126,10 @@ func (s *listInfo) list(cmd *cobra.Command, _ []string) error {
 looper:
 	for {
 		select {
-		case r := <-ch:
-			s.printList(ctx, r.Workloads, stdout, formattedOutput)
 		case <-ctx.Done():
 			break looper
+		case r := <-ch:
```
thallgren, line 131: This change seems unnecessary given that the order things appear i a `select` isn't significant.

pkg/client/cli/output/output.go
```diff
@@ -74,6 +74,18 @@ func Object(ctx context.Context, obj any, override bool) {
 	}
 }
 
+// StreamObject prints an object immediately.
+func StreamObject(ctx context.Context, obj any, override bool) {
```
thallgren, line 78: The idea with the `Object` function is that it panics if mixed with any other types of output. The whole point of designing it that way, was that we never should create commands that produces multiple formatted objects, because it leads to unpredictable JSON output.

pkg/client/cli/output/output.go
```diff
@@ -132,23 +141,51 @@ func Execute(cmd *cobra.Command) (*cobra.Command, bool, error) {
...
 	case formatJSON:
-		if encErr := json.NewEncoder(o.originalStdout).Encode(obj); encErr != nil {
-			panic(encErr)
+		// create the encoder once
+		encoder := json.NewEncoder(originalStdout)
+		return func(obj any) {
+			if err := encoder.Encode(obj); err != nil {
+				panic(err)
+			}
 		}
```
thallgren, line 173: This will produce invalid JSON on the encoder's output when the encoder is called more than once. It'll just print the objects, E.g.:
```
{"X":"x"}
{"Y":"y"}
```
The correct JSON must be a comma separated array:
```
[{"X":"x"},{"Y":"y"}]
```
And this is the reason why streaming is difficult.

thallgren (changes requested):
It's not clear to me why we would need to stream objects on output. It's difficult and the solution here won't work (and there's no test to prove that it does).

I think the current design, that force us to provide one single object to be formatted, is to be preferred, but give me a valid use case where streaming is necessary, and I'll change my mind.

pkg/client/cli/output/output.go
njayp: I should have provided more context. The problem comes from the display of services in the docker extension. If the user adds a service to the cluster, it should also appear in the table of intercept-able services shown by the extension. To do this, the extension calls `telepresence list --output json --watch`, which I wrote in the past to print a new list object whenever the list of services changed. Whenever a new object is printed, the extension would read it in and update the table accordingly. 

`{"X":"x"} \n
{"Y":"y"} \n
...`

is what I want.

thallgren: I understand. So how about this:
1. Add a new output format called "json-stream", which we document as having your desired characteristics.
2. Keep the original code as is, except ines 65-72, which, when that format is used, instead does:
```
		if encErr := json.NewEncoder(o.originalStdout).Encode(obj); encErr != nil {
			panic(encErr)
		}
		o.override  = true
```
Also, add a `case formatJSON_STREAM:` to the switch at line 136 that does nothing.

The `--watch` flag should not be combinable with `--format json`. I think `--format yaml` would be OK though, because the output is still valid yaml.

PR #2985, opened 2023-01-20
add retry to dialRequestWatcher

njayp:

## Description
Telepresence would shut down if it lost connection for a few minutes. This was caused by the `dialRequestWatcher` go-routine, which would exit with error, cascading into the death of the session. Adding a retry to `dialRequestWatcher` fixed the problem.

There was a secondary problem where the first message after reconnecting failed. This was due to the traffic manager attempting to re-use the dead connection. Now, on failure, the listenrequest is put back in the channel to be used by the next connection

### DIFF
--- a/cmd/traffic/cmd/manager/service.go
+++ b/cmd/traffic/cmd/manager/service.go
@@ -845,6 +845,14 @@ func (m *service) WatchDial(session *rpc.SessionInfo, stream rpc.Manager_WatchDi
 			}
 			if err := stream.Send(lr); err != nil {
 				dlog.Errorf(ctx, "WatchDial.Send() failed: %v", err)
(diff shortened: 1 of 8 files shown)


knlambert (commented):
LGTM, not sure I entirely get what a Dial is in this context though. Is it some sort of ping just before opening a tunnel ?

njayp:
> LGTM, not sure I entirely get what a Dial is in this context though. Is it some sort of ping just before opening a tunnel ?

The TUN is the network device. Inside the TUN we open one tunnel per connection, we open this tunnel by calling the manager.Tunnel grpc. Once we have a Tunnel, we wrap it in a dialer, the dialer will manage the messages that go through the tunnel.

knlambert (approved):

PR #2986, opened 2023-01-20
Add ability to install a released traffic-manager to the integration_tests.

thallgren:
Adds the method `InstallTrafficManagerVersion` to the integration_test framework so that tests can be written that exercises the current client against older traffic-managers.

### DIFF
--- a/integration_test/itest/cluster.go
+++ b/integration_test/itest/cluster.go
@@ -81,6 +81,7 @@ type cluster struct {
 	testVersion      string
 	compatVersion    string
 	registry         string
+	agentRegistry    string
 	kubeConfig       string
 	generalError     error
 	logCapturingPods sync.Map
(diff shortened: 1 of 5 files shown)


integration_test/itest/cluster.go
```diff
@@ -489,6 +537,20 @@ func (s *cluster) InstallTrafficManager(ctx context.Context, values map[string]s
 	return err
 }
 
+func (s *cluster) pullHelmChart(ctx context.Context, version string) (string, error) {
+	if err := Run(ctx, "helm", "repo", "add", "datawire", "https://app.getambassador.io"); err != nil {
+		return "", err
+	}
```
josecv, line 543: if the repo already exists this call will fail; maybe check the error?

josecv (changes requested):
lgtm just a note on the helm repo add

integration_test/itest/cluster.go
thallgren: I don't think that's true. You do get a message `datawire already exists with the same configuration, skipping`, but it doesn't fail.

josecv: hmm okay, i've seen it fail in the makefile, but maybe that's something else

josecv (approved):

PR #2989, opened 2023-01-24
Add unit tests on top of NewSession method

knlambert:
## Description

This is a proposal to start adding unit tests to some hard to test parts of the code. 

I would like to have your feedback on this particular case : if it makes sense, It can be a good example of how to make existing part of our code unit test friendly.

**Outline of the changes:**

I'm testing the `NewSession` function since it takes care of a lot of logic, following the [`also-proxy` parsing I did this week](https://github.com/telepresenceio/telepresence/pull/2987).

I had to replace several standalone function for struct methods, to make them compatible with interfaces (and so with the mocks). 

For example:

* `NewKubeconfig` -> `(*KubeConfigResolver).NewKubeconfig` (`pkg/client/k8s_config.go`).
* `ConnectToManager` -> `(*ManagerConnector).Connect` (`pkg/client/tm/connect.go`)

But I also had to add wrappers around the standard library for some functions to make them easy to mock too (See `pkg/std/os.go` and `pkg/std/user.go`.

As a result, you can see we now use interfaces to mock these methods in some strategical places.

Example with the `NewSession` function.

**Before:**
```go
func NewSession(
    ctx context.Context,
    sr *scout.Reporter,
    cr *rpc.ConnectRequest,
) (context.Context, userd.Session, *connector.ConnectInfo) {
   // [...]
}
```

`sr` is impossible to mock, since the object can only use the exact implementation. We need to have an interface instead to be able to pass a mock.

Also, some object initializations are made from the inside of the function, so it's not possible to inject mocks or different implementations:

```go
dnet.NewK8sPortForwardDialer(ctx, cluster.Kubeconfig.RestConfig, k8sapi.GetK8sInterface(ctx))
```

**Now:**

```go
type SessionManager struct {
	kcr  KubeConfigResolver
	p    PortForwardDialerBuilder
	mc   ManagerConnector
	sc   UserSessionCache
	os   OS
	user User
}

func (s *SessionManager) NewSession(
    ctx context.Context,
    sr userd.Reporter,
    cr *rpc.ConnectRequest,
) (context.Context, userd.Session, *connector.ConnectInfo) {
   // [...]
}
```

Several things:
* `scout` is now an interface, because we want to be able to pass different implementation, including mocks.
* `newSession`is now a method of a struct, since some dependencies that we want to mock should not be passed as parameter of the function each time we call it (basically all the `SessionManager` attributes like `os`, `user`, `mc`...).
* Some dependencies should not be initialized inside the function, so this is now done at `SessionManager` creation, using its attributes (it's all interfaces aswell of course).

Here an example of interface with `userd.Reporter`  : 

```go
//go:generate mockgen -destination=mocks/reporter_mock.go . Reporter
type Reporter interface {
	Report(ctx context.Context, action string, entries ...scout.Entry)
	SetMetadatum(ctx context.Context, key string, value any)
	InstallID() string
}
```

We can generate mocks from it, so it's easier to test it in our test suite. It could look like that: 

```go
// given
ctrl = gomock.NewController(s.T())
reporter := mock_userd.NewMockReporter(s.ctrl)
stdOS = mock_trafficmgr.NewMockOS(s.ctrl)

sessionManager := &SessionManager{
     os:   s.stdOS,
}

// We exepct a call to the report method.
reporter.EXPECT().Report(gomock.Any(), "connect")
// We expect a call to the `Current` method to get user information.
stdUser.EXPECT().Current().Return(&stduser.User{
		Uid:      "1000",
		Gid:      "1000",
		Username: "jdoe",
		Name:     "John doe",
		HomeDir:  "/home/jdoe",
	}, nil).AnyTimes()

// when
newCtx, sess, c := sessionManager.NewSession(ctx, reporter, cr)

// then
// some asserts
```

I also need to adapt a little the code of telepresence-pro to make it compatible. Please find the PR [here](https://github.com/datawire/telepresence-pro/pull/131).


pkg/client/userd/trafficmgr/mocks/clusterbuilder_mock.go
```diff
@@ -0,0 +1,52 @@
+// Code generated by MockGen. DO NOT EDIT.
+// Source: github.com/telepresenceio/telepresence/v2/pkg/client/userd/trafficmgr (interfaces: ClusterBuilder)
+
+// Package mock_trafficmgr is a generated GoMock package.
+package mock_trafficmgr
```
thallgren, line 5: Won't all this mock-code end up in the runtime binary? In other words,  shouldn't it all reside in xxx_test.go files?

thallgren (commented):
Making it easier to unit-test the code is always good. I'm not sure about introducing mockgen though. It generates a lot of stuff that then must be regenerated when the code change, which in turn will make all future PRs bigger and harder to read. So first question is, do we really need that extra layer of complexity, or can we create our own mocks as needed?

Second question is, if we do need mockgen, shouldn't all generated files be excluded from the runtime? I.e. include either a build instruction that excludes them, or reside in a `_test` package? I don't want the runtime binary to include pure test code.

knlambert:
> Making it easier to unit-test the code is always good. I'm not sure about introducing mockgen though. It generates a lot of stuff that then must be regenerated when the code change, which in turn will make all future PRs bigger and harder to read. So first question is, do we really need that extra layer of complexity, or can we create our own mocks as needed?
> 
> Second question is, if we do need mockgen, shouldn't all generated files be excluded from the runtime? I.e. include either a build instruction that excludes them, or reside in a `_test` package? I don't want the runtime binary to include pure test code.

Hello, in my experience, it's definitively a tradeoff, you're right, and I know that in the golang world, many say "Just write your own mocks !". 

But each time we saw the use of custom mocks, [in saas-app for example](https://github.com/datawire/saas_app/blob/master/internal/pkg/keycloak/keycloak_test.go#L1320-L1423), it was actually more interesting to use gomock : 

* We don't always meet the same implementations, so we have to learn the one the developper implemented. Sometimes it's a super simple mock, but it can have some additionnal custom logic.
* While it is a lot of code generation, it only generates what's specified in your interface (so only what you use since I define interfaces on the consumer side). You end up with the same amount of methods than with a custom mock. If you look at the code, [ it's not worse](https://github.com/datawire/saas_app/blob/master/internal/pkg/subscription/mocks/organizationservice_mock.go) than a [custom one ](https://github.com/datawire/saas_app/blob/master/internal/pkg/keycloak/keycloak_test.go#L1320-L1423)
* It's boring, and having to implement mocks on top of the tests also dissuades people from doing them since it's long and repetitive. 
* While it is a lot of code generation at first, when the mocks are generated, they don't change that much, since you just add somes methods from time to time, and the generation can even be done in one click from goland (with the go generate).

I just end up ignoring these files during the reviews.

For the second one, that's a good point, but it’s not an issue because the mocks are only imported from the test files (which themself are not included), so the compiler is smart enough to not include the mock files.

I’ve validated it with a simple example, but also with telepresence: 97M on release/v2, 97Mb on my branch too :) 

```bash
➜  telepresence git:(release/v2) ✗ ls -lh build-output/bin/telepresence
-rwxr-xr-x  1 kevinlambert  staff    97M Jan 30 15:05 build-output/bin/telepresence
```
```bash
➜  telepresence git:(knlambert/add-also-never-proxy-to-connect-with-tests-2) ✗ ls -lh build-output/bin/telepresence
-rwxr-xr-x  1 kevinlambert  staff    97M Jan 30 15:08 build-output/bin/telepresence
```

thallgren:
I'm not so concerned by the tradeoff between generated and hand-written mocks. What I am concerned about is that we need mocks at all. It's easy to write an integration test that validates the new flag, and by doing so, you don't just remove the need for the mocks, you also write tests that cover the real thing. Example:

```go
func (s *cliSuite) Test_AlsoProxyFlag() {
	ctx := s.Context()
	itest.TelepresenceQuitOk(ctx)
	itest.TelepresenceOk(ctx, "connect", "--also-proxy", "192.168.12.0/24")
	stdout, _, err := itest.Telepresence(ctx, "status", "--output", "json")
	s.Require().NoError(err)
	var status map[string]any
	s.NoError(json.Unmarshal([]byte(stdout), &status))
	// validate that the status map contains the correct proxy settings
}
```
If so desired, this can be extended with a `helm upgrade` that also sets the same property in the traffic-manager, and also settings in the local config, so that the priorities of those settings can be verified. It's only a few lines of code, but still a very powerful test.

thallgren:
In essence, I think it's about striking a balance. My preference would be if that balance is formulated as:

- Write unit tests for simple functionality.
- Write integration tests when it gets complicated.

And I rather we spend time improving the simplicity of writing and running integration tests, than introducing a mock generator.

thallgren:
And yes, I know that the compiler is smart enough to rule out things that aren't used. My concern is that we might do mistakes because the mocks are available in the runtime scope. They are also sometimes disturbing (the IDE doesn't rule them out when doing code-completion, considers them for auto-imports etc.).

knlambert:
I understand, but my logic is the following:

* Integrations tests already take 40 min to execute, and don't cover each possible edge cases (including unexpected errors, different setups, etc). Unit tests give the flexibility to emulate them without having to add extra infra / run time.
* Integration tests alone do not cover my needs. I can't TDD with integration tests, and as a new comer, they don't give me enough confidence. We should use both if it makes us more comfortable / proficient. I want to know if I've broken something and if it behaves like I expect. I doesn't mean we can't do an integration test if it seems like a quick win depending on the case.
* Unit tests without mocks are not possible except for simple things, as you said, so we need them.

* When it comes to importing mocks by mistake, I really never saw that issue. But worst case scenario, nomenclature and adding a test to validate that the binary isn't bigger than expected are valid solutions.

You can see that as an experiment. I really think that it could make our life easier, and in the end, the only risk is importing a mock by mistake. This is just giving us more options. You're still free to tell us if one particular unit test is cumbersome at the end of the day.

thallgren:
- An individual integration tests typically take 40 seconds to execute, and they are very easy to run locally. Please read my latest additions to DEVELOPERS.md. If you are concerned that the full set takes 40 minutes, then let's fix that. There are several tests that take far too long today because of unnecessary timeouts.
- Of course you can do TDD with integration tests. I do it all the time. Those tests are closer to what the user will experience, and hence closer to the end goal with the development. Not sure I understand the confidence argument. Mocks don't give me confidence because they don't test the real thing. Integration tests do.
- I would argue that unit tests *should* be intended for simple things, and that we don't need a solution for semi-complicated things that sits in between simple unit tests and integration tests.

I'm not at all convinced that the advantages using mocks outweigh its disadvantages in this code-base. Perhaps it's different in the saas-app where I don't think there is anything corresponding to our integration test framework.

thallgren:
If this was just an experiment that didn't require vital code to be refactored (introducing mocks often do), then it would be easy to slot in and then slot out again. But that isn't the case. Examples:

- A new `SessionManager` is introduced to create sessions. In addition to that, it also creates and deletes managers using helm (functions that doesn't need a session, just a k8sapi). So, it's not a just a session manager. It also manages helm installs. This is not a natural design because it's driven by the desire to be able to mock things that doesn't really exist, just for the sake of mocking.

- A new `ClusterBuilder` is introduced, just to get a way to replace the call to `CreateCluster`. That method is a misnomer in a sense, because it connects to a cluster, it doesn't create it. It creates the cluster-struct though. The effect of the misnomer gets worse when we introduce a `ClusterBuilder`, because that really sounds like something that builds a cluster from the ground up. It definitely doesn't sound like something related to a connect. The builder is there for one purpose only. To be able to mock the cluster connection.

- Then we have `KubeConfigResolver` just to be able to mock functions to deal with the kubeconfig, bringing NewConfig and NewClusterConfig together in one. Those two will never be used together in a binary. It's either or, but here the compiler is forced to bring in both. Just for the sake of mocking (the compiler might be forced to do that for other reasons too, but that's beside the point).

I can go on, but I think you understand what I'm getting at. The code gets locked down in undesired constructs that we will continue to build on, which will make future refactorings and/or code-splits more difficult.

PR #2998, opened 2023-01-30
Make it possible to run connect to remote user daemon.

thallgren:
## Description

This PR is makes it possible to split telepresence functionality so that a long lieved user daemon can run in one container and the CLI commands can  xecute from other containers. This is made possible by introducing:

1. The user daemon flag `--address`. When set, the user daemon will not
   use a unix socket (or named pipe on windows). Instead, it will use
   the address to configure a TCP listener.
2. The user daemon flag `--embed-network`. When set, the user daemon
   will embed the session that normally runs in the root daemon in the
   same process instead of contacting it using gRPC. This means that no
   root daemon needs to be started.
3. The environment variable `TELEPRESENCE_USER_DAEMON_ADDRESS`. When
   set, the CLI will assume that a user daemon has been started and can
   be connected using that address. It will also refrain from starting
   a root daemon.

## Checklist


pkg/client/cli/intercept/state.go
```diff
@@ -149,7 +150,10 @@ func create(sif State, ctx context.Context) (acquired bool, err error) {
 		// local-only
 		return true, nil
 	}
-	fmt.Fprintf(s.cmd.OutOrStdout(), "Using %s %s\n", r.WorkloadKind, s.AgentName)
+	detailedOutput := s.DetailedOutput && output.WantsFormatted(s.cmd)
+	if !detailedOutput {
+		fmt.Fprintf(s.cmd.OutOrStdout(), "Using %s %s\n", r.WorkloadKind, s.AgentName)
+	}
```
njayp, line 156: Instead of requiring that both flags are set, if `DetailedOutput` is set, can we default to yaml?

pkg/client/cli/cmds_misc.go
```diff
@@ -111,11 +112,16 @@ func InitConnectRequest(cmd *cobra.Command) (*connector.ConnectRequest, *pflag.F
...
 
-	kubeConfig := genericclioptions.NewConfigFlags(false)
-	kubeConfig.Namespace = nil // "connect", don't take --namespace
-	kubeFlags := pflag.NewFlagSet("Kubernetes flags", 0)
-	kubeConfig.AddFlags(kubeFlags)
-	flags.AddFlagSet(kubeFlags)
+	// Only include the kubernetes flags if the user daemon is started from the CLI. It's assumed that it is configured
+	// already using a default .kube/config if it is reachable using an address.
+	var kubeFlags *pflag.FlagSet
+	if client.GetEnv(ctx).UserDaemonAddress == "" {
```
njayp, line 118: I'm a little confused. Are we assuming that the daemon is already connected to the cluster if `UserDaemonAddress` is set?

njayp (approved):

pkg/client/cli/intercept/state.go
thallgren: I prefer to be able to choose between json (which I'm currently using) and yaml.

pkg/client/cli/cmds_misc.go
thallgren: No, but we assume that the daemon is started and that the CLI can connect to it.

njayp: I dont get why we can get rid of the kubeflags if we are not connected yet, Im sure im just missing something 🤷

PR #3010, opened 2023-02-06
Add support for CRD helm chart & integrate with helm install command

knlambert:
## Description

We want to add support for CRDs built on top of telepresence, although the current implementation doesn't allow it.

When using helm, [the best practice seems to isolate the CRDs in their own chart](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#method-2-separate-charts).

Since the code base to handle helm charts is already there, I've choosen to go this way.

This PR adds a new empty chart as a placeholder for future implementations, and the possibility to extend the code and enable these CRDs if required.

Outline of the changes:
* New CRD bool parameter for the Helm endpoint.
* Reuse of the functions to manage the helm chart by adding a `releaseName` parameter (Then we can use them for both the CRDs and the traffic-manager) since it was hardcoded before.
* Update of the charts package to reflect the two possible charts (CRDs and traffic-manager). It's mostly about adding a parameter for traffic manager vs crds.

## Checklist


charts/chart.go
```diff
@@ -4,29 +4,42 @@ import (
...
 	"embed"
+	"fmt"
 	"io"
 	"io/fs"
 	"sort"
 	"strings"
 
-	"github.com/spf13/afero"
 	"helm.sh/helm/v3/pkg/chart"
+
```
thallgren: This empty line indicates that you're not using the standard go-import rules.

thallgren (commented):
I was surprised to see this in the OSS repo given that the feature it enables is proprietary.

knlambert:
> I was surprised to see this in the OSS repo given that the feature it enables is proprietary.

It's basically empty though and not enabled for OSS. It was the easiest way to implement it (since the helm install code is here), but now if at some point we want to add CRDs to the OSS repo, it's possible.

thallgren (approved):

PR #3013, opened 2023-02-07
Add a Mount struct to the detailed intercept output.

thallgren:
## Description

When running the Telepresence daemon inside a container, it's desirable to prevent it from mounting remote directories because such mounts will not be visible to other containers unless the bind-mounted volume where the mount is made uses `bind-propagation=shared`. The bind propagation is not supported when using Docker Desktop (so in essence, it only works on Linux). In essence, this means that the mounts must be made by the host, based on information from the intercept, before the container that handles the intercepted traffic is started.

This PR adds a `mount` structure to the output from `telepresence intercept --detailed-output --output json xxx`. The struct will be filled in with mount details from the intercepted container even if the intercept was made using `--mount=false`. This enables an orchestrator to run the daemon and the intercept command in containers, capture this output, and then mount directories directory from the host before starting the intercept handlers.

## Checklist

### DIFF
--- a/pkg/client/cli/intercept/info.go
+++ b/pkg/client/cli/intercept/info.go
@@ -6,7 +6,7 @@ import (
 	"strings"
 
 	"github.com/telepresenceio/telepresence/rpc/v2/manager"
-	client2 "github.com/telepresenceio/telepresence/v2/pkg/client"
+	"github.com/telepresenceio/telepresence/v2/pkg/client"
 )
 
 type Ingress struct {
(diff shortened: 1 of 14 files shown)


pkg/client/cli/intercept/info.go
```diff
@@ -1,9 +1,12 @@
 package intercept
 
 import (
+	"context"
+	"path/filepath"
 	"strings"
 
 	"github.com/telepresenceio/telepresence/rpc/v2/manager"
+	client2 "github.com/telepresenceio/telepresence/v2/pkg/client"
```
knlambert: I guess a bad naming caused by goland refactoring feature.

knlambert (approved):

PR #3016, opened 2023-02-08
add ispec cache

njayp:

## Description

WIP add ispec functions to cache


pkg/client/cache/cache.go
```diff
@@ -10,43 +10,43 @@ import (
...
 	if err != nil {
 		return err
 	}
-	path := filepath.Join(dir, file)
-	jsonContent, err := os.ReadFile(path)
+	filePath := filepath.Join(dir, file)
+	return LoadFromFilePath(dest, filePath)
+}
+
+func LoadFromFilePath(dest any, filePath string) error {
```
josecv: This doesn't have to be public does it? if the idea is that SaveToUserCache will accept something like "ispect/foo-bar.json", then LoadFromUserCache should accept the same and there's no need to ever resort to an absolute path; I say we make this private

josecv (changes requested):

pkg/client/cache/cache.go
njayp: It has to be public so I can use it from the pro-repo 🤷

josecv (approved):

PR #3018, opened 2023-02-08
Fix build failures when dependabot updates a dependency

ghost:
## Description

This PR adds a new job to the `License verification` workflow that will check if a build is running because of a dependabot change.
This job checks if `make generate` made changes to DEPENDENCIES.md/DEPENDENCY_LICENSES.md, and will commit them and start a new build.

This will simplify merging PRs created by dependabot, that would otherwise fail because dependency information is out of date.

If the job detects changes to DEPENDENCIES.md/DEPENDENCY_LICENSES.md it will commit these files, as well as go.mod and go.sum, if applicable.

go.mo and go.sum can change because `go-mkgenerate` will update them if necessary. For example, when a module is updated, a package may be removed from the new version, and if other dependency uses the removed package, `go build` will fail.  `go-mkgenerate` fixes this by updating any dependencies that use removed packages.

Note: The workflow `License verification` was updated to run on `workflow_dispatch` because this is required for builds to be triggered after the action updates the dependency information.

## Checklist

## Tests:
1. [Test](https://github.com/telepresenceio/telepresence/actions/runs/4126635860) that workflow `License verification` fails if dependencies files are out of date and committer is not dependabot.
2. Test that when committer is dependabot, dependency information is updated and committed to the repo. This [workflow](https://github.com/telepresenceio/telepresence/actions/runs/4126734714) is when dependencies are updated, and [this](https://github.com/telepresenceio/telepresence/actions/runs/4126744536) one is when it runs again with the latest changes from the repo.

### DIFF
--- a/.github/workflows/licenses.yaml
+++ b/.github/workflows/licenses.yaml
@@ -24,7 +24,7 @@ jobs:
         shell: bash
         run: make generate
       - name: "Update dependency information after dependabot change"
-        uses: datawire/go-mkopensource/actions/save-dependabot-changes@v0.0.4
+        uses: datawire/go-mkopensource/actions/save-dependabot-changes@v0.0.7
         id: changed-by-dependabot
         with:
           branches_to_skip: 'release/v2'
--- a/.github/workflows/licenses.yaml
+++ b/.github/workflows/licenses.yaml
@@ -1,6 +1,7 @@
 name: "License verification"
 on:
   pull_request:
+  workflow_dispatch:
 jobs:
   dependency_info_linux:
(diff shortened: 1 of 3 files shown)


AidanWenzel (approved):
Looks great 👍 Now we just need dependabot to bump our action versions 🤔

josecv (approved):

ghost:
Updated the action and the `go-mkopensource` tool to the latest version to fix an issue that could make the license info out of sync with go.mod.

josecv (approved):

PR #3022, opened 2023-02-10
Lock podIP/port for FTP and SFTP for the duration of an intercept.

thallgren:
## Description

The telepresence client mounts remote directories based on a podIP and
port that are passed on in the `InterceptInfo`. An intercept will
survive a pod restart that will yield new podIP/port combinations and
the daemon will deal with this correctly. The mount code that now
executes outside the daemon doesn't, because it doesn't watch the
intercepts.

Rather that reimplementing that logic and force the client to watch
intercepts, this commit moves the concern of keeping the podIP and port
stable for the duration of an intercept by introducing an IP map
mechanism that hides the fact that pod IP is restarted.

Similar functionality could have been obtained by using a service and
well-known ports in the traffic-agent, but that will require new RBAC
privileges. The approach used here is completely hidden from the user.

## Checklist

### DIFF
--- a/cmd/traffic/cmd/manager/internal/cluster/info.go
+++ b/cmd/traffic/cmd/manager/internal/cluster/info.go
@@ -93,11 +93,15 @@ func NewInfo(ctx context.Context) Info {
 
 	apiSvc := "kubernetes.default"
 	var clusterDomain string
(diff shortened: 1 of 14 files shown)


thallgren:
Closing in favor of better approach.

PR #3024, opened 2023-02-13
Make Telepresence expose an DNS on port 53 and SFTP ports

thallgren:
## Description

Running Telepresence in a Docker container has its challenges, especially if Docker Desktop is used, the host isn't a Linux box, and mounts and DNS must be available from other containers as well. Using `--network host` is not really an option, because it isn't supported by Docker Desktop.

This PR makes Telepresence capable of two things:
1. A new flag, `--local-mount-port <port>` will make `telepresence intercept --detailed-output --output=[yaml|json]` create  a bridge to the remote SFTP service instead of starting an sshfs client. This enables the sshfs client to be started outside of the container, and thus, mount filesystems that then can be used as source for volumes that other containers will use.
2. When Telepresence detects that it runs in a docker container, it will now expose its DNS on localhost:53. This makes the container itself a DNS server. Very handy when other containers use `--network container:<tp-container>`. Not only do they get the TUN device, they also get a DNS resolver.

## Checklist


pkg/client/remotefs/bridge.go
```diff
@@ -0,0 +1,80 @@
...
+
+import (
+	"context"
+	"fmt"
+	"net"
+
+	"github.com/datawire/dlib/dgroup"
+	"github.com/datawire/dlib/dlog"
+	"github.com/telepresenceio/telepresence/rpc/v2/manager"
+	client2 "github.com/telepresenceio/telepresence/v2/pkg/client"
```
knlambert, line 11: client2 ? :)

knlambert (approved):

PR #3039, opened 2023-02-21
Add dependency submission workflow

shepz:
## Description

The alerts for `tools/src/ko/*`  are still there, my fix on dependabot config didn’t work because the [dependency graph](https://github.com/telepresenceio/telepresence/network/dependencies) still reference it.
I really wonder why github still thinking this folder exists on our default branch. There is an open [issue](https://github.com/dependabot/dependabot-core/issues/4364) in dependabot repo to have the option to ignore manifests in specific directories, but we’ll have to wait…
Another option will be to submit a new snapshot of [go dependencies to fix this dependency graph](https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/using-the-dependency-submission-api).

## Checklist


.github/workflows/go-dependency-submission.yaml
```diff
@@ -0,0 +1,26 @@
...
+jobs:
+  go-action-detection:
+    runs-on: ubuntu-latest
+    steps:
+      - name: 'Checkout Repository'
+        uses: actions/checkout@v3
+
+      - uses: actions/setup-go@v3
+        with:
+          go-version: ">=1.19.0"
```
thallgren:
```suggestion
          go-version: '~1.19.6'
```

thallgren (changes requested):
We should use go-version consistently.

thallgren (approved):

PR #3050, opened 2023-03-02
Make the CLI start or discover daemons running in docker

thallgren:
## Description

This PR changes the way the CLI finds, or starts, the user daemon. Previously, the CLI would either look for socket with a well-known name, or use the `TELEPRESENCE_USER_DAEMON_ADDRESS` to connect via TCP. This PR introduces a new boolean flag, `--docker` to the `telepresence connect` command. When set, the CLI will search the cache for a `DaemonInfo` entry that corresponds to the name of the current Kubernetes context. At most one such entry can exist at any one time. If found, the entry will have a designated port that the CLI can use for its connection.

If `telepresence connect --docker` is used, and no `DaemonInfo` entry is found in the cache, then the command will start a new daemon in a docker container and then connect to that daemon.

Other commands that talk to the daemon (like `telepresence status` or `telepresence list`, will automatically detect the `DaemonInfo` for the current Kuberentes context and connect to it automatically if it exists. If it doesn't, those commands will start a host-based daemon on demand, just like before (anything else would break backward compatibility). A future enhancement might be to alter this behavior using a configuration setting.

The PR also introduces some simplifications to the daemon. When it runs in docker, the daemon will just serve one single session. When the session ends, the daemon terminates. In other words, there's no difference between `telepresence quit` and `telepresence quit -s` when the daemon runs in a container.

When using `telepresence connect --docker`, it is possible to run multiple daemons simultaneously, provided that multiple kubernetes contexts exists.

## Checklist


knlambert:
List command doesn’t seem to detect the deamon in a container. Status works fine though.

I'll spend 30 min to investigate to see if that's not an issue with my setup. Otherwise LGTM :)

knlambert (changes requested):
List command doesn’t seem to detect the deamon in a container. Status works fine though.
I'll spend 30 min to investigate to see if that's not an issue with my setup. Otherwise LGTM :)

thallgren:
That's odd. Both status and list works for me and they use the exact same logic to detect the daemon. When you run list, what does `ls -l ~/Library/Caches/telepresence/daemons` show? If the correct image is running in docker, the timestamp of the file present there is updated every 5 seconds. The name end with the same port number as you see when you do `docker ps`.

In any case, I'll update the log output so that it shows something other than `connector/...`. That way, a `tail -F <log>` will reveal what daemon it is that logs output. The status and version commands also need some work (there's just one daemon).

thallgren:
The last commit takes care of the `telepresence list` not (always) working.

knlambert:
I was thinking, in this new docker mode, how do we start containers that we want to intercept ?

knlambert:
And do we need something to indicate what ports to expose (like for the spec).

thallgren:
There's just one port to expose from the daemon now, and it is auto-discovered by the CLI (the filename of the cached entry ends with the port number).

The way to start an intercept handler in a docker container is to use `telepresence intercept --docker-run`. I have some additional work to do before that connects seamlessly with a daemon running in a container. The spec runner contains the needed code (and will of course still work).

knlambert (approved):

PR #3051, opened 2023-03-03
Fix bug when doing a telepresence list with docker daemon

knlambert:
## Description

Just a draft to fix a bug with telepresence list not working with docker daemon mode.


thallgren:
Superseded by #3050.

PR #3052, opened 2023-03-03
Add authenticator middleware to integrate gke with docker mode

knlambert:
## Description

This PR adds the authenticator code to the OSS binary.

```go
// Start the server by calling it directly from your code:
lis, err := net.Listen("tcp", fmt.Sprintf("localhost:%d", port))
if err != nil {
    return "", fmt.Errorf("failed to listen on %d: %w", port, err)
}
var opts []grpc.ServerOption
grpcServer := grpc.NewServer(opts...)

originalKubeConfig := authenticator.LoadKubeConfig(kubeConfigPath)

rpcAuthenticator.RegisterAuthenticatorServer(grpcServer, originalKubeConfig)

if err := grpcServer.Serve(lis); err != nil {
    log.Fatal(err)
}

// Generate the patched kubeconfig file and provide it to your docker run command from a Go code
dockerKubeConfigStubFile, err := 
    patcher.GenerateTempKubeConfigStubFile(authenticator.LoadKubeConfig(kubeConfigPath))
if err != nil {
    return "", fmt.Errorf("failed to patch kubeconfig file: %w", err)
}
```

## Checklist


pkg/authenticator/patcher.go
```diff
@@ -0,0 +1,69 @@
+package authenticator
```
thallgren: Given that this isn't part of the server, perhaps it could live in a `authenticator/patcher` package.

pkg/authenticator/patcher.go
```diff
@@ -0,0 +1,69 @@
...
+func getKubeClientConfig(kubeConfig string) clientcmd.ClientConfig {
+	loadingRules := clientcmd.NewDefaultClientConfigLoadingRules()
+	loadingRules.ExplicitPath = kubeConfig
+
+	return clientcmd.NewNonInteractiveDeferredLoadingClientConfig(loadingRules, &clientcmd.ConfigOverrides{})
+}
+
+// GenerateTempKubeConfigStubFile go through the kubeconfig file and replace all users using the Exec auth method by
+// an invocation of the stub binary.
+func (k *KubeConfigPatcher) GenerateTempKubeConfigStubFile(kubeConfig string) (string, error) {
```
thallgren: What does the `KubeConfigPatcher` struct bring into the equation? Feels like this could be a plain function without a message receiver.

knlambert: As usual, testing :o) But I can simplify it yeah since I don't plan to test it considering it's low value.

thallgren (approved):

PR #3055, opened 2023-03-07
Use external kubernetes authenticator in container based daemon.

thallgren:
## Description

This PR incorporates the authentication method introduced in PR #3052 into the flow when connecting to a container based daemon using `telepresence connect --docker`.

New `telepresence kubeauth-foreground --portfile <path>` subcommand.
This command starts the gRPC service that performs the host-based authentication. The server is started on demand the first time a `telepresence connect --docker` uses a kubernetes context with a user configured with an `exec` type authentication. When started, it allocates a port and announces that port as  a two byte binary integer in the given port file. The server will keep on running until it is killed or until the portfile is removed, and while running, it will touch the port file every three seconds so that stale port files can be detected and removed.

The port is picked up by the CLI binary and the address `docker.host.internal:<port>` is injected into a stubbed kubeconfig, which is then mounted into the docker container that runs the daemon. This approach makes it unnecessary to use environment variables to pass and address and a port.

## Checklist

### DIFF
--- a/pkg/client/docker/kubeauth/cmd.go
+++ b/pkg/client/docker/kubeauth/cmd.go
@@ -1,0 +1,114 @@
+package kubeauth
+
+import (
+	"context"
+	"encoding/binary"
+	"net"
+	"os"
+	"time"
+
+	"github.com/spf13/cobra"
+	"google.golang.org/grpc"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+
+	"github.com/datawire/dlib/dhttp"
+	"github.com/datawire/dlib/dlog"
+	authGrpc "github.com/telepresenceio/telepresence/v2/pkg/authenticator/grpc"
+	"github.com/telepresenceio/telepresence/v2/pkg/client"
+	"github.com/telepresenceio/telepresence/v2/pkg/client/logging"
+	"github.com/telepresenceio/telepresence/v2/pkg/errcat"
+)
+
+const (
+	CommandName       = "kubeauth-foreground"
+	PortFileStaleTime = 3 * time.Second
+)
+
+type authService struct {
+	portFile  string
+	kubeFlags *genericclioptions.ConfigFlags
+}
+
+func Command() *cobra.Command {
+	as := authService{kubeFlags: genericclioptions.NewConfigFlags(false)}
+	c := &cobra.Command{
+		Use:    CommandName,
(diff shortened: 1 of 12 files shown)


shepz (approved):
LGTM, but @knlambert could you take a look since you've worked on that part

pkg/client/docker/kubeauth/cmd.go
```diff
@@ -0,0 +1,185 @@
...
+		select {
+		case <-ctx.Done():
+			return nil
+		case now = <-ticker.C:
+		}
+	}
+}
+
+func (as *authService) watchFiles(ctx context.Context) error {
+	// If any of the files that the current kubeconfig uses changes, then we die
```
knlambert:
```suggestion
	// If any of the files that the current kubeconfig uses change, then we die
```

pkg/client/docker/kubeauth/cmd.go
```diff
@@ -0,0 +1,185 @@
...
+	isOfInterest := func(s string, files []string) bool {
+		for _, file := range files {
+			if s == file {
+				return true
+			}
+		}
+		return false
+	}
+	for dir := range dirs {
+		// Can't watch things that doesn't exist. We want to know if files in there change though.
```
knlambert:
```suggestion
		// Can't watch things that don't exist. We want to know if files in there change though.
```

pkg/client/docker/kubeauth/cmd.go
```diff
@@ -0,0 +1,185 @@
...
+		}
+		select {
+		case <-ctx.Done():
+			return nil
+		case now = <-ticker.C:
+		}
+	}
+}
+
+func (as *authService) watchFiles(ctx context.Context) error {
```
knlambert, line 136: I'm not sure that I understand this, I assume it's to ensure that the kubeauth-foreground still uses the right kubeconfigs when we call a command ?

knlambert (approved):
nitpicks but LGTM

pkg/client/docker/kubeauth/cmd.go
thallgren: This is necessary because if the user adds, deletes, or modifies one of the files listed in the `KUBECONFIG` path, then a restart is necessary. A restart is also necessary if the client detects that the config has changed and removes the file.

PR #3056, opened 2023-03-08
Add the kube context and daemon address to the cache

josecv:
## Description

Add the kube context and daemon address to the cache

## Checklist

### DIFF
--- a/pkg/client/docker/daemon.go
+++ b/pkg/client/docker/daemon.go
@@ -378,8 +378,10 @@ func LaunchDaemon(ctx context.Context, name string) (conn *grpc.ClientConn, err 
 	case cid := <-cidFound: // Success, the daemon info file exists
 		err := cache.SaveDaemonInfo(ctx,
 			&cache.DaemonInfo{
-				Options:  map[string]string{"cid": cid},
-				InDocker: true,
+				Options:       map[string]string{"cid": cid},
+				InDocker:      true,
+				DaemonAddress: addr.String(),
+				KubeContext:   name,
 			}, cache.DaemonInfoFile(name, addr.Port))
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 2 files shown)


pkg/client/docker/daemon.go
```diff
@@ -378,8 +378,10 @@ func LaunchDaemon(ctx context.Context, name string) (conn *grpc.ClientConn, err
 	case cid := <-cidFound: // Success, the daemon info file exists
 		err := cache.SaveDaemonInfo(ctx,
 			&cache.DaemonInfo{
-				Options:  map[string]string{"cid": cid},
-				InDocker: true,
+				Options:       map[string]string{"cid": cid},
+				InDocker:      true,
+				DaemonAddress: addr.String(),
```
thallgren: I'd prefer if we just add the port here. The host differs depending on if you're on the host (127.0.0.1) or if you're in a container (DNS name of daemon).

thallgren (changes requested):

njayp (approved):

pkg/client/docker/daemon.go
josecv: understood; @njayp does that sound good to you? we should still be able to dial using only the port, yea?

thallgren (approved):

pkg/client/docker/daemon.go
njayp: it helps me to have the name of the `kubecontext`, it allows me to easily display the context the user is contected to, because `telepresence status --output json` does not do it anymore and I think Thomas has enough bugs to chase already

josecv: the kubecontext will still be here, it's just a question of whether the address includes the host or just the port

thallgren (changes requested):
Since this is used in the status output, the fields need a json annotation with snake case names.

pkg/client/docker/daemon.go
njayp: ah ok, sound good to me, ty

thallgren (approved):

PR #3057, opened 2023-03-08
Longer grace-period when expiring daemon caches

josecv:
## Description

This helps prevent issues with cache files being deleted prematurely when there's clock skew between the CLI and the docker container running the daemons

## Checklist

### DIFF
--- a/pkg/client/cache/daemons.go
+++ b/pkg/client/cache/daemons.go
@@ -10,6 +10,7 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/datawire/dlib/dlog"
 	"github.com/telepresenceio/telepresence/v2/pkg/filelocation"
 )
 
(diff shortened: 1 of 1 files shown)


njayp (approved):

pkg/client/cache/daemons.go
```diff
@@ -80,7 +81,8 @@ func daemonInfoFiles(ctx context.Context) ([]fs.DirEntry, error) {
 		if err != nil {
 			return nil, err
 		}
-		if fi.ModTime().Add(keepAliveInterval + 200*time.Millisecond).Before(time.Now()) {
+		if fi.ModTime().Add(keepAliveInterval + 600*time.Millisecond).Before(time.Now()) {
```
thallgren, line 84: I can see three reasons why a remove would be triggered here:

1. The file is really stale
2. There's a clock-skew
3. The daemon start-up is slow.
4. Clock-skew aside, this check is made just prior to when the daemon will update, but the daemon is busy and slowed down more than the added millisecond value

This fix will make the last case less likely, but I don't think it fixes the real problem.

My concern is that the problem we're facing is related to the second and third cases, because that would explain why one mac workstation got hit by this 100% (Nick's) and another (mine) never did. The skew on his Docker VM was worse than mine, and/or, his daemon takes longer to boot.

If I'm right, then this fix adjusts the tolerance by 400 ms, which might help in some cases.

Then it hit me, the daemon touches this file _after_ the session connects. That's bad. Connecting can take several seconds and perhaps it takes longer for Nick's workstation?

thallgren:
Superseded by #3059

PR #3061, opened 2023-03-09
Make telepresence helm xxx work with docker based daemon.

thallgren:
This commit ensures that the `telepresence helm xxx` commands work seamlessly with a container based daemon. The user can do:

    $ telepresence connect --docker
    $ telepresence helm xxx

or:

    $ telepresence helm --docker xxx

### DIFF
--- a/integration_test/helm_test.go
+++ b/integration_test/helm_test.go
@@ -51,10 +51,11 @@ func (s *helmSuite) Test_HelmCanInterceptInManagedNamespace() {
 
 func (s *helmSuite) Test_HelmCannotInterceptInUnmanagedNamespace() {
 	ctx := s.Context()
-	ctx = itest.WithUser(ctx, "default")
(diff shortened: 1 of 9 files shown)


knlambert (changes requested):
When I do an helm install but that the install fails, the daemon container stays up.

So when I restart the setup, I get : 

```cli
➜  telepresence git:(thallgren/helm-w-docker) ✗ telepresence helm install --docker
docker: Error response from daemon: Conflict. The container name "/klambert" is already in use by container "05ef6aeaea7130700829c01623403d413ca3292ab3ac65c2c54998bbd32df84a". You have to remove (or rename) that container to be able to reuse that name.
See 'docker run --help'.
telepresence helm install: error: daemon container exited with exit status 125
```

we probably miss a check instead of starting the daemon again.

knlambert (approved):

PR #3068, opened 2023-03-13
Refactor the cli. Get rid of cli/util.

thallgren:
This PR consists of a sequence of moves and renames, all under the pkg/client/cli and its subpackages. It's all about code hygiene and does not alter any logic. This part of the code has evolved more than two years to meet different needs. As a result, its structure has become more and more incomprehensive and spaghetti-like, and the naming of types and functions is inconsistent. The introduction of the `--docker` option was the last drop. Something had to be done.

There's no point in reviewing the PR as a whole. It must be done on a per-commit basis, or nothing makes much sense. Every significant move/rename is in a separate commit.

Also, package names like "util" [should be avoided](https://go.dev/blog/package-names).


josecv (approved):
love it.

PR #3084, opened 2023-03-21
Ensure that the docker container name is valid.

### LINKED ISSUE #14: launch of daemon container failed
When I try to connect to a cluster, I get this error:

```
launch of daemon container failed: docker: Error response from daemon: Invalid container name (tp-nick.santos@docker.com@nassau.us-east-1.eksctl.io), only [a-zA-Z0-9][a-zA-Z0-9_.-] are allowed. See 'docker run --help'.
```

Looks like it's trying to generate a container name based on the cluster context, but that won't work because they have different allowed characters?

thallgren:
## Description

The name used for `docker run --name <name>` must comply with the
regular expression `[a-zA-Z0-9][a-zA-Z0-9_.-]`. This is however not true
for the name of a Kubernetes context. So when using a context name as
the argument to `--name`, it must be mangled so that any illegal
characters are replaced.

This commit adds the `SafeContainerName` function which replaces all
illegal characters with an underscore unless it's the very first
character, in which case it is replaced with an 'a'.

## Checklist

### DIFF
--- a/pkg/client/docker/daemon.go
+++ b/pkg/client/docker/daemon.go
@@ -79,7 +79,7 @@ func DaemonOptions(ctx context.Context, name string) ([]string, *net.TCPAddr, er
 	addr := as[0]
 	port := addr.Port
 	opts := []string{
-		"--name", containerNamePrefix + name,
+		"--name", SafeContainerName(containerNamePrefix + name),
 		"--network", "telepresence",
 		"--cap-add", "NET_ADMIN",
 		"--device", "/dev/net/tun:/dev/net/tun",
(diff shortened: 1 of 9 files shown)


jacoblbeck (dismissed):

pkg/client/docker/daemon.go
```diff
@@ -100,11 +100,32 @@ func DaemonOptions(ctx context.Context, name string) ([]string, *net.TCPAddr, er
...
+	}
+	return n.String()
+}
+
 // DaemonArgs returns the arguments to pass to a docker run when starting a container daemon.
 func DaemonArgs(name string, port int) []string {
 	return []string{
 		"connector-foreground",
-		"--name", "docker-" + name,
+		"--name", SafeContainerName("docker-" + name),
```
knlambert, line 128: Is it the right place ? If I understand correctly, this is the name for the connector foreground, but we still have to use that saveContainerName function with the container name right ? `tp-something`

knlambert: In `DaemonOptions`

thallgren: Yes, I saw that too. It's added now.

jacoblbeck (approved):

PR #3116, opened 2023-04-03
Fix agent injection with huge workloads

### LINKED ISSUE #3025: Auto injection not working
**Describe the bug**

Telepresence auto-injection does not work when you are deploying a big set of deployments.

**To Reproduce**
1. When you enable traffic-agent auto-injection and you deploy a large number of deployments (+20)
2. I see that the auto-injection is not working
3. So I look at traffic manager
4. And I see the error

```
I0213 16:48:21.583342       1 request.go:682] Waited for 1.584687903s due to client-side throttling, not priority and fairness, request: GET:https://10.252.0.1:443/apis/apps/v1/namespaces/dev-tyler/deployments/auth-api 
```

**Expected behavior**

 - 2.9.5
 - MacOS
 - AWS EKS and local Kind

knlambert:
## Description

A user reported us that when there is a huge amount of pod being spinned up, the traffic manager can have trouble to inject the agent in all of them.

After investigation, this is due to the mutating webhook. It happens that it takes too much time to process the information, especially when it receives a lot of requests. Eventually, the 5s timeout is exceeded.

The reasons are multiple:
* Too many API calls (with throttle from the api when they're too many).
    * Some of them could be cached.
    * Those updating the configmap can provoke races conditions (`the object has been modified; please apply your changes to the latest version and try again`).
* Some mutex are maybe too restrictive and cause some traffic in the part of the code updating the map : 

This [line](https://github.com/telepresenceio/telepresence/blob/release/v2/cmd/traffic/cmd/manager/internal/mutator/watcher.go#L419) gets locked by this [one](https://github.com/telepresenceio/telepresence/blob/release/v2/cmd/traffic/cmd/manager/internal/mutator/watcher.go#L598).

## Checklist


cmd/traffic/cmd/manager/internal/mutator/watcher.go
```diff
@@ -389,66 +399,123 @@ func (c *configWatcher) Store(ctx context.Context, ac *agentconfig.Sidecar, upda
...
+	cu.Config[ac.AgentName] = yml
+
+	if updateSnapshot {
+		cu.AddToSnapshot[ac.AgentName] = struct{}{}
+	}
+
+	c.configUpdatersMapLock.Unlock()
+
+	if newGroup {
+		cu.Go(c.updateConfigMap(ctx, ns))
```
thallgren: Shouldn't this method get called using the errgroup's derived context, i.e. the context that is ignored at line 407 when the group is created?

cmd/traffic/cmd/manager/internal/mutator/watcher.go
```diff
@@ -389,66 +399,123 @@ func (c *configWatcher) Store(ctx context.Context, ac *agentconfig.Sidecar, upda
...
-		if !errors.IsNotFound(err) {
-			return fmt.Errorf("unable to get ConfigMap %s: %w", agentconfig.ConfigMap, err)
+	c.configUpdatersMapLock.Lock()
+	var cu *configUpdater
+
+	newGroup := false
+	if _, ok := c.configUpdaters[ns]; !ok {
+		grp, _ := errgroup.WithContext(ctx)
+		c.configUpdaters[ns] = &configUpdater{
+			grp,
```
thallgren: I would include the `ns` and `ctx` in the `configUpdater` struct and add them here. That way, the `updateConfigMap` no longer needs any arguments, which in turn means that it can be used directly as `cu.Go(c.updateConfigMap)` on line 426 instead of returning a no-args function. It would also address my comment about what context that gets passed on to that function.

cmd/traffic/cmd/manager/internal/mutator/watcher.go
```diff
@@ -389,66 +399,123 @@ func (c *configWatcher) Store(ctx context.Context, ac *agentconfig.Sidecar, upda
...
+	}
+
+	cu = c.configUpdaters[ns]
+	cu.Config[ac.AgentName] = yml
+
+	if updateSnapshot {
+		cu.AddToSnapshot[ac.AgentName] = struct{}{}
+	}
+
+	c.configUpdatersMapLock.Unlock()
```
thallgren: Why isn't this unlocked on line 418? Does this lock also protect the actual `cu`?

cmd/traffic/cmd/manager/internal/mutator/watcher.go
```diff
@@ -389,66 +399,123 @@ func (c *configWatcher) Store(ctx context.Context, ac *agentconfig.Sidecar, upda
...
-		if cm.Data[ac.AgentName] == yml {
-			// Race condition. Snapshot isn't updated yet, or we wouldn't have gotten here.
-			return nil
-		}
-	}
 
-	if updateSnapshot {
-		c.Lock()
+		// Can't add / remove things to the map at this point.
+		c.configUpdatersMapLock.Lock()
```
thallgren: So different namespaces cannot be updated concurrently? That strikes me as perhaps a bit too restrictive. What motivates this limitation?

cmd/traffic/cmd/manager/internal/mutator/watcher.go
```diff
@@ -389,66 +399,123 @@ func (c *configWatcher) Store(ctx context.Context, ac *agentconfig.Sidecar, upda
...
-			Data: map[string]string{
-				ac.AgentName: yml,
-			},
-		}
-		dlog.Debugf(ctx, "Creating new ConfigMap %s.%s with %s", agentconfig.ConfigMap, ns, ac.AgentName)
-		_, err = api.Create(ctx, cm, meta.CreateOptions{})
-	} else {
-		if _, ok := cm.Data[ac.AgentName]; ok {
-			dlog.Debugf(ctx, "Updating %s in ConfigMap %s.%s", ac.AgentName, agentconfig.ConfigMap, ns)
+			cmData = copyMap(nm)
```
thallgren:
```suggestion
			cmData = maps.Copy(nm)
```

cmd/traffic/cmd/manager/internal/mutator/watcher.go
```diff
@@ -389,66 +399,123 @@ func (c *configWatcher) Store(ctx context.Context, ac *agentconfig.Sidecar, upda
...
-	return err
+}
+
+func copyMap(m map[string]string) map[string]string {
+	mapCopy := make(map[string]string, len(m))
+	for k, v := range m {
+		mapCopy[k] = v
+	}
+	return mapCopy
 }
```
thallgren: Remove this in favor of [maps](https://github.com/telepresenceio/telepresence/tree/release/v2/pkg/maps).

cmd/traffic/cmd/manager/internal/mutator/watcher.go
```diff
@@ -230,6 +231,15 @@ type configWatcher struct {
 	data       map[string]map[string]string
 	modCh      chan entry
 	delCh      chan entry
+
+	configUpdatersMapLock sync.RWMutex
```
thallgren: My preference would be to name this `configUpdatersLock` to correspond with the attribute name of what it lockes. Mixing in the type in that attribute in the name is just confusing.

thallgren (changes requested):
I like the restructuring of the config map update, but I'm curious about the caching. Are we doing a lot of repeated get on the same resource? If so, is there a way to avoid that without using a generic cache? I'd like us to avoid patching what might be a design flaw.

cmd/traffic/cmd/manager/internal/mutator/watcher.go
knlambert: It kind of protect cu and its maps. If I don't put it there, when I run my tests with the `-race` flag, I get an error since the cu.AddToSnapshot map has a risk of being written by two routines at the same time.

With:

```go
	cu = c.configUpdaters[ns]
	cu.Config[ac.AgentName] = yml

	c.configUpdatersLock.Unlock()

	if updateSnapshot {
		cu.addToSnapshot[ac.AgentName] = struct{}{}
	}

	if newGroup {
		cu.Go(cu.updateConfigMap)
	}
```
I get

```cli 
==================
--- FAIL: TestSuiteConfigWatcher (0.06s)
    --- FAIL: TestSuiteConfigWatcher/TestStoreErrGroupBuffer (0.05s)
        testing.go:1319: race detected during execution of test
    testing.go:1319: race detected during execution of test
FAIL

```

cmd/traffic/cmd/manager/internal/mutator/watcher.go
knlambert: mmh let me check that again, not sure why I still need that

cmd/traffic/cmd/manager/internal/mutator/watcher.go
```diff
@@ -621,12 +683,25 @@ func (c *configWatcher) affectedConfigs(ctx context.Context, svc *core.Service,
 			wls = append(wls, stss...)
 		}
 	}
+
+	c.RLock()
```
knlambert, line 716: @thallgren do you have an opinion on this change ? It was the main bottle neck but I'm not sure if it can have an impact to reduce the scope of this lock like I did.

thallgren: I think moving the lock is a good improvement. There's no reason for having it while  querying for the workloads.

knlambert: Great :)

cmd/traffic/cmd/manager/internal/mutator/watcher.go
knlambert: It was preventing two updates to happen at the same time in some rare cases. I've added a lock which is scoped to the namespace with my last commit.

knlambert:
> I like the restructuring of the config map update, but I'm curious about the caching. Are we doing a lot of repeated get on the same resource? If so, is there a way to avoid that without using a generic cache? I'd like us to avoid patching what might be a design flaw.

This part is a recursive code so it was easier this way

```cli
2023-04-04 21:09:00.6607 info    agent-injector/conn=10.42.0.175:443 : cache echo-easy-79df8d46f5-ambassador-ReplicaSet miss
2023-04-04 21:09:01.2608 info    agent-injector/conn=10.42.0.175:443 : cache echo-easy-ambassador-Deployment miss
2023-04-04 21:09:01.2608 info    agent-injector/conn=10.42.0.175:443 : cache echo-easy-79df8d46f5-ambassador-ReplicaSet hit
2023-04-04 21:09:01.2608 info    agent-injector/conn=10.42.0.175:443 : cache echo-easy-ambassador-Deployment hit
```

I can take a look at this though since the resources seems to always be the same.

cmd/traffic/cmd/manager/internal/mutator/watcher_test.go
```diff
@@ -0,0 +1,214 @@
...
+//go:generate go run github.com/golang/mock/mockgen -package=mock_kubernetes -destination=mocks/k8s_corev1_mock.go k8s.io/client-go/kubernetes/typed/core/v1 CoreV1Interface
+//go:generate go run github.com/golang/mock/mockgen -package=mock_kubernetes -destination=mocks/k8s_configmap_mock.go k8s.io/client-go/kubernetes/typed/core/v1 ConfigMapInterface
+type suiteConfigWatcher struct {
+	suite.Suite
+
+	ctrl *gomock.Controller
+
+	kubeApiMock      *mock_kubernetes.MockInterface
+	coreV1ApiMock    *mock_kubernetes.MockCoreV1Interface
+	configMapApiMock *mock_kubernetes.MockConfigMapInterface
```
thallgren, line 32: Why aren't we using `k8s.io/client-go/kubernetes/fake` here like we do in the `agent_injector_tests`? Seems to me like that would put things more to the test than a simple mock.

cmd/traffic/cmd/manager/internal/mutator/watcher.go
```diff
@@ -431,23 +497,59 @@ func (c *configWatcher) Store(ctx context.Context, ac *agentconfig.Sidecar, upda
...
-	} else {
-		if _, ok := cm.Data[ac.AgentName]; ok {
-			dlog.Debugf(ctx, "Updating %s in ConfigMap %s.%s", ac.AgentName, agentconfig.ConfigMap, ns)
-		} else {
-			dlog.Debugf(ctx, "Adding %s to ConfigMap %s.%s", ac.AgentName, agentconfig.ConfigMap, ns)
+	}
+
+	if cm.Data == nil {
+		cm.Data = make(map[string]string)
+	}
```
thallgren: The extra initialization is unnecessary because `cm.Data` is only read until it is unconditionally assigned att lines 544 and 549. It's OK to read from a `nil` map.

cmd/traffic/cmd/manager/internal/mutator/watcher.go
```diff
@@ -431,23 +497,59 @@ func (c *configWatcher) Store(ctx context.Context, ac *agentconfig.Sidecar, upda
...
+				nm[agentName] = yml
+				c.cw.Unlock()
+			}
 		}
-		cm.Data[ac.AgentName] = yml
-		_, err = api.Update(ctx, cm, meta.UpdateOptions{})
 	}
+
+	if create {
+		cm.Data = cmData
```
thallgren: Move this assignment out of the `if` statement and remove it from line 549.

cmd/traffic/cmd/manager/internal/mutator/watcher.go
```diff
@@ -431,23 +497,59 @@ func (c *configWatcher) Store(ctx context.Context, ac *agentconfig.Sidecar, upda
...
-			dlog.Debugf(ctx, "Updating %s in ConfigMap %s.%s", ac.AgentName, agentconfig.ConfigMap, ns)
-		} else {
-			dlog.Debugf(ctx, "Adding %s to ConfigMap %s.%s", ac.AgentName, agentconfig.ConfigMap, ns)
+	}
+
+	if cm.Data == nil {
+		cm.Data = make(map[string]string)
+	}
+
+	cmData := map[string]string{}
```
thallgren:
```suggestion
	var cmData map[string]string
```
Nit for sure, but I think it brings clarity. Initialization is unnecessary. If there is a `c.cw.data[c.namespace]`, then `cmData` is assigned to that value, otherwise a `nil`, value is OK.

thallgren (approved):
Just some very minor nits and a question.

cmd/traffic/cmd/manager/internal/mutator/watcher_test.go
knlambert: Didn't know about this one ! Let me take a look.

knlambert: I was migrating to the fake client, but I noticed that it doesn't allow me to emulate a slow API for example, so I can't migrate my last test :(

PR #3126, opened 2023-04-06
Extract routing logic to a dedicated Router

josecv:
## Description

## Checklist


pkg/vif/device/device.go
```diff
@@ -1,4 +1,4 @@
-package vif
+package device
```
thallgren: I liked the name `vif.Device` a lot more than `device.Device` because it's clear and concise. `device.Device` stutter and says nothing about the type of the device.

I realize that the current package structure under `vif` is awkward. Static routing and the `routing` package isn't related to the `vif` for instance. But introducing stuttering like `device.Device` or `netstack.Stack` doesn't improve the situation. I also feel that the stack is a very integral part of the VIF so keeping it adjacent to the device is a good thing.

Suggestion:
Move the routing up directly under `pkg` to improve separation of concern. Keep your current split of the stack.go so that logging is a file of its own, but let them stay in the `vif` package. That way, the `vif` package is only the `Device`, the `Stack`, and the private `dlogEmitter` for the `Stack`. The stutter is gone, and the names are more descriptive.

This will also allow for the `InitLogger` to be unexported and called just before the stack is created in `NewTunnelVIF`, driving the encapsulation a step further.

pkg/vif/tunnel_vif.go
```diff
@@ -0,0 +1,41 @@
...
+	"context"
+
+	"gvisor.dev/gvisor/pkg/tcpip/stack"
+
+	"github.com/telepresenceio/telepresence/v2/pkg/tunnel"
+	"github.com/telepresenceio/telepresence/v2/pkg/vif/device"
+	"github.com/telepresenceio/telepresence/v2/pkg/vif/netstack"
+)
+
+type TunnelVIF struct {
```
thallgren: The name `vif.TunnelVIF` repeats the term `vif`. Suggest using `vif.Tunnel` or even `vif.DeviceWithStack`.

pkg/vif/routing/routing_linux.go
```diff
@@ -14,7 +14,7 @@ import (
 	"github.com/telepresenceio/telepresence/v2/pkg/iputil"
 )
 
-const findInterfaceRegex = `^(local\s)?[0-9.]+(\s+via\s+(?P<gw>[0-9.]+))?\s+dev\s+(?P<dev>[a-z0-9-]+)\s+src\s+(?P<src>[0-9.]+)`
+const findInterfaceRegex = `^(local\s)?[0-9.]+(\s+via\s+(?P<gw>[0-9.]+))?\s+dev\s+(?P<dev>[a-z0-9-]+)\s+(table [0-9]+\s+)?src\s+(?P<src>[0-9.]+)`
```
thallgren: My workstation have routes with `table local`, so `[0-9]` is too narrow. Example:
```
local 192.168.50.10 dev lo table local src 192.168.50.10 uid 0
```

pkg/client/rootd/session.go
```diff
@@ -859,20 +742,10 @@ func (s *Session) stop(c context.Context) {
...
-		err := np.RemoveStatic(cc)
-		if err != nil {
-			dlog.Warnf(c, "error removing route %s: %v", np, err)
-		}
-	}
-	if s.dev != nil {
-		if err := s.dev.Close(); err != nil {
-			dlog.Errorf(c, "unable to close %s: %v", s.dev.Name(), err)
+	if s.tunVif != nil {
+		cc = dcontext.WithoutCancel(c)
```
thallgren:
```suggestion
		cc, cancel := context.WithTimeout(dcontext.WithoutCancel(c), 1*time.Second)
		defer cancel()
```
to ensure that this never hangs.

thallgren (approved):

PR #3128, opened 2023-04-09
Ensure proper DNS function when no cluster CIDR is routed by the VIF.

thallgren:
## Description

Telepresence does not route cluster subnets that are already available on the workstation, which is good, but the DNS malfunctions on Linux and Windows unless there is a subnet covering the IP of the Telepresence DNS server, which is bad.

This PR ensures that a subnet is routed by the VIF even when none of the cluster's subnets are routed. In such situations, a tiny subnet is created with a 30-bit mask. It is guaranteed not to collide with any of the subnets used by existing network interfaces. The DNS server will then use an IP in that subnet.

On darwin (macOS), no special DNS subnet is needed because the DNS is managed by adding and removing entries under /etc/resolver that points directly to a port on localhost. So when no subnets are routed, there's actually no need for a VIF. This commit takes this into account, so that no VIF is started unless there's a need for it.

## Checklist

### DIFF
--- a/pkg/client/rootd/dns/server_darwin.go
+++ b/pkg/client/rootd/dns/server_darwin.go
@@ -130,7 +130,7 @@ func (r *resolveFile) setSearchPaths(paths ...string) {
 //	man 5 resolver
 //
 // or, if not on a Mac, follow this link: https://www.manpagez.com/man/5/resolver/
-func (s *Server) Worker(c context.Context, dev vif.Device, proxyCluster bool, configureDNS func(net.IP, *net.UDPAddr)) error {
+func (s *Server) Worker(c context.Context, dev vif.Device, configureDNS func(net.IP, *net.UDPAddr)) error {
 	resolverDirName := filepath.Join("/etc", "resolver")
 	resolverFileName := filepath.Join(resolverDirName, "telepresence.local")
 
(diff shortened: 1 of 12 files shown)


pkg/client/rootd/dns/server_darwin.go
```diff
@@ -188,7 +188,7 @@ func (s *Server) Worker(c context.Context, dev vif.Device, proxyCluster bool, co
 			return s.updateResolverFiles(c, resolverDirName, resolverFileName, dnsAddr, paths)
 		}, dev)
 		// Server will close the listener, so no need to close it here.
-		return s.Run(c, make(chan struct{}), []net.PacketConn{listener}, nil, s.resolveInCluster, proxyCluster)
+		return s.Run(c, make(chan struct{}), []net.PacketConn{listener}, nil, s.resolveInCluster)
```
josecv, line 191: i'm so happy you got rid of that proxyCluster parameter, i've tried to once or twice

pkg/client/rootd/session.go
```diff
@@ -533,6 +541,33 @@ func (s *Session) watchClusterInfo(ctx context.Context) {
...
+func (s *Session) createSubnetForDNSOnly(ctx context.Context, mgrInfo *manager.ClusterInfo) {
+	// Avoid alsoProxied and neverProxied
+	avoid := make([]*net.IPNet, 0, len(s.alsoProxySubnets)+len(s.neverProxyRoutes))
+	avoid = append(avoid, s.alsoProxySubnets...)
+	for _, r := range s.neverProxyRoutes {
+		avoid = append(avoid, r.RoutedNet)
+	}
+
+	// Avoid the service subnet. It might be mapped with iptables (if running bare-metal) and
+	// hence invisible when listing known routes.
```
josecv, line 555: great info

josecv (approved):

PR #3130, opened 2023-04-12
Make tele intercept --docker-run play well with tele connect --docker

thallgren:
## Description

A lot of improvements were made in how we run the Telepresence daemon in a Docker container when the [Telepresence Intercept Specification](https://www.getambassador.io/docs/telepresence/latest/reference/intercepts/specs#configuring-intercept-using-specifications) was introduced. A major one was that an intercept handler that runs in docker, now can talk to the daemon without using the host network. Another that volumes could be mounted using the [Telemount plugin](https://github.com/datawire/docker-volume-telemount), removing the need for FUSE software like macFUSE and Winfsp. We didn't touch the `telepresence intercept --docker-run` flag at that time though, which lead to weird results when combining it with a daemon started with `telepresence connect --docker`. This PR rectifies that.

If the daemon is started in a container with `telepresence connect --docker` (or implicitly by passing `--docker` directly to `telepresence intercept`), and the `--docker-run` is passed to the intercept command, then Telepresence will automatically connect the dots between the intercept handler and the daemon container. It will also automatically mount the remote filesystems that are made available by the traffic-agents SFTP server.

Two new flags are also added in this PR. The `--docker-build <docker context>` and the repeatable `docker-build-opt key=value`, to enable container's to be build on the fly by the intercept command. The `--docker-build` implies `--docker-run`.

The PR also contains some other improvements (details in the commits).

## Checklist


DEPENDENCIES.md
```diff
@@ -30,7 +30,7 @@ following Free and Open Source software:
     github.com/davecgh/go-spew                                                   v1.1.1                                ISC license
     github.com/docker/cli                                                        v20.10.22+incompatible                Apache License 2.0
     github.com/docker/distribution                                               v2.8.1+incompatible                   Apache License 2.0
-    github.com/docker/docker                                                     v20.10.22+incompatible                Apache License 2.0
+    github.com/docker/docker                                                     v20.10.24+incompatible                Apache License 2.0
```
knlambert, line 33: I think you forgot to update the changelog

pkg/client/cli/intercept/docker_run.go
```diff
@@ -0,0 +1,218 @@
...
+		return docker.PullImage(ctx, imageName)
+	}
+
+	// DockerBuild will produce an image-ID that must be injected into the docker run argument
+	// list. It must be injected between arguments intended for docker run and arguments intended
+	// for the container, so we require that a placeholder is present. E.g.
+	//
+	// telepresence intercept hello --docker-build ./some/path -- -it --add-host foo IMAGE --port 8080
+	if (idx < 0 || imageName != "IMAGE") && len(s.Cmdline) > 0 {
+		return errcat.User.New(`` +
```
knlambert, line 38: Empty `` ?

knlambert (changes requested):

pkg/client/cli/intercept/docker_run.go
thallgren: Yeah, because otherwise the formatting of the lines that follow gets messed up.

knlambert (approved):

PR #3140, opened 2023-04-19
Replace deprecated info in the JSON output of the telepresence list command

### LINKED ISSUE #3092: Replace deprecated info in the JSON output of the telepresence list command.
The JSON output of the `telepresence list` command contains information from structures that have since been deprecated. A workload can have multiple services and multiple ports and that is not reflected in a good way with the current info.

The following should happen in order to fix this:
1. Today, the session function [workloadInfoSnapshot()](/telepresenceio/telepresence/blob/release/v2/pkg/client/userd/trafficmgr/session.go#L730) calls the deprecated function `getCurrentAgentsInNamespace(ns)` to obtain a list of `AgentInfo` structs. That struct is outdated and should not be used. Instead, we must retrieve the configmap `telepresence-agents` in the namespace, and use the entries found there. See [this code](/telepresenceio/telepresence/blob/release/v2/pkg/client/cli/cmd/genyaml.go#L114-L131) for an example on how to read this configmap and parse its map of `Sidecar`. The `Sidecar` is what replaces the `AgentInfo`.
2. The [WorkloadInfo](/telepresenceio/telepresence/blob/release/v2/rpc/connector/connector.proto#L290-L291) message must change so that it contains a `Sidecar` instead of an `AgentInfo`. This will also render the `ServiceReference` obsolete, so it can be removed. Please note that there is no rpc message for `Sidecar`, so the easiest way to pass it is probably as `bytes` containing a JSON blob (similar to what we do with the `ClientConfig`).
3. The [Client Rbac](/telepresenceio/telepresence/blob/release/v2/charts/telepresence/templates/_helpers.tpl#L72) must include a rule to allow a `get` on the `telepresence-agents` configmap.
4. Necessary updates in the list command output.

shepz:
## Description
Update the JSON output of `telepresence list` :
- Add Sidecar information
- Remove AgentConfig
- ServiceReference is now a map (easier to validate if already present)
- Remove legacyUninstall and code associated

I wasn't sure how the output should look like at the end. Here is a example with `echo` intercepted

```json
{
  "cmd": "list",
  "stdout": [
    {
      "name": "dataprocessingservice",
      "namespace": "default",
      "workload_resource_type": "Deployment",
      "services": {
        "1a22362f-5708-4abc-90ae-58e002980acd": {
          "name": "dataprocessingservice",
          "namespace": "default",
          "ports": [
            {
              "port": 3000
            }
          ]
        }
      },
      "uid": "49c9b342-63cc-4864-b5d4-fce6460b4750"
    },
    {
      "name": "echo",
      "namespace": "default",
      "intercept_infos": [...],
      "workload_resource_type": "Deployment",
      "services": {
        "7eaa2974-8c9a-49c4-8333-bee89e1cdbfc": {
          "name": "echo-headless",
          "namespace": "default",
          "ports": [
            {
              "port": 80
            }
          ]
        },
        "f5cfbd68-3c39-4897-8272-3509e7bf423d": {
          "name": "echo",
          "namespace": "default",
          "ports": [
            {
              "port": 80
            }
          ]
        }
      },
      "uid": "c9b4d403-a1b9-420f-af69-b456dab2dcf4",
      "sidecar": {
        "agentImage": "docker.io/datawire/ambassador-telepresence-agent:1.13.13",
        "agentName": "echo",
        "namespace": "default",
        "logLevel": "info",
        "workloadName": "echo",
        "workloadKind": "Deployment",
        "managerHost": "traffic-manager.ambassador",
        "managerPort": 8081,
        "tracingPort": 15766,
        "envoyLogLevel": "warning",
        "envoyServerPort": 18000,
        "envoyAdminPort": 19000,
        "containers": [
          {
            "name": "echo-server",
            "intercepts": [
              {
                "serviceName": "echo-headless",
                "serviceUID": "7eaa2974-8c9a-49c4-8333-bee89e1cdbfc",
                "targetPortNumeric": true,
                "protocol": "TCP",
                "containerPort": 8080,
                "servicePort": 80,
                "agentPort": 9900
              },
              {
                "serviceName": "echo",
                "serviceUID": "f5cfbd68-3c39-4897-8272-3509e7bf423d",
                "targetPortNumeric": true,
                "protocol": "TCP",
                "containerPort": 8080,
                "servicePort": 80,
                "agentPort": 9900
              }
            ],
            "envPrefix": "A_",
            "mountPoint": "/tel_app_mounts/echo-server",
            "Mounts": null
          }
        ]
      }
    },
    {
      "name": "verylargedatastore",
      "namespace": "default",
      "workload_resource_type": "Deployment",
      "services": {
        "fc24b1f1-e738-427b-aec8-8bee60d3dc9d": {
          "name": "verylargedatastore",
          "namespace": "default",
          "ports": [
            {
              "port": 8080
            }
          ]
        }
      },
      "uid": "f2c436d0-d346-478e-b79f-b2f7e3d15b05"
    },
    {
      "name": "verylargejavaservice",
      "namespace": "default",
      "workload_resource_type": "Deployment",
      "services": {
        "b10f947b-c28b-4cf1-a703-0f4bb65b1348": {
          "name": "verylargejavaservice",
          "namespace": "default",
          "ports": [
            {
              "port": 8080
            }
          ]
        }
      },
      "uid": "214ac1cf-79bd-4028-8462-6c4a1b480955"
    }
  ]
}

```

## Checklist


rpc/connector/connector.proto
```diff
@@ -291,28 +291,18 @@ message WorkloadInfo {
   // Reason why workload cannot be intercepted, or empty if it can.
   string not_interceptable_reason = 2;
 
-  // AgentInfo reported from the traffic manager in case the workload has a traffic-agent installed
-  telepresence.manager.AgentInfo agent_info = 3;
+  message Sidecar {
+    bytes json = 1;
+  }
+  // Sidecar
+  Sidecar sidecar = 3;
```
knlambert: is it an endpoint which may be requested by let's say system-a ? If yes, it's important to have a new number.

https://stackoverflow.com/questions/26826421/protocol-buffers-unique-numbered-tag-clarification

If it's not in use, that's fine :)

pkg/client/cli/cmd/list.go
```diff
@@ -167,7 +174,21 @@ func (s *listCommand) printList(ctx context.Context, workloads []*connector.Work
 	}
 
 	if formattedOut {
-		output.Object(ctx, workloads, false)
+		var o []*workloadJsonOutput
+		for _, v := range workloads {
+			var l workloadJsonOutput
```
knlambert:
```suggestion
			l := workloadJSONOutput{WorkloadInfo: v}
```

pkg/client/userd/trafficmgr/agents.go
```diff
@@ -51,6 +58,30 @@ func (s *session) getCurrentAgentsInNamespace(ns string) map[string]*manager.Age
...
+	// Load configmap entry from the telepresence-agents configmap
+	cm, err := k8sapi.GetK8sInterface(ctx).CoreV1().ConfigMaps(ns).Get(ctx, agentconfig.ConfigMap, meta.GetOptions{})
+	if err != nil {
+		if !k8sErrors.IsNotFound(err) {
+			dlog.Error(ctx, errcat.User.New(err))
+		}
+		return sidecars
+	}
+
+	for workload, sidecar := range cm.Data {
```
knlambert, line 62: Can data be nil ?

pkg/client/userd/trafficmgr/agents.go
```diff
@@ -51,6 +58,30 @@ func (s *session) getCurrentAgentsInNamespace(ns string) map[string]*manager.Age
...
+		var cfg agentconfig.Sidecar
+		if err = yaml.Unmarshal([]byte(sidecar), &cfg); err != nil {
+			dlog.Errorf(ctx, "Unable to parse entry for %q in configmap %q: %v", workload, agentconfig.ConfigMap, err)
+			return sidecars
```
knlambert: Should it return right away ? are we done going through the list of sidecars ?

pkg/client/userd/trafficmgr/session.go
```diff
@@ -739,13 +727,19 @@ nextIs:
...
+	sMap := make(map[string]*rpc.WorkloadInfo_Sidecar)
 	for _, ns := range nss {
-		for k, v := range s.getCurrentAgentsInNamespace(ns) {
-			aMap[k] = v
+		for k, v := range s.getCurrentSidecarsInNamespace(ctx, ns) {
+			data, err := json.Marshal(v)
+			if err != nil {
+				continue
+			}
+			sMap[k] = &rpc.WorkloadInfo_Sidecar{Json: data}
```
knlambert, line 818: `JSON` should be in uppercase

knlambert (changes requested):

pkg/client/cli/cmd/list.go
```diff
@@ -167,7 +174,21 @@ func (s *listCommand) printList(ctx context.Context, workloads []*connector.Work
 	}
 
 	if formattedOut {
-		output.Object(ctx, workloads, false)
+		var o []*workloadJsonOutput
```
thallgren:
```suggestion
		o := make([]*workloadJsonOutput, len(workloads))
```

pkg/client/cli/cmd/list.go
```diff
@@ -167,7 +174,21 @@ func (s *listCommand) printList(ctx context.Context, workloads []*connector.Work
...
+		for _, v := range workloads {
```
thallgren:
```suggestion
		for i, v := range workloads {
```

pkg/client/cli/cmd/list.go
```diff
@@ -167,7 +174,21 @@ func (s *listCommand) printList(ctx context.Context, workloads []*connector.Work
...
+			var l workloadJsonOutput
+			l.WorkloadInfo = v
+
+			if v.Sidecar != nil {
+				var sidecar agentconfig.Sidecar
+				_ = json.Unmarshal(v.Sidecar.Json, &sidecar)
+				l.Sidecar = &sidecar
+			}
+
+			o = append(o, &l)
```
thallgren:
```suggestion
			o[i] = &l
```

thallgren (changes requested):
This looks good to me aside from my nit about slice initialization. I would like you to add (or rather remove) one more thing. I think the `getCurrentAgents()` and everything associated with it can be removed now, which means that we no longer need the agent listener. That'd be a big win.

pkg/client/userd/trafficmgr/agents.go
```diff
@@ -51,6 +58,30 @@ func (s *session) getCurrentAgentsInNamespace(ns string) map[string]*manager.Age
 	return agents
 }
 
+func (s *session) getCurrentSidecarsInNamespace(ctx context.Context, ns string) map[string]*agentconfig.Sidecar {
+	sidecars := make(map[string]*agentconfig.Sidecar)
```
thallgren: Place this line after successful retrieval of `cm`. Then preallocate so that the map doesn't need to grow.
```suggestion
	sidecars := make(map[string]*agentconfig.Sidecar, len(cm.Data))
```

pkg/client/userd/trafficmgr/session.go
shepz: the linting prevent to use uppercase : ```Field name "JSON" must be underscore_separated_names like "json"```

pkg/client/userd/trafficmgr/agents.go
shepz: yes, but it does NOT panic the cli. I added a debug log

shepz:
> This looks good to me aside from my nit about slice initialization. I would like you to add (or rather remove) one more thing. I think the `getCurrentAgents()` and everything associated with it can be removed now, which means that we no longer need the agent listener. That'd be a big win.

The `getCurrentAgents()` function is exclusively employed for the legacyUninstall process. To clarify, should we consider removing this function as well?

pkg/client/userd/trafficmgr/session.go
```diff
@@ -566,34 +567,37 @@ func (s *session) getInfosForWorkloads(
 		}
 		for _, workload := range wls {
 			if _, ok := wiMap[workload.GetUID()]; ok {
+				if _, ok := wiMap[workload.GetUID()].Services[string(svc.UID)]; !ok {
+					wiMap[workload.GetUID()].Services[string(svc.UID)] = &rpc.WorkloadInfo_ServiceReference{
```
thallgren: Why do three hash lookups of the same value?
```suggestion
			if wlInfo, ok := wiMap[workload.GetUID()]; ok {
				if _, ok := wlInfo.Services[string(svc.UID)]; !ok {
					wlInfo.Services[string(svc.UID)] = &rpc.WorkloadInfo_ServiceReference{
```

thallgren (changes requested):

thallgren (approved):

thallgren:
Yes, please remove the legacy uninstall as well. If users want to uninstall versions that are older than 2.6.0, then they can do so using an older version of the client. I don't see a problem with that.

thallgren (changes requested):
Still needs the aforementioned deletions.

CHANGELOG.yml
```diff
@@ -37,6 +37,10 @@ items:
         title: Add ASLR to telepresence binaries
         body: >-
           ASLR hardens binary sercurity against fixed memory attacks.
+      - type: bugfix
```
thallgren, line 50: I guess this'll go under 2.15.0 now?

thallgren (approved):

knlambert (approved):

CHANGELOG.yml
shepz: yes

PR #3148, opened 2023-04-24
Added timeouts config `agentArrival`

xzhang-brex:
## Description

Currently, the time to wait for the agent to arrive is hardcoded 30 seconds, however, it is not sufficient for some slow pods. Here we add `agentArrival` to the client config to tell the traffic manager how long it should wait for the agent to arrive.

## Checklist


CHANGELOG.md
```diff
@@ -10,7 +16,7 @@
 - Feature: There's a new --address flag to the intercept command allowing users to set the target IP of the intercept.
 
 - Feature: The new flags `--docker-build`, and `--docker-build-opt` was added to `telepresence intercept` to facilitate a
-  docker run directly from a docker context.
```
xzhang-brex: typo fix that is unrelated to the subject of this PR.

pkg/client/config.go
```diff
@@ -142,7 +144,7 @@ type Timeouts struct {
 	PrivateRoundtripLatency time.Duration `json:"roundtripLatency,omitempty" yaml:"roundtripLatency,omitempty"`
 	// PrivateProxyDial is how long to wait for the proxy to establish an outbound connection
 	PrivateProxyDial time.Duration `json:"proxyDial,omitempty" yaml:"proxyDial,omitempty"`
-	// PrivateTrafficManagerConnect is how long to wait for the traffic-manager API to connect
```
xzhang-brex: typo fix that is unrelated to the subject of this PR.

thallgren (changes requested):
Thanks @xzhang-brex for this PR. It's well made. I do however think that the config ended up in the wrong place, which is my fault. I didn't have time to go through this with you in depth.

The `config.yml` file and it's settings are intended for the client. The traffic-manager must be configured using the Helm chart, and values in charts/telepresence/values.yaml. A value like this would typically be configured there and then passed in via an environment variable setting. The [InterceptDisableGlobal](https://github.com/telepresenceio/telepresence/blob/release/v2/cmd/traffic/cmd/manager/managerutil/envconfig.go#L43) can serve as a good example. It is passed in as the [INTERCEPT_DISABLE_GLOBAL](https://github.com/telepresenceio/telepresence/blob/release/v2/charts/telepresence/templates/deployment.yaml#L80) env, and originally set [here](https://github.com/telepresenceio/telepresence/blob/release/v2/charts/telepresence/templates/deployment.yaml#L80). I think the `agentArrivalTimeout` value should be placed adjacent to the `disableGlobal` (and hence become `intecept.agentArrivalTimeout`.

Again, sorry for not providing enough info around this.

xzhang-brex:
@thallgren Thank you for the feedback! I've addressed the comment!

I added `timeouts` group to be consistent with the client end config and for future timeouts-related config. Feel free to remove it and move agentArrivalTimeout to `intecept` group if you want.

thallgren (approved):
Perfect! Thanks.

charts/telepresence/values.yaml
```diff
@@ -171,6 +171,11 @@ managerRbac:
 intercept:
   disableGlobal: false
 
+timeouts:
+  # The duration the traffic manager should wait for an agent to arrive (i.e., to be registered in the traffic manager's state)
+  # Default: 30s
+  agentArrival: 30s
```
thallgren, line 177: This should be added to the README.md in this folder that contains a description of all values.

PR #3161, opened 2023-05-05
fix(chart): specify port in hook urls

maximee-leroyy:
## Description

Specify port in hook URLs when we want to use a different port with `agentInjector.webhook.port`

## Checklist


thallgren (changes requested):
We have a `.Values.agentInjector.webhook.port` setting. Shouldn't that be used here instead of introducing a new value?

maximee-leroyy:
> We have a `.Values.agentInjector.webhook.port` setting. Shouldn't that be used here instead of introducing a new value?

Ah yes my bad i meant to use this value, it should be good now!

thallgren (approved):

thallgren:
Thanks @maxime1907 for your contribution!

PR #3172, opened 2023-05-09
Add mapping ability to the Telepresence DNS resolver

### LINKED ISSUE #3162: Add mapping ability to the Telepresence DNS resolver
In some scenarios, it is desirable to map the hostname that an application uses locally to a service with a different name in a cluster. Consider the following use-case.

A user wants to use an application which is dependent on a "postgres" service. The current setup is Docker Compose and in there, this service is named "postgres" and hence, also reachable through DNS by that name. Now the user wants this container to act as an intercept handler for a service running in a cluster. In this cluster, the "postgres" service resides in another namespace "big-data" and is named "psql".

The Telepresence DNS server is configured using a [kubeconfig extension](https://github.com/telepresenceio/telepresence/blob/release/v2/pkg/client/k8s_config.go#L28-L49), or by using the Helm chart [client structure](https://github.com/telepresenceio/telepresence/blob/release/v2/charts/telepresence/values.yaml#L343-L348). We should add a mappings element adjacent to the `includeSuffixes` and `excludeSuffixes` that will allow us to map hostnames to aliases, such that the user in the example can add:

```yaml
  dns:
    mappings:
      - name: postgres
        aliasFor: psql.big-data
```

The DNS-server should handle this as a CNAME record.

Another use-case is the opposite use-case where the user really wants to avoid that the intercept handler application uses a service in the cluster. Instead, any reference to a dependent service should go to one that runs locally. Let's assume that in the above example, the application also uses a "redis" service. We could prevent that Telepresence resolves this by adding "redis" as an `excludeSuffix`, but this isn't really a suffix. It's a hostname. So let's add:

```yaml
  dns:
    excludes:
      - redis
    mappings:
      - name: postgres
        aliasFor: psql.big-data
```

When the DNS-server is queried for one of the names in the `excludes` list, it will respond with `NXNAME`, or, in case of the overriding resolver, delegate to its fallback resolver.

knlambert:
## Description

This PR adds two festures:
* Being able to exclude some hostnames from the DNS resolution which can be useful to avoid targeting things which are running locally.
* With the same idea, some of these things running locally can have a remote equivalent that we may want to be able to target using their local name. For that we need a mapping configuration to link the local name to its remote equivalent.

This PR adds two new fields `excludes` and `mappings` to the kubeconfig configuration (see [issue](https://github.com/telepresenceio/telepresence/issues/3162) for examples).

## Checklist


pkg/client/rootd/dns/server.go
```diff
@@ -249,6 +260,24 @@ func (s *Server) resolveInCluster(c context.Context, q *dns.Question) (result dn
 	return result, rCode, nil
 }
 
+func (s *Server) ResolveAlias(query string) *string {
+	for i := range s.config.Mappings {
+		mappingName := s.config.Mappings[i].Name + "."
+		if strings.Count(mappingName, ".") == 1 {
+			mappingName += s.clusterDomain
+		}
```
thallgren: What use-case does this cover. The name just before the cluster domain is almost always "svc".

pkg/client/rootd/dns/server.go
```diff
@@ -249,6 +260,24 @@ func (s *Server) resolveInCluster(c context.Context, q *dns.Question) (result dn
...
+
+		if mappingName == query {
+			dotName := s.config.Mappings[i].AliasFor + "."
+			if strings.Count(dotName, ".") == 1 {
+				dotName += s.clusterDomain
+			}
```
thallgren: Same question here.

thallgren (changes requested):
I think the approach taken for the mappings is wrong. The DNS resolver should return a CNAME record when it encounters a mapping. That'll give the caller a chance to reissue the query with the new name, which in turn might or might not be resolved by another server.

pkg/client/rootd/dns/server.go
knlambert: When I receive a query for let's say "echo-service", then `query` = 'echo-service.cluster.local' since it's using the default resolver, so to make it "match" to the mapping, I need to add the domain. Do I miss something ?

pkg/client/rootd/dns/server.go
knlambert: Related to thing before, but that's probably not necessary.

pkg/client/rootd/dns/server.go
thallgren: "echo-service" is most likely a reference to `echo-service.<namespace>.svc.<clusterDomain>`. It's never `echo-service.<clusterDomain>`, so it's wrong in two ways.
1. We must not assume that the mapping includes a namespace.
2. The `svc` part is missing regardless.

thallgren:
Given that these are new features, there should be integration tests that protects regressions.

pkg/client/rootd/dns/server.go
```diff
@@ -249,6 +256,17 @@ func (s *Server) resolveInCluster(c context.Context, q *dns.Question) (result dn
 	return result, rCode, nil
 }
 
+func (s *Server) ResolveAlias(q *dns.Question) *string {
+	for i := range s.config.Mappings {
+		mappingName := s.config.Mappings[i].Name + "."
+		if mappingName == q.Name || mappingName == strings.TrimSuffix(q.Name, s.clusterDomain) {
```
thallgren:
```suggestion
	if len(s.config.Mappings) == 0 {
		return nil
	}
	qName := q.Name[:len(q.Name)-1] // strip trailing dot
	for i := range s.config.Mappings {
		if mappingName == qName  {
			dotName := s.config.Mappings[i].AliasFor + "."
			return &dotName
		}
		if strings.HasPrefix(q.Name, mappingName) && q.Name == mappingName + ".svc." + s.clusterDomain {
			dotName := s.config.Mappings[i].AliasFor + ".svc." + s.clusterDomain
			return &dotName
		}
```
Motivation:
- no repeated string concatenation and also no string concatenation unless `q.Name` is prefixed with `mappingName` (saves some CPU-cycles).
- if a search path ending with the cluster domain has been applied to `q.Name`, then it will most likely be `.svc.<cluster domain>`, not just `<cluster domain>`.

I'm not 100% convinced that a cluster domain test is necessary, but if such a test matches, then the returned CNAME must also have the `.svc.<cluster domain>` part appended.

Thinking more about this, the `.svc.<cluster domain>` check is in fact a synthetic mapping and as such, it can be added in advance to the mapping table so that no extra test is needed here. Such an addition could be conditioned to only happen on platforms where the resolver is subjected to search paths (Windows and Linux with systemd-resolved). I don't think it is ever needed when using the overriding resolver (Linux in docker uses that) or on macOS (where all namespaces become domain names).

knlambert: I'm not sure we're talking about the same thing. Currently for this function [here](https://github.com/telepresenceio/telepresence/blob/afdf8eba6efeeb537bbc8f2c3b5dedd25e96b830/pkg/client/rootd/dns/server.go#L662),

```cli
curl echo-service.blue
```

```golang
func (s *Server) resolveQuery(q *dns.Question, dv *cacheEntry) (dnsproxy.RRs, int, error) {
    // Here, q.name = `echo-service.blue.`
}
```

In the other hand:

```cli
curl echo-service
```

```golang
func (s *Server) resolveQuery(q *dns.Question, dv *cacheEntry) (dnsproxy.RRs, int, error) {
    // Here, q.name = `echo-service.cluster.local.`
}
```

There is a default root domain added when we only specify a name that we need to get ride of / ignore. 

Here the default DNS resolver on Mac:

```cli
➜  ~ cat /etc/resolver/telepresence.local
# Generated by telepresence

port 59553
domain cluster.local # Means that cluster.local is added by default.
nameserver 127.0.0.1
search cluster.local
```

thallgren: The name `echo-service.cluster.local.` is nonsensical. It will never resolve to anything in the cluster.

It's the `search cluster.local` that adds  `.cluster.local`. And it should probably not be there. The reason it works anyway, is that we add one file in this directory for each mapped namespace during connect.

thallgren: @knlambert can you please try and remove the `search` from the emitted `telepresence.local`  file? I'm pretty sure it's bogus.

knlambert: Pretty sure that It won't be there anymore :D that's what this field is supposed to do. The search field is only added in the telepresence.local file. The others don't have it, but sure let me try :)

knlambert: There are two places in the code trying to inject that line : 

https://github.com/telepresenceio/telepresence/blob/ae0bd806e2e95e04117de621bfe070c11f6a3ea6/pkg/client/rootd/dns/server_darwin.go#L201
https://github.com/telepresenceio/telepresence/blob/ae0bd806e2e95e04117de621bfe070c11f6a3ea6/pkg/client/rootd/dns/server_darwin.go#L194

If I remove the search line, `curl echo-easy` stops working. The requests never reach the local server :(

thallgren: And you only remove it from the `telepresence.local` file? Not from any other generated file?

thallgren: Try this:
1. Create a successful intercept.
2. Verify that `curl echo-easy` works.
3. Manually edit the `telepresence.local` file and remove the `search` line.
4. Verify that `curl echo-easy` still works

knlambert:
Ok, what about this [change](https://github.com/telepresenceio/telepresence/pull/3172/commits/795199504d18986c1f4083d650e43f421ef12151) to avoid having the defaut cluster.local domain being sent to the server in case of a unqualified hostname ?

Basically I keep the telepresence.local dns resolver, but I remove the "search" field which should only be in the one intended to process unqualified hostnames (it can only be one of these through all the files): 

```cli
➜  ~ cat /etc/resolver/telepresence.local
# Generated by telepresence

port 63653
domain cluster.local
nameserver 127.0.0.1
# search cluster.local // This line isn't there anymore.
```
Then I introduce a new resolver named tel2-search, which isn't a real hostname but avoid the issue of sending them with `cluster.local`

```cli
➜  telepresence git:(knlambert/dns-resolver-mapping) ✗ cat /etc/resolver/tel2-search.local      
# Generated by telepresence

port 63653
domain tel2-search
nameserver 127.0.0.1
search tel2-search
```
And If I intercept, that's this file which is modified now : 

```cli
➜  telepresence git:(knlambert/dns-resolver-mapping) ✗ telepresence intercept echo-easy --port 9090 -n blue
Using Deployment echo-easy
   Intercept name         : echo-easy-blue
   State                  : ACTIVE
   Workload kind          : Deployment
   Destination            : 127.0.0.1:9090
   Service Port Identifier: proxied
   Volume Mount Point     : /var/folders/x_/4x_4pfvx2j3_94f36x551g140000gp/T/telfs-1138152778
   Intercepting           : all TCP requests
➜  telepresence git:(knlambert/dns-resolver-mapping) ✗ cat /etc/resolver/tel2-search.local
# Generated by telepresence

port 63653
domain tel2-search
nameserver 127.0.0.1
search blue.svc.cluster.local tel2-search

```

thallgren:
I think using `tel2-search` in the search path is great, but it shouldn't be used as the domain. The domain is there so that queries that don't need a search path, e.g. `eho-easy.default.svc.cluster.local` are sent to the correct resolver.

integration_test/not_connected_test.go
```diff
@@ -20,10 +22,20 @@ func (s *notConnectedSuite) SuiteName() string {
...
+		s.serviceName = "echo"
+		s.appSpace, _ = itest.AppAndMgrNSName(suffix)
+		return s
 	})
 }
 
+func (s *notConnectedSuite) SetupSuite() {
+	s.Suite.SetupSuite()
+	ctx := s.Context()
+	itest.ApplyEchoService(ctx, s.serviceName, s.appSpace, 80)
```
thallgren: This will slow down all other tests in the `notConnectedSuite` when running them as singleton tests. So please don't do this.

integration_test/not_connected_test.go
```diff
@@ -12,6 +12,8 @@ import (
 type notConnectedSuite struct {
 	itest.Suite
 	itest.NamespacePair
+
+	appSpace, serviceName string
```
thallgren: If the intent is to use a single service here, why not use the single-service suite?

pkg/client/cli/cmd/status.go
```diff
@@ -288,7 +288,7 @@ func printDNS(kvf *ioutil.KeyValueFormatter, d *client.DNSSnake) {
 	if len(d.RemoteIP) > 0 {
 		dnsKvf.Add("Remote IP", d.RemoteIP.String())
 	}
-	dnsKvf.Add("Exclude suffixes", fmt.Sprintf("%v", d.ExcludeSuffixes))
+	dnsKvf.Add("Excludes suffixes", fmt.Sprintf("%v", d.ExcludeSuffixes))
```
thallgren: Did the IDE get a bit too ambitious when substituting?

pkg/client/k8s_config.go
```diff
@@ -44,6 +65,12 @@ type DnsConfig struct {
 	// a lookup. Includes have higher priority than excludes.
 	IncludeSuffixes []string `json:"include-suffixes,omitempty"`
 
+	// Excludes are a list of hostname that the DNS resolver will not resolve even if they exist.
+	Excludes []string `json:"exclude,omitempty"`
+	// Mappings contains a list of DNS Mappings. Each item references a hostname, and an associated alias. If a
```
thallgren, line 71:
```suggestion

	// Mappings contains a list of DNS Mappings. Each item references a hostname, and an associated alias. If a
 ```

File: pkg/client/rootd/dns/server_darwin.go [hunk 8]
```diff
@@ -50,7 +50,8 @@ func (r *resolveFile) write(fileName string) error {
 // or, if not on a Mac, follow this link: https://www.manpagez.com/man/5/resolver/
 func (s *Server) Worker(c context.Context, dev vif.Device, configureDNS func(net.IP, *net.UDPAddr)) error {
 	resolverDirName := filepath.Join("/etc", "resolver")
-	resolverFileName := filepath.Join(resolverDirName, "telepresence.local")
+	localResolverFileName := filepath.Join(resolverDirName, "telepresence.local")
+	tel2SearchResolverFileName := filepath.Join(resolverDirName, "tel2-search.local")
```
thallgren: This suggests a domain named `tel2-search`. I don't think we want that. The `tel2-search` is intended as a search path only. It's not a domain.

thallgren (changes requested):

integration_test/not_connected_test.go
knlambert: Should I create a new suite for this ?

Because right now there is :
* Not connected suite: -> No service, not connected
* Single service -> Service, but connected.

So I can't add it to kubeconfig_extension_test.go because it uses the not connected, and I can't switch it for a single service because it breaks the other tests.

knlambert:
> I think using `tel2-search` in the search path is great, but it shouldn't be used as the domain. The domain is there so that queries that don't need a search path, e.g. `eho-easy.default.svc.cluster.local` are sent to the correct resolver.

The issue is that if the search path doesn't match the domain, it doesn't work.

I've tried this : 
```cli
port 63653
domain cluster.local
nameserver 127.0.0.1
search tel2-search
```
But then the DNS queries are not even sent to the server, so I don't have a good solution to both route the non qualified names and trim the cluster.local at the same time (except using my initial solution).

What do you suggest ?

thallgren:
Oh, I see. The macOS deducts that the domain will not get hit by the given search path. Ok, so instead of just adding `tel2-search`, perhaps add `tel2-search.cluster.local` ? That will still give us something recognizable that we can remove.

knlambert:
> Oh, I see. The macOS deducts that the domain will not get hit by the given search path. Ok, so instead of just adding `tel2-search`, perhaps add `tel2-search.cluster.local` ? That will still give us something recognizable that we can remove.

OK so just to get this right:
I go back to having only the telepresence.local resolver, but instead of `search cluster.local`, I get `search tel2-search.cluster.local` so I can identify the unqualified hostnames ?

thallgren (changes requested):
I don't think the "tel2-search" is handled correctly still. The purpose of that name is to deduct that a search path has been appended for the sake of reaching the resolver. A search path that hence must be removed.  The "tel2-search." (and everything to the right of it) must be stripped off. There's no difference in "X.tel2-search." and "X.tel2-search.\<cluster domain>." for instance. They are both just "X".

Ideally, this strip (and restore) of the "tel2-search" should happen early, and in one place only.

pkg/client/rootd/dns/server.go
```diff
@@ -666,6 +728,28 @@ func (s *Server) resolveQuery(q *dns.Question, dv *cacheEntry) (dnsproxy.RRs, in
...
 
+	// Returns a CNAME pointing to the mapping when there is a hit.
+	if mappingAlias := s.ResolveMappingAlias(q.Name); mappingAlias != nil {
+		dv.answer = dnsproxy.RRs{&dns.CNAME{
+			Hdr:    dns.RR_Header{Name: q.Name, Rrtype: dns.TypeCNAME, Class: dns.ClassINET, Ttl: dnsTTL},
+			Target: *mappingAlias,
+		}}
+		// On Windows, or in a Linux container, just returning the cname isn't enough, and the DNS resolver won't try
+		// to get the record with a subsequent request, so we need to resolve the record until we get a better solution.
+		if runtime.GOOS == "windows" || proc.RunningInContainer() {
```
knlambert, line 739: I ended up having the same problem in a container than on Windows. Could it be because in both cases, we intercept the DNS queries to redirect them to our local resolver, so it doesn't work well with Cnames ?

thallgren (approved):

PR #3183, opened 2023-05-15
IPv6 enable the telepresence docker network.

### LINKED ISSUE #3179: Enable IPv6 on the telepresence docker network
An apache running in a docker based intercept handler might fail with a DNS lookup failure when it dials other services in the cluster even though the DNS is resolved and returned. The reason appears to be that IPv6 isn't enabled, because when adding a network where it is, that resolves the problem.

The solution is to add IPv6 to the `telepresence` network, but doing that isn't straightforward because it requires that IPv6 has been enabled in `/etc/docker/daemon.json`, or that the network is created with an explicitly added IPv6 subnet and gateway. Let's say we create the `telepresence` network using `--ipv6` . If it fails, I can see the following options:
1. inform the user that ipv6 must be added to `/etc/docker/daemon.json`
2. make another attempt without `--ipv6`, and log a warning that some DNS functionality might not be available.
3. make another attempt, but add an explicit subnet/gateway using `--subnet` and `--gateway` flag.

Number 3 is probably the most user-friendly, but it also requires Telepresence to select a random subnet, which defeats the purpose of defining this in `/etc/docker/daemon.json`.

thallgren:
## Description

DNS AAAA queries will never reach a Telepresence daemon that runs in a Docker container unless that container is attached to an IPv6 enabled network. This commit therefore ensures that the "telepresence" network is IPv6 enabled by first trying to create it that way, and if that fails due to the lack of an IPv6 pool in `/etc/docker/daemon.json`, by adding a random IPv6 subnet and gateway to force the enablement without the existence of such a pool.

## Checklist

### DIFF
--- a/pkg/client/docker/network.go
+++ b/pkg/client/docker/network.go
@@ -1,0 +1,119 @@
+package docker
+
+import (
+	"context"
+	"fmt"
+	"math/rand"
+	"strings"
+
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/api/types/network"
+	dockerClient "github.com/docker/docker/client"
+
+	"github.com/datawire/dlib/dlog"
+)
+
+// EnsureNetwork checks if a network with the given name exists, and creates it if that is not the case.
+func EnsureNetwork(ctx context.Context, name string) error {
+	cli, err := dockerClient.NewClientWithOpts(dockerClient.FromEnv, dockerClient.WithAPIVersionNegotiation())
+	if err != nil {
+		return err
+	}
+	defer cli.Close()
+	resource, err := cli.NetworkInspect(ctx, name, types.NetworkInspectOptions{})
+	if err != nil {
+		if !dockerClient.IsErrNotFound(err) {
+			return err
+		}
+	} else {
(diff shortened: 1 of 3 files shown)


pkg/client/docker/network.go
```diff
@@ -0,0 +1,119 @@
...
+	ipam := &resource.IPAM
+
+	// Save original Config to use a source for copy
+	numConfigs := len(ipam.Config)
+	ipamConfig := make([]network.IPAMConfig, numConfigs)
+	copy(ipamConfig, ipam.Config)
+
+	// Make attempts to add a random IPv6 subnet and gateway to IPAM
+	for retry := 0; retry < 5; retry++ {
+		ipn := fmt.Sprintf("%016x", rand.Int63())
```
knlambert, line 95: I guess the probability of a conflict is super low. Would it generate an error ? Otherwise LGTM.

knlambert (approved):

pkg/client/docker/network.go
thallgren: The last attempt will be reported as an error if it gets to that.

PR #3188, opened 2023-05-19
Fixed `.timeouts.agentArrival` template.

xzhang-brex:
## Description

We are using the `default` function wrongly, as a result, the value passed in is not being honored.

## Checklist

### DIFF
--- a/charts/telepresence/templates/deployment.yaml
+++ b/charts/telepresence/templates/deployment.yaml
@@ -80,7 +80,7 @@ spec:
           - name: INTERCEPT_DISABLE_GLOBAL
             value: {{ quote (default .intercept.disableGlobal false) }}
           - name: AGENT_ARRIVAL_TIMEOUT
-            value: {{ quote (default .timeouts.agentArrival "30s") }}
+            value: {{ quote (default "30s" .timeouts.agentArrival) }}
         {{- /*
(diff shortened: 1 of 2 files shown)


charts/telepresence/templates/deployment.yaml
```diff
@@ -80,7 +80,7 @@ spec:
           - name: INTERCEPT_DISABLE_GLOBAL
             value: {{ quote (default .intercept.disableGlobal false) }}
```
knlambert: I think this one is written wrong too and only works because its set in values.yaml (clearly the one you took as an example), can you fix it since you're at it and add it to the changelog ? :D

knlambert (changes requested):

charts/telepresence/templates/deployment.yaml
xzhang-brex: @knlambert done!!!

knlambert (approved):

xzhang-brex:
@knlambert @thallgren hey guys could you apply ok to test label?

xzhang-brex:
@knlambert @thallgren guys could you help rerun the test? It seems like a race condition is happening again.

thallgren (approved):

PR #3203, opened 2023-05-25
Refactor routing APIs to be based on routing tables

josecv:
## Description

Refactors most of telepresence's routing logic. Best read commit-by-commit

## Checklist


pkg/vif/device_windows.go
```diff
@@ -40,7 +40,24 @@ func openTun(ctx context.Context) (td *nativeDevice, err error) {
 			dlog.Errorf(ctx, "%+v", err)
 		}
 	}()
-	interfaceName := "tel0"
+	interfaceFmt := "tel%d"
```
knlambert, line 44: Just curious, why should we increase it now ?

pkg/routing/routing.go
```diff
@@ -43,6 +43,14 @@ func Subnets(routes []*Route) []*net.IPNet {
 	return ns
 }
 
+func SubnetsOverlap(one, other *net.IPNet) bool {
```
knlambert: I see these as not used ?

image

pkg/vif/router.go
```diff
@@ -23,12 +22,26 @@ type Router struct {
 	staticOverrides []*routing.Route
 	// The subnets that are currently being routed
 	routedSubnets []*net.IPNet
+	// The subnets that are allowed to be routed even in the presence of conflicting routes
+	whitelistedSubnets []*net.IPNet
```
knlambert, line 28: Maybe I would change it for something like allowList instead, especially since we push the project to CNCF. :P

pkg/routing/routing_linux.go
```diff
@@ -164,6 +173,75 @@ func GetRoute(ctx context.Context, routedNet *net.IPNet) (*Route, error) {
...
+	rule.Priority = priority
+	rule.Family = netlink.FAMILY_V4
+	if err := netlink.RuleAdd(rule); err != nil {
+		return nil, err
+	}
+	return &table{
+		index: index,
+		rule:  rule,
+	}, nil
+}
```
knlambert, line 213: Still curious, why do we want the priority to be under the existing rules ?

knlambert (dismissed):

pkg/vif/router.go
thallgren: @knlambert why would we change that? "whitelisted" is much more concise. It doesn't mean just "allowed", it means that you *override* the fact that it isn't.

pkg/vif/device_windows.go
```diff
@@ -40,7 +40,24 @@ func openTun(ctx context.Context) (td *nativeDevice, err error) {
...
+	ifaceNumber := 0
+	ifaces, err := net.Interfaces()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get interfaces: %w", err)
+	}
+	for _, iface := range ifaces {
+		dlog.Infof(ctx, "Found interface %s", iface.Name)
```
thallgren: This looks like debug logging to me.

pkg/routing/routing.go
```diff
@@ -130,3 +132,58 @@ func interfaceLocalIP(iface *net.Interface, ipv4 bool) (net.IP, error) {
 	}
 	return nil, nil
 }
+
+func compareRoutes(ctx context.Context, osRoute, tableRoute *Route) (bool, error) {
+	dlog.Tracef(ctx, "Comparing OS route %s to table route %s", osRoute, tableRoute)
+	if runtime.GOOS == "linux" {
```
thallgren: Please move this into routing_linux.go instead.

pkg/routing/routing_linux.go
thallgren: @knlambert Because we don't want to magically change the existing rules without approval from the user.

pkg/vif/router.go
thallgren: @knlambert Given that the terms blacklist and whitelist are very common in the code base (and in the networking domain as a whole), and also used by modules we depend on, I suggest we keep changes like that to public facing material, such as docs and release notes. We can change it later if CNCF complains about it.

pkg/vif/tunneling_device.go
```diff
@@ -5,36 +5,52 @@ import (
 
 	"gvisor.dev/gvisor/pkg/tcpip/stack"
 
+	"github.com/hashicorp/go-multierror"
```
thallgren, line 6:
```suggestion
	"github.com/hashicorp/go-multierror"
```

pkg/vif/testdata/router/main.go
```diff
@@ -8,20 +8,31 @@ import (
...
+	bCtx := client.WithConfig(context.Background(), &cfg)
+	logger := logrus.New()
+	logger.SetLevel(logrus.DebugLevel)
+	bCtx = dlog.WithLogger(bCtx, dlog.WrapLogrus(logger))
+	vif.InitLogger(bCtx)
+
+	ctx, cancel := context.WithCancel(client.WithConfig(bCtx, &cfg))
 	defer cancel()
-	vif.InitLogger(ctx)
 	dev, err := vif.NewTunnelingDevice(ctx, nil)
```
thallgren:
```suggestion
	dev, err := vif.NewTunnelingDevice(ctx, func(context.Context, tunnel.ConnID) (tunnel.Stream, error) {
		return nil, errors.New("stream routing not enabled; refusing to forward"")
	})
```
This way, you don't need a specific test in `vif.dispatchToStream()`.

thallgren (changes requested):

pkg/routing/routing_linux.go
josecv: well having the priority _under_ the existing rules will mean our table gets queried _first_
we do this so that overrides actually work; we do the check for existing routes elsewhere

pkg/vif/device_windows.go
josecv: yes it is sorry

pkg/vif/router.go
josecv: I'm not sure what the CNCF bit has to do with this? I think this is a whitelist

pkg/routing/routing.go
josecv: yeah there's probably a few of those, I'll take a look

pkg/routing/routing.go
```diff
@@ -135,22 +102,14 @@ func interfaceLocalIP(iface *net.Interface, ipv4 bool) (net.IP, error) {
...
-				dlog.Tracef(ctx, "Checking address %s against %s", addr.String(), osRoute.RoutedNet.IP.String())
-				if addr.(*net.IPNet).IP.Equal(osRoute.LocalIP) {
-					return true, nil
-				}
-			}
-		}
+	if osRoute.Interface.Index == tableRoute.Interface.Index {
+		return true, nil
+	}
+	osCompare, err := osCompareRoutes(ctx, osRoute, tableRoute)
```
thallgren: Just do:
```
	return osCompareRoutes(ctx, osRoute, tableRoute)
```

thallgren (approved):
Approving, feel free to fix my nit.

pkg/vif/router.go
thallgren: @josecv I agree, it's a whitelist, but nowadays, some people think that by changing the vocabulary, we'll address systemic racism:
https://www.cncf.io/announcements/2021/10/13/inclusive-naming-initiative-announces-new-community-resources-for-a-more-inclusive-future/

pkg/routing/routing_linux.go
thallgren: Ah, sorry.

pkg/routing/routing.go
josecv: omg morning brain yes

pkg/vif/router.go
josecv: ah

pkg/vif/device_windows.go
josecv: we should've increased it always; it's just that now the unit tests are actually creating multiple devices, and so for the first time having two devices named tel0 on the system doesn't cut it anymore

josecv:
woooo 🥳

PR #3206, opened 2023-05-26
Created a way to exclude certain Environment Variables from being propagated to the client

jacoblbeck:
## Description
Added a new configmap that is customizable with an array of env var keys, that we then remove when reviewing an intercept at the traffic-manager so that administrators of the cluster can forbid certain variables.

## Checklist

### DIFF
--- a/charts/telepresence/templates/excluded-variables-configmap.yaml
+++ b/charts/telepresence/templates/excluded-variables-configmap.yaml
@@ -1,0 +1,10 @@
+{{- if not .Values.rbac.only }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: telepresence-excluded-variables
+  namespace:  {{ include "traffic-manager.namespace" . }}
+  data: {{ range .Values.variables.excluded }}
+    {{ . }}: "true"
+  {{- end }}
+{{- end }}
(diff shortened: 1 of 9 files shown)


charts/telepresence/templates/excluded-variables-configmap.yaml
```diff
@@ -0,0 +1,10 @@
+{{- if not .Values.rbac.only }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: telepresence-excluded-variables
+  namespace:  {{ include "traffic-manager.namespace" . }}
+data: {{ range .Values.variables.excluded }}
```
thallgren: Even if it is a long name, I'd suggest using `.Values.intercept.environment.excluded`. It's more concise and will also be extendable if we want to add more functionality in the future (substitutions, default values, etc.).

charts/telepresence/templates/trafficManagerRbac/cluster-scope.yaml
```diff
@@ -56,6 +56,7 @@ rules:
   - delete
   resourceNames:
   - telepresence-agents
+  - telepresence-excluded-variables
```
thallgren: Using a configmap here is quite powerful. Could be be made more generic? Perhaps name it `telepresence-intercept-env` and then have an `excluded` top-level key under `data`?

thallgren (commented):
I like the solution but would like to see it more generic. What do you think?

thallgren (approved):

thallgren:
This looks good now, but it would be great if you could add an integration test for it.

PR #3207, opened 2023-05-29
Add GRPC endpoints & status for mappings / excludes features

knlambert:
## Description

We need expose the [mapping / excludes features ](https://github.com/telepresenceio/telepresence/pull/3172) when running in Docker mode.

Two things:

* `mappings` & `excludes` were added to the status command.
* Two new GRPC functions were added to Set the values remotely.

## Checklist


pkg/client/rootd/dns/server.go
```diff
@@ -11,9 +11,10 @@ import (
 	"sync/atomic"
 	"time"
 
-	"github.com/miekg/dns"
 	"google.golang.org/protobuf/types/known/durationpb"
 
+	"github.com/miekg/dns"
```
thallgren: This is wrong.

pkg/client/rootd/dns/server.go
```diff
@@ -346,6 +349,25 @@ func (s *Server) SetSearchPath(ctx context.Context, paths, namespaces []string)
 	}
 }
 
+// SetExcludes sets the excludes list in the config.
+func (s *Server) SetExcludes(ctx context.Context, excludes []string) {
+	s.config.Excludes = excludes
```
thallgren, line 394: The `Excludes` and `Mappings` must be protected using a mutex. Preferably and RWMutex since writes will be very rare and concurrent reads very common.

thallgren (changes requested):

pkg/client/rootd/dns/server.go
```diff
@@ -213,6 +216,19 @@ func (s *Server) shouldDoClusterLookup(query string) bool {
...
+	s.configLock.RLock()
+	defer s.configLock.RUnlock()
+	// When intercepting, this function will potentially receive the hostname of any search param, so their
+	// unqualified hostname should be evaluated too.
+	unqualifiedHostnames := make([]string, 0)
+	for i := range s.search {
+		withoutDomain := strings.TrimSuffix(query, s.search[i])
+		unqualifiedHostnames = append(unqualifiedHostnames, strings.TrimSuffix(withoutDomain, "."))
+	}
+	return slice.ContainsAny(s.config.Excludes, append(unqualifiedHostnames, query))
```
thallgren: This is a slow way of doing this, involving building a slice that contains elements that perhaps don't have a matching suffix, which is then scanned sequentially multiple times, only to return a boolean if there's a hit. Also, the last line append `query` without stripping its trailing dot. 

How about something like this?
```suggestion
	if slice.Contains(s.config.Excludes, query[:len(query)-1]) {
		return true
	}
	for _, sp := range s.search {
		if strings.HasSuffix(query, sp) && slice.Contains(s.config.Excludes, query[:len(query)-len(sp)-1]) {
			return true
		}
	}
	return false
```

thallgren (changes requested):

thallgren (approved):

PR #3252, opened 2023-06-28
add Agent Pull Policy to helm chart

njayp:
## Description

add Agent Pull Policy to helm chart. It is defaulted to "IfNotPresent" in the values file, meaning it will always be present in the traffic manager as an env var

## Checklist


charts/telepresence/values.yaml
```diff
@@ -213,6 +213,7 @@ agent:
     name:
     tag:
     pullSecrets: []
+    pullPolicy: IfNotPresent
```
thallgren, line 165: This should be documented in the chart's README.md.

cmd/traffic/cmd/manager/managerutil/envconfig.go
```diff
@@ -52,6 +52,7 @@ type Env struct {
 
 	AgentRegistry             string                      `env:"AGENT_REGISTRY,           parser=nonempty-string"`
 	AgentImage                string                      `env:"AGENT_IMAGE,              parser=string,         default="`
+	AgentImagePullPolicy      string                      `env:"AGENT_IMAGE_PULL_POLICY,  parser=string,		 default="`
```
thallgren:
```suggestion
	AgentImagePullPolicy      string                      `env:"AGENT_IMAGE_PULL_POLICY,  parser=string,         default="`
```
because using tabs inside a string makes the string misaligned unless you have a specific tab-with setting.

thallgren (changes requested):

thallgren (approved):

PR #3272, opened 2023-07-20
Pass the scout.Reporter in the context.

thallgren:
A lot of structures kept a ``*scout.Reporter` entry that was passed in as a constructor parameter. This commit removes all of that in favor of using a context value. This makes it easier to configure or mock the reporter, and also to completely disable it, should we ever want to do that. It also makes the reporter available everywhere there's a context.


pkg/client/cli/cmd/status.go
```diff
@@ -142,8 +142,8 @@ func BasicGetStatusInfo(ctx context.Context) (ioutil.WriterTos, error) {
 		}
 		wt = &si
 	}
-	reporter := scout.NewReporter(ctx, "cli")
-	us.InstallID = reporter.InstallID()
+	ctx = scout.NewReporter(ctx, "cli")
```
knlambert, line 145: Is this one started down the line ?

knlambert (approved):
LGTM. I guess it comes with another PR in the pro repo :)

pkg/client/cli/cmd/status.go
thallgren: No, neither started or closed. It's just used for the InstallID().

PR #3273, opened 2023-07-21
Add consumption state to the traffic-manager

knlambert:
## Description

This PR extends the traffic manager state to include consumption metrics, so we can later expose them to prometheus.

Benchmark:

**Before**
```cli
➜  ~ wrk -t12 -c400 -d30s http://echo-easy.blue
Running 30s test @ http://echo-easy.blue
  12 threads and 400 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency    62.96ms   12.13ms 276.85ms   88.41%
    Req/Sec   319.44    134.07   666.00     66.77%
  114157 requests in 30.09s, 20.03MB read
  Socket errors: connect 157, read 0, write 0, timeout 0
Requests/sec:   3794.00
Transfer/sec:    681.73KB
```
**After**
```cli
~ wrk -t12 -c400 -d30s http://echo-easy.blue
Running 30s test @ http://echo-easy.blue
  12 threads and 400 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency    64.58ms   15.99ms 322.08ms   92.70%
    Req/Sec   311.97    185.42   670.00     58.31%
  111306 requests in 30.06s, 19.53MB read
  Socket errors: connect 157, read 0, write 0, timeout 0
Requests/sec:   3702.29
Transfer/sec:    665.26KB
```

No major difference.
## Checklist

### DIFF
--- a/cmd/traffic/cmd/manager/state/consumption.go
+++ b/cmd/traffic/cmd/manager/state/consumption.go
@@ -9,7 +9,7 @@ import (
 
 // SessionConsumptionMetricsStaleTTL is the duration after which we consider the metrics to be staled, meaning
 // that they should not be updated anymore since the user doesn't really use Telepresence at the moment.
-const SessionConsumptionMetricsStaleTTL = 15 * time.Minute
+const SessionConsumptionMetricsStaleTTL = 60 * time.Minute
 
 func NewSessionConsumptionMetrics() *SessionConsumptionMetrics {
 	return &SessionConsumptionMetrics{
@@ -45,8 +45,8 @@ func (s *state) GetSessionConsumptionMetrics(sessionID string) *SessionConsumpti
 	s.mu.RLock()
 	defer s.mu.RUnlock()
 	for i := range s.sessions {
-		if i == sessionID {
-			return s.sessions[i].ConsumptionMetrics()
+		if css, ok := s.sessions[i].(*clientSessionState); i == sessionID && ok {
+			return css.ConsumptionMetrics()
 		}
 	}
 	return nil
@@ -57,7 +57,9 @@ func (s *state) GetAllSessionConsumptionMetrics() map[string]*SessionConsumption
 	s.mu.RLock()
 	defer s.mu.RUnlock()
 	for sessionID := range s.sessions {
-		allSCM[sessionID] = s.sessions[sessionID].ConsumptionMetrics()
+		if css, ok := s.sessions[sessionID].(*clientSessionState); ok {
+			allSCM[sessionID] = css.ConsumptionMetrics()
+		}
 	}
 	return allSCM
 }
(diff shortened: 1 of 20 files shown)


jacoblbeck (approved):

cmd/traffic/cmd/manager/state/state.go
```diff
@@ -218,7 +218,10 @@ func (s *state) RemoveSession(ctx context.Context, sessionID string) {
 	defer s.mu.Unlock()
 	dlog.Debugf(ctx, "Session %s removed. Explicit removal", sessionID)
 
-	s.unlockedRemoveSessionConsumption(sessionID)
+	if _, isClientSession := s.sessions[sessionID].(*clientSessionState); isClientSession {
+		s.unlockedRemoveSessionConsumption(sessionID)
```
josecv: can we move this check into the method? it seems confusing to have one check in one file and another in another, if I'm looking at consumption.go I'm wondering why `Refresh` needs a client session but `unlockedRemove` doesn't

cmd/traffic/cmd/manager/state/consumption.go
```diff
@@ -1,44 +1,81 @@
...
 
-func (s *state) unlockedAddSessionConsumption(sessionID string) {
-	s.sessionConsumptionMetrics[sessionID] = &SessionConsumptionMetrics{
-		Duration:   0,
-		LastUpdate: time.Now(),
+func (sc *SessionConsumptionMetrics) RunCollect(ctx context.Context) {
+	for {
+		select {
+		case <-ctx.Done():
+			sc.closeChannels()
```
josecv: should we like defer this? i'm concerned there's a channel leak here if you just close the one, and `closeChannels` is v easy to make idempotent

pkg/tunnel/bidipipe.go
```diff
@@ -40,8 +52,9 @@ func (p *bidiPipe) Start(ctx context.Context) {
 		wg.Add(2)
 		dlog.Debugf(ctx, "   FWD connect %s", p.name)
 		atomic.AddInt32(p.counter, 1)
-		go doPipe(ctx, p.a, p.b, &wg)
-		go doPipe(ctx, p.b, p.a, &wg)
+		// p.pm collects metrics only for one stream (since the same data is going through both streams)
```
josecv: p.pm?

pkg/tunnel/stream.go
```diff
@@ -91,6 +91,10 @@ func ReadLoop(ctx context.Context, s Stream) (<-chan Message, <-chan error) {
 
 		for {
 			m, err := s.Receive(ctx)
+			if m != nil && b != nil {
+				b <- uint64(len(m.Payload()))
```
josecv: This is a potentially blocking channel write in the middle of hot network code, you should probably have a timeout (millis or nanos) of some sort here just in case

cmd/traffic/cmd/manager/state/consumption.go
```diff
@@ -1,44 +1,81 @@
...
 
 // SessionConsumptionMetricsStaleTTL is the duration after which we consider the metrics to be staled, meaning
 // that they should not be updated anymore since the user doesn't really use Telepresence at the moment.
 const SessionConsumptionMetricsStaleTTL = 1 * time.Minute // TODO: Increase.
 
+func NewSessionConsumptionMetrics() *SessionConsumptionMetrics {
+	return &SessionConsumptionMetrics{
+		ConnectDuration:     0,
+		FromClientBytesChan: make(chan uint64),
+		ToClientBytesChan:   make(chan uint64),
```
josecv: please buffer these channels to at least 1024 messages. There's one write to the channels for every network packet that we see, which could potentially block really really badly.

josecv (changes requested):
concerned about the performance implications of this; i suggest you just hit it a few times with [wrk](https://github.com/wg/wrk) to make sure it's all okay

josecv:
Also I think from a design perspective I'd feel much better if the channels were private and you just had methods to report one packet; this is similar to how prometheus does metrics, when you increment a counter that's an `Increment()` method call that immediately writes to a metrics channel behind the scenes

pkg/tunnel/probe.go
```diff
@@ -0,0 +1,73 @@
...
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case b, ok := <-p.channel:
+			if !ok {
+				p.Close()
+				return
+			}
+			p.value += b
```
josecv: is this thread safe with GetValue? should it be locked or atomic?

josecv (approved):
just that lock question -- if it doesn't need a lock just explain why and merge it :)

pkg/tunnel/probe.go
knlambert: Mmh, no you're right. 

I mean, it's not really an issue in this case if the counter isn't perfectly synchronized, but it's still a race condition. I'll fix it :)

PR #3298, opened 2023-08-10
Fix potential process deadlock when watching workload

knlambert:
## Description

Using these commands was creating a deadlock :

```
tp list --output json-stream
tp connect
```

Impossible to quit the daemons either. The reason was that we watch the session context, instead of the request context, so the resource is never released.

This PR fixes that issue by looking at both the session, and the request contexts.

## Checklist


integration_test/testdata/echo-server/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:alpine3.15 AS builder
+FROM golang:1.20.7-alpine as builder
```
thallgren: Revert this and ensure use of datawire/dlib 1.3.1

integration_test/testdata/udp-echo/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:alpine3.15 AS builder
+FROM golang:1.20.7-alpine as builder
```
thallgren: Please revert.

pkg/client/cli/cmd/list.go
```diff
@@ -119,8 +120,10 @@ func (s *listCommand) list(cmd *cobra.Command, _ []string) error {
 	ch := make(chan *connector.WorkloadInfoSnapshot)
 	go func() {
 		for {
-			r, err := stream.Recv()
+			var r *connector.WorkloadInfoSnapshot
+			r, err = stream.Recv()
```
thallgren: So `err` here was private because this is a separate go routine. Unless its private, you'll need to protect it with a mutex outside of this go routine (a context cancellation might lead to concurrent access). It would probably be cleaner to send both `r` and `err` in a struct on the channel.

pkg/client/userd/trafficmgr/session.go
```diff
@@ -679,7 +679,7 @@ func (s *session) WatchWorkloads(c context.Context, wr *rpc.WatchWorkloadsReques
 	for {
 		select {
 		case <-c.Done():
-			return nil
+			return fmt.Errorf("the session was closed")
```
thallgren: It's not an error to close the session with a context cancellation.

thallgren (changes requested):

integration_test/list_watch_test.go
```diff
@@ -1,29 +1 @@
 package integration_test
-
-import (
-	"context"
-	"time"
-
-	"github.com/telepresenceio/telepresence/v2/integration_test/itest"
-)
-
-func (s *connectedSuite) Test_ListWatch() {
```
thallgren: Why is this test removed?

knlambert: I'm trying to debug :( sometime the tests pass, all green, sometime they don't for an entire day. Don't pay attention to the last commit. I try to find out what's causing it.

thallgren (approved):

Issue #1608, opened 2021-03-28
[v2] - Feature request - An ability to intercept k8s service


andreykrupskii:
First of all, I'm a big fan of telepresence as it's making the development process more convenient and saves my time. While I'm still not able to migrate to Telepresence 2 because of two problems. And this issue is about one of them. 

**Problem**: In our infrastructure micro-services besides the HTTP interface also communicate via message broker (rabbitmq). It was really easy to apply the previous version of Telepresence into our flow, because it just replaces an original container with the proxy, so the local instance is able to consume all messages from worker queues from the message broker. While Telepresence 2 injects sidecar container and the original instance stays alive, so messages from worker queues will be balanced between local and remote processes. 

**Feature request**: It would be a nice feature to be able to intercept traffic from a specific k8s service. So for our case, we would need just to scale the original deployment to zero and to start a local instance that will receive both HTTP traffic and message broker messages.

AndreyKrupskii:
Another option is to extend `intercept` command with an option `--swap-original`, that will start the proxy container (as before) instead of injecting the sidecar.

cujo:
This is must have feature for sure. My workaround for now is to replace image with stub while developing and then undo it:

set up
```
kubectl set image deployment/my-deployment my-container=ibmcom/pause:3.0
telepresence intercept my-service
```
tear down
```
telepresence quit
kubectl rollout undo deployment/my-deployment
```

AndreyKrupskii:
@cujo, I've also tried to stub the original container while then I was faced with another issue: inability to intercept several container ports in parallel - https://github.com/telepresenceio/telepresence/issues/1607

Splitside:
That would be a really useful feature!

petergardfjall:
Wanted to pitch in that the current design of Telepresence v2 does not work well for the large class of applications that operate with work queue semantics.

For example where incoming requests get placed in a job queue later to be picked up by an available worker. This worked perfectly with Telepresence v1, which entirely swapped out the deployment for the local process. With v2, it's a fifty-fifty chance that a job gets handled by the (still-running) original pod inside the cluster, which makes the debugging value of telepresence very limited (in fact it almost adds confusion).

Having the option to either (1) _replace_ the deployment (as was done with `--swap-deployment` in v1) or (2) being able to intercept a _service_ (as suggested here), combined with the ability to scale down the backing workload to zero replicas, would also bring v2 to feature parity with v1 for this large class of applications.

[...]

PR #3330, opened 2023-08-30
Infrastructure to enable container replacement

josecv:
## Description

This PR adds changes to the agent injector and traffic manager to enable the creation of intercepts that replace the application container. The CLI flag is not implemented here.

## Checklist


petergardfjall:
@josecv Is this PR what I think it is, a `telepresence intercept --replace` command?
(That's almost exactly what I wished for [here](https://github.com/telepresenceio/telepresence/issues/1608#issuecomment-1091614626). I'm still stuck on v1 of telepresence since I have an application with work queue semantics).

Would love to see it completed!

rpc/manager/manager.proto
```diff
@@ -75,50 +75,47 @@ message InterceptSpec {
...
+
+  string target_host = 6;
+
+  // The port on the workstation that the intercept is redirected to
+  int32 target_port = 7;
 
   // Same as AgentInfo.Namespace of the Workload
   string namespace = 8;
 
-  // How to decide which subset of requests to that agent to intercept.
```
thallgren: I'm not in favor or reordering based on the gRPC number for each attribute. That number is to ensure backward compatibility when things move around, and it's only used by gRPC internals.

I'm of course fine with reordering attributes, but when we do, then the order should either be alphabetical or based on logical grouping of attributes (both rules broken here, because `mechanism` is moved away from `mechanism_args`).

pkg/agentconfig/sidecar.go
```diff
@@ -153,6 +158,9 @@ type Sidecar struct {
 	// InitResources is the resource requirements for the initContainer sidecar
 	InitResources *core.ResourceRequirements `json:"initResources,omitempty"`
 
+	// User-supplied configuration for the intercept.
+	UserConfig *UserConfig `json:"userConfig,omitempty"`
```
thallgren: A `Sidecar` has many `Containers`, that has many `Intercepts`. The `--replace` flag affects an `Intercept`. So placing this in the `Sidecar` is a bit inconsistent. I realise that the flag must affect all `Intercepts` for a specific `Container`, because the affected `Container` is removed, but why must it affect all `Containers` managed by a `Sidecar`?

I would like to have a `Replace` attribute on the `Container` here, and no `UserInfo`. Once set, that attribute should be considered persistent.

Moving forward (not in this PR) I think it might be a good idea to also enable control over that attribute via an annotation in the workload's pod-template, because this information is really fairly static. It's not something that users are likely to set in different ways. Topic for discussion for sure.

cmd/traffic/cmd/manager/mutator/agent_injector.go
```diff
@@ -230,6 +231,27 @@ func needInitContainer(config *agentconfig.Sidecar) bool {
 	return false
 }
 
+func deleteAppContainer(ctx context.Context, pod *core.Pod, config *agentconfig.Sidecar, patches patchOps) patchOps {
```
thallgren, line 230: This should delete the containers that are marked for replace, not all containers that can be intercepted.

thallgren (changes requested):

pkg/agentconfig/sidecar.go
josecv: The problem is that the injector doesn't know whether or not the workload is currently intercepted. It tries to be idempotent with respect to the configmap. If The attribute is set persistently to true, then how does the agent know whether or not to do the replacement?

cmd/traffic/cmd/manager/service.go
```diff
@@ -443,7 +446,11 @@ func (s *service) PrepareIntercept(ctx context.Context, request *rpc.CreateInter
 	span := trace.SpanFromContext(ctx)
 	tracing.RecordInterceptSpec(span, request.InterceptSpec)
 
-	return s.state.PrepareIntercept(ctx, request)
+	replacePolicy := agentconfig.ReplacePolicyNever
+	if request.InterceptSpec.Replace {
+		replacePolicy = agentconfig.ReplacePolicyInactive
```
thallgren, line 451: Shouldn't this be ReplacePolicyActive?

pkg/agentconfig/sidecar.go
```diff
@@ -49,6 +49,17 @@ const (
...
 )
 
+type ReplacePolicy int
+
+const (
+	// --replace is false.
+	ReplacePolicyNever ReplacePolicy = iota
+	// --replace is true, the intercept is active.
+	ReplacePolicyActive
+	// -replace is true, the intercept is inactive.
```
thallgren:
```suggestion
	// --replace is true, the intercept is inactive.
```

thallgren (changes requested):

cmd/traffic/cmd/manager/service.go
josecv: Nope; it's just the prepare call, the intercept isn't live yet. The CreateIntercept call will then switch it to Active. This way Prepare is just used for the setup of the configmap, i.e. to ensure that it's impossible to later say --replace=false for instance

josecv: If this were Active, I think it'd violate the API for PrepareIntercpept. After all, then it'd be possible to call Prepare without a subsequent Create and effectively bring down the service (since there's no listener for the agent to relay the request to)

thallgren (approved):

petergardfjall:
Any ETA for a `--replace` CLI flag? :star_struck:

PR #3331, opened 2023-08-30
Make the telepresence CLI flags more consistent

thallgren:
## Description

When removing implicit connects, it is safe to assume that calls that follow a `telepresence connect` will use the same context and the same daemon. This makes global options like `--context` and `--docker` redundant. They are only meaningful during connect.

This PR deprecates those options (implicit connect is still allowed, but deprecated), and also introduces a `--name` option for the `telepresence connect` that can be used to override the default generated name for the connection `<kubecontext name>-<namespace>`.  When using docker, this name, prefixed with "tp-", is used for the containerized daemon. A new global option `--use <name match>` is also added so that a connection can be easily identified, should multiple daemons run simultaneously.

## Checklist


CHANGELOG.yml
```diff
@@ -35,6 +35,23 @@ items:
...
+          with commands that accept the full set of Kubernetes flags (e.g. <code>telepresence connect</code>).
+      - type: change
+        title: Deprecation of the --namespace flag for the intercept command.
+        body: >-
+          The <code>--namespace</code> flag is now deprecated for <code>telepresence intercept</code> command. The flag can instead
+          be used with all commands that accept the full set of Kubernetes flags (e.g. <code>telepresence connect</code>).
+      - type: change
         body: >-
           The behavior changed so that a connected Telepresence client is bound to a namespace. The namespace can then
           not be changed unless the client disconnects and reconnects.
```
shepz: Should this block (with no title) be removed since we already talk about it in the first block ?

thallgren: Yes. The block below the added one should be removed. Fixing.

shepz (approved):

PR #3334, opened 2023-09-01
Inject SecurityContext into traffic agent

shepz:
## Description

This PR fix the openshift CFI issue.

In OpenShift, SecurityContextConstraints (SCCs) are assigned to ServiceAccounts (SAs) that run the pods, which means that different applications may not necessarily use the same SCC, resulting in variations in SecurityContexts across applications.

OpenShift sets the SecurityContext during the deployment of the pod. However, utilizing the SecurityContext from an existing container has been found to resolve the issue and reinstate the proper functioning of the traffic agent.

## Checklist

### DIFF
--- a/pkg/agentconfig/container.go
+++ b/pkg/agentconfig/container.go
@@ -141,7 +141,7 @@ func AgentContainer(
 	if r := config.Resources; r != nil {
 		ac.Resources = *r
 	}
-	// use the primary container's constraints to ensure psp compliance for the agent
+	// use the primary container's sc to ensure psp compliance for the agent
 	if sc := pod.Spec.Containers[0].SecurityContext; sc != nil {
 		ac.SecurityContext = sc
 	}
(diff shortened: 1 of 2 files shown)


josecv (approved):

josecv (approved):

pkg/agentconfig/container.go
```diff
@@ -142,6 +142,25 @@ func AgentContainer(
...
+			continue
+		}
+
+		for _, app := range pod.Spec.Containers {
+			if app.Name == cc.Name {
+				if app.SecurityContext != nil {
+					ac.SecurityContext = app.SecurityContext
+					break outerLoop
+				}
+				break
```
thallgren, line 159:
```suggestion
				}
				break outerLoop
```

thallgren (changes requested):
Looks good to me.

thallgren (approved):

jcornick-alabs:
@shepz, it looks like this is all green, lets's get the RC cut as soon as possible today so we can get it to AmDocs.  We would love for them to have been able to try it before their regular meeting tomorrow.

Thanks!!

PR #3343, opened 2023-09-09
Add explicit environment propagation between client and connector

thallgren:
## Description

This commit removes the code that passed environment variables in as kubernetes flags, to instead pass them explicitly where needed.

Also adds propagation of the kubernetes flags to the client when it reconnects, so that those flags can be used for other purposes.

The commit removes the propagation of kubernetes flags to the root daemon. It has no use for them.

## Checklist


knlambert (approved):
LGTM but you forgot to update the changelog :o)

thallgren:
@knlambert I didn't forget the changelog. I just don't think internal fixes like this deserve an entry there. IMO, the changelog should contain things that are of interest to the users.

PR #3352, opened 2023-09-25
Allow caches to have readable permissions.

josecv:
## Description

## Checklist

### DIFF
--- a/pkg/client/cache/cache.go
+++ b/pkg/client/cache/cache.go
@@ -31,11 +31,7 @@ func SaveToUserCache(ctx context.Context, object any, file string, perm Permissi
 	fullFilePath := filepath.Join(filelocation.AppUserCacheDir(ctx), file)
 	// get dir of joined path
 	dir := filepath.Dir(fullFilePath)
-	var dirPerm fs.FileMode = 0o700
-	if perm == Public {
-		dirPerm = 0o755
-	}
-	if err := dos.MkdirAll(ctx, dir, dirPerm); err != nil {
+	if err := dos.MkdirAll(ctx, dir, 0o755); err != nil {
 		return err
 	}
 	return dos.WriteFile(ctx, fullFilePath, jsonContent, (fs.FileMode(perm)))
--- a/pkg/client/cache/cache.go
+++ b/pkg/client/cache/cache.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io/fs"
 	"os"
 	"path/filepath"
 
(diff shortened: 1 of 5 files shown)


pkg/client/userd/trafficmgr/session_info_cache.go
```diff
@@ -27,7 +27,8 @@ func SaveSessionInfoToUserCache(ctx context.Context, daemonID *daemon.Identifier
 		KubeContext: daemonID.KubeContext,
 		Namespace:   daemonID.Namespace,
 		Session:     session,
-	}, sessionInfoFile(daemonID))
+		// Private because the kubecontext has tokens.
+	}, sessionInfoFile(daemonID), cache.Private)
```
thallgren: The `KubeContext` in the `SavedSession` is just the name of the context, not the context itself, so I don't think this is sensitive.

The kubeconfig file that contains tokens is [saved](https://github.com/telepresenceio/telepresence/blob/release/v2/pkg/client/docker/daemon.go#L293C58-L293C58) under the `<user-cache>/kube` folder.

It just strikes me that using /root/.cache/telepresence as the [user cache](https://github.com/telepresenceio/telepresence/blob/release/v2/pkg/client/docker/daemon.go#L48) in the containerized daemon will be weird if the container's user isn't root. I think we should mount that cache in a different folder.

pkg/client/cache/cache.go
```diff
@@ -23,10 +31,14 @@ func SaveToUserCache(ctx context.Context, object any, file string) error {
 	fullFilePath := filepath.Join(filelocation.AppUserCacheDir(ctx), file)
 	// get dir of joined path
 	dir := filepath.Dir(fullFilePath)
-	if err := dos.MkdirAll(ctx, dir, 0o700); err != nil {
+	var dirPerm fs.FileMode = 0o700
+	if perm == Public {
+		dirPerm = 0o755
+	}
+	if err := dos.MkdirAll(ctx, dir, dirPerm); err != nil {
```
thallgren: This is not going to work because we currently save private and public files in the same directory. The `apikeys.json` (saved from pro) must be private for instance. Other files in that directory are not private. We can't have a solution where the first saved file dictates the permissions of the directory.

thallgren (changes requested):

pkg/client/cache/cache.go
josecv: Okay; I'll make the directory readable then, since I don't want to implement the change to have a private directory we mount separately right now

pkg/client/userd/trafficmgr/session_info_cache.go
josecv: okay, then, I'll change that constant then?

thallgren (approved):

PR #3354, opened 2023-09-26
Add certmanager certificate to use with agentInjectorWebhook #3353

### LINKED ISSUE #3353: Support using Certmanager to generate mutator-webhook-tls secret

I use ArgoCD to deploy the Telepresence Helm chart in a Kubernetes cluster. As ArgoCD use the Helm template function to render manifests, the `lookup` function used in the template `agentInjectorWebhook.yaml` does not work as intended. 

The result in ArgoCD is a Telepresence app unsynced every 3 minutes because the TLS certificate is regenerated by Helm. 

Also, when the app is synchronized, the `upgrade-legacy` post-upgrade hook fails to run successfully because the `traffic-manager` pod did not restart to ensure using the newly generated TLS certificate. 

I want to allow using Certmanager features to generate the certificate and use it with the `agentInjectorWebhook`.

Do nothing and wait until ArgoCD include the Helm build including this feature: https://github.com/argoproj/argo-cd/issues/5202. (It has been merged: https://github.com/argoproj/argo-cd/issues/5202#issuecomment-1644432182)

  - Output of `telepresence version`: `v2.15.1`
  - Kubernetes Environment and Version: `v1.26.7`

PLWagner:
## Description

I want to use Certmanager to generate `mutator-webhook-tls` secret. I use ArgoCD and I want to bypass certificate generation by Helm every 3 minutes. The Telepresence app in ArgoCD is always out of sync.

## Checklist


charts/telepresence/values.yaml
```diff
@@ -136,6 +136,25 @@ agentInjector:
...
+    #
+    # Possible options: helm ou certmanager
+    #
+    # Default: helm
+    method: helm
+
+    # The certmanager configuration block
+    #
+    # Default: {}
+    certmanager: {}
```
thallgren: Please add a brief description for these new values to the [charts/telepresence/README.md](https://github.com/telepresenceio/telepresence/blob/release/v2/charts/telepresence/README.md).

charts/telepresence/values.yaml
```diff
@@ -136,6 +136,25 @@ agentInjector:
     name: mutator-webhook-tls
   certificate:
     regenerate: false
+
+    # The method used to generate the TLS certificate for the agent-injector.
+    #
+    # Possible options: helm ou certmanager
```
thallgren:
```suggestion
    # Possible options: helm or certmanager
```

charts/telepresence/templates/certificate.yaml
```diff
@@ -0,0 +1,32 @@
...
+kind: Certificate
+metadata:
+  name: {{ .Values.agentInjector.secret.name }}
+spec:
+  secretName: {{ .Values.agentInjector.secret.name }}
+  commonName: {{ .Values.agentInjector.name }}
+  dnsNames:
+  - {{ (printf "%s.%s" .Values.agentInjector.name .Release.Namespace ) }}
+  - {{ (printf "%s.%s.svc" .Values.agentInjector.name .Release.Namespace ) }}
+  {{- with .Values.agentInjector.certificate.certmanager.duration }}
```
thallgren: Besides renaming `issuer` to `issuerRef` and tweaking the `duration` format, the fields below are just copied from an otherwise anonymous map `.Values.agentInjector.certificate.certmanager`. This raises some questions:
1. Why not start all this with a `{{- with .Values.agentInjector.certificate.certmanager }}` and avoid repeating that over and over?
2. Why is the original duration format different and why using `issuer` instead of `issurerRef`?

I guess the main question is, why not just expand the whole certmanager map here and require that it uses a correct duration format and names to start with?

thallgren (changes requested):
Thanks for providing this.

charts/telepresence/values.yaml
PLWagner: I will, thx!

charts/telepresence/values.yaml
PLWagner: Good catch.

charts/telepresence/templates/certificate.yaml
PLWagner: You did bring an interesting point. I like it. I did submit another commit regarding this. My previous commit has been done with inspiration on the Cilium and ArgoCd Helm chart.

charts/telepresence/README.md
```diff
@@ -24,77 +24,81 @@ Notable chart changes are listed in the [CHANGELOG](./CHANGELOG.md)
...
-| client.connectionTTL                           | The time that the traffic-manager will retain a client connection without any sign of life from the workstation             | `24h`                                                                       |
-| client.routing.alsoProxySubnets                | The virtual network interface of connected clients will also proxy these subnets                                            | `[]`                                                                        |
-| client.routing.neverProxySubnets               | The virtual network interface of connected clients never proxy these subnets                                                | `[]`                                                                        |
-| client.routing.allowConflictingSubnets         | Allow the specified subnets to be routed even if they conflict with other routes on the local machine.                      | `[]`                                                                        |
```
thallgren, line 96: This line seems to have been inadvertently removed (it was added just recently, so perhaps a merge mistake?)

PLWagner: Yes, my mistake. I'll fix this during the day.

thallgren (approved):

thallgren:
@PLWagner Thank you very much for your contribution.

PR #3370, opened 2023-10-03
Fix wrongly sent connect_error metric

knlambert:
## Description

The `connect_error` metric is sent even if the connect was successful. This PR fixes it by checking the error status`UNSPECIFIED`  which means [success](https://github.com/telepresenceio/telepresence/blob/a212c3b84d55c140406a80a99296ff4832c04cb2/rpc/connector/connector.proto#L169).

## Checklist

### DIFF
--- a/pkg/client/userd/trafficmgr/session.go
+++ b/pkg/client/userd/trafficmgr/session.go
@@ -134,21 +134,21 @@ func NewSession(
 	ctx context.Context,
 	cr *rpc.ConnectRequest,
 	config *client.Kubeconfig,
-) (_ context.Context, _ userd.Session, cErr *connector.ConnectInfo) {
+) (_ context.Context, _ userd.Session, ret *connector.ConnectInfo) {
 	dlog.Info(ctx, "-- Starting new session")
 	scout.Report(ctx, "connect")
 
 	defer func() {
-		if cErr != nil {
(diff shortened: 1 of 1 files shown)


thallgren:
So I went ahead and created a fix for this too, before I discovered this PR. It fixes what this PR fixes, but it also [improves the metrics](https://github.com/telepresenceio/telepresence/pull/3372) a bit. What do you think?

pkg/client/userd/trafficmgr/session.go
```diff
@@ -134,21 +134,21 @@ func NewSession(
 	ctx context.Context,
 	cr *rpc.ConnectRequest,
 	config *client.Kubeconfig,
-) (_ context.Context, _ userd.Session, cErr *connector.ConnectInfo) {
+) (_ context.Context, _ userd.Session, ret *connector.ConnectInfo) {
 	dlog.Info(ctx, "-- Starting new session")
 	scout.Report(ctx, "connect")
```
thallgren, line 139: I think we should either report `connect` or `connect_error`. Not both.

thallgren (changes requested):

knlambert:
> So I went ahead and created a fix for this too, before I discovered this PR. It fixes what this PR fixes, but it also [improves the metrics](https://github.com/telepresenceio/telepresence/pull/3372) a bit. What do you think?


Let's go with your :)

PR #3386, opened 2023-10-14
Configurable images

thallgren:
## Description

This PR ensures that all Telepresence images (client in docker mode, traffic-manager, and traffic-agent) are configurable. The change is mainly intended to make the integration test framework more flexible.

The PR also incorporates some commits added by dependabot.

## Checklist


integration_test/itest/image.go
```diff
@@ -1,13 +1,54 @@
...
+	nb := strings.Builder{}
+	if img.Registry != "" {
+		nb.WriteString(img.Registry)
+		nb.WriteByte('/')
+	}
+	nb.WriteString(img.Name)
+	if img.Tag != "" {
+		nb.WriteByte(':')
+		nb.WriteString(img.Tag)
+	}
```
josecv, line 26: can we write `:latest` here if no tag? i just like the explicitness of it; sometimes you run into a repo that doesn't have a latest tag and it's confusing as hell to get a message like `dockerhub has no mysql image`

integration_test/itest/image.go
```diff
@@ -1,13 +1,54 @@
...
+			img.Name = imgQN[:i]
+			img.Tag = imgQN[i+1:]
+		} else {
+			img.Name = imgQN
+			img.Tag = defaultTag
+		}
+		return img
+	}
+	return nil
+}
```
josecv, line 50: are there docker client libraries we can maybe use to parse this? i just get queasy at this kind of character-wise manipulation of a formatted string

josecv (approved):
looks gucci, just a couple of nits

PR #3388, opened 2023-10-16
Push client image

shepz:
## Description

## Checklist

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -41,6 +41,7 @@ jobs:
         run: |
           docker login -u="${{ secrets.DOCKERHUB_USERNAME }}" -p="${{ secrets.DOCKERHUB_PASSWORD }}"
           make push-tel2-image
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yaml
```diff
@@ -41,6 +41,7 @@ jobs:
         run: |
           docker login -u="${{ secrets.DOCKERHUB_USERNAME }}" -p="${{ secrets.DOCKERHUB_PASSWORD }}"
           make push-tel2-image
+          make push-client-image-x
```
thallgren: This is great, but it seems a tad inconsistent to push buildx images for the client while only pushing a plain build image for the traffic-manager and traffic-agent.

Perhaps add a the `-x` variants to the `tel2-image` too,  and  then a `push-tel2-image-x` here? I know that would close a couple of issues complaining about lack of arm support in the cluster.

thallgren (approved):

PR #3389, opened 2023-10-17
Improve outbound traffic performance.

thallgren:
## Description

The root-daemon now communicates directly with the traffic-manager
instead of routing all outbound traffic through the user-daemon. The
root-daemon uses a patched kubeconfig where `exec` configurations to
obtain credentials are dispatched to the user-daemon. This to ensure
that all authentication plugins will execute by the user and not by root. The old
behavior or routing everything through the user-daemon can be restored
by setting `cluster.connectFromUserDaemon` to `true` in `config.yml`

## Checklist


CHANGELOG.yml
```diff
@@ -40,6 +40,20 @@ items:
           the setting <code>images.clientImage</code> and will default first to the value of the environment <code>
           TELEPRESENCE_CLIENT_IMAGE</code>, and then to the value preset by the telepresence binary. This
           configuration setting is primarily intended for testing purposes.
+      - type: feature
+        title: Improve outbound traffic performance.
+        body: >-
+          The root-daemon now communicates directly with the traffic-manager instead of routing all outbound traffic
+          through the user-daemon. The root-daemon uses a patched kubeconfig where <code>exec</code> configurations to
+          obtain credentials are dispatched to the user-daemon. This to ensure that all authentication plugins will
+          execute in user-space. The old behavior or routing everything through the user-daemon can be restored by
```
sarabrajsingh:
```suggestion
          execute in user-space. The old behavior of routing everything through the user-daemon can be restored by
```

sarabrajsingh (approved):

PR #3397, opened 2023-10-24
Introduce direct port-forwards for outbound connections

thallgren:
## Description

An outbound connection established using the TUN-device would typically use the existing port-forward to the the traffic-manager to create a tunnel. The tunnel would route traffic through the traffic-manager to a traffic-agent, which would then connect to the final destination.

This PR adds what's needed to instead create a new port-forward directly to the destination using the following flow:

1. A TCP connect request arrives to the TUN device.
2. The IP and port are sent to the traffic-manager for resolution.
3. The traffic-manager resolves the IP to a service or a pod.
4. If the IP resolved to a service, then the traffic-manager finds the selected pod.
5. The name, namespace, and port to port-forward to is returned to the client.
6. The client creates the port-forward.

If the attempt to create a port-forward fails at some point, then the TUN-device will fall back to its old behavior and instead create a tunnel.

The direct port-forward feature is configurable using the `config.yml` setting `cluster.directPortForward`. It's default setting is `true`.

## Checklist


knlambert (approved):

thallgren:
Closing this because the solution doesn't work as expected. A port forward to a pod apparently bypasses iptable rules on that pod.

PR #3402, opened 2023-10-26
Add an --allow-conflicting-subnets flag locally.

josecv:
This was only configurable in the traffic-manager before, but we've had user feedback that sometimes the persona setting up that traffic-manager is not the same as the persona setting up VPNs, and different users connecting to the same traffic manager may have different routes. So it's been exposed as a CLI flag.

## Checklist


sarabrajsingh (approved):
lgtm!

PR #3408, opened 2023-10-30
Add prometheus metrics to track intercepts usage

shepz:
## Description

Add prometheus metrics to track intercepts and connect usage 

## Checklist


cmd/traffic/cmd/manager/service.go
```diff
@@ -530,13 +586,16 @@ func (s *service) RemoveIntercept(ctx context.Context, riReq *rpc.RemoveIntercep
 	ctx = managerutil.WithSessionInfo(ctx, riReq.GetSession())
 	sessionID := riReq.GetSession().GetSessionId()
 	name := riReq.Name
+	client := s.state.GetClient(sessionID)
```
knlambert, line 632: Maybe reuse that one for line 593

cmd/traffic/cmd/manager/service.go
```diff
@@ -149,6 +186,9 @@ func (s *service) ArriveAsClient(ctx context.Context, client *rpc.ClientInfo) (*
 	}
 
 	installId := client.GetInstallId()
+
+	s.metrics.ConnectCounter.With(prometheus.Labels{"client": client.Name, "install_id": client.InstallId}).Inc()
```
knlambert: Maybe I would just attach methods to that metric object so it's cleaner to read for all functions

Like s.metrics.OnConnect(client.Name, client.InstallID)

knlambert (dismissed):
I think this code handles the graceful cases.

You have to also deal with 
* Connect expires
* Intercept expires

cmd/traffic/cmd/manager/service.go
```diff
@@ -69,6 +72,53 @@ type service struct {
 
 var _ rpc.ManagerServer = &service{}
 
+type Metrics struct {
+	InterceptGlobalCounter     *prometheus.CounterVec
+	InterceptPersonalCounter   *prometheus.CounterVec
```
thallgren: There's no such thing as personal intercepts in OSS.

cmd/traffic/cmd/manager/service.go
```diff
@@ -500,6 +562,14 @@ func (s *service) CreateIntercept(ctx context.Context, ciReq *rpc.CreateIntercep
 		}
 	}
 
+	SetGauge(s.metrics.InterceptActiveStatusGauge, client.Name, client.InstallId, spec.Name, 1)
+
+	if spec.Mechanism == "tcp" {
+		IncrementCounter(s.metrics.InterceptGlobalCounter, client.Name, client.InstallId)
+	} else {
+		IncrementCounter(s.metrics.InterceptPersonalCounter, client.Name, client.InstallId)
+	}
```
thallgren: This distinction is not valid i OSS. A non "tcp" mechanism doesn't mean that it's personal (because there's no such thing).

thallgren (changes requested):
Shouldn't this be added to pro?

thallgren:
Why is this PR establishing a completely new pattern for Prometheus gauges instead of following the pattern we already have. I'm adding new gauges to that in my PR [here](https://github.com/telepresenceio/telepresence/blob/thallgren/bypass-tm/cmd/traffic/cmd/manager/manager.go#L126-L154).

cmd/traffic/cmd/manager/state/state.go
```diff
@@ -86,14 +89,21 @@ type interceptFinalizerCall struct {
 	ctx   context.Context
 }
 
+type (
```
thallgren, line 101: We don't usually group types. I suspect that if you run `make format`, this grouping will be removed.

cmd/traffic/cmd/manager/manager.go
```diff
@@ -152,6 +195,21 @@ func (s *service) servePrometheus(ctx context.Context) error {
...
 	})
 
+	labels := []string{"client", "install_id"}
+	InterceptCounter = newCounterVecFunc("intercept_count", "The total number of global intercepts by user", labels)
+	InterceptActiveStatusGauge = newGaugeVecFunc("intercept_active_status",
+		"Flag to indicate when an intercept is active. 1 for active, 0 for not active.", append(labels, "workload"))
+	ConnectCounter = newCounterVecFunc("connect_count", "The total number of connects by user", labels)
+	ConnectActiveStatusGauge = newGaugeVecFunc("connect_active_status", "Flag to indicate when a connect is active. 1 for active, 0 for not active.", labels)
+
+	s.state.SetConnectionEstablishedFinalizer(func(client, installId string) {
```
thallgren: The use of the term "connection" is somewhat ambiguous and the term "established" here seems a bit redundant. Could this be named `SetAllClientSessionsFinalizer`? Perhaps pass the `*rpc.ClientInfo` instead of passing its name and install-id explicitly. Less to change, should we want to use some other info.

cmd/traffic/cmd/manager/manager.go
```diff
@@ -152,6 +195,21 @@ func (s *service) servePrometheus(ctx context.Context) error {
...
+		SetGauge(ConnectActiveStatusGauge, client, installId, nil, 0)
+	})
+
+	s.state.SetInterceptEstablishedFinalizer(func(client, installId string, workload *string) {
```
thallgren: and this one `SetAllInterceptsFinalizer` and also get a `*rpc.ClientInfo` rather than a name + install-id.

cmd/traffic/cmd/manager/manager.go
```diff
@@ -35,6 +35,10 @@ var (
 	DisplayName                 = "OSS Traffic Manager"               //nolint:gochecknoglobals // extension point
 	NewServiceFunc              = NewService                          //nolint:gochecknoglobals // extension point
 	WithAgentImageRetrieverFunc = managerutil.WithAgentImageRetriever //nolint:gochecknoglobals // extension point
+	InterceptCounter            *prometheus.CounterVec                //nolint:gochecknoglobals // prometheus metric
+	InterceptActiveStatusGauge  *prometheus.GaugeVec                  //nolint:gochecknoglobals // prometheus metric
+	ConnectCounter              *prometheus.CounterVec                //nolint:gochecknoglobals // prometheus metric
+	ConnectActiveStatusGauge    *prometheus.GaugeVec                  //nolint:gochecknoglobals // prometheus metric
```
thallgren: Why must these be globals and not contained in the state like other counters?

thallgren (changes requested):

cmd/traffic/cmd/manager/state/state.go
shepz: hum, when I don't group them. The lint is [failing](https://github.com/telepresenceio/telepresence/actions/runs/6748003557/job/18345398661)

thallgren (approved):

PR #3409, opened 2023-10-31
Use traffic-agent port-forwards for outbound and intercepted traffic

thallgren:
Use traffic-agent port-forwards for outbound and intercepted traffic.

The telepresence TUN-device is now capable of establishing direct port-forwards to a traffic-agent in the connected namespace. That port-forward is then used for all outbound traffic to the device, and also for all traffic that arrives from intercepted workloads. Getting rid of the extra hop via the traffic-manager improves performance and reduces the load on the traffic-manager. The feature can only be used if the client has Kubernetes port-forward permissions to the connected namespace. It can be disabled by setting `cluster.agentPortForward` to `false` in `config.yml`.


cmd/traffic/cmd/manager/manager.go
```diff
@@ -115,24 +115,27 @@ func MainWithEnv(ctx context.Context) error {
...
-			Help: h,
-		}, func() float64 { return float64(f()) })
-	}
 	newGaugeFunc("agent_count", "Number of connected traffic agents", s.state.CountAgents)
 	newGaugeFunc("client_count", "Number of connected clients", s.state.CountClients)
 	newGaugeFunc("intercept_count", "Number of active intercepts", s.state.CountIntercepts)
 	newGaugeFunc("session_count", "Number of sessions", s.state.CountSessions)
 	newGaugeFunc("tunnel_count", "Number of tunnels", s.state.CountTunnels)
+	newGaugeFunc("tunnel_ingress_bytes", "Number of bytes tunneled from clients", s.state.CountTunnelIngress)
+	newGaugeFunc("tunnel_egress_bytes", "Number bytes tunneled to clients", s.state.CountTunnelEgress)
```
josecv: these should be counters since they only go up. There are different promql semantics for gauges and counters, and we wanna make sure we're doing what's standard

cmd/traffic/cmd/manager/service.go
```diff
@@ -212,10 +212,86 @@ func (s *service) Depart(ctx context.Context, session *rpc.SessionInfo) (*empty.
...
+	interceptsCh := s.state.WatchIntercepts(ctx, func(_ string, info *rpc.InterceptInfo) bool {
+		return info.ClientSession.SessionId == clientSession
+	})
+	sessionDone, err := s.state.SessionDone(clientSession)
+	if err != nil {
+		return err
+	}
 
+	var interceptInfos map[string]*rpc.InterceptInfo
+	intercepted := func(name, namespace string) bool {
```
josecv: can we call this `isIntercepted` please?

cmd/traffic/cmd/agent/server.go
```diff
@@ -0,0 +1,125 @@
...
+// for logging (it can be cancelled). Errors are logged but not fatal.
+func (s *state) ReportMetrics(ctx context.Context, metrics *rpc.TunnelMetrics) {
+	go func() {
+		mCtx, mCancel := context.WithTimeout(context.WithoutCancel(ctx), time.Second)
+		defer mCancel()
+		_, err := s.manager.ReportMetrics(mCtx, metrics)
+		if err != nil {
+			dlog.Errorf(ctx, "ReportMetrics failed: %v", err)
+		}
+	}()
```
josecv, line 124: could we kick off this goroutine when we instantiate the state and then talk to it through a channel? that way we're not creating a new goroutine here for every single request. If you give the channel a big old buffer and then just do a select with a default to write to it then it won't ever block

josecv (changes requested):
mainly the prometheus counters

cmd/traffic/cmd/agent/server.go
thallgren: This function is called once of when a channel closes and is supposed to be very short-lived. The go routine is not strictly necessary. I just didn't want things to wait for the traffic-manager to accept the metrics. I'd like to keep it as is.

cmd/traffic/cmd/manager/manager.go
thallgren: I guess that goes for all of them, not just the ones I added here?

thallgren:
@josecv only the last two commits were added after your review. The rest is a rebase on top of release/v2.

cmd/traffic/cmd/manager/service.go
```diff
@@ -212,10 +212,86 @@ func (s *service) Depart(ctx context.Context, session *rpc.SessionInfo) (*empty.
...
+			}
+		case is, ok := <-interceptsCh:
+			if !ok {
+				return nil
+			}
+			interceptInfos = is.State
+			for i, a := range agents {
+				a.Intercepted = intercepted(agentNames[i], a.Namespace)
+			}
+		}
```
njayp: should there be a `case <-ctx.Done()` here?

njayp (approved):

cmd/traffic/cmd/manager/service.go
thallgren: I don't think that's necessary given that both the `agentsCh` and `interceptsCh` will be closed as soon as the context is cancelled. We don't look for `<-ctx.Done()` in `WatchAgents` and `WatchAgentsNS` either, for the same reason.

josecv (approved):

PR #3438, opened 2023-11-24
The DNS search path on Windows is now restored when Telepresence quits

thallgren:
## Description

The DNS search path that Telepresence uses to simulate the DNS lookup functionality in the connected cluster namespace was not removed by a `telepresence quit`, resulting in connectivity problems from the workstation. Telepresence will now remove the entries that it has added to the search list when it quits.

## Checklist

### DIFF
--- a/pkg/vif/device_windows.go
+++ b/pkg/vif/device_windows.go
@@ -203,22 +203,26 @@ func (t *nativeDevice) setDNS(ctx context.Context, clusterDomain string, server 
 	if err != nil {
 		return err
 	}
-	// Windows does not use a dot suffix in the search path.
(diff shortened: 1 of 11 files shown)


pkg/vif/device_windows.go
LanceEa, line 1: Do we have any test coverage for this behavior? I'm guessing since it is windows, we would need to have a windows CI environment setup which I'm not sure if we do.

thallgren: Good point. We don't have any tests to verify that the search path is restored, and we do have a CI-environment. I'll add a test.

LanceEa (approved):
LGTM

PR #3446, opened 2023-12-04
Keep telepresence-agents configmap in sync with pod states.

### LINKED ISSUE #3441: Failure to inject traffic-agent leaves the intercept configuration in a bad state.
The first time a user calls `telepresence intercept xxx`, the traffic-manager will inject an entry for the workload "xxx" into the "telepresence-agents" configmap. This triggers the mutating webhook to call on the agent-injector to add the traffic-agent sidecar to the intercepted pod(s). If this injection fails for some reason, then the entry in the "telepresence-agents" will still be present. If the user now tries to intercept again, then the configmap will not change, and hence no attempt will be made to inject an agent. The intercept attempt then times out.

The quick-fix to make the intercept work again is to do `telepresence uninstall --agent xxx`.

This must be fixed so that Telepresence recovers from this bad state.

thallgren:
An intercept attempt that resulted in a timeout due to failure of injecting the traffic-agent left the `telepresence-agents` configmap with a state that indicated that an agent had been added.

The next time an intercept was made, presumably after fixing what caused the failure, the traffic-manager would detect the existing entry in the configmap and leave it untouched and hence, not trigger the necessary rollout needed for the injection to take place.

This commit ensures that the configmap entry is deleted when there is a failure to inject the traffic-agent. This keeps the configmap in sync with the actual state of the pods.

Closes #3441


cmd/traffic/cmd/manager/state/intercept.go
```diff
@@ -88,6 +88,11 @@ func (s *state) PrepareIntercept(
 		return interceptError(err)
 	}
 	if err = s.waitForAgent(ctx, ac.AgentName, ac.Namespace, failedCreateCh); err != nil {
+		// If no agent arrives, then drop its entry from the configmap. This ensures that there
```
LanceEa, line 91: Will the `err` returned here potentially include the timeout error? 

If so, is there a chance the `waitForAgent` times out, the AgentConfig is dropped but then the Agent comes available anyways? If that happened does that put us in a bad state or it doesn't matter at that point or would get cleaned up by some other reconciliation?

_this could be a silly question, due to my lack of knowledge on these components._

LanceEa (approved):

cmd/traffic/cmd/manager/state/intercept.go
thallgren: If there's a timeout, then the `err` here will have the `DealineExceeded` status set. It will be propagated to the user with a suggestion to take a look at the events generated in the intercepted namespace (failure to pull the image can be one reason, another has been tls errors between the webhook and the injector, in which case there will be no events at all).

The worst thing that can happen if we remove the `AgentConfig` is that it will be regenerated on the next intercept attempt. That's then taken care of by the new `isRolloutNeeded()` included in this PR which will discover the existing agent, skip the rollout, and then we're in sync again.

LanceEa: Sounds good, thanks for explaining!

PR #3448, opened 2023-12-06
Improvements related to use of multiple connections

thallgren:
This PR contains a set of bug-fixes and improvements that became necessary by the introduction of the ability to use multiple `telepresence connect`.

Our CI was a bit outdated so the last two commits are improvements in that area. Please note that those changes cannot be tested by this PR because our CI uses `pull_request_target`, so I created a [separate PR](/telepresenceio/telepresence/pull/3449) for that with an empty commit that targets this PR.


pkg/client/cli/cmd/quit.go
```diff
@@ -1,52 +1,32 @@
...
+					return err
+				}
 				connect.Disconnect(cmd.Context())
 			}
 			return nil
 		},
 	}
 	flags := cmd.Flags()
 	flags.BoolVarP(&quitDaemons, "stop-daemons", "s", false, "stop all local telepresence daemons")
-	flags.BoolVarP(&quitRootDaemon, "root-daemon", "r", false, "stop daemons")
```
LanceEa, line 45: In the release notes, we didn't have any thing about removing these flags, what is the typical deprecation policy/process for these?

pkg/client/cli/cmd/status.go
```diff
@@ -354,12 +533,17 @@ func (cs *UserDaemonStatus) print(kvf *ioutil.KeyValueFormatter) {
 	if cs.ContainerNetwork != "" {
 		kvf.Add("Container network", cs.ContainerNetwork)
 	}
-	kvf.Add("Connection name", cs.ConnectionName)
```
LanceEa, line 357: Since this is being removed, is it replaced by something? 

Is the  `Name` on the `TrafficManagerStatus` replace the need for it? Or, does the new hostname provide a better description?

LanceEa (approved):
lgtm, just left two questions on removals to make sure we are handling the removals accordingly.

Alice-Lilith (approved):

pkg/client/cli/cmd/quit.go
thallgren: Whenever we want to deprecate a flag, we ensure that a deprecation warning is printed. We then wait at least two minor releases before removing it. So a flag deprecated in 2.17.x will not be removed until 2.20.x.

These two flags have been deprecated since version 2.8.0 (released October last year) .

pkg/client/cli/cmd/status.go
thallgren: Yes, the `Name` replaces the `ConnectionName`. This is Telepresence name for the daemon, and even a host-based daemon will have a name.

The `hostname` is the actual hostname used by a container (and returned by the shell `hostname` command). Docker will provide a default (the container ID IIRC) unless it's provided by the user.

Issue #1646, opened 2021-04-07
[v2] Add a way to stop the main container


LukeShu:
Some workloads (such as controllers) don't respond to incoming traffic, but instead are oriented around a `watch` to the Kubernetes apiserver.  For these workloads, you want to actually stop the main container and only run the copy on the laptop.

This use-case is well-served by Telepresence 1 swap-deployment.

As I write this, an idea occurs to me: It might be worth it to override the main pod's `KUBERNETES_SERVICE_HOST` and `KUBERNETES_SERVICE_PORT` env-vars and point them at our own apiserver proxy, this way the main container could keep running but we can stop feeding it events while an intercept is active.  However, this only addresses the workloads that get their input from the apiserver; it wouldn't address controllers that get their input from elsewhere (perhaps from Consul).

thallgren:
Perhaps this is also the solution needed to implement the T1 option `--from-pod <port>`? Not sure what happens if we simply override ports that the main container is listening to though. I think it must be stopped to avoid conflicts.

thallgren:
Relates to #1708.

jamesbattersby:
I have the same situation where the container I am debugging does things that are not triggered by incoming traffic, but from timers or other events.  I would need this to stop, so it doesn't interfere with my local debugging version.  In V1, swap-deployment was ideal.

olegsu:
This is a good feature I used with telepresence 1 and miss in the new version.
In my use case I have an application that pulls data from DB or other 3rd parties APIs, so just overriding the default Kubernetes env vars won't help, but a most complete solution required I guess.

olegsu:
Hey, I see this [comment](https://github.com/telepresenceio/telepresence/issues/1708#issuecomment-876701156) from #1708 saying that this was released. In fact in the case that your application does not expose service this is not working. 
Do you guys have a plan to support this at some point? I love working with telepresence 2 but cant really continue to work with both version :-(

thallgren:
@olegsu not sure I understand. #1708 is about the `--to-pod` flag. Not about stopping the main container.

olegsu:
@thallgren Thanks for the clarification.
I am still wondering how to use Telepresence 2 to develop Kubernetes Controller for example :-)

scottillogical:
[...]

PR #3469, opened 2023-12-28
Add --replace flag to intercept command

petergardfjall:
## Description

This PR adds a `--replace` flag to the `telepresence intercept` command which, when specified, will replace application containers in the intercepted workload.

This PR builds on the server-side machinery implemented in https://github.com/telepresenceio/telepresence/pull/3330.

Having the possibility to replace workload containers should provide a solution for tickets like https://github.com/telepresenceio/telepresence/issues/1646 and https://github.com/telepresenceio/telepresence/issues/1608.

## Checklist

### DIFF
--- a/integration_test/intercept_flags_test.go
+++ b/integration_test/intercept_flags_test.go
@@ -1,0 +1,72 @@
+package integration_test
+
+import (
+	"regexp"
+	"strconv"
+	"time"
+
+	"github.com/datawire/dlib/dlog"
+	"github.com/telepresenceio/telepresence/v2/integration_test/itest"
+)
+
+type interceptFlagSuite struct {
+	itest.Suite
+	itest.SingleService
+}
+
+func (s *interceptFlagSuite) SuiteName() string {
+	return "InterceptFlag"
+}
+
+func init() {
+	itest.AddSingleServiceSuite("-intercept-flag", "echo", func(h itest.SingleService) itest.TestingSuite {
+		return &interceptFlagSuite{Suite: itest.Suite{Harness: h}, SingleService: h}
+	})
+}
+
+func (s *interceptFlagSuite) SetupSuite() {
+	if s.CompatVersion() != "" {
+		s.T().Skip("Not part of compatibility suite")
+	}
+	s.Suite.SetupSuite()
+}
+
+func (s *interceptFlagSuite) Test_ContainerReplace() {
+	ctx := s.Context()
+	require := s.Require()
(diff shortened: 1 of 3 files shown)


thallgren:
Hi @petergardfjall . Thanks for providing this. Can you please sign your commit and also please fix the [lint complaint](https://github.com/telepresenceio/telepresence/actions/runs/7346076447/job/20000253186?pr=3469)?

petergardfjall:
> Hi @petergardfjall . Thanks for providing this. Can you please sign your commit and also please fix the [lint complaint](https://github.com/telepresenceio/telepresence/actions/runs/7346076447/job/20000253186?pr=3469)?

Hi @thallgren, I've addressed your comments.

One thing I noted is that in its present shape `--replace` won't preserve volume mounts of the original workload pod at the local mount point. I suppose that is to be expected since the `replace` will simply delete the workload pod (it is implemented by the [deleteAppContainer](https://github.com/telepresenceio/telepresence/blob/3155355294f4dfe1fa3762b83decd13ee4491c65/cmd/traffic/cmd/manager/mutator/agent_injector.go#L199) patch if I'm reading it right). I suppose that patch would need to do finer-grained surgery on the application containers if the intercept is to preserve volume mounts. Perhaps something along the lines of:

- replace `image` with `k8s.gcr.io/pause` (or similar)
- remove `readinessProbe` and `livenessProbe`

That is perhaps a bit more similar in spirit to the `swap-deployment` command in Telepresence v1?

Either way I think that should not pollute this PR.
If you can provide some guidance I might take a stab at it in a follow-up PR.

thallgren:
Agree about not polluting this PR. If mounts doesn't work, then that's a separate issue. Not sure why that doesn't work. It really should, even when the container is replaced.

The `readinessProbe`/`livenessProbe` intended for the deleted container is also something that needs to be fixed. Not sure about replacing the image with `k8s.gcr.io/pause` though. The agent really should take over whatever it's doing (including mounts).

thallgren:
This option is already present in our enterprise edition, so rather than asking you to provide an integration test, I pushed a commit to your branch with the test that we have already have.

thallgren (approved):

petergardfjall:
> Agree about not polluting this PR. If mounts doesn't work, then that's a separate issue.

Opened an issue https://github.com/telepresenceio/telepresence/issues/3483 with a proposed fix in https://github.com/telepresenceio/telepresence/pull/3484.

PR #3476, opened 2024-01-05
Fix traffic-manager graceful shutdown

lazybetrayer:
Currently, when the `traffic-manager` pod is terminated, the http server is not closed properly, causing it to block indefinitely, and will eventually be killed after the grace period.

`Shutdown` should be called and waited to gracefully shut down the server, and there's no need to close the listener since `Serve`  always closes it.


cmd/traffic/cmd/manager/mutator/service.go
```diff
@@ -186,17 +185,30 @@ func serveAndWatchTLS(ctx context.Context, s *http.Server, addr string) error {
...
+	}()
+
+	<-ctx.Done()
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*30)
+	defer cancel()
+	if err := s.Shutdown(ctx); err != nil {
+		dlog.Errorf(ctx, "error shutting down http server: %v", err)
+		return err
+	}
+	return nil
```
thallgren: This change seems like overkill to me, and it does not take into consideration that we already have a Hard/Soft context setup active (hard to know, but it stems from the dgroup.Group returned from the `NewService` func in manager/service.go).

It should be enough to just add:
```
	go func() {
		<-ctx.Done()
		_ = s.Shutdown(dcontext.HardContext(ctx))
	}()
```
just prior to the call to `s.Serve()`. The good thing about doing that is that any error from `s.Serve()` will then be propagated to the `Group`, and hence also gracefully terminate the other routines started by that group.

thallgren: Another problem with this change is this:
```
ctx, cancel := context.WithTimeout(context.Background(), time.Second*30)
```
because that effectively removes all context values. One side effect of doing that is that logging that the subsequent `dlog.Errorf` is using, won't be configured. We use `context.WithoutCancel(ctx)` instead of `context.Background()` for situations like this.

thallgren (changes requested):
Looks good on most part, but the actual shutdown logic can be improved.

lazybetrayer:
Hi @thallgren, I have updated the code, please take a look. 
I also add a channel to wait for `Shutdown` to return according to the [doc](https://pkg.go.dev/net/http#Server.Shutdown)
> When Shutdown is called, Serve, ListenAndServe, and ListenAndServeTLS immediately return ErrServerClosed. Make sure the program doesn't exit and waits instead for Shutdown to return.

What do you think? Thanks.

thallgren (approved):

thallgren:
Even better than the changes I proposed. Thanks!

PR #3484, opened 2024-01-12
Make sure volume mounts are properly mounted on the laptop for replace intercepts.

### LINKED ISSUE #3483: Volume mounts are broken for replace intercepts
**Describe the bug**

When doing a replace intercept (`telepresence intercept --replace`) volume mounts are not properly "forwarded" to the laptop.

An attempt to solve the issue is provided in https://github.com/telepresenceio/telepresence/pull/3484

**To Reproduce**

1. When I run (NOTE: [sample-deployment.yaml](https://github.com/telepresenceio/telepresence/files/13922965/sample-deployment.yaml.txt) is attached to the issue):

   ```shell
   $ kind create cluster  --name tp

   # ... Wait for cluster to come up ...

   #
   # Deploy sample application.
   # NOTE: sample-deployment.yaml is attached to the issue.
   #
   $ kubectl apply -f sample-deployment.yaml

   #
   # Build telepresence from latest release/v2 commit
   #
   $ export TELEPRESENCE_REGISTRY=docker.io/<myaccount>
   $ make protoc build push-images

   #
   # Use docker registry where we published image.
   #
   $ cat > ~/.config/telepresence/config.yml <<EOF
   images:
     registry: docker.io/<myaccount>
   EOF

   #
   # Install traffic-manager.
   #
   $ ./build-output/bin/telepresence helm install --set image.pullPolicy=Always
   $ ./build-output/bin/telepresence connect
   $ ./build-output/bin/telepresence loglevel debug

   #
   # Intercept workload.
   #
   $ ./build-output/bin/telepresence intercept --replace my-web-server
   Using Deployment my-web-server
   Intercept name         : my-web-server
   State                  : ACTIVE
   Workload kind          : Deployment
   Destination            : 127.0.0.1:8080
   Service Port Identifier: http
   Volume Mount Point     : /tmp/telfs-2373721965
   Intercepting           : all TCP connections

   #
   # NOTE: volume mount is empty!
   #
   $ ls -al /tmp/telfs-2373721965
   total 0
   ```

2. I see that the local volume mount did not mount the volume. I expected to see:

   ```shell
   $ ls -al /tmp/telfs-2373721965
   total 8
   drwxr-xr-x 1 root root 4096 jan 12 19:23 etc
   drwxr-xr-x 1 root root 4096 jan 12 19:23 var
   ```

   But instead got:

   ```shell
   $ ls -al /tmp/telfs-2373721965
   total 0
   ```

3. I have the full error logs (`telepresence gather-logs`) locally. Let me know if they are needed.

**Expected behavior**

I expected the volumes mounted on the replaced workload containers to be present
on the local volume mount point.

- Output of `telepresence version`: the latest `release/v2` commit 443119746fc2ad3d13e33a00ce1532be092a942e
- Operating system of workstation running `telepresence` commands: `Ubuntu 22.04.3 LTS`
- Kubernetes environment and Version: `kind v0.17.0 go1.21.2 linux/amd64`

**Additional context**

I submitted a PR [#3484](https://github.com/telepresenceio/telepresence/pull/3484) which includes an integration test case that reproduces the issue when run against the main `release/v2` branch. That PR is an attempt at fixing the issue, but I'm not familiar enough with the code base (nor all possible usage scenarios of Telepresence) to know if the fix is sound (it passes the aforementioned test-case but I may have, unintentionally, broken something else).

petergardfjall:
## Description

An attempt to fix https://github.com/telepresenceio/telepresence/issues/3483

The biggest contribution of the PR might be an integration test case that reproduces the issue (and fails for `release/v2`).

I'm uncertain about the soundness of the attempted fix. I'm not intimately familiar with the Telepresence codebase and don't fully understand the ins and outs of Kubernetes admission hooks. But at least it makes the code pass the provided test case.

## Checklist

### DIFF
--- a/cmd/traffic/cmd/manager/mutator/agent_injector.go
+++ b/cmd/traffic/cmd/manager/mutator/agent_injector.go
@@ -191,11 +191,16 @@ func (a *agentInjector) Inject(ctx context.Context, req *admission.AdmissionRequ
 		return nil, fmt.Errorf("invalid value %q for annotation %s", ia, agentconfig.InjectAnnotation)
 	}
 
+	config := scx.AgentConfig()
+	if agentAlreadyInjected(ctx, pod, config) {
+		dlog.Infof(ctx, "Pod %s.%s already has container %s", pod.Name, pod.Namespace, agentconfig.ContainerName)
+		return nil, nil
+	}
+
 	// Create patch operations to add the traffic-agent sidecar
 	dlog.Infof(ctx, "Injecting %s into pod %s.%s", agentconfig.ContainerName, pod.Name, pod.Namespace)
 
 	var patches PatchOps
-	config := scx.AgentConfig()
 	patches = deleteAppContainer(ctx, pod, config, patches)
 	patches = addInitContainer(pod, config, patches)
 	patches = addAgentContainer(ctx, pod, config, patches)
@@ -224,6 +229,19 @@ func (a *agentInjector) Inject(ctx context.Context, req *admission.AdmissionRequ
 func (a *agentInjector) Uninstall(ctx context.Context) {
 	atomic.StoreInt64(&a.terminating, 1)
 	a.agentConfigs.DeleteMapsAndRolloutAll(ctx)
+}
+
+func agentAlreadyInjected(ctx context.Context, pod *core.Pod, config *agentconfig.Sidecar) bool {
+	acn := agentconfig.AgentContainer(ctx, pod, config)
+	if acn == nil {
+		return false
+	}
+	for _, pcn := range pod.Spec.Containers {
+		if pcn.Name == agentconfig.ContainerName {
+			return true
+		}
+	}
+	return false
 }
 
 func needInitContainer(config *agentconfig.Sidecar) bool {
--- a/integration_test/intercept_mount_test.go
+++ b/integration_test/intercept_mount_test.go
@@ -134,6 +134,45 @@ func (s *singleServiceSuite) Test_InterceptMountRelative() {
 	require.True(st.IsDir(), "<mount point>/var is not a directory")
 }
 
+// Verifies proper forwarding of volume mounts when running a telepresence intercept with the
+// "--replace" flag.
+func (s *singleServiceSuite) Test_InterceptMountReplace() {
+	if runtime.GOOS == "darwin" {
+		s.T().Skip("Mount tests don't run on darwin due to macFUSE issues")
+	}
+	if runtime.GOOS == "windows" {
+		s.T().Skip("Windows mount on driver letters. Relative mounts are not possible")
+	}
+	require := s.Require()
+
+	ctx := s.Context()
+	port, cancel := itest.StartLocalHttpEchoServer(ctx, s.ServiceName())
+	defer cancel()
+
+	mountPoint, err := os.MkdirTemp("", "mount-") // Don't use the testing.Tempdir() because deletion is delayed.
+	require.NoError(err)
+	ctx = itest.WithWorkingDir(ctx, mountPoint)
+	stdout := itest.TelepresenceOk(ctx,
+		"intercept", "--replace", s.ServiceName(), "--mount", mountPoint, "--port", strconv.Itoa(port))
+	defer func() {
+		itest.TelepresenceOk(ctx, "leave", s.ServiceName())
+	}()
+	s.Contains(stdout, "Using Deployment "+s.ServiceName())
+
+	s.Eventually(func() bool {
+		stdout, _, err := itest.Telepresence(ctx, "list", "--intercepts")
+		return err == nil && regexp.MustCompile(s.ServiceName()+`\s*: intercepted`).MatchString(stdout)
+	}, 10*time.Second, time.Second)
+
+	time.Sleep(200 * time.Millisecond) // List is really fast now, so give the mount some time to become effective
+	st, err := os.Stat(mountPoint)
+	require.NoError(err, "Stat on <mount point> failed")
+	require.True(st.IsDir(), "Mount point is not a directory")
+	st, err = os.Stat(filepath.Join(mountPoint, "var"))
+	require.NoError(err, "Stat on <mount point>/var failed")
+	require.True(st.IsDir(), "<mount point>/var is not a directory")
+}
+
 func (s *singleServiceSuite) Test_InterceptDetailedOutput() {
 	ctx := s.Context()
 	port, cancel := itest.StartLocalHttpEchoServer(ctx, s.ServiceName())
--- a/pkg/agentconfig/container.go
+++ b/pkg/agentconfig/container.go
@@ -73,9 +73,13 @@ func AgentContainer(
 			dlog.Errorf(ctx, "unable to parse agent version from image name %s", config.AgentImage)
 		}
 	}
+	// TODO could this be where we get volumes wrong?
+	dlog.Debugf(ctx, "looking for volume mounts on pod %s.%s ...", pod.Name, pod.Namespace)
 	EachContainer(pod, config, func(app *core.Container, cc *Container) {
 		var volPaths []string
+		dlog.Debugf(ctx, "looking for volume mounts on container %s ...", cc.Name)
 		volPaths, mounts = appendAppContainerVolumeMounts(app, cc, mounts, pod.ObjectMeta.Annotations, agentVersion)
+		dlog.Debugf(ctx, "found %d volume mounts on container %s: %v", len(volPaths), cc.Name, volPaths)
 		if len(volPaths) > 0 {
 			evs = append(evs, core.EnvVar{
 				Name:  cc.EnvPrefix + EnvInterceptMounts,
--- a/pkg/agentconfig/container.go
+++ b/pkg/agentconfig/container.go
@@ -73,13 +73,9 @@ func AgentContainer(
 			dlog.Errorf(ctx, "unable to parse agent version from image name %s", config.AgentImage)
 		}
 	}
-	// TODO could this be where we get volumes wrong?
-	dlog.Debugf(ctx, "looking for volume mounts on pod %s.%s ...", pod.Name, pod.Namespace)
 	EachContainer(pod, config, func(app *core.Container, cc *Container) {
 		var volPaths []string
-		dlog.Debugf(ctx, "looking for volume mounts on container %s ...", cc.Name)
 		volPaths, mounts = appendAppContainerVolumeMounts(app, cc, mounts, pod.ObjectMeta.Annotations, agentVersion)
-		dlog.Debugf(ctx, "found %d volume mounts on container %s: %v", len(volPaths), cc.Name, volPaths)
 		if len(volPaths) > 0 {
 			evs = append(evs, core.EnvVar{
 				Name:  cc.EnvPrefix + EnvInterceptMounts,


cmd/traffic/cmd/manager/mutator/agent_injector.go
```diff
@@ -191,11 +191,16 @@ func (a *agentInjector) Inject(ctx context.Context, req *admission.AdmissionRequ
 		return nil, fmt.Errorf("invalid value %q for annotation %s", ia, agentconfig.InjectAnnotation)
 	}
 
+	config := scx.AgentConfig()
+	if agentAlreadyInjected(ctx, pod, config) {
+		dlog.Infof(ctx, "Pod %s.%s already has container %s", pod.Name, pod.Namespace, agentconfig.ContainerName)
+		return nil, nil
+	}
+
```
thallgren, line 199: Not sure what this code adds. The `addInitContainer` called later on will detect if the agent is injected already, and if it is, check if it is unchanged. That check is bypassed here.

thallgren (commented):
I tested this, and the added test fails. Looking at the actual modification, I don't understand why that would address the mount problem.

I take it that you did some debugging and found something that would be rectified by this. Would you care to elaborate on those findings so that we can find a working solution?

petergardfjall:
Thanks for looking into this! 

First off, were you able to reproduce the issue?
I did provide fairly detailed steps to reproduce the issue in (https://github.com/telepresenceio/telepresence/issues/3483). The local volume mount failed consistently for me when doing a `telepresence intercept --replace`.

> I tested this, and the added test fails. Looking at the actual modification, I don't understand why that would address the mount problem.

Does the added `--replace` test also fail for you when run against the main `release/v2` branch?

> I take it that you did some debugging and found something that would be rectified by this. Would you care to elaborate on those findings so that we can find a working solution?

What I noticed by examining log output was that the admission hook was getting called more than once during the course of a `--replace` intercept. 

I don't remember exactly but I think that during the first "round" the hook did patch as expected (removing the application container, preserving volumes, etc) but in a later "round" the hook once again ran this time without having the application around and (seemingly) therefore not discovering its volumes.

This is log output from one of those runs (note: diff formatting for highlighting). As you can see patching of the pod takes place twice with different results (at one point I added more debug output and I think it was then I saw that volumes got removed by the second patch, I could have a more thorough look if you like).

My changes was a rather clumsy attempt to avoid any additional changes (like removing the volumes) after the agent had already been successfully injected.

```diff
2024-01-12 18:15:15.8886 debug   agent-injector : Handling admission request CREATE my-web-server-6dbd7bcbd8-.default
2024-01-12 18:15:15.8963 debug   agent-injector : The my-web-server-6dbd7bcbd8-.default pod has not enabled traffic-agent container injection through "telepresence-agents" configmap or through "telepresence.getambassador.io/inject-traffic-agent" annotation; skipping
2024-01-12 18:15:15.8963 debug   agent-injector : Webhook request handled successfully
2024-01-12 18:16:02.3197 debug   httpd/conn=127.0.0.1:8081 : PrepareIntercept called : session_id="bd7f6071-41fd-4c0e-bf10-91ce763e76b3"
2024-01-12 18:16:02.3272 debug   agent-configs : ADDED telepresence-agents.default
2024-01-12 18:16:02.3334 debug   agent-configs : MODIFIED telepresence-agents.default
2024-01-12 18:16:02.3335 debug   agent-configs : add my-web-server.default
2024-01-12 18:16:02.3380 debug   agent-configs : Rollout of my-web-server.default is necessary. An agent is desired but the pod my-web-server-6dbd7bcbd8-t4pd9 doesn't have one
2024-01-12 18:16:02.3457 info    agent-configs : Successfully rolled out my-web-server.default
2024-01-12 18:16:02.3511 debug   agent-injector : Received webhook request...
2024-01-12 18:16:02.3517 debug   agent-injector : Handling admission request CREATE my-web-server-667fc58b79-.default
2024-01-12 18:16:02.3573 info    agent-injector : Injecting traffic-agent into pod my-web-server-667fc58b79-.default
2024-01-12 18:16:02.3574 info    agent-injector : Injecting 8 patches into pod my-web-server-667fc58b79-.default
2024-01-12 18:16:02.3598 debug   agent-injector : Webhook request handled successfully
2024-01-12 18:16:02.3624 debug   agent-injector : Received webhook request...
2024-01-12 18:16:02.3632 debug   agent-injector : Handling admission request CREATE my-web-server-667fc58b79-.default
2024-01-12 18:16:02.3685 info    agent-injector : Injecting traffic-agent into pod my-web-server-667fc58b79-.default
2024-01-12 18:16:02.3688 info    agent-injector : Pod my-web-server-667fc58b79-.default already has container traffic-agent and it isn't modified
2024-01-12 18:16:02.3690 debug   agent-injector : Webhook request handled successfully
2024-01-12 18:16:03.2167 debug   httpd/conn=10.244.0.6:8081 : ArriveAsAgent called
2024-01-12 18:16:03.2185 debug   httpd/conn=10.244.0.6:8081 : WatchLookupDNS called : session_id="agent:17fbce86-b152-41f8-a8ca-d3a2ae30841e"
2024-01-12 18:16:03.2187 debug   httpd/conn=127.0.0.1:8081 : CreateIntercept called : session_id="bd7f6071-41fd-4c0e-bf10-91ce763e76b3"
2024-01-12 18:16:03.2187 debug   httpd/conn=10.244.0.6:8081 : WatchLogLevel called
2024-01-12 18:16:03.2187 debug   httpd/conn=10.244.0.6:8081 : WatchDial called : session_id="agent:17fbce86-b152-41f8-a8ca-d3a2ae30841e"
2024-01-12 18:16:03.2187 debug   httpd/conn=10.244.0.6:8081 : WatchIntercepts called : session_id="agent:17fbce86-b152-41f8-a8ca-d3a2ae30841e"
2024-01-12 18:16:03.2191 debug   httpd/conn=10.244.0.6:8081 : WatchIntercepts sending update : session_id="agent:17fbce86-b152-41f8-a8ca-d3a2ae30841e"
2024-01-12 18:16:03.2264 debug   agent-configs : MODIFIED telepresence-agents.default
2024-01-12 18:16:03.2264 debug   agent-configs : add my-web-server.default
2024-01-12 18:16:03.2268 debug   httpd/conn=10.244.0.6:8081 : Intercept my-web-server.default valid. Disposition: WAITING : session_id="agent:17fbce86-b152-41f8-a8ca-d3a2ae30841e"
2024-01-12 18:16:03.2271 debug   httpd/conn=10.244.0.6:8081 : WatchIntercepts sending update : session_id="agent:17fbce86-b152-41f8-a8ca-d3a2ae30841e"
2024-01-12 18:16:03.2269 debug   httpd/conn=127.0.0.1:8081 : WatchIntercepts sending update : session_id="bd7f6071-41fd-4c0e-bf10-91ce763e76b3"
2024-01-12 18:16:03.2285 debug   httpd/conn=10.244.0.6:8081 : ReviewIntercept called: bd7f6071-41fd-4c0e-bf10-91ce763e76b3:my-web-server - ACTIVE : session_id="agent:17fbce86-b152-41f8-a8ca-d3a2ae30841e"
2024-01-12 18:16:03.2308 debug   httpd/conn=127.0.0.1:8081 : WatchIntercepts sending update : session_id="bd7f6071-41fd-4c0e-bf10-91ce763e76b3"
2024-01-12 18:16:03.2310 debug   httpd/conn=10.244.0.6:8081 : Intercept my-web-server.default valid. Disposition: ACTIVE : session_id="agent:17fbce86-b152-41f8-a8ca-d3a2ae30841e"
2024-01-12 18:16:03.2311 debug   httpd/conn=10.244.0.6:8081 : WatchIntercepts sending update : session_id="agent:17fbce86-b152-41f8-a8ca-d3a2ae30841e"
2024-01-12 18:16:03.2363 debug   agent-configs : Rollout of my-web-server.default is necessary. An agent is desired but the pod my-web-server-6dbd7bcbd8-t4pd9 doesn't have one
2024-01-12 18:16:03.2561 info    agent-configs : Successfully rolled out my-web-server.default
2024-01-12 18:16:03.2687 debug   agent-injector : Received webhook request...
2024-01-12 18:16:03.2696 debug   agent-injector : Handling admission request DELETE my-web-server-667fc58b79-c86gn.default
2024-01-12 18:16:03.2848 debug   agent-injector : Webhook request handled successfully
2024-01-12 18:16:03.2867 debug   agent-injector : Received webhook request...
2024-01-12 18:16:03.2872 debug   agent-injector : Handling admission request CREATE my-web-server-7b9b7d689b-.default
2024-01-12 18:16:03.3023 info    agent-injector : Injecting traffic-agent into pod my-web-server-7b9b7d689b-.default
2024-01-12 18:16:03.3024 debug   agent-injector : Deleted container nginx
- 2024-01-12 18:16:03.3024 info    agent-injector : Injecting 9 patches into pod my-web-server-7b9b7d689b-.default
2024-01-12 18:16:03.3026 debug   agent-injector : Webhook request handled successfully
2024-01-12 18:16:03.3053 debug   agent-injector : Received webhook request...
2024-01-12 18:16:03.3063 debug   agent-injector : Handling admission request CREATE my-web-server-7b9b7d689b-.default
2024-01-12 18:16:03.4584 debug   httpd/conn=127.0.0.1:8081 : LookupDNS my-web-server.default.svc.cluster.local. A : session_id="bd7f6071-41fd-4c0e-bf10-91ce763e76b3"
2024-01-12 18:16:03.4595 debug   httpd/conn=10.244.0.6:8081 : AgentLookupDNSResponse called my-web-server.default.svc.cluster.local. : session_id="agent:17fbce86-b152-41f8-a8ca-d3a2ae30841e"
2024-01-12 18:16:03.4596 debug   httpd/conn=127.0.0.1:8081 : LookupDNS on agents: my-web-server.default.svc.cluster.local. A -> [10.96.209.132] : session_id="bd7f6071-41fd-4c0e-bf10-91ce763e76b3"
2024-01-12 18:16:03.4628 debug   httpd/conn=127.0.0.1:8081 : LookupDNS my-web-server.default.svc.cluster.local. AAAA : session_id="bd7f6071-41fd-4c0e-bf10-91ce763e76b3"
2024-01-12 18:16:03.4637 debug   httpd/conn=10.244.0.6:8081 : AgentLookupDNSResponse called my-web-server.default.svc.cluster.local. : session_id="agent:17fbce86-b152-41f8-a8ca-d3a2ae30841e"
2024-01-12 18:16:03.4637 debug   httpd/conn=127.0.0.1:8081 : LookupDNS on agents: my-web-server.default.svc.cluster.local. AAAA -> NOERROR : session_id="bd7f6071-41fd-4c0e-bf10-91ce763e76b3"
2024-01-12 18:16:03.5169 debug   httpd/conn=10.244.0.6:8081 : WatchIntercepts context cancelled : session_id="agent:17fbce86-b152-41f8-a8ca-d3a2ae30841e"
2024-01-12 18:16:03.5170 debug   httpd/conn=10.244.0.6:8081 : Depart called : session_id="agent:17fbce86-b152-41f8-a8ca-d3a2ae30841e"
2024-01-12 18:16:03.5170 debug   httpd/conn=10.244.0.6:8081 : Session agent:17fbce86-b152-41f8-a8ca-d3a2ae30841e removed. Explicit removal : session_id="agent:17fbce86-b152-41f8-a8ca-d3a2ae30841e"
2024-01-12 18:16:03.5173 debug   httpd/conn=127.0.0.1:8081 : WatchIntercepts sending update : session_id="bd7f6071-41fd-4c0e-bf10-91ce763e76b3"
- 2024-01-12 18:16:03.5380 info    agent-injector : Injecting traffic-agent into pod my-web-server-7b9b7d689b-.default
- 2024-01-12 18:16:03.5383 debug   agent-injector : Pod my-web-server-7b9b7d689b-.default already has container traffic-agent but it is modified
- 2024-01-12 18:16:03.5383 info    agent-injector : Injecting 1 patches into pod my-web-server-7b9b7d689b-.default
2024-01-12 18:16:03.5384 debug   agent-injector : Webhook request handled successfully
2024-01-12 18:16:04.1802 debug   httpd/conn=10.244.0.6:8081 : ArriveAsAgent called
```

cmd/traffic/cmd/manager/mutator/agent_injector.go
petergardfjall: In that later check it will see the agent as modified. I _believe_ (may be off) that it thinks it is modified because the application container is gone and it tries to discover volumes from that container. I supplied some logs in [this comment](https://github.com/telepresenceio/telepresence/pull/3484#issuecomment-1911569638) that shows that the agent is seen as modified the second time the admission hook gets run.

thallgren:
I agree that it's odd that a second injection should be needed here. That means that the first injection causes some modification that in turn triggers the injector again. Qualified guess of the flow:
1. The first injection removes the original container.
2. The container removal is considered somewhat intrusive, so a second injection is triggered.
3. The second injection doesn't contain mounts nor environments from the now non-existing container, so the agent is modified to drop them.

Just a guess though. I think that in order to nail this down, it will be necessary to see what the second patch contains (perhaps by logging it).

petergardfjall:
> Just a guess though. I think that in order to nail this down, it will be necessary to see what the second patch contains (perhaps by logging it).

Note: this is from running latest commit on `release/v2` with extra debug logging. Yet again, tried to highlight the interesting parts.

```diff
2024-01-26 11:00:05.8040 debug   httpd/conn=127.0.0.1:8081 : PrepareIntercept called : session_id="e1c2e135-63ba-480b-9214-27057de1f287"
2024-01-26 11:00:05.8108 debug   agent-configs : ADDED telepresence-agents.default
2024-01-26 11:00:05.8161 debug   agent-configs : MODIFIED telepresence-agents.default
2024-01-26 11:00:05.8161 debug   agent-configs : add my-web-server.default
2024-01-26 11:00:05.8205 debug   agent-configs : Rollout of my-web-server.default is necessary. An agent is desired but the pod my-web-server-6dbd7bcbd8-fgqj7 doesn't have one
2024-01-26 11:00:05.8279 info    agent-configs : Successfully rolled out my-web-server.default
2024-01-26 11:00:05.8342 debug   agent-injector : Received webhook request...
2024-01-26 11:00:05.8352 debug   agent-injector : Handling admission request CREATE my-web-server-556d9fd5d9-.default
2024-01-26 11:00:05.8458 info    agent-injector : Injecting traffic-agent into pod my-web-server-556d9fd5d9-.default
2024-01-26 11:00:05.8472 info    agent-injector : Injecting 8 patches into pod my-web-server-556d9fd5d9-.default:
[
  {
    "op": "add",
    "path": "/spec/containers/-",
    "value": {
      "name": "traffic-agent",
      "image": "docker.io/petergardfjall/tel2:2.19.0-dev",
      "args": [
        "agent"
      ],
      "ports": [
        {
          "name": "http",
          "containerPort": 9900,
          "protocol": "TCP"
        }
      ],
      "env": [
        {
          "name": "_TEL_AGENT_POD_IP",
          "valueFrom": {
            "fieldRef": {
              "apiVersion": "v1",
              "fieldPath": "status.podIP"
            }
          }
        },
        {
          "name": "_TEL_AGENT_NAME",
          "valueFrom": {
            "fieldRef": {
              "apiVersion": "v1",
              "fieldPath": "metadata.name"
            }
          }
        },
        {
          "name": "A_TELEPRESENCE_MOUNTS",
          "value": "/etc/config:/var/run/secrets/kubernetes.io/serviceaccount"
        }
      ],
      "resources": {},
      "volumeMounts": [
        {
          "name": "config-volume",
          "mountPath": "/tel_app_mounts/nginx/etc/config"
        },
        {
          "name": "kube-api-access-7hkcv",
          "readOnly": true,
          "mountPath": "/tel_app_mounts/nginx/var/run/secrets/kubernetes.io/serviceaccount"
        },
        {
          "name": "traffic-annotations",
          "mountPath": "/tel_pod_info"
        },
        {
          "name": "traffic-config",
          "mountPath": "/etc/traffic-agent"
        },
        {
          "name": "export-volume",
          "mountPath": "/tel_app_exports"
        },
        {
          "name": "tel-agent-tmp",
          "mountPath": "/tmp"
        }
      ],
      "readinessProbe": {
        "exec": {
          "command": [
            "/bin/stat",
            "/tmp/agent/ready"
          ]
        }
      },
      "imagePullPolicy": "IfNotPresent"
    }
  },
  {
    "op": "add",
    "path": "/spec/volumes/-",
    "value": {
      "name": "traffic-annotations",
      "downwardAPI": {
        "items": [
          {
            "path": "annotations",
            "fieldRef": {
              "apiVersion": "v1",
              "fieldPath": "metadata.annotations"
            }
          }
        ]
      }
    }
  },
  {
    "op": "add",
    "path": "/spec/volumes/-",
    "value": {
      "name": "traffic-config",
      "configMap": {
        "name": "telepresence-agents",
        "items": [
          {
            "key": "my-web-server",
            "path": "config.yaml"
          }
        ]
      }
    }
  },
  {
    "op": "add",
    "path": "/spec/volumes/-",
    "value": {
      "name": "export-volume",
      "emptyDir": {}
    }
  },
  {
    "op": "add",
    "path": "/spec/volumes/-",
    "value": {
      "name": "tel-agent-tmp",
      "emptyDir": {}
    }
  },
  {
    "op": "replace",
    "path": "/spec/containers/0/ports/0/name",
    "value": "tm-http"
  },
  {
    "op": "replace",
    "path": "/metadata/annotations",
    "value": {
      "telepresence.getambassador.io/inject-traffic-agent": "enabled",
      "telepresence.getambassador.io/restartedAt": "2024-01-26T11:00:05Z"
    }
  },
  {
    "op": "replace",
    "path": "/metadata/labels",
    "value": {
      "app": "my-web-server",
      "pod-template-hash": "556d9fd5d9",
      "telepresence.io/workloadEnabled": "true",
      "telepresence.io/workloadName": "my-web-server"
    }
  }
]
2024-01-26 11:00:05.8473 debug   agent-injector : Webhook request handled successfully
2024-01-26 11:00:05.8497 debug   agent-injector : Received webhook request...
2024-01-26 11:00:05.8509 debug   agent-injector : Handling admission request CREATE my-web-server-556d9fd5d9-.default
2024-01-26 11:00:05.8576 info    agent-injector : Injecting traffic-agent into pod my-web-server-556d9fd5d9-.default
2024-01-26 11:00:05.8579 info    agent-injector : Pod my-web-server-556d9fd5d9-.default already has container traffic-agent and it isn't modified
2024-01-26 11:00:05.8579 debug   agent-injector : Webhook request handled successfully
2024-01-26 11:00:06.6315 debug   httpd/conn=10.244.0.5:8081 : ArriveAsAgent called
2024-01-26 11:00:06.6329 debug   httpd/conn=10.244.0.5:8081 : WatchLookupDNS called : session_id="agent:b62054db-6253-4aac-8eca-28476cb910a1"
2024-01-26 11:00:06.6330 debug   httpd/conn=10.244.0.5:8081 : WatchIntercepts called : session_id="agent:b62054db-6253-4aac-8eca-28476cb910a1"
2024-01-26 11:00:06.6330 debug   httpd/conn=10.244.0.5:8081 : WatchIntercepts sending update : session_id="agent:b62054db-6253-4aac-8eca-28476cb910a1"
2024-01-26 11:00:06.6331 debug   httpd/conn=10.244.0.5:8081 : WatchDial called : session_id="agent:b62054db-6253-4aac-8eca-28476cb910a1"
2024-01-26 11:00:06.6331 debug   httpd/conn=10.244.0.5:8081 : WatchLogLevel called
2024-01-26 11:00:06.6334 debug   httpd/conn=127.0.0.1:8081 : CreateIntercept called : session_id="e1c2e135-63ba-480b-9214-27057de1f287"
2024-01-26 11:00:06.6411 debug   agent-configs : MODIFIED telepresence-agents.default
2024-01-26 11:00:06.6412 debug   agent-configs : add my-web-server.default
2024-01-26 11:00:06.6420 debug   httpd/conn=127.0.0.1:8081 : WatchIntercepts sending update : session_id="e1c2e135-63ba-480b-9214-27057de1f287"
2024-01-26 11:00:06.6421 debug   httpd/conn=10.244.0.5:8081 : Intercept my-web-server.default valid. Disposition: WAITING : session_id="agent:b62054db-6253-4aac-8eca-28476cb910a1"
2024-01-26 11:00:06.6421 debug   httpd/conn=10.244.0.5:8081 : WatchIntercepts sending update : session_id="agent:b62054db-6253-4aac-8eca-28476cb910a1"
2024-01-26 11:00:06.6445 debug   httpd/conn=10.244.0.5:8081 : ReviewIntercept called: e1c2e135-63ba-480b-9214-27057de1f287:my-web-server - ACTIVE : session_id="agent:b62054db-6253-4aac-8eca-28476cb910a1"
2024-01-26 11:00:06.6482 debug   agent-configs : Rollout of my-web-server.default is necessary. An agent is desired but the pod my-web-server-6dbd7bcbd8-fgqj7 doesn't have one
2024-01-26 11:00:06.6539 debug   httpd/conn=127.0.0.1:8081 : WatchIntercepts sending update : session_id="e1c2e135-63ba-480b-9214-27057de1f287"
2024-01-26 11:00:06.6544 debug   httpd/conn=10.244.0.5:8081 : Intercept my-web-server.default valid. Disposition: ACTIVE : session_id="agent:b62054db-6253-4aac-8eca-28476cb910a1"
2024-01-26 11:00:06.6545 debug   httpd/conn=10.244.0.5:8081 : WatchIntercepts sending update : session_id="agent:b62054db-6253-4aac-8eca-28476cb910a1"
2024-01-26 11:00:06.6583 info    agent-configs : Successfully rolled out my-web-server.default
2024-01-26 11:00:06.6844 debug   agent-injector : Received webhook request...
2024-01-26 11:00:06.6852 debug   agent-injector : Handling admission request DELETE my-web-server-556d9fd5d9-qvglf.default
2024-01-26 11:00:06.6948 debug   agent-injector : Webhook request handled successfully
2024-01-26 11:00:06.7104 debug   agent-injector : Received webhook request...
2024-01-26 11:00:06.7115 debug   agent-injector : Handling admission request CREATE my-web-server-6f9d7b547b-.default
2024-01-26 11:00:06.7272 info    agent-injector : Injecting traffic-agent into pod my-web-server-6f9d7b547b-.default
2024-01-26 11:00:06.7272 debug   agent-injector : Deleted container nginx
- 2024-01-26 11:00:06.7275 info    agent-injector : Injecting 9 patches into pod my-web-server-6f9d7b547b-.default:
- [
-   {
-     "op": "remove",
-     "path": "/spec/containers/0"
-   },
-   {
-     "op": "add",
-     "path": "/spec/containers/-",
-     "value": {
-       "name": "traffic-agent",
-       "image": "docker.io/petergardfjall/tel2:2.19.0-dev",
-       "args": [
-         "agent"
-       ],
-       "ports": [
-         {
-           "name": "http",
-           "containerPort": 9900,
-           "protocol": "TCP"
-         }
-       ],
-       "env": [
-         {
-           "name": "_TEL_AGENT_POD_IP",
-           "valueFrom": {
-             "fieldRef": {
-               "apiVersion": "v1",
-               "fieldPath": "status.podIP"
-             }
-           }
-         },
-         {
-           "name": "_TEL_AGENT_NAME",
-           "valueFrom": {
-             "fieldRef": {
-               "apiVersion": "v1",
-               "fieldPath": "metadata.name"
-             }
-           }
-         },
-         {
-           "name": "A_TELEPRESENCE_MOUNTS",
-           "value": "/etc/config:/var/run/secrets/kubernetes.io/serviceaccount"
-         }
-       ],
-       "resources": {},
-       "volumeMounts": [
-         {
-           "name": "config-volume",
-           "mountPath": "/tel_app_mounts/nginx/etc/config"
-         },
-         {
-           "name": "kube-api-access-xtnhg",
-           "readOnly": true,
-           "mountPath": "/tel_app_mounts/nginx/var/run/secrets/kubernetes.io/serviceaccount"
-         },
-         {
-           "name": "traffic-annotations",
-           "mountPath": "/tel_pod_info"
-         },
-         {
-           "name": "traffic-config",
-           "mountPath": "/etc/traffic-agent"
-         },
-         {
-           "name": "export-volume",
-           "mountPath": "/tel_app_exports"
-         },
-         {
-           "name": "tel-agent-tmp",
-           "mountPath": "/tmp"
-         }
-       ],
-       "readinessProbe": {
-         "exec": {
-           "command": [
-             "/bin/stat",
-             "/tmp/agent/ready"
-           ]
-         }
-       },
-       "imagePullPolicy": "IfNotPresent"
-     }
-   },
-   {
-     "op": "add",
-     "path": "/spec/volumes/-",
-     "value": {
-       "name": "traffic-annotations",
-       "downwardAPI": {
-         "items": [
-           {
-             "path": "annotations",
-             "fieldRef": {
-               "apiVersion": "v1",
-               "fieldPath": "metadata.annotations"
-             }
-           }
-         ]
-       }
-     }
-   },
-   {
-     "op": "add",
-     "path": "/spec/volumes/-",
-     "value": {
-       "name": "traffic-config",
-       "configMap": {
-         "name": "telepresence-agents",
-         "items": [
-           {
-             "key": "my-web-server",
-             "path": "config.yaml"
-           }
-         ]
-       }
-     }
-   },
-   {
-     "op": "add",
-     "path": "/spec/volumes/-",
-     "value": {
-       "name": "export-volume",
-       "emptyDir": {}
-     }
-   },
-   {
-     "op": "add",
-     "path": "/spec/volumes/-",
-     "value": {
-       "name": "tel-agent-tmp",
-       "emptyDir": {}
-     }
-   },
-   {
-     "op": "replace",
-     "path": "/spec/containers/0/ports/0/name",
-     "value": "tm-http"
-   },
-   {
-     "op": "replace",
-     "path": "/metadata/annotations",
-     "value": {
-       "telepresence.getambassador.io/inject-traffic-agent": "enabled",
-       "telepresence.getambassador.io/restartedAt": "2024-01-26T11:00:06Z"
-     }
-   },
-   {
-     "op": "replace",
-     "path": "/metadata/labels",
-     "value": {
-       "app": "my-web-server",
-       "pod-template-hash": "6f9d7b547b",
-       "telepresence.io/workloadEnabled": "true",
-       "telepresence.io/workloadName": "my-web-server"
-     }
-   }
- ]
2024-01-26 11:00:06.7289 debug   agent-injector : Webhook request handled successfully
2024-01-26 11:00:06.7322 debug   agent-injector : Received webhook request...
2024-01-26 11:00:06.7336 debug   agent-injector : Handling admission request CREATE my-web-server-6f9d7b547b-.default
2024-01-26 11:00:07.0085 info    agent-injector : Injecting traffic-agent into pod my-web-server-6f9d7b547b-.default
2024-01-26 11:00:07.0089 debug   agent-injector : Pod my-web-server-6f9d7b547b-.default already has container traffic-agent but it is modified
- 2024-01-26 11:00:07.0090 info    agent-injector : Injecting 1 patches into pod my-web-server-6f9d7b547b-.default:
- [
-   {
-     "op": "replace",
-     "path": "/spec/containers/0",
-     "value": {
-       "name": "traffic-agent",
-       "image": "docker.io/petergardfjall/tel2:2.19.0-dev",
-       "args": [
-         "agent"
-       ],
-       "ports": [
-         {
-           "name": "http",
-           "containerPort": 9900,
-           "protocol": "TCP"
-         }
-       ],
-       "env": [
-         {
-           "name": "_TEL_AGENT_POD_IP",
-           "valueFrom": {
-             "fieldRef": {
-               "apiVersion": "v1",
-               "fieldPath": "status.podIP"
-             }
-           }
-         },
-         {
-           "name": "_TEL_AGENT_NAME",
-           "valueFrom": {
-             "fieldRef": {
-               "apiVersion": "v1",
-               "fieldPath": "metadata.name"
-             }
-           }
-         }
-       ],
-       "resources": {},
-       "volumeMounts": [
-         {
-           "name": "traffic-annotations",
-           "mountPath": "/tel_pod_info"
-         },
-         {
-           "name": "traffic-config",
-           "mountPath": "/etc/traffic-agent"
-         },
-         {
-           "name": "export-volume",
-           "mountPath": "/tel_app_exports"
-         },
-         {
-           "name": "tel-agent-tmp",
-           "mountPath": "/tmp"
-         }
-       ],
-       "readinessProbe": {
-         "exec": {
-           "command": [
-             "/bin/stat",
-             "/tmp/agent/ready"
-           ]
-         }
-       },
-       "imagePullPolicy": "IfNotPresent"
-     }
-   }
- ]
2024-01-26 11:00:07.0090 debug   agent-injector : Webhook request handled successfully
2024-01-26 11:00:07.2187 debug   httpd/conn=10.244.0.5:8081 : WatchIntercepts context cancelled : session_id="agent:b62054db-6253-4aac-8eca-28476cb910a1"
2024-01-26 11:00:07.2203 debug   httpd/conn=10.244.0.5:8081 : Depart called : session_id="agent:b62054db-6253-4aac-8eca-28476cb910a1"
2024-01-26 11:00:07.2203 debug   httpd/conn=10.244.0.5:8081 : Session agent:b62054db-6253-4aac-8eca-28476cb910a1 removed. Explicit removal : session_id="agent:b62054db-6253-4aac-8eca-28476cb910a1"
2024-01-26 11:00:07.2204 debug   httpd/conn=127.0.0.1:8081 : WatchIntercepts sending update : session_id="e1c2e135-63ba-480b-9214-27057de1f287"
2024-01-26 11:00:07.6826 debug   httpd/conn=10.244.0.5:8081 : ArriveAsAgent called

```

thallgren:
Interesting. The diff between the two agent-image injections is this, so yeah. That would definitely invalidate the mounts:
```
-           "name": "A_TELEPRESENCE_MOUNTS",
-           "value": "/etc/config:/var/run/secrets/kubernetes.io/serviceaccount"
36a41,49
-           "name": "config-volume",
-           "mountPath": "/tel_app_mounts/nginx/etc/config"
-         },
-         {
-           "name": "kube-api-access-xtnhg",
-           "readOnly": true,
-           "mountPath": "/tel_app_mounts/nginx/var/run/secrets/kubernetes.io/serviceaccount"
```

thallgren:
Perhaps we can solve this using annotations. The first injection would replace containers that have been removed with annotations holding a json value of their original value. Then the agent container generator could look at those annotations (in addition to containers that are left intact) in order to get the full set of containers to act on. That would then yield an agent container with the exact same content and make the second injection attempt a no-op.

The reason the second inject attempt must go through the motions, is that other sidecar injectors (Istio, LinkerD, etc.) might have changed things that will force other changes to the agent.

What I suggest here is fairly complex. Let me know if you feel it's beyond what you set out to do here, and I'll look into it when I get some time over.

petergardfjall:
> What I suggest here is fairly complex. Let me know if you feel it's beyond what you set out to do here, and I'll look into it when I get some time over.

I'd happily hand it over to you (I'd also be happy to help out, but I would probably need some guidance).
Perhaps the test case could still be of use?

thallgren:
Superseded by #3500

PR #3565, opened 2024-04-04
fix(chart): add option to disable the webhook

arthlr:
## Description

In my use case of telepresence, the mutating webhook is not needed. To stick to the kiss programming principle, I would like to add an option to disable the webhook.

## Checklist


thallgren (changes requested):
This looks promising, and I like the idea of being able to completely disable the webhook.

I think there must be some additional logic in cmd/traffic/cmd/manager/manager.go that disables the agent-injector http-service from starting (line 139), and prevents the creation of the `injectorCertGetter` (line 118)  because it will otherwise make attempts to either watch or mount the mutator-webhook-tls secret, which is no longer generated.

thallgren:
@capuche2412 I didn't hear back from you so I decided to use your commit as the base for a more elaborated change that is in line with what I requested. Thanks for contributing and sparking this improvement.

thallgren (approved):

arthlr:
This looks amazing ! I was busy and I haven't found the time to complete this PR. 
Thanks a lot for this improvement :+1:

PR #3570, opened 2024-04-10
 Fix panic due to root daemon not running.

thallgren:
If a `telepresence connect` was made at a time when the root daemon was
not running (an abnormal condition) and a subsequent intercept was then
made, a panic would occur when the port-forward to the agent was set up.
This is now fixed so that the initial `telepresence connect` is refused
unless the root daemon is running.


thallgren:
Merged into release/v2

PR #3618, opened 2024-06-13
Chart: Allow users to supply their own webhook secret

### LINKED ISSUE #3601: chart: add option to supply own cert bundle
The helm chart currently uses the `genCA` and `genSignedCert` helm functions. These are non-declarative, and produce different results with each render.

I am deploying this chart via terraform's `helm_release` resource, which renders the chart several times during a workflow, and errors if the renders are different. In this case, that fails, because the certs are different.

ArgoCD, or similar, would likely have their own problems related to this.

Add a new `agentInjector.certificate.method` option, that allows me to specify my own certificate components. I'm happy for this to take whatever form is needed, but I need to be able to provide values generated from outside of helm.

I _cannot_ install the traffic manager via the CLI method: organization policy, and best practices of declarative / infrastructure-as-code, **require** me to use `terraform` or similar.

`helm_release` _does_ have some capacity to modify manifests post-render, but that would be cumbersome, and I'd like to avoid it if at all possible.

 - Chart version 2.19.5

**Additional context**
Add any other context about the feature request here.

mkantzer:
## Description

Creates a new value `supplied` for `agentInjector.certificate.method`, that fully disables the generation of a secret. This prevents an apply error when helm would try to create an already-existing secret with annotations indicating ownership by a system other than this chart.

## Checklist

I'm not sure on the process of testing helm charts here, so I'd appreciate some help there.


thallgren (approved):
Looks good to me. Thank you for contributing this.

mkantzer:
@thallgren I don't have write access, so I can't merge. Is there anything else I need to do, or will you pull this in when ready?

Issue #3640, opened 2024-07-02
Add support for argo rollouts


mkantzer:
**Please describe your use case / problem.**
I'm using Argo Rollouts to control my replicasets / pods.
They have a similar shape to a `Deployment`, but with some additional fields in their `spec` that make them function. Fundamentally though, they are still using a Pod Template Spec to spin up replicasets.

**Describe the solution you'd like**
Telepresence detects a Rollout in the same way it would a Deployment, and use it in the same way.

**Describe alternatives you've considered**
I've instead been targeting replicasets, but those do not have stable names, and it's causing issues for my developers who need to start keeping track of which ones are live.

**Versions (please complete the following information)**
OSS Client             : v2.18.0
OSS Daemon in container: v2.18.0
Traffic Manager        : v2.20.0-test.4
Traffic Agent          : docker.io/datawire/ambassador-telepresence-agent:1.14.5

**Additional context**
Add any other context about the feature request here.

cindymullins-dw:
@mkantzer, thanks for the suggestion.  I'll mark this as a feature request.  Currently, [supported workloads](https://www.getambassador.io/docs/telepresence/latest/reference/intercepts#supported-workloads) are deployments, replica sets and stateful sets.

thallgren:
I believe this is a duplicate of #2406.

thallgren:
Closed by #3651

PR #3651, opened 2024-07-18
Support Argo Rollout workloads

### LINKED ISSUE #2406: Add support for Argo Rollout workloads
Currently, workloads that can be intercepted are Deployments, ReplicaSets and StatefulSets. It would be great to also support Argo Rollouts for users that use these workload types to leverage rollout strategies like blue/green and canary.

As of now, for a service `foo` that points to Rollout controlled pods, the following error is printed:
```
> telepresence intercept foo
telepresence: error: No interceptable deployment, replicaset, or statefulset matching foo found
```

Support adding the traffic-agent configuration to a Rollout object. Its specs are mainly the same as a Deployment with some additional properties.

It is currently possible to intercept the ReplicaSet associated to a Rollout object like this (`foo-57bc7c4959` being a ReplicaSet created by the Rollout controller):
```
> telepresence intercept foo --workload foo-57bc7c4959
```

This will work for a first time intercept, but as soon as the Rollout object gets updated, the updated ReplicaSet will be scaled down to 0 and a new one will be scaled up, replacing the traffic-agent-enabled pods.

Telepresence 2.5.0
Kubernetes 1.20

P0lip:
Closes #2406 #3640 

## Description

I'm a bit on the edge with having a separate path for rollouts when relevant crds are not available. We'll always have that scenario in some way, but could approach it differently.
I went with a simple resource check that we do once upon the start, but the alternative path I was considering (and I still am, to be frank) is to simply handle the errors gracefully and that should be sufficient in most cases.

I'm also torn when it comes to those CRDs overall Since we have a different behavior depending on the availability of the aforementioned CRDs as they are cluster-wide, we isolating the proper test isolation is trickier when we re-use the same cluster.

Argo Rollouts controller is installed "globally" under `argo-rollouts` namespace. It could be potentially installed per namespace, but that shouldn't generally matter in our case.

I added `Test_SuccessfullyInterceptsArgoRollout` to `WorkloadsSuite` in integration_test/workloads_test.go which should test things pretty thoroughly for this case, but if you think I should add specific tests for workload watcher or agent injection, LMK.

This PR needs https://github.com/datawire/k8sapi/pull/21. I tried to combine `k8sFactory` and the argo addition in some way, to have a single function one calls, but there's more there.

`integration_test/testdata/k8s/argorollout.goyaml` is unused, I know. I'll either drop it or use it in some tests

## Checklist


CHANGELOG.yml
```diff
@@ -30,6 +30,15 @@ docDescription: >-
...
   customizable development environments.
 items:
+  - version: 2.20.0
+    date: (TBD)
+    notes:
+      - type: feature
+        title: Add support for Argo Rollout workloads.
+        body: >-
+          Telepresence now supports Argo Rollout workloads.
+          It is recommended to set the following annotation `telepresence.getambassador.io/inject-traffic-agent: disabled`
```
thallgren:
```suggestion
          It is recommended to set the following annotation <code>telepresence.getambassador.io/inject-traffic-agent: enabled</code>
```
It should be enabled, and believe it or not, but this text is html, not markdown.

pkg/informers/context.go
```diff
@@ -0,0 +1,51 @@
...
+func getOpts(ns string) (k8sOpts []informers.SharedInformerOption, argoOpts []argorolloutsinformer.SharedInformerOption) {
+	if ns != "" {
+		k8sOpts = append(k8sOpts, informers.WithNamespace(ns))
+		argoOpts = append(argoOpts, argorolloutsinformer.WithNamespace(ns))
+	}
+
+	return k8sOpts, argoOpts
+}
+
+func WithInformers(ctx context.Context, ns string) context.Context {
```
thallgren: I'm not happy with the naming here, because it introduces stuttering and provides less information. In general, I think we should be careful using plural in package names.

Consider `informers.WithInformers` compared to `informer.WithFactory` or `informers.GetInformers` compared to `informer.GetFactory`

Perhaps rename the `GlobalInformers` to `Factory`? That will result in `informer.Factory` when used, which IMO provides more hints than `informers.GlobalInformers`. `Global` is actually not true, because in a namespaced setting, the informer will only look at one namespace.

cmd/traffic/cmd/manager/mutator/watcher.go
```diff
@@ -534,6 +559,14 @@ func (c *configWatcher) StartWatchers(ctx context.Context) error {
 			return err
 		}
 	}
+	for _, si := range c.rls {
+		if si == nil {
+			continue
+		}
```
thallgren: Why would `si` ever be nil here?

cmd/traffic/cmd/manager/mutator/watcher.go
```diff
@@ -789,16 +822,25 @@ func (c *configWatcher) Start(ctx context.Context) {
 	c.dps = make([]cache.SharedIndexInformer, len(nss))
 	c.rss = make([]cache.SharedIndexInformer, len(nss))
 	c.sss = make([]cache.SharedIndexInformer, len(nss))
+	c.rls = make([]cache.SharedIndexInformer, len(nss))
```
thallgren: I think this should be `nil` if `k8sapi.GetArgoRolloutCRDState(ctx)` returns `false`. So maybe break everything related to that condition into a separate loop.

thallgren (changes requested):
This looks good to me. I like the fact that the informers are started conditionally, but perhaps we should also give the user the ability to control this using a Helm chart value? That way, no additional RBAC permissions are needed unless you want to use this feature (many users are sensitive to such additions). It would also make it possible for users of argo-rollouts to explicitly disable this new feature of Telepresence.

Using a Helm chart value would also remove the different behavior depending on CRD's that you mention.

pkg/informers/context.go
P0lip: I agree, I didn't like that name either, but couldn't think of a better name.
`Factory` is good.

cmd/traffic/cmd/manager/mutator/watcher.go
P0lip: It would since we allocate an array with a non-zero length, so the default value would be nil for a workload interface, but this condition will become redundant once the following comment is addressed  https://github.com/telepresenceio/telepresence/pull/3651#discussion_r1683947920

charts/telepresence/README.md
```diff
@@ -101,6 +101,7 @@ The following tables lists the configurable parameters of the Telepresence chart
 | client.routing.allowConflictingSubnets               | Allow the specified subnets to be routed even if they conflict with other routes on the local machine.                      | `[]`                                                                        |
 | client.dns.excludeSuffixes                           | Suffixes for which the client DNS resolver will always fail (or fallback in case of the overriding resolver)                | `[".com", ".io", ".net", ".org", ".ru"]`                                    |
 | client.dns.includeSuffixes                           | Suffixes for which the client DNS resolver will always attempt to do a lookup. Includes have higher priority than excludes. | `[]`                                                                        |
+| argoRollouts.enabled                                 | Enable/Disable the argo-rollouts integration.                                                                               | `false`                                                                     |
```
mkantzer: Just a thought: I can imagine a future where multiple `Deployment` alternatives are supported. Might be worth nesting this as `alternateWorkloads.argoRollouts.enabled`.

thallgren: I like this idea, but I don't think it should be "alternative". It should be just "workloads". Let me explain why.

Each workload used today adds overhead. Both to the client-  and cluster-side of Telepresence. If we add an element that defaults to:
```yaml
workloads:
  deployments:
    enabled: true
  replicaSets:
    enabled true
  statefulSets:
    enabled true
  argoRollouts:
    enabled false
```
Then a user that just wants `deployments` to be considered a workload could control this and get rid of the current overhead for replicaSets and statefulSets. This will require some more work on our shared-indexer logic and client code,  and the complete implementation of what I'm suggesting here is beyond the scope of this PR, but adding the `workloads` top level element could be done now as a preparation.

Thoughts?

pkg/client/userd/trafficmgr/workloads.go
```diff
@@ -139,37 +144,45 @@ func workloadEquals(oa, ob runtime.Object) bool {
...
 	}
 }
 
 func (nw *namespacedWASWatcher) hasSynced() bool {
 	return nw.svcWatcher.HasSynced() &&
 		nw.wlWatchers[0].HasSynced() &&
 		nw.wlWatchers[1].HasSynced() &&
-		nw.wlWatchers[2].HasSynced()
+		nw.wlWatchers[2].HasSynced() &&
+		(nw.wlWatchers[3] == nil || nw.wlWatchers[3].HasSynced())
```
thallgren: Those should probably use the `deployments`, `replicasets`, `statefulsets`, and `rollouts` constants.

cmd/traffic/cmd/manager/managerutil/envconfig.go
```diff
@@ -69,6 +69,8 @@ type Env struct {
 	ClientDnsExcludeSuffixes             []string      `env:"CLIENT_DNS_EXCLUDE_SUFFIXES,        		parser=split-trim"`
 	ClientDnsIncludeSuffixes             []string      `env:"CLIENT_DNS_INCLUDE_SUFFIXES,       		parser=split-trim,  default="`
 	ClientConnectionTTL                  time.Duration `env:"CLIENT_CONNECTION_TTL,              		parser=time.ParseDuration"`
+
+	ArgoRolloutsEnabled bool `env:"ARGO_ROLLOUTS_ENABLED, parser=bool, default=0"`
```
thallgren: This default should be "false", not "0".

cmd/traffic/cmd/manager/managerutil/envconfig_test.go
```diff
@@ -65,10 +65,12 @@ func TestEnvconfig(t *testing.T) {
 		},
 		"simple": {
 			Input: map[string]string{
-				"AGENT_REGISTRY": "docker.io/datawire",
+				"AGENT_REGISTRY":        "docker.io/datawire",
+				"ARGO_ROLLOUTS_ENABLED": "1",
```
thallgren: Use "true" here, not "1".

thallgren (changes requested):
Just some minor nits, and perhaps add the top level `workloads` element for the Helm value.

thallgren (approved):
Approving this, but I'd still like to see a `workloads` top-level element for the Helm value.

thallgren:
@P0lip FYI a rebase from the latest release/v2 will get rid of the errors involving the "thhal" docker repository.

P0lip:
> Approving this, but I'd still like to see a workloads top-level element for the Helm value.

Sounds good, I like that change and will add it.

P0lip:
It looks like [this](https://github.com/telepresenceio/telepresence/blob/c2e1e363437f6e5eba9aabe42101f668eb17f7e3/.github/workflows/dev.yaml#L101L105) GH action step is not executed causing the integration test to fail

thallgren:
Oh, that's because the CI is triggered using `pull_request_target`, which in essence means that it runs using the `.github` folder from the target `release/v2` branch (security feature). In essence, this PR cannot be green unit it's been merged (catch 22).

What you can do, is create another branch from this PR's branch, and push an intentionally empty commit. Then create a new PR that targets this PR. It's CI will then use the `.github` folder from this PR. If that's green, then it's OK to merge this PR.

charts/telepresence/README.md
mkantzer: I'm not familiar enough with the codebase to comment on the implementation side, but from a user perspective that sounds great. (especially the "beyond the scope of this PR" bit so I can get Rollout support sooner 😛)

P0lip:
@thallgren I created a [small branch](https://github.com/telepresenceio/telepresence/tree/jrozek/dev/argo-ci-changes) containing the GH actions changes and made this one point at it.
I need to tweak PATH on Windows, because we're getting `exec: "kubectl-argo-rollouts": executable file not found in %PATH%` there.
However, the tests pass on Linux (both platforms) & MacOS, so from that standpoint we should be good.
I'll ping you once the Windows test is passing.

We can then just change the target branch on this PR back to v2 and merge it.

P0lip:
@thallgren I fixed the issue with PATH on Windows and as such this PR can be merged.
Feel free to change the target branch to v2 and merge it at your earliest convenience.

PR #3668, opened 2024-08-18
Fix backward incompatibility regression.

thallgren:
The introduction of argo-rollouts introduced a call to the new traffic-manager function `GetKnownWorkloadKinds` from the user client, and a client connecting to older traffic-managers that don't have this method would therefore fail with an error.

This commit checks if the returned error has code `Unimplemented`, and if so, assumes the default set of workload kinds.


P0lip (approved):
LGTM

PR #3681, opened 2024-09-02
Add timestamp to telepresence_logs.zip filename

### LINKED ISSUE #2997: Add a date/time stamp to telepresence_logs.zip filename
Feature request

Add a date/time stamp to the zip file that gets output by `telepresence gather-logs`.

Context

Re-running `telepresence gather-logs` multiple times over during the course of debugging means that you end up with a ton of `telepresence_logs/ (N)/ .zip` when downloading them either via slack or email, and sometimes you can easily lose track of which is the one you're interested in.

kodsurfer:
## Description

Add timestamp to telepresence_logs.zip filename.

Closes #2997

### DIFF
--- a/CHANGELOG.yml
+++ b/CHANGELOG.yml
@@ -34,6 +34,10 @@ items:
     date: (TBD)
     notes:
       - type: feature
+        title: Add timestamp to telepresence_logs.zip filename.
+        body: >-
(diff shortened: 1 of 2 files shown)


thallgren (approved):
Looks good to me. Thanks for providing this

PR #3716, opened 2024-10-29
Dont validate namespace in CI

fordneild:
## Description

Currently there exists a deadlock in the following situation:
- you are rendering the helm chart without a release name (as is common in CI tests)
- you are using managerRbac.namespaces

First, lets acknowledge that if isCI is set, we assume we dont have a release name/namespace. Therefore the chart skips the validation that release name is traffic-manager, which makes sense. The chart then defaults the traffic manager namespace to be ambassador. However this will throw if ambassador is not included in the list of namespaces.

## Checklist

### DIFF
--- a/charts/telepresence/templates/trafficManagerRbac/namespace-scope.yaml
+++ b/charts/telepresence/templates/trafficManagerRbac/namespace-scope.yaml
@@ -7,7 +7,7 @@ require less permissions in clientRbac.yaml
 */}}
 {{- if .Values.managerRbac.namespaced }}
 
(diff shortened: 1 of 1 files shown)


thallgren:
@fordneild , can you give a bit more context to this? Our CI tests will indeed provide a chart release-name when running, and for an external CI where Telepresence is involved, I'd imaging that you'd use the Helm chart that is published [here](https://artifacthub.io/packages/helm/telepresence-oss/telepresence-oss).

thallgren:
Closing due to unresponsive reporter.

PR #3718, opened 2024-11-04
Add deployments, statefulSets, replicaSets to workloads Helm chart values

P0lip:
## Description

This PR adds the remaining workload kinds Telepresence is able to intercept to the `workloads` option in Helm chart. This based on the initial suggestion made by @thallgren in 
https://github.com/telepresenceio/telepresence/pull/3651#discussion_r1689392081.
To retain backwards compatibility all existing workload kinds that is Deployment, StatefulSet and ReplicaSet are enabled by default unless configured otherwise.

## Checklist

### DIFF
--- a/charts/telepresence/templates/deployment.yaml
+++ b/charts/telepresence/templates/deployment.yaml
@@ -81,9 +81,26 @@ spec:
             value: {{ .grpc.maxReceiveSize }}
           {{- end }}
           {{- end }}
-          {{- if .workloads.argoRollouts }}
-          - name: ARGO_ROLLOUTS_ENABLED
-            value: {{ .workloads.argoRollouts.enabled | quote }}
+          {{- if .workloads }}
+          {{- with .workloads }}
+          - name: ENABLED_WORKLOAD_KINDS
+            value: >-
+              {{- if or (not .deployments) .deployments.enabled }}
+              Deployment
+              {{- end }}
+              {{- if or (not .statefulSets) .statefulSets.enabled }}
+              StatefulSet
+              {{- end }}
+              {{- if or (not .replicaSets) .replicaSets.enabled }}
+              ReplicaSet
+              {{- end }}
+              {{- if and .argoRollouts .argoRollouts.enabled }}
+              Rollout
+              {{- end }}
+          {{- end }}
+          {{- else }}
+          - name: ENABLED_WORKLOAD_KINDS
+            value: Deployment StatefulSet ReplicaSet
           {{- end }}
       {{- if .agentInjector.enabled }}
         {{- /*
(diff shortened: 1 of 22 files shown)


charts/telepresence/templates/deployment.yaml
```diff
@@ -81,9 +81,26 @@ spec:
...
+              {{- if or (not .replicaSets) .replicaSets.enabled }}
+              replicaset
+              {{- end }}
+              {{- if and .argoRollouts .argoRollouts.enabled }}
+              rollout
+              {{- end }}
+          {{- end }}
+          {{- else }}
+          - name: ENABLED_WORKLOAD_KINDS
+            value: deployment statefulset rollout
```
thallgren: This implies that `Rollout` is enabled by default and `ReplicaSet` is disabled. I don't think we want that, given that `Rollout` is a third-party extension.

charts/telepresence/templates/deployment.yaml
```diff
@@ -81,9 +81,26 @@ spec:
...
+              {{- end }}
+              {{- if or (not .statefulSets) .statefulSets.enabled }}
+              statefulset
+              {{- end }}
+              {{- if or (not .replicaSets) .replicaSets.enabled }}
+              replicaset
+              {{- end }}
+              {{- if and .argoRollouts .argoRollouts.enabled }}
+              rollout
+              {{- end }}
```
thallgren: Can we please use `Deployment`, `StatefulSet`, `ReplicaSet`, and `Rollout` here for consistency? I believe that two representations are enough (the plural with leading lowercase being the other).

pkg/workload/watcher.go
```diff
@@ -34,6 +35,19 @@ type WorkloadEvent struct {
 	Workload k8sapi.Workload
 }
 
+type WorkloadKind string
+
+const (
+	DeploymentWorkloadKind  WorkloadKind = "deployment"
+	StatefulSetWorkloadKind WorkloadKind = "statefulset"
+	ReplicaSetWorkloadKind  WorkloadKind = "replicaset"
+	RolloutWorkloadKind     WorkloadKind = "rollout"
```
thallgren: Same here, please use CamelCase.

thallgren (changes requested):
This looks great overall, but I'm reluctant to introduce the lowercase singular kind.

charts/telepresence/templates/deployment.yaml
P0lip: ah, it's a typo. I meant to include `replicaset` there and not `rollout`.
Good catch.

thallgren (approved):

PR #3727, opened 2024-11-21
add is_license_status field to provide validation status

FuYu3699:
## Description

Added a new field in License object to provide the if a license is valid.

## Checklist


rpc/manager/manager.proto
```diff
@@ -429,6 +429,8 @@ message License {
   // cluster_id is really the id of the namespace where the traffic-manager is installed.
   string cluster_id = 3;
   string err_msg = 4;
+  // is_license_valid indicates whether the license is valid.
+  bool is_license_valid = 5;
```
thallgren:
```suggestion
  bool valid = 5;
```
This is a field in the structure named "License" so repeating the word "license" in the field name is redundant. Also, since it's a `bool`, it either is or isn't, which makes the "is" word redundant too.

Based on that, I think the name "valid" is better. This in turn makes the comment "valid indicates whether the license is valid" somewhat redundant too.

FuYu3699: Thanks @thallgren, I renamed it to `valid`

thallgren (approved):

FuYu3699:
I need to make this change on top of release/v2.20.3 instead.

PR #3736, opened 2024-12-02
Introduce telepresence ingest

thallgren:
This PR introcudes the new `telepresence ingest` command along with a lot of other
improvements and fixes that were made during the implementation work. Details are
in the commit comments.


pkg/client/remotefs/sftp.go
```diff
@@ -42,11 +43,22 @@ func (m *sftpMounter) Start(ctx context.Context, id, clientMountPoint, mountPoin
...
-		defer dlog.Infof(ctx, "Unmounting SFTP file system for intercept %q (pod %s) at %q", id, podIP, clientMountPoint)
+		dlog.Infof(ctx, "Mounting SFTP file system for container %s[%s] (pod %s) at %q", workload, container, podIP, clientMountPoint)
+		defer func() {
+			dlog.Infof(ctx, "Unmounting SFTP file system for container %s[%s] (pod %s) at %q", workload, container, podIP, clientMountPoint)
+			time.Sleep(time.Second)
+
+			// sshfs sometimes leave the mount point in a bad state. This will clean it up
+			ctx, cancel := context.WithTimeout(context.WithoutCancel(ctx), time.Second)
+			defer cancel()
+			umount := proc.CommandContext(ctx, "fusermount", "-uz", clientMountPoint)
```
P0lip: `fusermount` is unavailable on MacOS. One would need to use plain `umount` instead.

pkg/client/userd/trafficmgr/podaccess.go
```diff
@@ -0,0 +1,272 @@
...
+	"github.com/telepresenceio/telepresence/v2/pkg/agentconfig"
+	"github.com/telepresenceio/telepresence/v2/pkg/client"
+	"github.com/telepresenceio/telepresence/v2/pkg/client/remotefs"
+	"github.com/telepresenceio/telepresence/v2/pkg/forwarder"
+	"github.com/telepresenceio/telepresence/v2/pkg/iputil"
+)
+
+type podAccess struct {
+	// ctx is a context canceled by the cancel attribute. It must be used by
+	// services that should be canceled when the intercept ends
```
P0lip: a nitpick, but some comments in this file weren't updated when after the move
```suggestion
	// services that should be canceled when the intercept/ingest ends
```

P0lip (approved):
I gave it a try locally and it works very well.

Issue #3731, opened 2024-11-22
Telepresence chart cannot be included as a child Helm chart unless parent chart is called "traffic-manager"


dansimone:
**Please describe your use case / problem.**

I saw this mentioned in another [issue](https://github.com/telepresenceio/telepresence/issues/2329) that was closed almost 2 years ago regarding the ability to include Telepresence as the child chart of another Helm chart.

It would be useful to be able to include the Telepresence Helm chart as the child Helm chart.  Today, this is only possible if the parent chart is literally called "traffic-manager", due to [this](https://github.com/telepresenceio/telepresence/blob/release/v2/charts/telepresence/templates/_helpers.tpl#L14) check in `_helpers.tpl`.

I saw a justification for why this was done in https://github.com/telepresenceio/telepresence/issues/2329, but it seems that an alternative would be to:
- Allow a top-level `name` to be specified in the [values.yaml](https://github.com/telepresenceio/telepresence/blob/release/v2/charts/telepresence/values.yaml), defaulting to "".
- $name can be set to the value of `.name` if set, otherwise fall back to `.Release.Name`.

But actually, in the RBAC setup yamls, I don't see anywhere where it matters what the Release.Name was, unless I am missing something.  Is there some other use for this constraint?

**Versions (please complete the following information)**
- v1.20.3

thallgren:
This seems very reasonable. @dansimone would you be able to help with this and create a PR?

dansimone:
@thallgren, absolutely.  Please have a look at this: https://github.com/telepresenceio/telepresence/pull/3737/files.  I'll send this to the Slack channel shortly.

dansimone:
Ok, looks like I can't join the Slack channel (datawire-oss) mentioned in the PR template because of: 
```
You can use any account with the domain:
canstar.com.au
box.com
```

@thallgren, would you mind sending this out for review?

thallgren:
The link in the PR template was out of date. I've changed it to:

> in #telepresence-oss channel on the [CNCF Slack](https://slack.cncf.io/)

Closing this now, since your PR has been merged and will be included in the next release.

PR #3737, opened 2024-12-02
Lift requirement of .Release.Name == traffic-manager

dansimone:
## Description

Addresses https://github.com/telepresenceio/telepresence/issues/3731, which is an apparently unnecessary restriction that the .Release.Name under which telepresence is installed is called "traffic-manager".  This restriction was preventing telepresence from being included as a sub-chart in a parent chart called anything but "traffic-manager".

This change just does the same for "traffic-manager.name" that we already do [here](https://github.com/telepresenceio/telepresence/compare/release/v2...dansimone:telepresence:release/v2?expand=1#diff-4df592d4dba16f53e48009e7161d000c794bbdf00e385406a2e491fa8ac6b956L4) for "telepresence.name": it has a default value, but can otherwise be controlled by the `nameOverride` value.

The rest of the Helm chart is already not referring to the literal "Release.Name" in any meaningful way - everything meaningful is referring to the "traffic-manager.name".

## Checklist

### DIFF
--- a/CHANGELOG.yml
+++ b/CHANGELOG.yml
@@ -70,7 +70,7 @@ items:
           Telepresence will now use WebSockets instead of SPDY when creating port-forwards to the Kubernetes Cluster, and
           will fall back to SPDY when connecting to clusters that don't support SPDY. Use of the deprecated SPDY can be
           forced by setting `cluster.forceSPDY=true` in the `config.yml`.
-          
+
           See [Streaming Transitions from SPDY to WebSockets](https://kubernetes.io/blog/2024/08/20/websockets-transition/)
           for more information about this transition.
       - type: feature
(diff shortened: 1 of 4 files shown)


thallgren (changes requested):
Thanks for providing this.

I think this PR needs more work, because if I read it correctly, it doesn't just lift the restriction on the ReleaseName, it will also allow the actual kubernetes service to be named something other than "traffic-manager". This is problematic, because both client and the traffic-agent will use the service name "traffic-manager" verbatim in multiple places in the code when establishing gRPC connections.

I think the simple solution would be to let the service name stay fixed at "traffic-manager" (perhaps also the deployment and container name), but an improvement where the code is rewritten to instead use a selector matching the labels would be a very good addition

Did you test this PR with another name and managed to connect?

dansimone:
Good point, you're right.  Let me test out a simpler solution tomorrow, basically just what you said: just keep the traffic-manager.name hard-coded to "traffic-manager", and removing the reference the Release.Name (which I don't think has any bearing otherwise).

The following could be a way to address the hard-coding overall:
1) Merge the above simple fix
2) Update the actual code to use a selector instead of looking for a hard-coded "traffic-manager" service.  I am up for helping with this fix as well.

dansimone:
Ok, updated with the simpler solution for 1).  With this, I can install Telepresence and connect to it using the primary Helm chart or through a parent chart with an arbitrary name.

thallgren (approved):
Looks great. Thanks.

PR #3740, opened 2024-12-07
Release/v2.21.0

thallgren:


docs/howtos/intercepts.md
```diff
@@ -65,7 +65,7 @@ network telepresence, and remote mounts must be made relative to a specific moun
 
 4. Intercept all traffic going to the application in your cluster:
     ```
-    telepresence intercept <workload-name> --port [<local-port>][:<remote-port>] --env-file <path-to-env-file>`.
+    telepresence intercept &lt;workload-name&gt; --port [&lt;local-port&gt;][:&lt;remote-port&gt;] --env-file &lt;path-to-env-file&gt;`.
```
P0lip: any reason for these HTML character entities?
They aren't really compliant with GFM, so the preview on GH is somewhat broken - less than/greater than inside of a fenced code block should already be escaped as `&lt;` and `&gt;`.
Do we need that due to the telepresence.io app?

cmd/traffic/cmd/manager/config/config.go
```diff
@@ -115,9 +124,27 @@ func (c *config) refreshFile(ctx context.Context, data map[string]string) {
 		c.clientYAML = nil
 		dlog.Debugf(ctx, "Cleared client config")
 	}
+
+	c.agentEnv = AgentEnv{}
+	if yml, ok := data[agentEnvConfigFileName]; ok {
+		data, err := yaml.YAMLToJSON([]byte(yml))
+		if err == nil {
+			err = json.Unmarshal(data, &c.agentEnv)
+		}
```
P0lip, line 133:
```
nit - don't need to address it now, but at some point. YAML 1.2 much compliant with JSON assuming JSON or core schemas are used (core is the recommended default for processors). There are a few differences that should be irrelevant in this context.
As such, one can do
```suggestion
		data, err := yaml.Unmarshal([]byte(yml))
```
to handle both file types.

pkg/client/cli/cmd/uninstall.go
```diff
@@ -2,88 +2,76 @@ package cmd
...
 func uninstall() *cobra.Command {
 	ui := &uninstallCommand{}
 	cmd := &cobra.Command{
-		Use:  "uninstall [flags] { --agent <agents...> | --all-agents }",
-		Args: ui.args,
-
-		Short:             "Uninstall telepresence agents",
-		RunE:              ui.run,
-		ValidArgsFunction: cobra.NoFileCompletions,
+		Use:   "uninstall [flags] <workloads...>",
```
P0lip, line 30: it might be good to mention that positional argument in a changelog perhaps?

pkg/client/userd/trafficmgr/session.go
```diff
@@ -1037,9 +1034,10 @@ func (s *session) connectRootDaemon(ctx context.Context, nc *rootdRpc.NetworkCon
 		}()
 		rd = rootdRpc.NewDaemonClient(conn)
 
+		tmTimeout := client.GetConfig(ctx).Timeouts().Get(client.TimeoutTrafficManagerConnect)
 		for attempt := 1; ; attempt++ {
 			var rootStatus *rootdRpc.DaemonStatus
-			tCtx, tCancel := context.WithTimeout(ctx, 15*time.Second)
+			tCtx, tCancel := context.WithTimeout(ctx, tmTimeout/2)
```
P0lip, line 1040: I assume we divide by 2 to effectively keep the previous value (assuming the default 60s was not overriden).

docs/reference/client.md
```diff
@@ -13,15 +13,24 @@ The [Telepresence CLI client](../quick-start.md) is used to connect Telepresence
...
+| `docker-run`     | run a docker image in a container that shares the network established by a connect.  Especially useful when using `connect --docker`.                                                                                                                                                                                                                                                                              |
+| `gather-logs`    | Gather logs from traffic-manager, traffic-agents, user, and root daemons, and export them into a zip file that can be shared with others or included with a github issue. Use `--get-pod-yaml` to include the yaml for the `traffic-manager` and `traffic-agent`s. Use `--anonymize` to replace the actual pod names + namespaces used for the `traffic-manager` and pods containing `traffic-agent`s in the logs. |
```
P0lip, line 23:
```suggestion
```

P0lip (approved):

docs/howtos/intercepts.md
thallgren: Yeah, I made that change after a build failure from docusaurus suggesting that this was treated as mdx, which it shouldn't be given the extension. I'll investigate this further, because the preview on GH should be as good as it can be.

cmd/traffic/cmd/manager/config/config.go
thallgren: There's no yaml that can handle the JSON functionality in the json package that we're using now. (such as the `json` tags and `omitzero` See https://github.com/go-json-experiment/json for more info.

pkg/client/cli/cmd/uninstall.go
thallgren: Yep, will address that.

pkg/client/userd/trafficmgr/session.go
thallgren: No, it actually changes value from 15s to 30s :-) (and makes it dependent on the `trafficManagerConnect`).

P0lip: ah yeah, I meant 60 (the default value as I mentioned there) divided by 2 which is indeed 30. I don't know why but I saw `15*time.Second` as `2*15*time.Second` in my head

PR #3746, opened 2024-12-15
Allow serverless ingest without specifying a ports annotation

thallgren:
The ability to intercept a workload without a service is built around the `telepresence.getambassador.io/inject-container-ports` annotation, and it was also required to ingest such a workload. This was counterintuitive and the requirement was removed. An ingest doesn't use a port.


go.mod
```diff
@@ -172,21 +172,22 @@ require (
...
 	sigs.k8s.io/json v0.0.0-20241014173422-cfa47c3a1cc8 // indirect
 	sigs.k8s.io/kustomize/api v0.18.0 // indirect
 	sigs.k8s.io/kustomize/kyaml v0.18.1 // indirect
-	sigs.k8s.io/structured-merge-diff/v4 v4.4.3 // indirect
+	sigs.k8s.io/structured-merge-diff/v4 v4.5.0 // indirect
 )
 
 replace github.com/telepresenceio/telepresence/rpc/v2 => ./rpc
+
+replace github.com/datawire/k8sapi => github.com/thallgren/k8sapi v0.0.0-20241214214806-eba2ff884c63
```
P0lip: now that k8sapi change is merged, we can update it

P0lip (approved):

PR #3749, opened 2024-12-17
Get rid of old code that retrieves "clusterID".

thallgren:
The clusterID was replaced with installID some time ago, the difference being that the clusterID was the UID of the "default" namespace whereas the installID is the UID of the namespace where the traffic-manager is installed.


cmd/traffic/cmd/manager/cluster/info.go
```diff
@@ -377,6 +360,12 @@ func (oi *info) watchPodSubnets(ctx context.Context, namespaces []string) {
 func (oi *info) setSubnetsFromEnv(ctx context.Context) bool {
 	subnets := managerutil.GetEnv(ctx).PodCIDRs
 	if len(subnets) > 0 {
+		mgrIp, _ := netip.AddrFromSlice(oi.ManagerPodIp)
```
P0lip, line 363: IMHO this change may deserve a separate commit.

P0lip (approved):

cmd/traffic/cmd/manager/cluster/info.go
thallgren: Doh, I missed this comment. But yeah, I agree. It should have been a separate commit.

PR #3759, opened 2024-12-29
Dynamic namespace support

### LINKED ISSUE #3247: Namespace selected via selector
I set up a namespace-scoped traffic-manager, by inserting all the names of the namespace in which I wanted it to have permissions (same for `clientsRbac`)
This works fine, but if I will have more namespace I will update `values.yaml` everytime.

Since the namespaces in my clusters have common labels, it would be nice to allow us to select namespaces using selectors.

-

```
Client         : v2.13.1
Root Daemon    : v2.13.1
User Daemon    : v2.13.1
Traffic Manager: v2.14.0

```
  - Kubernetes Environment and Version: 1.24

thallgren:
This PR changes how the Telepresence traffic-manager and mutating webhook deals with namespaces so that everything is controlled by one single namespace selector.

The top level Helm chart value `namespaceSelector` was added. It is a verbatim Kubernetes namespace selector, and deprecates the `agentInjector.webhook.namespaceSelector` and `managerRbac.namespaces`. It also acts as the default for `clientRbac.namespaces`.

The short form value `namespaces` was also added. It is a list of names and mutually exclusive to `namespaceSelector`. The list will get transformed into:
```yaml
namespaceSelector:
  matchExpressions:
  - key: kubernetes.io/metadata.name
    operator: In
    values: [ <namespaces> ]
```

A selector can be _dynamic_ or _static_. This in turn controls if telepresence is "cluster-wide" or "namespaced". A _dynamic_ selector
requires cluster-wide permissions for the traffic-manager, and only a _static_ selector can serve as base when installing `Role`/`RoleBinding` pairs.

A selector is considered static if it meets the following conditions:
- The selector must have exactly one element in the `matchLabels` or the
  `matchExpression` list (if the element is in the `matchLabels` list,
   it is normalized into "key in [value]").
- The element must meet the following criteria:
  - The `key` of the element must be "kubernetes.io/metadata.name".
  - The `operator` of the element must be "In" (case-sensitive).
  - The `values` list of the element must contain at least one value.


charts/telepresence/templates/trafficManagerRbac/cluster-scope.yaml
```diff
@@ -60,7 +54,7 @@ rules:
   - list
   - get
   - watch
-{{- if .Values.agentInjector.enabled }}
+{{- if .agentInjector.enabled }}
```
P0lip, line 57: I know it's an older code, but I believe this could be merged with the above. We'd only need to add `create` here and the previous definition could be dropped.

cmd/traffic/cmd/manager/cluster/info.go
```diff
@@ -304,6 +296,21 @@ func clusterDomainFromResolvConf(confFile, namespace string) (string, error) {
 }
 
 func (oi *info) watchNodeSubnets(ctx context.Context, mustSucceed bool) bool {
+	ok, err := k8sapi.CanI(ctx, &auth.ResourceAttributes{
```
P0lip: A loose suggestion, but it'd be nice if `k8sapi.CanI` took an slice of `*auth.ResourceAttributes` or (make it a variadic function), so we could dedupe some of the code here and one or two other places.

cmd/traffic/cmd/manager/config/config.go
```diff
@@ -1,60 +1,81 @@
...
+	// ForceEvent is for testing purposes only.
+	ForceEvent(ctx context.Context) error
 }
 
 type AgentEnv struct {
 	Excluded []string `json:"excluded,omitempty"`
 }
 
+func (ae AgentEnv) Equal(o AgentEnv) bool {
+	return slices.Equal(ae.Excluded, o.Excluded)
```
P0lip, line 46: a minor nit, but the order of elements shouldn't really matter in this case, so we could potentially account for it like we do in some other places.
We don't seem to sort them anywhere, we could do that after the data is unmarshaled.

cmd/traffic/cmd/manager/namespaces/watcher.go
```diff
@@ -0,0 +1,297 @@
...
+			dlog.Debug(ctx, "Listing namespaces is not permitted")
+			staticErr := h.computeNamesStatic(ctx)
+			if staticErr == nil {
+				return false, nil
+			}
+			dlog.Error(ctx, staticErr)
+		}
+		if err != nil {
+			return false, fmt.Errorf("error listing namespaces: %v", err)
+		}
```
P0lip:
```suggestion
	  return false, fmt.Errorf("error listing namespaces: %v", err)
```

P0lip (approved):
LGTM. I like this change.

charts/telepresence/templates/trafficManagerRbac/cluster-scope.yaml
thallgren: No, I don't think that's correct. This entry is specifically for the configmap named "telepresence-agents".  The entry above isn't (using create isn't possible for a named resource).

PR #3761, opened 2025-01-05
Remove the dormant container during intercept with --replace.

### LINKED ISSUE #3758: Allow customisation of sleeperImage in `--replace` mode for compliance with cluster restrictions
When using Telepresence in `--replace` mode to intercept a deployment, the service container is replaced with a "sleeperImage" ([currently hardcoded](https://github.com/telepresenceio/telepresence/blob/e8675c034447ed3bfdbf90ab402ead9e9e1deea4/cmd/traffic/cmd/manager/mutator/agent_injector.go#L228) to `alpine:latest`). However, in my environment, I face restrictions on the Docker registries and versions of images allowed in the cluster. This hardcoding prevents me from using Telepresence effectively in `--replace` mode due to compliance and compatibility constraints.

I need the `--replace` mode specifically for services that run background tasks, as the ordinary interception via sidecar only redirects traffic to the local process but does not handle the original container's background jobs, which continue working. While sidecar-based interception is fast and sufficient for most use cases, it is inadequate for scenarios involving tasks such as batch processing or scheduled jobs.

I would like the ability to override the sleeperImage (including the image registry and tag) used in `--replace` mode. This could be implemented in a way that allows flexibility, such as:
- Adding configurable values in the Telepresence Helm chart.
- Providing CLI parameters to specify the image, registry, and tag during the interception setup.

Either option would work, though the former might be better for environments with stricter cluster-wide policies.

Manually stopping the original container before each interception. 
While this could stop the background processes temporarily, it introduces challenges with liveness probes configured on the deployment, which may trigger pod restarts or other issues.

  - Output of `telepresence version`:
```
OSS Client         : v2.21.1
OSS Root Daemon    : v2.21.1
OSS User Daemon    : v2.21.1
OSS Traffic Manager: v2.21.1
Traffic Agent      : telepresenceio/tel2:2.21.1
```  

  - Kubernetes Environment and Version
all versions

**Additional context**
This feature would significantly enhance Telepresence's usability for services with background tasks and in restricted environments. Example of the command: `telepresence intercept <name> --port <ports> --env-file <env-file> --replace`

### LINKED ISSUE #3807: When using the replace option in an intercept, the 'command' for the new pod should not be the same as for the original deployment
**Describe the bug**
We are using the replace option when intercepting. However, the pod to intercept is configured with the following command:
```
command:
    - /bin/sh
    - '-c'
```
This command is then used by the pod which is replaced as well, leading it to crash:

Image
Image

This causes the old deployment to never be terminated, and so not correctly replaced.

**To Reproduce**
1. Have a deployment which has a 'command' configured
2. Intercept the deployment with the replace option
3. See that the replacement pod does not start up

**Expected behavior**
The replacement pod should start up, and the old pod should be terminated.

Telepresence version v2.21.3
Windows 10
K3D running Server Version: v1.31.4+k3s1

thallgren:
There's no need to have a dormant container because the traffic-agent
can serve as the replacement itself. That way, there's no need to
redirect ports either, which means that there's no need to rename ports
or use an init-container.

An annotation is added to the pod to cater for when multiple injections
happen due to replace-policy, so that injections that follow the one
when an app-container is removed can reproduce the traffic-agent
container, taking its ports, mounts, and environment into account.


pkg/forwarder/interceptor.go
```diff
@@ -40,14 +43,14 @@ type interceptor struct {
...
-		return newUDP(addr, targetHost, targetPort)
+func NewInterceptor(from agentconfig.PortAndProto, targetHost string, targetPort uint16) Interceptor {
+	switch from.Proto {
+	case core.ProtocolTCP:
+		return newTCP(from.Port, targetHost, targetPort)
+	case core.ProtocolUDP:
+		return newUDP(from.Port, targetHost, targetPort)
 	default:
-		panic(fmt.Errorf("unsupported net.Addr type %T", addr))
+		panic(fmt.Errorf("unsupported protocodl %s", from.Proto))
```
P0lip:
```suggestion
		panic(fmt.Errorf("unsupported protocol %s", from.Proto))
```
typo

cmd/traffic/cmd/manager/mutator/agent_injector.go
```diff
@@ -417,8 +397,18 @@ func compareVolumeMounts(a, b []core.VolumeMount) bool {
 	return eq
 }
 
-func containerEqual(a, b *core.Container) bool {
+func containerEqual(ctx context.Context, a, b *core.Container) bool {
 	// skips contain defaults assigned by Kubernetes that are not zero values
+	if dlog.MaxLogLevel(ctx) >= dlog.LogLevelDebug {
+		diff := cmp.Diff(a, b,
+			cmp.Comparer(compareProbes),
```
P0lip: a minor one too, but we already have options below, so could dedupe it a bit

```go
func containerEqual(ctx context.Context, a, b *core.Container) bool {
	// skips contain defaults assigned by Kubernetes that are not zero values
	options := cmp.Options{
		cmp.Comparer(compareProbes),
		cmp.Comparer(compareVolumeMounts),
		cmpopts.IgnoreFields(core.Container{}, "ImagePullPolicy", "Resources", "TerminationMessagePath", "TerminationMessagePolicy"),
	}
	if dlog.MaxLogLevel(ctx) >= dlog.LogLevelDebug {
		diff := cmp.Diff(a, b, options...)
		if diff != "" {
			dlog.Debug(ctx, diff)
		}
		return diff == ""
	}
	return cmp.Equal(a, b, options...)
}
```

pkg/agentconfig/container.go
```diff
@@ -313,6 +332,41 @@ func appendSecretVolume(env dos.Env, annotation, volumeName string, pod *core.Po
...
+				break
+			}
+		}
+	}
+	return result
+}
+
+func eachConfiguredContainer(configureContainers []*core.Container, config *Sidecar, f func(*core.Container, *Container)) {
+	for i, cn := range configureContainers {
+		f(cn, config.Containers[i])
```
P0lip: is it possible that a container will be nil at runtime?
Technically from the perspective of `ConfiguredContainers` function that might be the case if we're unable to find the configured container.
If so, we may want to filter out nil values here as the callbacks we pass to `eachConfiguredContainer` will panic due to them expecting a non-nil value.

P0lip (approved):
Just one typo.
I like the swap of `net.IP` with `netip.AddrPort`

pkg/agentconfig/container.go
thallgren: Good point, A nil entry in the `configContainers` should never be possible, but looking at the `ConfiguredContainers` function, I realize that if the `pod.Spec.Containers` for some reason is out of sync with the `config.Containers`, and no longer contain all the containers that were present at the time when the `config.Containers` was generated, then that could actually happen.

PR #3762, opened 2025-01-05
One single invocation of intercept can now intercept multiple ports.

thallgren:
It is now possible to intercept multiple ports with one single invocation of `telepresence intercept` by just repeating the `--port` flag.


pkg/client/userd/trafficmgr/intercept.go
```diff
@@ -333,21 +335,110 @@ func (s *session) ensureNoInterceptConflict(ir *rpc.CreateInterceptRequest) *rpc
...
+	mv := s.managerVersion
+	n := mv.Major - major
+	if n == 0 {
+		if n = mv.Minor - minor; n == 0 {
+			n = mv.Patch - patch
+		}
+	}
+	return int(n)
+}
+
```
P0lip, line 441: It might be better to turn it into `requireManagerVersion` and return the boolean value, or use `semver.Version` instead to make it clean.

```suggestion
func (s *session) satisfiesManagerVersion(r semver.Range) bool {
	return r(s.managerVersion)
}
```

and usage

```go
session.satisfiesManagerVersion(semver.MustParseRange(">=2.22.0"))
session.satisfiesManagerVersion(semver.MustParseRange("2.22.0"))
```
etc.


Alternatively, the function could return the severity as well, so major/minor/patch to be more reliable.

Some checks like the one in `ensureWatchers` may yield unexpected results if we have another patch release of traffic-manager 2.20.x (unlikely to happen, but you know what I mean).
As an example, we'll always have the result of `1` if the the traffic-manager version is `2.x.1` while inputs are `2.x.0` and `2.x-1.1`, as an example: `2.21.1` (version), `2.21.0`, `2.20.0`.

P0lip (approved):

pkg/client/userd/trafficmgr/intercept.go
thallgren: The problem with semver comparison is that a pre-release or build-release is less than the final release (hence the word `Finalized` in the function). The traffic-manager version isn't final during development.

So let's say I want to check that the traffic-manager is >=2.22.0. I can't just do that with a version range, because `2.22.0-alpha.10` is actually less. The alternative would be to use `>=2.21.n` where `n` is a very high value.

Also, it seems a bit redundant to parse well known numbers from a string. But perhaps that's just me being overly CPU-cycle conservative :-)

PR #3764, opened 2025-01-08
New telepresence replace command.

thallgren:
The new `telepresence replace` command simplifies and clarifies container replacement.

Previously, the `--replace` flag within the `telepresence intercept` command was used to replace containers. However, this approach introduced inconsistencies and limitations:

* **Confusion:** Using a flag to modify the core function of a command designed for traffic interception led to ambiguity.
* **Inaccurate Behavior:** Replacement was not possible when no incoming traffic was intercepted, as the command's design focused on traffic routing.

To address these issues, the `--replace` flag within `telepresence intercept` has been deprecated. The new `telepresence replace` command provides a dedicated and consistent method for replacing containers, enhancing clarity and reliability.

Key differences between `replace` and `intercept`:

1. **Scope:** The `replace` command targets and affects an entire container, impacting all its traffic, while an `intercept` targets specific services and/or service/container ports.
2. **Port Declarations:** Remote ports specified using the `--port` flag are container ports.
3. **No Default Port:** A `replace` can occur without intercepting any ports.
4. **Container State:** During a `replace`, the original container is no longer active within the cluster.

The deprecated `--replace` flag still works, but is hidden from the `telepresence intercept` command help, and will print a deprecation warning when used.


thallgren:
Still a draft. Integration and docs are in the works.

pkg/workload/watcher.go
```diff
@@ -106,19 +94,13 @@ func NewWatcher(ctx context.Context, ns string, enabledWorkloadKinds []Kind) (Wa
...
-				}
-
-			case "Rollout":
-				if slices.Contains(enabledKinds, RolloutKind) {
-					return true
-				}
+			kind := k8sapi.Kind(ref.Kind)
+			switch kind {
+			case k8sapi.DeploymentKind, k8sapi.RolloutKind:
+				return enabledKinds.Contains(kind)
```
P0lip, line 103: this introduces a tiny change in the behavior that might be unwanted.
Previously, if the kind wasn't listed as enabled, we'd continue the iteration as possibly there would be another kind that's enabled.
In other words, the old behavior would be


```go
  if enabledKinds.Contains(kind) {
    return true
  }
```

thallgren: There's no point in continuing the iteration. An object can only have one controller, and once the condition on line 111/99 is true, that controller has been found.

docs/reference/intercepts/cli.md
P0lip, line 1: I'm not sure how the hosted documentation URLs are built, but if they are inferred from the directory structure, we could rename the `intercepts` directory here

cmd/traffic/cmd/manager/mutator/service.go
P0lip, line 1: we still have panic recovery in we still have `serveMutatingFunc` - didn't you intend to remove it in https://github.com/telepresenceio/telepresence/commit/59fb3f70b84b9bb935308205a85cf3d3a0eb2f4c#diff-e6df1931e03199997ae8c5c0ec6325293de7862dd39912c2995d2099cfa06032?

pkg/client/cli/cmd/genyaml.go
```diff
@@ -421,6 +419,55 @@ func (g *genInitContainerInfo) run(cmd *cobra.Command, kubeFlags map[string]stri
...
+		Args:  cobra.NoArgs,
+		Short: "Generate YAML for the pod template metadata annotations.",
+		Long:  "Generate YAML for the pod template metadata annotations. See genyaml for more info on what this means",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return info.run(cmd, flags.Map(kubeFlags))
+		},
+	}
+	flags := cmd.Flags()
+	flags.StringVarP(&info.workloadName, "workload", "w", "",
+		"Name of the workload. If given, the configmap entry will be retrieved telepresence-agents configmap, mutually exclusive to --config")
```
P0lip:
```suggestion
		"Name of the workload. If given, the configmap entry will be retrieved from the telepresence-agents configmap, mutually exclusive to --config")
```

integration_test/workloads_test.go
```diff
@@ -24,6 +24,7 @@ func (s *connectedSuite) successfulIntercept(tp, wl, port string) {
 		2*time.Second, // polling interval
 	)
 
+	itest.TelepresenceOk(ctx, "loglevel", "trace")
```
P0lip: It'd be good to restore to the previous state
```suggestion
	itest.TelepresenceOk(ctx, "loglevel", "trace")
	defer itest.TelepresenceOk(ctx, "loglevel", "debug")
```

P0lip (approved):
Nice, I like the removal of tp agents configmap

PR #3797, opened 2025-02-15
Backward and forward compatibility testing

thallgren:
This PR enables the integration tests to run with an traffic-manager/traffic-agent pair of a different version than the binary being tested, so that compatibility tests can be added for release candidates.


thallgren:
Closing this because it will be split into more than one PR.

PR #3798, opened 2025-02-17
Port forward stability and race conditions

thallgren:
This PR improves the port-forward logic by switching from caching dialers using pod-name and namespace to instead use the pod-id. The old way of doing it didn't work well with stateful sets, because they reuse the pod name when pods are recreated.

The PR also contains a couple of stability improvements to the agent port-forward (agentpf) mechanism, and a bugfix in the traffic-manager related to intercept disposition changes from NO_AGENT to INTERCEPTED.

An update to the latest golangci-lint version was also added.


njayp (approved):
Seems ok, there are a couple pieces I'm not familiar with. The agentpf switching the dial-watcher on and off feels weird, could you explain why?

thallgren:
> The agentpf switching the dial-watcher on and off feels weird, could you explain why?

There are two things to consider:
1. Each "client" in the agentpf corresponds to a traffic-agent in the cluster, and there might be a large number of traffic-agents. The Telepresence daemon must communicate only with those clients that it engage with, or if there are no engagements, with at least one traffic-agent (to establish tunnels). The logic tries to keep the number of active clients to a minimum in order to save resources (each active client has an open kubernetes port-forward).
2. Only traffic-agents serving an intercept or a replace will ever dial the client, so there's no need for a dial-watcher unless the client status is "intercepted". Therefore, a client's dial-watcher is switched on and off in accordance with this status. If the agent is dormant (as is the case when ingesting or doing nothing at all with it), then there's no need to watch for dials.

PR #3802, opened 2025-02-19
Rename charts/telepresence to charts/telepresence-oss.

### LINKED ISSUE #3801: Helm chart path "fix"
This is rather an improvements than a bug fix. 
To allow telepresence-oss to be installed  using argocd will require dir name change.

currently if you pull `helm pull oci://ghcr.io/telepresenceio/telepresence-oss` oci chart and untar it
you'll find `telepresence` dir and not `telepresence-oss`. While it isn't RFC that small change will allow to use gitops to install  telepresence-oss.

**To Reproduce**
add argocd application
```
---
apiVersion: argoproj.io/v1alpha1
kind: ApplicationSet
metadata:
  name: telepresence
spec:
  generators:
    - list:
        elements:
          - env: dev
            cluster-name: cluster
  template:
    metadata:
      name: "telepresence-{{env}}"
    spec:
      project: telepresence
      sources:
        # Chart with updated dir/path  >> `ghcr.io/mike-sirs/helm-oci`
        - repoURL: elepresence-oss:2.21.3/telepresence
          chart: telepresence-oss
          targetRevision: 2.21.3
          helm:
            releaseName: telepresence
            values: |
              agentInjector:
                enabled: false
```
argo wont be able to find Chart.yaml since it expects `telepresence-oss/Chart.yaml instead of telepresence/Chart.yaml`

thallgren:
The Helm chart name "telepresence-oss" was inconsistent with its contained folder "telepresence". As a result, attempts to install the chart using an argo ApplicationSet failed. The contained folder was renamed to match the chart name.

This commit renames the folder and also removes old code related to "telepresence-crds". Telepresence OSS has no CRDs.

Closes #3801


pkg/client/cli/cmd/helm.go
```diff
@@ -108,12 +106,6 @@ func (ha *HelmCommand) addValueSettingFlags(flags *pflag.FlagSet) {
 	}
 }
 
-func (ha *HelmCommand) addCRDsFlags(flags *pflag.FlagSet) {
-	if HelmExtendFlagsFunc != nil {
```
P0lip, line 112: At this point, we should probably just remove `HelmExtendFlagsFunc` altogether as it's not consumed anymore.

P0lip (approved):

PR #3803, opened 2025-02-21
fix: set security context for init intercept

### LINKED ISSUE #3804: backoff restarting of the agent-init when PodSecurityContext is set
Hello,

First, thank you very much for this awesome work.

Unfortunately, we won't be able to force all users in our clusters quickly enough to avoid injecting the init container to deploy telepresence and use interception.

Documentation :
https://www.telepresence.io/docs/troubleshooting/#injected-init-container-doesnt-function-properly

Also for security reasons, we recommend deploying applications with a PodSecurityContext with a high UID number.

For example : 

```
pod.spec.securityContext:
  runAsUser: 10001
```

In this case the securityContext.runAsUser inherit of PodSecurityContext.runAsUser if securityContext.runAsUser is not set, so by wanting to intercept the traffic of an application, we obtain the errors :

```
2025-02-18 10:30:44.7234 info    Traffic Agent Init v2.21.2
2025-02-18 10:30:44.7263 error   failed to clear chain TEL_PREROUTING_TCP: running [/usr/sbin/iptables -t nat -N TEL_PREROUTING_TCP --wait]: exit status 4: iptables v1.8.10 (nf_tables): Could not fetch rule set generation id: Permission denied (you must be root)

2025-02-18 10:30:44.7263 error   quit: failed to clear chain TEL_PREROUTING_TCP: running [/usr/sbin/iptables -t nat -N TEL_PREROUTING_TCP --wait]: exit status 4: iptables v1.8.10 (nf_tables): Could not fetch rule set generation id: Permission denied (you must be root)
```

To avoid these configuration issues, a PR could resolve this dilemma : https://github.com/telepresenceio/telepresence/pull/3803

Thanks in advance.
Rgs

smehboub:
Fixes: https://github.com/telepresenceio/telepresence/issues/3804


thallgren:
I suspect that this might be a breaking change. I would prefer to see an option to inject the securityContext as a whole from the helm-chart, similar to how we handle the `resources` for the init-container. We already have this in the chart's values.yaml:
```yaml
agent:
  logLevel:
  resources: { }
  initResources: { }
```
Adding a
```yaml
  initSecurityContext: {}
```
here, inject its JSON as an environment variable unless empty in the deployment.yaml, and then add the logic needed to use that environment variable, would make the securityContext fully configurable.

cmd/traffic/cmd/manager/mutator/agent_injector.go
```diff
@@ -258,8 +258,7 @@ func addInitContainer(pod *core.Pod, config *agentconfig.Sidecar, patches PatchO
 		if ic.Name == oc.Name {
 			if ic.Image == oc.Image &&
 				slices.Equal(ic.Args, oc.Args) &&
-				compareVolumeMounts(ic.VolumeMounts, oc.VolumeMounts) &&
-				compareCapabilities(ic.SecurityContext, oc.SecurityContext) {
```
thallgren: I don't think this is correct. If the securityContext changes for some reason, then the init_container needs to be patched.

cmd/traffic/cmd/manager/mutator/agent_injector.go
```diff
@@ -324,29 +323,6 @@ func compareProbes(a, b *core.Probe) bool {
 	return eq
 }
 
-func compareCapabilities(a *core.SecurityContext, b *core.SecurityContext) bool {
```
thallgren: Please keep this in place.

cmd/traffic/cmd/manager/mutator/agent_injector_test.go
```diff
@@ -1518,10 +1518,6 @@ matchExpressions:
     image: ghcr.io/telepresenceio/tel2:2.13.3
     name: tel-agent-init
     resources: {}
-    securityContext:
```
thallgren: A good way to verify that this isn't a breaking change is that existing tests doesn't need to be modified. This test must be kept, and should succeed, because it verifies the default security context for the init_container. The same is true for other test changes.

You can add a test to verify that the default can be overridden.

pkg/agentconfig/container.go
```diff
@@ -268,15 +268,13 @@ func InitContainer(config *Sidecar) *core.Container {
 				},
 			},
 		},
-		SecurityContext: &core.SecurityContext{
```
thallgren: This is the default. It can be removed here, but then a default securityContext must be added to the chart's values.yaml. We can't just remove the default altogether.

thallgren (changes requested):
Looks promising, but it removes the default securityContext with the `NET_ADMIN` capability. That's still a breaking change. The default must remain.

smehboub:
> Looks promising, but it removes the default securityContext with the `NET_ADMIN` capability. That's still a breaking change. The default must remain.

Hello @thallgren 
Thank you very very much for the review.

cmd/traffic/cmd/manager/mutator/agent_injector.go
smehboub: Ok i understand, i made the change

cmd/traffic/cmd/manager/mutator/agent_injector.go
smehboub: Ok i understand, i made the change

cmd/traffic/cmd/manager/mutator/agent_injector_test.go
smehboub: Ok i understand, i made the change

pkg/agentconfig/container.go
smehboub: Ok i understand, i made the change

thallgren (approved):
Looks great. Thanks for providing this!

PR #3805, opened 2025-02-23
Add json-schema for the telepresence-oss Helm chart

thallgren:
Add json-schema for the telepresence-oss Helm chart

Helm can validate a chart using a json-schema using the command
`helm lint`, and this schema can be part of the actual Helm chart. The
telepresence-oss Helm chart now includes such a schema.

A new `telepresence helm lint` command was added so that linting can be
performed using the embedded chart.

More info [here](https://helm.sh/docs/topics/charts/#schema-files).


charts/telepresence-oss/values.schema.yaml
```diff
@@ -0,0 +1,580 @@
...
+        type: object
+        additionalProperties: false
+        properties:
+          enabled:
+            description: Enable/Disable the support for StatefulSets
+            type: boolean
+
+$defs:
+  duration:
+    pattern: "^[+-]?(\\d+(\\.\\d+)?(h|m|s|ms|us|µs))+"
```
P0lip, line 539: is there any reason we're not enforcing an instance type here too?
`pattern` is only applicable to string instances, so every other instance type is allowed unless restricted otherwise like at `#/properties/client/properties/connectionTTL`.

charts/telepresence-oss/values.schema.yaml
```diff
@@ -0,0 +1,580 @@
...
+
+  readinessProbe:
+    description: Define readinessProbe for the Traffic Manger.
+    $ref: "#/$defs/probe"
+
+  replicaCount:
+    description: Number or replicas for the traffic-manager. The Traffic Manager only support running with one replica at the moment.
+    type: integer
+    minimum: 1
+    maximum: 1
```
P0lip:
```suggestion
    const: 1
```
we could use `const` unless there's some tooling that doesn't support that keyword.

charts/telepresence-oss/values.schema.yaml
```diff
@@ -0,0 +1,580 @@
...
+            description: Enable/Disable the support for StatefulSets
+            type: boolean
+
+$defs:
+  duration:
+    pattern: "^[+-]?(\\d+(\\.\\d+)?(h|m|s|ms|us|µs))+"
+  rfc1123Label:
+    description: RFC 1123 label name
+    type: string
+    pattern: "^[a-z0-9][a-z0-9-]{1,62}$"
```
P0lip: I believe this regex isn't fully valid for rfc1123, as IIRC the label must end with an alphanumeric character, so it should be

```suggestion
    pattern: "^[a-z0-9](?:[a-z0-9-]{,61}[a-z0-9])?$"
```
this should be more valid

P0lip (approved):
Frankly speaking, I'm not sure whether `validate` isn't a more suitable command name than `lint`. I always think of `lint` as a set of actions to assert conformance with a certain opinionated ruleset, while `validate` is more commonly used for asserting compliance with a spec.
Either way, the difference is quite subtle, and both naming should do as good.

charts/telepresence-oss/values.schema.yaml
thallgren: Agree, but why use a non-capturing group? We don't use the result anyway so it just makes the regexp harder to read.

charts/telepresence-oss/values.schema.yaml
thallgren: Heh, I need to refresh my json-schema skills :-)

charts/telepresence-oss/values.schema.yaml
thallgren: No reason, the instance type should be added here.

thallgren:
Agree on "validate" and "lint". The name "lint" was chosen for consistency. It's the name of the helm command.

PR #3822, opened 2025-03-14
docs(chart): adds replicasCount to README

wrbbz:
## Description

Adds to Chart docs `replicaCount` as it was not mentioned there

## Checklist


thallgren:
Thanks for submitting this, but the `replicaCount` is omitted for a reason. It's defined as a constant value of 1, so it cannot be changed. Not sure what the value would be to add it.

wrbbz:
I've managed to change it via helm values

Nevertheless, why is it 1? Is there any reason for disabling HA mode?

thallgren:
Yes, there's a very good reason for disabling HA mode. It will simply not work, because the Traffic Manager keeps a fair amount of state in memory.

wrbbz:
So, there is no point in load balancing between different pods and all requests should be managed by the single pod?
Is there any performance benchmark? I'm wondering what resources request should be to handle, for example, a hundred simultaneous connections to traffic-manager

thallgren:
The traffic between the clients and the pods is not routed via the traffic-manager, so that "all requests are managed by a single pod" is a misconception. The traffic-manager just keeps track of the state of the pods, and in cases where there is no traffic-agent installed in any pod, it may act as a fallback for DNS resolution.

A client connection will typically not make that many requests to the manager, so performance should never be an issue unless you have a very large cluster. When that's the case, it's probably wise to install several managers with different namespace selectors anyway.

We've never seen any reason to make the traffic-manager use a persistent store (which would be required for HA). If you have use-cases that clearly shows that our thinking here is wrong, and if you're willing to sponsor the effort (with contributions or funding), then we'll of course be interested in revisiting this.

wrbbz:
Got it. Thank you for the clarification!

Closing this MR then

Issue #3836, opened 2025-04-04
Allow Environment Variable Overrides in `--docker-run`


kbatalin:
**Please describe your use case / problem.**
When using `telepresence intercept --docker-run`, Telepresence automatically fetches environment variables from the targeted pod and sets them as-is in the local Docker container. However, there are many cases where developers need more control over these environment variables. For example, you might want to test a feature that relies on a modified or additional variable, or remove a variable to see how the application behaves. Or when you need to temporarily remove multi-line environment variables because they're not supported. Currently, there's no direct way to override these environment variables without resorting to workarounds (like manually ingesting and editing the environment variables or creating a custom image with a special startup script). This limits ability to quickly experiment with different configurations.

**Describe the solution you'd like**
Provide a way to directly manipulate the environment variables passed to the Docker container. This could be done via flags or a configuration file that allows adding new variables, overriding existing values, or removing certain variables before `telepresence intercept --docker-run` spins up the container.

**Describe alternatives you've considered**
1. Manual ingest and container run: Using `telepresence ingest --env-file` to fetch environment variables, modifying them, then running a container manually. However, this bypasses Telepresence's automatic volume mounting and other interception features.
2. Custom startup script: Creating a custom image that modifies or adds environment variables upon startup. This requires additional maintenance of a separate image just for testing purpose.
3. Temporarily deleting a problematic env variable (e.g., a multi-line one) from the pod in the cluster, running Telepresence, then re-adding the env var afterward. This affects other developers.

**Versions (please complete the following information)**
  - Output of `telepresence version`: v2.21.1

**Additional context**
Allowing environment variable manipulation would greatly streamline local development and testing scenarios. It's a common need to temporarily adjust configuration to replicate edge cases or test new features without permanently modifying upstream configurations or building custom images.

thallgren:
Thanks for raising this issue @kbatalin.

It is possible to send flags to the underlying `docker run`. You must place those flags *after* the stand-alone `--` argument but *before* the name of the docker image. So you can add `-e x=y` flags here, and I believe you could even add `--env-file <your own env file>`, here because a `docker run` allows multiple such flags.

Hope this helps.

P.S.
I'm aware that the help text for `telepresence ingest` and friends could be improved to make this clearer, and I'm working on a PR for that.

thallgren:
The fix with improved help texts was included in the 2.22.3 version.

PR #3838, opened 2025-04-04
Change the help text for the engagement sub-commands

thallgren:
The new help text clarifies what is expected after the stand-alone double dash, so that the user understands that options to docker run must come before the name of the image to run.


thallgren:
Relates to #3836

PR #3850, opened 2025-04-22
docs: troubleshooting for GitLab pipelines

Blarc:
## Description

Described a solution for troubleshooting DNS in GitLab pipelines.

## Checklist

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -150,3 +150,16 @@ for Telepresence to connect to your cluster.
 ## `too many files open` error when running `telepresence connect` on Linux
 
 If `telepresence connect` on linux fails with a message in the logs `too many files open`, then check if `fs.inotify.max_user_instances` is set too low. Check the current settings with `sysctl fs.inotify.max_user_instances` and increase it temporarily with `sudo sysctl -w fs.inotify.max_user_instances=512`. For more information about permanently increasing it see [Kernel inotify watch limit reached](https://unix.stackexchange.com/a/13757/514457).
+
+## DNS does not resolve in GitLab pipeline
+
+If services are not resolved after running `telepresence connect` in GitLab pipeline, check if they can be reached directly via the IP. If that is the case, the nameserver is probably not configured correctly and can be fixed by setting the kube-dns as the nameserver in `/etc/resolv.conf`:
+```bash
+# namespace variable needs to be set beforehand
+kube_dns_service_ip="$(kubectl get svc -n kube-system kube-dns -o jsonpath='{.spec.clusterIP}')"
+cat > /etc/resolv.conf <<EOF
+nameserver ${kube_dns_service_ip}
+search ${namespace}.svc.cluster.local svc.cluster.local cluster.local
+options ndots:5
+EOF
+```


docs/troubleshooting.md
```diff
@@ -150,3 +150,16 @@ for Telepresence to connect to your cluster.
...
+## DNS does not resolve in GitLab pipeline
+
+If services are not resolved after running `telepresence connect` in GitLab pipeline, check if they can be reached directly via the IP. If that is the case, the nameserver is probably not configured correctly and can be fixed by setting the kube-dns as the nameserver in `/etc/resolv.conf`:
+```bash
+# namespace variable needs to be set beforehand
+kube_dns_service_ip="$(kubectl get svc -n kube-system kube-dns -o jsonpath='{.spec.clusterIP}')"
+cat > /etc/resolv.conf <<EOF
+nameserver ${kube_dns_service_ip}
+search ${namespace}.svc.cluster.local svc.cluster.local cluster.local
+options ndots:5
```
thallgren, line 163: This will effectively bypass Telepresence's DNS server, so I'm a bit reluctant to include this tip without investigating further why that server isn't working.

How is DNS originally configured in the GitLab pipeline? I.e. what does the original `/etc/resolv.conf` look like?
Also, if you turn on debug-logging on the client, what does the daemon.log file say about DNS?

thallgren (commented):
Need more info

docs/troubleshooting.md
Blarc: Here are all Telepresence logs and the log from GitLab pipeline: [logs.zip](https://github.com/user-attachments/files/19903089/logs.zip).

In GitLab log you can see the original `/etc/resolv.conf`:
```
$ cat /etc/resolv.conf
# Generated by Docker Engine.
# This file can be edited; Docker Engine will not make further changes once it
# has been modified.
nameserver 10.1.10.1
search medius
# Based on host file: '/etc/resolv.conf' (legacy)
# Overrides: []
```

thallgren: I'm sorry. I miss the most relevant log for the DNS resolver. The daemon.log. Can you please include that too?

Blarc: How do I get it? To get the logs I ran `telepresence gather-logs -o logs.zip` at the end of the job, as can be seen in the GitLab log.

Keep in mind, I'm running this inside of a docker container, since that's how GitLab pipelines work:
```
$ telepresence status
OSS Daemon in container: Running
  Version           : 2.22.3
  Executable        : /usr/local/bin/telepresence
  Install ID        : 7e969871-b215-4b34-89c3-a9584db088c9
  Status            : Connected
  Kubernetes server : https://10.4.10.3:6443
  Kubernetes context: kubetest5
  Namespace         : gamayun-ui-447999
  Manager namespace : operators
  Mapped namespaces : [gamayun-ui-447999]
OSS Daemon in container: Running
  Version: v2.22.3
  DNS    : 
    Exclude suffixes: [.com .io .net .org .ru]
    Include suffixes: []
    Timeout         : 4s
  Subnets: (9 subnets)
    - 10.233.0.0/18
    - 10.233.64.0/24
    - 10.233.65.0/24
    - 10.233.66.0/24
    - 10.233.67.0/24
    - 10.233.68.0/24
    - 10.233.69.0/24
    - 10.233.70.0/24
    - 10.233.71.0/24
OSS Traffic Manager: Connected
  Version      : v2.22.3-rc.0
  Traffic Agent: ghcr.io/telepresenceio/tel2:2.22.3-rc.0
  ```

Comment (thallgren) [reply docs/troubleshooting.md:163]: Sorry, you're absolutely right. I missed that, and I do indeed see DNS being configured in the connector.log (there's only no daemon.log when running as root). DNS seems to be alive and kickin' in this log. It hooks into the existing DNS on IP 10.1.10.1 using `iptables`. In essence, everything sent to that IP will first go to the Telepresence DNS server and that server will then use the original DNS at that IP as its fallback.

The only thing I can think of that would might be problematic is the `search: medius` entry in the `/etc/resolv.conf`. Can you try replacing that with `search: tel2-search` or simply just remove it?

Unfortunately, the connector.log only covers the time spent during the `gather-logs` command. It would be very interesting if you could massage the DNS a bit (using nslookup, curl, or whatever triggers a lookup) with telepresence connected before you execute that command. The lookups should be visible in the both the connector.log and the log from the traffic-manager.

Comment (Blarc) [reply docs/troubleshooting.md:163]: Sorry for a late response, I was away on holiday.

If I remove all lines from `/etc/resolv.conf`, it works. Here are the logs: [logs2.zip](https://github.com/user-attachments/files/20158165/logs2.zip).

I've also ran two curls which hopefully show up in the logs.

**Comment by vsimon** [comment]
I came across this issue in the GitLab SaaS version, related to DNS and we don't have control over how the CI-runner/Docker is configured. Before and after Telepresence connects, the `resolve.conf` file contents are the same. The `resolve.conf` file appears to be bind-mounted at runtime so we can't replace/delete/move the file but we can overwrite its contents. 

For anyone that might run into this in the future, this helped:

```yaml
job:
  ...
  script:
    - telepresence connect
    - echo "nameserver 127.0.0.1" > /etc/resolv.conf # Telepresence runs a DNS server on port 53 but it can't update the bind-mounted resolv.conf file
```

thallgren:
Thanks @vsimon . I'll add this to our troubleshooting guide.

PR #3854, opened 2025-04-29
Helm: Fix nodeSelector type

peterjenkins1:
## Description

Fix incorrect type for [nodeSelector](https://github.com/smartlyio/telepresence/blob/release/v2/charts/telepresence-oss/values.schema.yaml#L578-L579) in Helm chart.

Kubernetes defines different types for `nodeSelector` (inside `PodSpec objects`) and `NodeSelector` (inside `NodeAffinity`, `VolumeNodeAffinity` and a bunch of other places). This PR changes the type to be the right one.

Upstream type definitions:
 * [nodeSelector](https://github.com/kubernetes/api/blob/master/core/v1/types.go#L3963)
 * [NodeSelector](https://github.com/kubernetes/api/blob/master/core/v1/types.go#L3523-L3527)

`values.yaml` example
```yaml
# chart schema requires type io.k8s.api.core.v1.NodeSelector see: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.33/#nodeselector-v1-core
#
# nodeSelector:
#   nodeSelectorTerms:
#   - matchExpressions:
#     - key: example.com/example_label
#       operator: In
#       values:
#       - "true"
#  
# However that's not valid for a Deployment object which has a diffent syntax https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.30/#podspec-v1-core
#
nodeSelector:
  example.com/example_label: "true"
```

## Checklist


thallgren (approved):
Thanks for providing this. As I understand it, the `nodeSelector` should probably have been named something like `podTemplateNodeSelector` to begin with, but doing that now would be a breaking change, so this is all good.

I would like you to add a note in the CHANGELOG.yml though, and do a subsequent `make docs-files`. It's OK to target this change for the next minor release. We can always move it to a patch release if need be later on.

thallgren (approved):

PR #3859, opened 2025-05-16
ensures duplicate container port names don't cause intercept to fail.

### LINKED ISSUE #3858: `telepresence intercept` unsuccessful when two container port definitions share the same name
**Describe the bug**
I have multiple containers in one pod definition within my Deployment spec. Both containers have a port named `http`. This causes an error when I try to run `telepresence intercept`. To fix it, all I had to do was name each container port differently.

```
telepresence intercept: error: connector.CreateIntercept: rpc error: code = Unknown desc = Error creating: Pod "testpod-71f5d984fd-mjb02" is invalid: spec.containers[3].ports[1].name: Duplicate value: "http"
```

**To Reproduce**
1. Create a deployment with two containers
2. Name a port in each container the same name
3. Attempt to intercept one of those ports using `telepresence intercept`
4. See error.

**Expected behavior**
I expected telepresence intercept to handle the duplicate names automatically if it happens.

```
OSS Client         : v2.22.4
OSS Root Daemon    : v2.22.4
OSS User Daemon    : v2.22.4
OSS Traffic Manager: v2.22.4
Traffic Agent      : ghcr.io/telepresenceio/tel2:2.22.4
```
 
operating system: macOS 15.4.1

output of kubectl version
```
Client Version: v1.30.5
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: v1.24.11+k3s1
WARNING: version difference between client (1.30) and server (1.24) exceeds the supported minor version skew of +/-1
```

kfreezen:
## Description

This PR fixes #3858 by ensuring duplicate names have a `-{num}` appended to them. This prevents the "Duplicate value" I encountered.

## Checklist

 - [x] I made sure to add any documentation changes required for my change. **none needed AFAIK**

### DIFF
--- a/pkg/agentconfig/container.go
+++ b/pkg/agentconfig/container.go
@@ -36,7 +36,9 @@ func (a *ContainerBuilder) AgentContainer(ctx context.Context) (*core.Container,
 			for _, ic := range PortUniqueIntercepts(cc) {
 				name := ic.ContainerPortName
 
-				if _, ok := names[name]; ok {
+				// We don't want to apply duplication logic to empty strings
+				// as - is not a valid starting character for port names.
+				if _, ok := names[name]; ok && ic.ContainerPortName != "" {
 					// if name already exists, append a number to it
 					names[ic.ContainerPortName]++
 					// convert number to string
@@ -48,9 +50,11 @@ func (a *ContainerBuilder) AgentContainer(ctx context.Context) (*core.Container,
 					} else {
 						name = ic.ContainerPortName + number
 					}
-				} else {
+				} else if ic.ContainerPortName != "" {
 					name = ic.ContainerPortName
 					names[ic.ContainerPortName] = 1
+				} else {
+					name = ic.ContainerPortName
 				}
 
 				ports = append(ports, core.ContainerPort{
--- a/pkg/agentconfig/container.go
+++ b/pkg/agentconfig/container.go
@@ -26,6 +26,7 @@ func (a *ContainerBuilder) AgentContainer(ctx context.Context) (*core.Container,
 	ports := make([]core.ContainerPort, 0, 5)
 	confCns := a.configuredContainers(ctx)
 
+	names := make(map[string]int)
 	a.eachConfiguredContainer(confCns, func(app *core.Container, cc *Container) {
 		switch cc.Replace {
 		case ReplacePolicyContainer:
@@ -33,8 +34,27 @@ func (a *ContainerBuilder) AgentContainer(ctx context.Context) (*core.Container,
 			ports = append(ports, app.Ports...)
 		case ReplacePolicyIntercept:
 			for _, ic := range PortUniqueIntercepts(cc) {
+				name := ic.ContainerPortName
+
+				if _, ok := names[name]; ok {
+					// if name already exists, append a number to it
+					names[ic.ContainerPortName]++
+					// convert number to string
+					number := "-" + strconv.Itoa(names[ic.ContainerPortName])
+					// if string length of name plus number is greater than 15
+					if len(ic.ContainerPortName)+len(number) > 15 {
+						// truncate name to 15 characters
+						name = ic.ContainerPortName[:15-len(number)] + number
+					} else {
+						name = ic.ContainerPortName + number
+					}
+				} else {
+					name = ic.ContainerPortName
+					names[ic.ContainerPortName] = 1
+				}
+
 				ports = append(ports, core.ContainerPort{
-					Name:          ic.ContainerPortName,
+					Name:          name,
 					ContainerPort: int32(ic.AgentPort),
 					Protocol:      ic.Protocol,
 				})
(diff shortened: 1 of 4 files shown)


thallgren (approved):
I like this solution. It's simple and solves the problem. I'm a bit surprised that Kubernetes allow non-unique port names though, especially since it also allows that a service targets the port using a `targetPort: <name>`. That's surely a bug, because there's no way to determine what container that service would use.

Apparently there have been some attempts to fix it. It was reported in
https://github.com/kubernetes/kubectl/issues/1048 and later addressed in PR https://github.com/kubernetes/kubernetes/pull/105416. Both closed by the bot due to lack of resources and inactivity.

I'm mentioning this because it's probably not good practice to use non-unique port names, and maybe the bugfix entry in our changelog should mention that this is a fix to circumvent a bug in Kubernetes, should you do so anyway.

thallgren:
You need to run `make docs-files` when you make additions to the CHANGELOG.yml. Hence the build error.

kfreezen:
For some reason `make docs-files` creates an executable at `tools/bin` rather than `tools/bin/relnotesgen`. I was able to work around it by creating the `tools/bin` directory manually.

pkg/agentconfig/container.go
```diff
@@ -26,15 +26,35 @@ func (a *ContainerBuilder) AgentContainer(ctx context.Context) (*core.Container,
...
 	a.eachConfiguredContainer(confCns, func(app *core.Container, cc *Container) {
 		switch cc.Replace {
 		case ReplacePolicyContainer:
 			// Simply inherit the ports of the replaced container
 			ports = append(ports, app.Ports...)
 		case ReplacePolicyIntercept:
 			for _, ic := range PortUniqueIntercepts(cc) {
+				name := ic.ContainerPortName
+
+				if _, ok := names[name]; ok {
```
thallgren: Integration tests revealed a flaw. Unnamed ports must not be subjected to the duplication avoidance logic. As things stands now, the first unnamed port is OK, but the second is named "-2", and that's not a valid port name.

thallgren (changes requested):

pkg/agentconfig/container.go
kfreezen: I believe this is fixed.

pkg/agentconfig/container.go
```diff
@@ -36,7 +36,9 @@ func (a *ContainerBuilder) AgentContainer(ctx context.Context) (*core.Container,
 			for _, ic := range PortUniqueIntercepts(cc) {
 				name := ic.ContainerPortName
 
-				if _, ok := names[name]; ok {
+				// We don't want to apply duplication logic to empty strings
```
thallgren, line 39: This seems a bit contrived. Why do a lookup if the name is "" and then check if it is ""?

I'd rather you kept your old implementation but just fence it with:
```
if name != "" {
  your old impl here
}

**Review by thallgren — CHANGES_REQUESTED** [review]

File: pkg/agentconfig/container.go [hunk 3]
```diff
@@ -26,15 +26,39 @@ func (a *ContainerBuilder) AgentContainer(ctx context.Context) (*core.Container,
...
+						if len(ic.ContainerPortName)+len(number) > 15 {
+							// truncate name to 15 characters
+							name = ic.ContainerPortName[:15-len(number)] + number
+						} else {
+							name = ic.ContainerPortName + number
+						}
+					} else {
+						name = ic.ContainerPortName
+						names[ic.ContainerPortName] = 1
+					}
```
thallgren: I'm aware that this is a nit, but I believe this can be simplified further:
- The `name` is already assigned the value of `ic.ContainerPortName`, and can be used consistently in the rest of the logic. 
- Instead of discarding the number on the first lookup, it, it can be increased and just stored. This saves two hash lookups. Currently your `names[ic.ContainerPortName]++` construct will perform the same lookup again, and then a third lookup is made when converting the number to a string.
```suggestion
					if n, ok := names[name]; ok {
						// if name already exists, append a number to it
						n++
						names[name] = n
						// convert to numeric name suffix
						suffix := "-" + strconv.Itoa(n)
						// if string length of name plus suffix is greater than 15
						if len(name)+len(suffix) > 15 {
							// truncate name to 15 characters
							name = name[:15-len(suffix)]
						}
						name += suffix
					} else {
						names[name] = 1
					}
```

thallgren (changes requested):
This looks good now, aside from my nit about simplifying further.

Please sign your commits. The project requires signing to accept your contribution.

kfreezen:
Aaaaahhh sorry about that

thallgren (approved):

thallgren:
Merging this even though the CI-tests were failing. The reason for the failure is an upgrade to Kubernetes 1.33 which causes the traffic-manager to fail when it tries to establish the service CIDR. I'm working on a fix for that.

Issue #3863, opened 2025-05-21
[Linux] Telepresence fails to connect with many namespaces due to DBus message size limit


nahuel11500:
**Describe the bug**

On Linux machines (Ubuntu-based), Telepresence fails to connect when the number of accessible Kubernetes namespaces becomes large. This results in a cryptic error message about not being able to determine the traffic-manager namespace, or the user/root daemons not staying up. Upon deeper investigation, the root cause appears to be hitting a dbus message size limit due to an overly long argument list generated when querying all namespaces implicitly.

A workaround involving explicitly mapping only a subset of namespaces (--mapped-namespaces) alleviates the problem but introduces instability, requiring frequent manual reconnections. It also reduces usability across teams, as every namespace change would require modifying scripts.

**To Reproduce**

Run telepresence connect on a Linux (Ubuntu) workstation with access to a large number of Kubernetes namespaces.

Observe errors such as:

telepresence connect: error: connector.Connect: unable to determine the traffic-manager namespace

client rate limiter Wait returned an error: context canceled

User Daemon: Not running, Root Daemon: Not running, Traffic Manager: Not connected

Observe that telepresence status shows the daemons are not staying up.

Checking logs or gather-logs may show truncated or unclear errors. Further inspection reveals the real issue: Argument list too long—likely caused by DBus or system limits being exceeded.

[telepresence_logs_2025-05-21T15:59:42+02:00.zip](https://github.com/user-attachments/files/20369513/telepresence_logs_2025-05-21T15.59.42%2B02.00.zip)

**Expected behavior**

Telepresence should be able to handle environments with many namespaces without exceeding system argument/message size limits. Ideally, this would be done by chunking requests internally or otherwise managing resource queries to prevent overwhelm

**Versions (please complete the following information):**
 telepresence version : v2.20.3
Operating System: Ubuntu 22.04, Ubuntu 24.04 (Linux-only issue; not reproducible on macOS)
Kubernetes environment: Bare metal cluster with kube 1.31.4

**Additional context**

I would be happy to provide a PR if the issue is accepted.

PR #3864, opened 2025-05-22
Fix: Argument-list-to-long when connecting to cluster with lots of namespaces

nahuel11500:
## Description

PR to fix this [issue](https://github.com/telepresenceio/telepresence/issues/3863). 

## Checklist

### DIFF
--- a/pkg/client/rootd/dbus/resolved.go
+++ b/pkg/client/rootd/dbus/resolved.go
@@ -80,6 +80,9 @@ func SetLinkDNS(c context.Context, networkIndex int, ips ...net.IP) error {
 	})
 }
 
+// Set at an arbitrary value.
+const maxDomainsPerBatch = 100
+
 func SetLinkDomains(c context.Context, networkIndex int, domains ...string) error {
 	return withDBus(c, func(conn *dbus.Conn) error {
 		dds := make([]resolvedDomain, 0, len(domains))
@@ -94,8 +97,20 @@ func SetLinkDomains(c context.Context, networkIndex int, domains ...string) erro
 			}
 			dds = append(dds, resolvedDomain{Name: domain, RoutingOnly: routing})
 		}
-		return conn.Object("org.freedesktop.resolve1", "/org/freedesktop/resolve1").CallWithContext(
-			c, "org.freedesktop.resolve1.Manager.SetLinkDomains", 0, int32(networkIndex), dds).Err
+		for i := 0; i < len(dds); i += maxDomainsPerBatch {
+			end := i + maxDomainsPerBatch
+			if end > len(dds) {
+				end = len(dds)
+			}
+			if err := conn.Object("org.freedesktop.resolve1", "/org/freedesktop/resolve1").CallWithContext(
+				c, "org.freedesktop.resolve1.Manager.SetLinkDomains", 0, int32(networkIndex), dds[i:end]).Err; err != nil {
+				return fmt.Errorf(
+					"SetLinkDomains failed for network index %d, batch %d-%d (count=%d): %w",
+					networkIndex, i, end, end-i, err,
+				)
+			}
+		}
+		return nil
 	})
 }
 
--- a/CHANGELOG.yml
+++ b/CHANGELOG.yml
@@ -82,6 +82,20 @@ items:
           nodeSelector (inside PodSpec objects) and NodeSelector (inside NodeAffinity, VolumeNodeAffinity and a
           bunch of other places). The schema was changed to use the correct type. The name `nodeSelector` is still
           used in the Helm chart so this change is backwards compatible.
+      - type: bugfix
+        title: Telepresence fails to connect with many namespaces due to DBus message size limit
+        body: >-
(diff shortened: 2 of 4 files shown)


pkg/client/rootd/dbus/resolved.go
```diff
@@ -80,9 +80,12 @@ func SetLinkDNS(c context.Context, networkIndex int, ips ...net.IP) error {
 	})
 }
 
+// Set at an arbitrary value.
+const maxDomainsPerBatch = 100
+
 func SetLinkDomains(c context.Context, networkIndex int, domains ...string) error {
 	return withDBus(c, func(conn *dbus.Conn) error {
-		dds := make([]resolvedDomain, 0, len(domains))
+		var dds []resolvedDomain
```
thallgren: Why should this slice not be preallocated to its expected size?

pkg/client/rootd/dbus/resolved.go
```diff
@@ -94,8 +97,17 @@ func SetLinkDomains(c context.Context, networkIndex int, domains ...string) erro
...
-		return conn.Object("org.freedesktop.resolve1", "/org/freedesktop/resolve1").CallWithContext(
-			c, "org.freedesktop.resolve1.Manager.SetLinkDomains", 0, int32(networkIndex), dds).Err
+		for i := 0; i < len(dds); i += maxDomainsPerBatch {
+			end := i + maxDomainsPerBatch
+			if end > len(dds) {
+				end = len(dds)
+			}
+			if err := conn.Object("org.freedesktop.resolve1", "/org/freedesktop/resolve1").CallWithContext(
+				c, "org.freedesktop.resolve1.Manager.SetLinkDomains", 0, int32(networkIndex), dds[i:end]).Err; err != nil {
+				return fmt.Errorf("batch %d-%d failed: %w", i, end, err)
```
thallgren: Could this error message be more informative?

thallgren (commented):
Looks good overall, but I have some questions.

pkg/client/rootd/dbus/resolved.go
nahuel11500: You're right, it's a mistake to change that, we can keep it like it was. Thanks for catching it

pkg/client/rootd/dbus/resolved.go
nahuel11500: Yes of course, I changed it, is it fine like that ? I added the networkIndex and count number in the batch.

nahuel11500: Thanks for the review

pkg/client/rootd/dbus/resolved.go
```diff
@@ -94,8 +97,20 @@ func SetLinkDomains(c context.Context, networkIndex int, domains ...string) erro
...
 		}
-		return conn.Object("org.freedesktop.resolve1", "/org/freedesktop/resolve1").CallWithContext(
-			c, "org.freedesktop.resolve1.Manager.SetLinkDomains", 0, int32(networkIndex), dds).Err
+		for i := 0; i < len(dds); i += maxDomainsPerBatch {
+			end := i + maxDomainsPerBatch
+			if end > len(dds) {
+				end = len(dds)
+			}
+			if err := conn.Object("org.freedesktop.resolve1", "/org/freedesktop/resolve1").CallWithContext(
+				c, "org.freedesktop.resolve1.Manager.SetLinkDomains", 0, int32(networkIndex), dds[i:end]).Err; err != nil {
```
thallgren, line 106: This implementation suggests that `SetLinkDomains` will append the domains to an existing set. I'm under the impression that it overwrites any previous setting, and the [documentation](https://www.freedesktop.org/software/systemd/man/latest/org.freedesktop.resolve1.html) does not mention anything about appending. If I'm right, then this implementation will just set the last batch and throw away any previous batches.

Do you have any evidence proving that an append happens here?

nahuel11500: You're right, SetLinkDomains use [this method](https://github.com/systemd/systemd/blob/main/src/resolve/resolved-link-bus.c#L308) which does overwrite the entire list of DNS search/routing domains. This implementation is flawed and my locals tests didn't catch it. Thanks you for pointing it out. 

Then, given this behavior, that likely means there's no straightforward solution to [telepresenceio/telepresence#3863](https://github.com/telepresenceio/telepresence/issues/3863). 

Thank for your review

thallgren: Sorry I didn't catch this when I reviewed the first time.

Do  you know if there's any way to determine what the maximum limit is? Or perhaps track that the setting didn't set all the necessary domains? Being able to somehow convey that info to the user would indeed be a nice improvement.

nahuel11500: Well, I can look to add error handling if the Dbus limit is hit. I don't know yet how I could implement that but let me some time to look into it, I'll try to come up with something.

thallgren:
Closing this as it didn't fix the problem.

Issue #3873, opened 2025-06-09
Telepresence DNS Fallback stripping CNAME information from DNS Records


Yuheng3107:
**Describe the bug**
TLDR: Telepresence fallback logic for DNS without systemd-resolved strips crucial information about CNAME
Without telepresence:
```
nslookup example.company.local
Server:         10.0.0.53
Address:        10.0.0.53#53

example.company.local canonical name = alias.company.local.
Name:   alias.company.local
Address: 10.10.10.10
```

After running telepresence connect:
```
nslookup example.company.local
Server:         10.0.0.53
Address:        10.0.0.53#53

Name:   example.company.local
Address: 10.10.10.10
```
**Expected behavior**

After running telepresence connect, when using the fallback logic, I want to get the record with the CNAME information preserved if it has a CNAME
```
nslookup example.company.local
Server:         10.0.0.53
Address:        10.0.0.53#53

example.company.local canonical name = alias.company.local.
Name:   alias.company.local
Address: 10.10.10.10
```
**Versions (please complete the following information):**
 - Output of `telepresence version` (preferably while telepresence is connected)
OSS Client : v2.22.4
OSS User Daemon : v2.22.4
OSS Traffic Manager: v2.22.4
 - Operating system of workstation running `telepresence` commands: RHEL
 - Kubernetes environment and Version [e.g. Minikube, bare metal, Google Kubernetes Engine:
Openshift
Client Version: 4.15.0
Kustomize Version: v5.0.4
Kubernetes Version: v1.28.15
**Additional context**
Applications that rely on cnames being returned to it will fail, for e.g kerberos.

thallgren:
Can you please provide a daemon.log that covers a lookup made after you've changed the loglevel to trace (e.g. `telepresence loglevel --local-only trace`).

Yuheng3107:
2025-06-10 01:34:16.6188 debug   daemon/session/dns/legacy/Server : ServeDNS 51839 A      example.company.local. (from pkg/client/rootd/dns/server.go:794)
2025-06-10 01:34:16.6189 debug   daemon/session/dns/legacy/Server : Cluster DNS included for single label name "example" (from pkg/client/rootd/dns/server.go:229)
2025-06-10 01:34:16.6190 debug   daemon/session/dns/legacy/Server : Lookup A "example." (from pkg/client/rootd/session.go:397)
[...]

PR #3877, opened 2025-06-13
fix: removed logic to drop search domains from suffix, which causes F…

Yuheng3107:
…QDN domains that consist of multiple labels to have their suffix dropped

## Description

To fix Issue #3873:  Telepresence DNS Fallback stripping CNAME information from DNS Records
 The logic that was dropping search domains from the suffix has been removed. This prevents FQDN DNS queries from being forwarded to the cluster for resolution, which fixes the bug where only an A record is returned despite the original nameservers having both CNAME and A records for the entry

## Checklist

 - [X] I made sure to update `./CHANGELOG.yml` (our release notes are generated from this file).
 - [X] I made sure to add any documentation changes required for my change.
 - [X] My change is adequately tested.
 - [X] I updated `CONTRIBUTING.md` with any special dev tricks I had to use to work on this code efficiently.


thallgren (approved):

thallgren:
Big thanks!

Issue #3883, opened 2025-06-16
Enable running root daemon as a service


Yuheng3107:
**Describe the bug**
```
telepresence daemon-foreground -h
...
Usage:
  telepresence daemon-foreground <logging dir> <config dir> [flags]
```
Inside the telepresence daemon-foreground commands, it states that it only requires two arguments. However, if you pass it the two arguments it requires, we will get telepresence daemon-foreground: error: accepts 3 arg(s), received 2 

**Expected behavior**
Clearer documentation regarding telepresence daemon-foreground command. When making any extra changes, it would be good if they could be optional rather than required parameters to avoid having breaking changes.
For instance, the third arg which is the rootDaemonPath should have a default value of /var/run/telepresence-daemon.socket that can be modified optionally

**Versions (please complete the following information):**
 - Output of `telepresence version` (preferably while telepresence is connected) v2.23.0-rc.0
 - Operating system of workstation running `telepresence` commands RHEL
 - Kubernetes environment and Version [e.g. Minikube, bare metal, Google Kubernetes Engine]
Client Version: 4.15.0
Kustomize Version: v5.0.4
Kubernetes Version: v1.28.15
**Additional context**
I use telepresence daemon-foreground to run telepresence root daemon in the background so that users can use telepresence without requiring root perms. It would be better for this API to be stable, and if not, it should have clearer documentation on what it expects.

thallgren:
Values under `/var/run/` won't work under Windows. And using `C:\Users\Administator\...` won't work either, because the administrator may have a different name (often has, actually since many systems only have one single user, which also is the administrator).

The third argument was added as a solution to #3875. Not sure I see any reason why it should be made configurable more than it already is (by changing the users cache dir).

thallgren:
The telepresence daemon is generally started by the telepresence CLI, and we don't allow the versions of the daemon and the CLI to diverge, so using optional arguments here wouldn't help much. Your use-case, where you start the daemon manually as root, is not really a supported one. In order to support it fully, we should extend the CLI so that:

1. Somehow make the CLI aware that the root daemon is managed externally so that:
  a. `telepresence connect` errors with a sensible message when the root daemon isn't running.
  b. `telepresence quit -s` never kills the root daemon.

thallgren:
Relates to #2958

Yuheng3107:
[...]

Issue #2958, opened 2022-12-29
Telepresence systemd service(s)


spnngl:
**Please describe your use case / problem.**
We'd like to use `telepresence connect` as any others VPN/networking tools with systemd to monitor/reboot it and avoid launching root-level daemon by hand.
Possibility to send logs/errors to stdout/stderr to have them in systemd-journald.

**Describe the solution you'd like**
One systemd service/template to connect to one or multiple cluster.

**Describe alternatives you've considered**
I've made some services, one root & one user, to do it. They are basic and needs improvements/suggestions.

Root daemon:
```
# /etc/systemd/system/telepresence.service
[Unit]
Description=Telepresence daemon
Documentation=https://www.telepresence.io/docs/latest/quick-start/
After=network-online.target multi-user.target
Wants=network-online.target

[Service]
# Ensure we're clean and no old socket is used
ExecStartPre=rm -f /var/run/telepresence-daemon.socket
StandardInput=null
ExecStart=/usr/local/bin/telepresence daemon-foreground %L/telepresence %T
ExecStop=/usr/local/bin/telepresence quit -s --no-report
LockPersonality=yes
MemoryDenyWriteExecute=yes
NoNewPrivileges=yes
ProtectProc=invisible
ProtectClock=yes
DeviceAllow=/dev/net/tun
Environment=XDG_CACHE_HOME=%T
Environment=XDG_CONFIG_HOME=%T
ProtectControlGroups=yes
ProtectHome=yes
ProtectKernelLogs=yes
ProtectKernelModules=yes
ProtectSystem=full
Restart=always
RestartSec=0
RestrictNamespaces=yes
RestrictRealtime=yes
RestrictSUIDSGID=yes

[Install]
WantedBy=multi-user.target
```

User daemon:
```
# $HOME/.config/systemd/user/telepresence-connect@.service
[Unit]
Description=Telepresence connect service

[Service]
Type=forking
ExecStart=/usr/local/bin/telepresence connect --cache-dir=%T/telepresence --context=%i --request-timeout=2m --no-report
ExecStop=/usr/local/bin/telepresence quit --stop-daemons --no-report
PrivateTmp=yes
MemoryDenyWriteExecute=yes
NoNewPrivileges=yes
ProtectProc=invisible
ProtectControlGroups=yes
ProtectHome=read-only
ProtectSystem=full
RestrictNamespaces=yes
RestrictRealtime=yes
RestrictSUIDSGID=yes
```

[...]

PR #3889, opened 2025-06-19
feat(telepresence): optionally write logs to stdout/stderr if /dev/stdout …

Yuheng3107:
…or /dev/stderr is passed as the first argument to telepresence daemon-foreground

## Description

To add the option to write logs to stderr/stdout to get picked up by journald when running telepresence using systemd with telepresence daemon-foreground 
## Checklist


Yuheng3107:
@thallgren if you think this PR is fine I will add the relevant changelogs etc

pkg/client/rootd/service.go
```diff
@@ -404,9 +406,9 @@ func run(cmd *cobra.Command, args []string) error {
 	rootDaemonPath := args[2]
 	c := cmd.Context()
 
-	// Spoof the AppUserLogDir and AppUserConfigDir so that they return the original user's
-	// directories rather than directories for the root user.
-	c = filelocation.WithAppUserLogDir(c, loggingDir)
+	// Check if we should log to stdout or stderr (which will be captured by journald)
+	useJournald := loggingDir == "/dev/stdout" || loggingDir == "/dev/stderr"
```
Antvirf: This isn't really about "logging to journald", it is about logging to standard out/error which may or may not be picked up by journald later. Would rename this accordingly, otherwise it's a bit misleading

Yuheng3107:
#3883 #2958

pkg/client/rootd/service.go
```diff
@@ -427,9 +429,43 @@ func run(cmd *cobra.Command, args []string) error {
...
-	if err != nil {
-		return err
+
+	// If we're logging to stdout or stderr, set up logging to journald
+	if useJournald {
+		// Use stdout/stderr directly for logging
+		logger := logrus.StandardLogger()
+		logger.SetLevel(logrus.InfoLevel)
+		logger.ReportCaller = false
+		logger.Formatter = log.NewFormatter("2006-01-02 15:04:05.0000")
```
Antvirf: Logic related to logging would probably belong here https://github.com/telepresenceio/telepresence/blob/e21f4dc2ed4f5c198b50ecab4633e7a10189ba81/pkg/client/logging/initcontext.go#L28 in this file/package instead? Looking at the code there, it already seems to have an option to duplicate things to stdout/err

thallgren (commented):
The loggingDir is the directory where logfiles in general are written. Commands like `telepresence gather-logs` expects to find files there. Those logs are extremely useful in bug-reports. If we redirect logs to stdout/stderr, then we need to modify that command so that it uses journalctl or similar to collect the log. How do we do that on macOS or windows? Perhaps we don't? But if so, then the logic controlling this should be conditioned on platform.

The InitContext already has a flag to make it log to stdout. Would a default of using stderr be better? It doesn't feel ideal to use `loggingDir` as magic token to redirect the output. I would rather make the choice between stdout/stderr configurable, perhaps using an environment variable?

PR #3893, opened 2025-06-22
Update installation documentation links in release workflow

markrity:
Fix broken architecture link from quick-start page
Problem
The architecture link from the quick-start page breaks when accessed from URLs with trailing slashes, resulting in 404 errors.
Root Cause
Relative links behave differently depending on whether the current URL has a trailing slash - the browser resolves them relative to different base paths.
Solution
Update the relative link path to ensure consistent resolution behavior regardless of trailing slashes.
Impact
Fixes broken navigation from quick-start to architecture documentation
Ensures the link works reliably for all users
Improves overall documentation UX

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -126,7 +126,7 @@ jobs:
           tag: ${{ github.ref_name }}
           body: |
             ## Draft Release
-            For more information, visit our [installation docs](https://www.telepresence.io/docs/latest/quick-start/).
+            For more information, visit our [installation docs](https://www.telepresence.io/docs/latest/quick-start).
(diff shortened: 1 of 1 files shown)


markrity:
long story short , I was reading the documentation and seen that architecture hrefs doesn't work if you came from 
https://telepresence.io/docs/quick-start/ 

fixed at `telepresenceio`-homepage repo and `telepresence` main repo.

thallgren (approved):
Thanks! I guess no tests ever try those links in succession.

thallgren:
@markrity Can you please sign your commit so that the DCO check passes?

PR #3903, opened 2025-07-03
feat: add option to remove agents if they have been unused for longer than agents.maxIdleTime

Yuheng3107:
Signed-off-by: Kuang Yu Heng <yuheng3107@gmail.com>

feat: add logic to update config map to store latest time whenever an active engagement is created

feat: added logic in agent polling mechanism to remove agents if they have been unused for longer than AGENT_MAX_IDLE_TIME

feat: add default value for maxIdleTime to be basically infinite

add agent.maxIdleTime which can be configured for the traffic-manager to remove agents who did not have any active engagements to be removed to reduce overhead

## Checklist


charts/telepresence-oss/README.md
```diff
@@ -30,6 +30,7 @@ The following tables lists the configurable parameters of the Telepresence chart
...
 | agent.initResources                                  | The resources for the injected init container                                                                                                                       |                                                                             |
 | agent.logLevel                                       | The logging level for the traffic-agent                                                                                                                             | defaults to logLevel                                                        |
+| agent.maxIdleTime                                    | maximum time the agent is idle (no engagements to the workload) before it is cleaned up by the traffic manager                                                      |                                                                             |
```
thallgren: Add "0h = infinite" as the default.

cmd/traffic/cmd/manager/service.go
```diff
@@ -268,10 +272,98 @@ func (s *service) Remain(ctx context.Context, req *rpc.RemainRequest) (*empty.Em
...
-
+	err := s.UpdateLastEngagementTime(ctx, sessionID)
 	// There's no reason for the caller to wait for this removal to complete.
 	go s.state.RemoveSession(context.WithoutCancel(ctx), sessionID)
-	return &empty.Empty{}, nil
+	return &empty.Empty{}, err
+}
+
+func (s *service) removeUnusedAgent(ctx context.Context, sessionID tunnel.SessionID) error {
+	maxIdleTime := managerutil.GetEnv(ctx).AgentMaxIdleTime
```
thallgren, line 307: I think this method should begin with `agentKey, err := s.getAgentKey(sessionID)` and simply return `nil` when there is no agent (i.e. the `sessionID` is a client session). No reason to go through the motion of unmarshaling the yaml first.

cmd/traffic/cmd/manager/service.go
```diff
@@ -268,10 +272,98 @@ func (s *service) Remain(ctx context.Context, req *rpc.RemainRequest) (*empty.Em
...
+	}
+
+	lastEngagementTime := agentStates[agentKey].LastEngagementTime
+	idleTime := time.Since(lastEngagementTime)
+	if idleTime > maxIdleTime {
+		s.state.RemoveAgentSession(ctx, sessionID)
+	}
+	return nil
+}
+func (s *service) UpdateLastEngagementTime(ctx context.Context, sessionID tunnel.SessionID) error {
```
thallgren: This method must also be called when an engagement starts, and then periodically during the time when the engagement is active, or you risk that an active engagement suddenly gets removed.

cmd/traffic/cmd/manager/service.go
```diff
@@ -268,10 +272,98 @@ func (s *service) Remain(ctx context.Context, req *rpc.RemainRequest) (*empty.Em
...
+	namespace := managerutil.GetEnv(ctx).ManagerNamespace
+
+	client := k8sapi.GetK8sInterface(ctx).CoreV1()
+	configMap, err := client.ConfigMaps(namespace).Get(ctx, agentconfig.ManagerAppName, metav1.GetOptions{})
+	if err != nil {
+		return fmt.Errorf("Error in getting traffic-manager configmap, err: %w", err)
```
thallgren: Instead of "Error in", which starts with a capital letter, use "failure when". See [Errors](https://go.dev/wiki/Errors) in the Go Wiki.

cmd/traffic/cmd/manager/service.go
```diff
@@ -268,10 +272,98 @@ func (s *service) Remain(ctx context.Context, req *rpc.RemainRequest) (*empty.Em
...
+	}
+
+	agentStateYAML, ok := configMap.Data[config.AgentStateFileName]
+	if !ok {
+		return fmt.Errorf("Agent data not found in traffic-manager ConfigMap")
```
thallgren:
```suggestion
		return fmt.Errorf("%s not found in traffic-manager ConfigMap", config.AgentStateFileName)
```

cmd/traffic/cmd/manager/service.go
```diff
@@ -268,10 +272,98 @@ func (s *service) Remain(ctx context.Context, req *rpc.RemainRequest) (*empty.Em
...
+	if agentStateYAML != "" {
+		err = yaml.Unmarshal([]byte(agentStateYAML), &agentStates)
+		if err != nil {
+			return fmt.Errorf("err in unmarshalling YAML: %w", err)
+		}
+	} else {
+		agentStates = make(map[string]AgentState)
+	}
+
+	agentKey, err := s.getAgentKey(sessionID)
```
thallgren: Method should do this before anything else. No use continuing if the agent isn't found.

cmd/traffic/cmd/manager/service.go
```diff
@@ -268,10 +272,98 @@ func (s *service) Remain(ctx context.Context, req *rpc.RemainRequest) (*empty.Em
...
+	agentStateYAML, ok := configMap.Data[config.AgentStateFileName]
+	if !ok {
+		return fmt.Errorf("Agent data not found in traffic-manager ConfigMap")
+	}
+
+	var agentStates map[string]AgentState
+	if agentStateYAML != "" {
+		err = yaml.Unmarshal([]byte(agentStateYAML), &agentStates)
+		if err != nil {
+			return fmt.Errorf("err in unmarshalling YAML: %w", err)
```
thallgren:
```suggestion
			return fmt.Errorf("error unmarshalling agent states: %w", err)
```
(for consistency with "error marshalling ..." further down)

cmd/traffic/cmd/manager/service.go
```diff
@@ -268,10 +272,98 @@ func (s *service) Remain(ctx context.Context, req *rpc.RemainRequest) (*empty.Em
...
+		return fmt.Errorf("error updating ConfigMap: %w", err)
+	}
+
+	return nil
+}
+
+func (s *service) getAgentKey(sessionID tunnel.SessionID) (string, error) {
+	agent := s.state.GetAgent(sessionID)
+	if agent == nil {
+		return "", fmt.Errorf("Cannot find agent")
```
thallgren: Given that this isn't an error, perhaps the return value should be boolean?

thallgren (changes requested):

thallgren:
Please don't forget to sign your commits.

cmd/traffic/cmd/manager/service.go
Yuheng3107: thanks, will modify

cmd/traffic/cmd/manager/service.go
Yuheng3107: would a check that no engagement is active in the removal step inside the Remain function work

cmd/traffic/cmd/manager/service.go
Yuheng3107: fair, I changed it to return a string, empty string means not found

Yuheng3107:
> Please don't forget to sign your commits.

got it, was planning to merge and tidy the commits in the end, thanks!

cmd/traffic/cmd/manager/service.go
thallgren: Yes, that'll work and it's much simpler than what I suggested. If the engagement isn't active, and hasn't been active for some time, then it's OK to remove it. I don't see any flaws in that logic.

cmd/traffic/cmd/manager/service.go
```diff
@@ -260,18 +263,118 @@ func (s *service) Remain(ctx context.Context, req *rpc.RemainRequest) (*empty.Em
...
+	configMap, err := client.ConfigMaps(namespace).Get(ctx, agentconfig.ManagerAppName, metav1.GetOptions{})
+	if err != nil {
+		return fmt.Errorf("failure when getting traffic-manager configmap, err: %w", err)
+	}
+
+	agentStateYAML, ok := configMap.Data[config.AgentStateFileName]
+	if !ok {
+		return fmt.Errorf("%s not found in traffic-manager ConfigMap", config.AgentStateFileName)
+	}
+
```
thallgren: This shouldn't be needed, given that the traffic-manager maintains an in-memory state of the agentStateYAML using a watcher. So please just use `agentStateYAML := s.configWatcher.GetAgentStateYaml(ctx)` here too. It's important to limit the number of API calls to the k8s API.

Yuheng3107: you are right, am assuming I will PATCH the configmap since I don't have the full configmap

Yuheng3107: fixed it, will test the full functionality sometime later and inform you on Slack, thanks for reviewing

Yuheng3107:
functionality still doesn't work, am in the midst of debugging the issue

Yuheng3107:
Not sure why configmap is not being updated properly, will fix when im back from my trip

also need to decide where to add the update latest time functionality, doesn't seem to be one function that is called after telepresence leave runs

charts/telepresence-oss/README.md
```diff
@@ -30,6 +30,7 @@ The following tables lists the configurable parameters of the Telepresence chart
...
 | agent.initResources                                  | The resources for the injected init container                                                                                                                       |                                                                             |
 | agent.logLevel                                       | The logging level for the traffic-agent                                                                                                                             | defaults to logLevel                                                        |
+| agent.maxIdleTime                                    | maximum time the agent is idle (no engagements to the workload) before it is cleaned up by the traffic manager                                                      |                                                                             |                                                                                   "0h"
```
thallgren: I think there's a point in adding "0h (infinite)" or "0h (never cleaned up)" so that the reader understands that the default isn't cleaning it up immediately.

thallgren (approved):
This looks good now. Just a minor nit about the doc. Also, there are several unsigned commits in the PR. Perhaps you can squash all of them together into one that is signed?

build-aux/docker/images/Dockerfile.client
```diff
@@ -48,4 +48,4 @@ RUN apk add --no-cache ca-certificates iptables
 COPY --from=telepresence-build /usr/local/bin/telepresence /usr/local/bin
 
 ENTRYPOINT ["telepresence"]
-CMD []
+CMD []
```
thallgren: Please refrain from removing the last newline here.

build-aux/docker/images/Dockerfile.traffic
```diff
@@ -55,4 +55,4 @@ RUN \
   chmod 0777 /home/telepresence
 
 ENTRYPOINT ["traffic"]
-CMD []
+CMD []
```
thallgren: Please refrain from removing the last newline here.

build-aux/docker/images/Dockerfile.client
Yuheng3107: fixed

Yuheng3107:
@thallgren  currently the latestEngagementTime is only recorded upon agent arrival currently, can't find a good way to record it when the intercept is left as that is client side

CHANGELOG.yml
```diff
@@ -24,34 +24,16 @@
...
-        title: Requests lost when using wiretap
+      - type: feature
+        title: add ability to optionally clean up sidecars that have been idle above a specified duration
         body: >-
-          Wiretap connection `Close` and `Write` could sometimes be out of sync, so that the `Close` would be executed
-          before the `Write`, causing a "read/write on closed pipe" error and loss of data.
+          Added configuration parameter `agent.maxIdleTime` to the Helm Chart, to control how long a sidecar can be idle before it is cleaned up.
+          The updating of latestEngagementTime is done every 1m in the Remain Call, which is now called every 1min, compared to every 5s in the past 
+          Removal of the sidecar is also done in the Remain Call, if the sidecar has been idle for longer than the configured `agent.maxIdleTime`.         
+
```
thallgren: This change doesn't look quite right. The error is probably caused by the recent update of the release/v2 branch.

cmd/traffic/cmd/manager/service.go
```diff
@@ -260,20 +264,159 @@ func (s *service) Remain(ctx context.Context, req *rpc.RemainRequest) (*empty.Em
 	if ok := s.state.MarkSession(req, s.clock.Now()); !ok {
 		return nil, status.Errorf(codes.NotFound, "Session %q not found", sessionID)
 	}
+
+	err := s.removeUnusedAgent(ctx, sessionID)
```
thallgren: Don't call this function unless the caller of `Remain` is an agent, i.e. move it down to line 277.

cmd/traffic/cmd/manager/service.go
```diff
@@ -260,20 +264,159 @@ func (s *service) Remain(ctx context.Context, req *rpc.RemainRequest) (*empty.Em
...
 	s.state.RefreshSessionConsumptionMetrics(sessionID)
-	return &empty.Empty{}, nil
+
+	agent := s.state.GetAgent(sessionID)
+	if agent == nil {
+		dlog.Debugf(ctx, "No agent found for session %s", sessionID)
```
thallgren: This will cause the message to tick every time a client calls `Remain`. It will be confusing for our users so I'd rather you just remove it.

cmd/traffic/cmd/manager/service.go
```diff
@@ -260,20 +264,159 @@ func (s *service) Remain(ctx context.Context, req *rpc.RemainRequest) (*empty.Em
...
 	ctx = managerutil.WithSessionInfo(ctx, session)
-	sessionID := tunnel.SessionID(session.GetSessionId())
 
+	sessionID := tunnel.SessionID(session.GetSessionId())
 	// There's no reason for the caller to wait for this removal to complete.
 	go s.state.RemoveSession(context.WithoutCancel(ctx), sessionID)
 	return &empty.Empty{}, nil
 }
 
+func (s *service) removeUnusedAgent(ctx context.Context, sessionID tunnel.SessionID) error {
```
thallgren: I suggest the following change:

1. Export the `workloadKey` struct in the mutator package.
2. Let `Remain` create the such a key when it's called by an agent.
3. Use this key instead of a sessionID as an argument to `removeUnusedAgent`.
4. Use this key as an argument to `CountActiveIntereptsForAgent().
5. Add a `mutator.WorkloadKey.String()`. The result is used as a key in the `agentState` map.
6. Remove the `getAgentKey` function altogether (no longer needed).

cmd/traffic/cmd/manager/service.go
```diff
@@ -260,20 +264,159 @@ func (s *service) Remain(ctx context.Context, req *rpc.RemainRequest) (*empty.Em
...
+	agentKey := s.getAgentKey(ctx, sessionID)
+	if agentKey == "" {
+		return nil
+	}
+
+	agent := s.state.GetAgent(sessionID)
+	if agent == nil {
+		return nil
+	}
+
```
thallgren: Remove this. There's no need for an `agent` here.

thallgren (changes requested):

CHANGELOG.yml
Yuheng3107: added the missing version 2.23.3 update back

cmd/traffic/cmd/manager/service.go
Yuheng3107: the agent key was taken from the agent itself, so shouldn't it be a method of the agent (for standardisation of the name, I don't understand the rationale of why we want to include workloadKey, exporting its fields would result in a ripple effect to other parts of the code that use workloadKey)

I agree with your overall idea of refactoring though, I think I will modify the state.AgentSession struct instead to have a method for the agentKey

Yuheng3107: have added what I think makes a bit more sense in the latest PR, thanks for the review, I think after refactoring its much more cohesive

Yuheng3107:
tested and it works, but the granularity is in minutes now

cmd/traffic/cmd/manager/state/state.go
```diff
@@ -852,3 +854,12 @@ func (s *state) allInterceptsFinalizerCall(client *ClientSession, workload *stri
 		s.allInterceptsFinalizer(client, workload)
 	}
 }
+
+func (s *state) CountActiveInterceptsForAgent(agent *AgentSession) int {
```
thallgren: This function is now misleading. The `AgentSession` denotes the agent of a _specific pod_, but the count is for _all agents_, i.e. all pods for the workload denoted by the spec (this is partly why I suggest using the `WorkloadKey`).

The whole concept of removing agents must be based on workloads, not agent sessions. A workload with a replicaCount > 1 will have several agent sessions. So please don't base the logic on agent sessions. Use a workload key.

thallgren (changes requested):
The agent session is useful when determining the workload key in the `Remain` function, but shouldn't otherwise be used.

cmd/traffic/cmd/manager/state/state.go
Yuheng3107: I understand why we need to use the WorkloadKey now, sorry for the misunderstanding, will get it fixed first thing tomorrow

cmd/traffic/cmd/manager/service.go
```diff
@@ -1028,7 +1147,7 @@ func (s *service) WatchWorkloads(request *rpc.WorkloadEventsRequest, stream rpc.
 	return ww.Watch(ctx, stream)
 }
 
-const agentSessionTTL = 15 * time.Second
+const agentSessionTTL = 1 * time.Minute
```
thallgren: This duration must be slightly longer than the poll time used by the traffic-agent so that we don't risk that it's being garbage collected due to a race.
```suggestion
const agentSessionTTL = 70 * time.Second
 ```

**Comment by Yuheng3107** [comment]
Have updated the code to use workloadKey for better semantics, even though Remain works on an agent level, the agent removal logic is on a workload level (i.e if any replicas have an active engagement, the whole workload cannot be removed)

I think change is more semantic rather than functional, do correct me if I am wrong

**Comment by Yuheng3107** [comment]
also, would like to address some issues with this PR:

The first being the potential undefined state of lastEngagementTime, I feel that lastEngagementTime should be updated in ArriveAsAgent as a agent is only created when an engagement happens. We want to prevent the case where an agent is killed/intercept is interrupted using CtrlC and the sidecar is spawned with no lastEngagementTime, as lastEngagementTime is currently only updated when there are active engagements. If it were uninitialised, the logic to remove it is skipped, which could result in sidecars that have no active engagements, no latestEngagementTime, and hence not being removed. Thus I believe we need to initialise the lastEngagementTime inside the ArriveAsAgent function when the sidecar agents are created.

The second concern I have is regarding the frequency of api calls to the k8s server. With the current implementation, there is at least 1 Patch update to the configmap per workload, even more (unnecessary calls) if there are mutliple agents in the same workload making the Patch Update in succession (no point updating the configmap multiple times in succession, timestamp will be very similar).

The proposed change I have to fix this would involve updating the in memory representation of the configmap rather than the configmap itself, and having a separate goroutine that runs every 2mins to update the state of the configmap with the in memory representation for persistence.

I will add them as separate commits and revert them if you think they are redundant.

File: cmd/traffic/cmd/manager/service.go [hunk 20]
```diff
@@ -260,20 +264,138 @@ func (s *service) Remain(ctx context.Context, req *rpc.RemainRequest) (*empty.Em
...
+		return &empty.Empty{}, nil
+	}
+
+	workloadKey := &mutator.WorkloadKey{
+		Name:      agent.Name,
+		Namespace: agent.Namespace,
+		Kind:      k8sapi.Kind(agent.Kind),
+	}
+
+	err := s.removeUnusedAgent(ctx, workloadKey)
```
Antvirf: Missing error check? Also perhaps `UpdateLastEngagement` *first* and then run the `removeUsedAgent`

cmd/traffic/cmd/manager/service.go
```diff
@@ -260,20 +264,138 @@ func (s *service) Remain(ctx context.Context, req *rpc.RemainRequest) (*empty.Em
...
+	}
+
+	agentStateFileYAML := s.configWatcher.GetAgentStateYaml(ctx)
+	var agentStateFile AgentStateFile
+	err := yaml.Unmarshal(agentStateFileYAML, &agentStateFile)
+	if err != nil {
+		return fmt.Errorf("err in unmarshalling YAML: %w", err)
+	}
+
+	lastEngagementTime := agentStateFile.AgentStates[*workloadKey].LastEngagementTime
```
Antvirf: Worth doing the map lookup safely in case the state has any issue?

cmd/traffic/cmd/manager/service.go
```diff
@@ -237,6 +242,15 @@ func (s *service) ArriveAsAgent(ctx context.Context, agent *rpc.AgentInfo) (*rpc
 		return nil, err
 	}
 
+	err = s.UpdateLastEngagementTime(ctx, &mutator.WorkloadKey{
```
thallgren: I think it's enough to just call this from `Remain`. The first call to `Remain` will follow shortly after the `ArriveAsAgent`.

cmd/traffic/cmd/manager/manager.go
```diff
@@ -313,3 +315,19 @@ func (s *service) runSessionGCLoop(ctx context.Context) error {
...
+
+func (s *service) runUpdateTrafficManagerConfigMapLoop(ctx context.Context) error {
+	// Loop updating the agentState every 2mins
+	ticker := time.NewTicker(2 * time.Minute)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ticker.C:
+			err := s.updateTrafficManagerConfigMap(ctx)
```
thallgren: There should be a flag to avoid unnecessary updates. You could for instance add a

```go
  tmConfigMapUpdated atomic.Bool
```
 to the `service` struct

thallgren (commented):
This is looking good now. Just a couple of minor things.

cmd/traffic/cmd/manager/service.go
Yuheng3107: but there is an edge case since the time only gets updated when there are active intercepts. It could well be the case that a sidecar is created but the intercept was cancelled before Remain was called. Then the sidecar will remain there indefinitely and not be cleaned up

cmd/traffic/cmd/manager/manager.go
Yuheng3107: thanks for the suggestion! have added it in the latest update, in the midst of testing now

Yuheng3107:
when everything is finalized I will modify the changelog and squash everything into one commit

Yuheng3107:
@thallgren  sorry for the long wait, have verified that all the functionality works as intended, if all is ok I will squash the commits and update the CHANGELOG

thallgren (approved):
Looks great now.

cmd/traffic/cmd/manager/service.go
thallgren: `Remain` is called by the agent regardless of if it is currently intercepting or not. If the agent exists, it will call `Remain`.

thallgren: So, please just remove this call

Yuheng3107: yes, remain will be called, but in the Remain call, the UpdateLastEngagementTime function will NOT update the engagement time if there are no active intercepts, and there is a chance that the lastEngagementTime is uninitialised, which would leave the agent hanging indefinitely if it never receieves any engagements.

In the ArriveAsAgent call, I skipped the check for active intercepts (as I assume ArriveAsAgent indicates there is an engagement created), which GUARANTEES that there will be a lastEngagementTime and the sidecar will not be left hanging

cmd/traffic/cmd/manager/service.go
```diff
@@ -260,20 +276,162 @@ func (s *service) Remain(ctx context.Context, req *rpc.RemainRequest) (*empty.Em
...
+		Name:      agent.Name,
+		Namespace: agent.Namespace,
+		Kind:      k8sapi.Kind(agent.Kind),
+	}
+
+	err := s.removeUnusedAgent(ctx, workloadKey)
+	if err != nil {
+		dlog.Errorf(ctx, "error removing unused agent: %v", err)
+	}
+	if s.state.CountActiveInterceptsForWorkload(workloadKey) > 0 {
```
Yuheng3107: without UpdateLastEngagementTime being called in ArriveAsAgent, in the case where an agent is created and the intercept is cancelled before Remain is called, subsequent Remain calls will NOT trigger UpdateLastEngagementTime. The IsZero() check for the removeUnusedAgent function will mean that the sidecar is NEVER removed

thallgren: It shouldn't matter if the intercept is cancelled before `Remain` is called. The `Remain` call establishes the fact that the agent exists. If it does, then it's engagement time is updated accordingly. If there is no timestamp present, then obviously the agent was just created, so just set it to the current time.

Yuheng3107:
@thallgren updated, hope it is satisfactory

Yuheng3107:
the other commits have serious merge conflicts if I squash them, can I just leave them, otherwise it will take me a long while to get it resolved

cmd/traffic/cmd/manager/service.go
```diff
@@ -260,20 +267,165 @@ func (s *service) Remain(ctx context.Context, req *rpc.RemainRequest) (*empty.Em
...
+	dlog.Tracef(ctx, "Logging workloadKey for last engagement time: %s", *workloadKey)
+
+	agentStateFileYAML := s.configWatcher.GetAgentStateYaml(ctx)
+	dlog.Tracef(ctx, "Logging agentStateFileYAML: %s", agentStateFileYAML)
+
+	var agentStateFile AgentStateFile
+	if string(agentStateFileYAML) != "" {
+		if s.state.CountActiveInterceptsForWorkload(workloadKey) == 0 {
+			// don't update last engagement time if there are no active intercepts and it is not the first time
+			return nil
```
thallgren: I think you need to check if the time for the workloadKey is the zero time and, if so, don't return here. How would you otherwise know that this "it is not the first time"? The `agentStateFileYAML` may contain other workloads.

thallgren (changes requested):

thallgren:
You can just leave the commits.

thallgren (approved):

thallgren:
Thanks again for being so patient, and sorry for being such a nit-picker. The main reason why I wanted the `Remain` to be the only function that updates, is to avoid any weird issues caused by race-conditions between the gRPC calls, no matter what changes we might do to the traffic-agent in the future.

Yuheng3107:
> Thanks again for being so patient, and sorry for being such a nit-picker. The main reason why I wanted the `Remain` to be the only function that updates, is to avoid any weird issues caused by race-conditions between the gRPC calls, no matter what changes we might do to the traffic-agent in the future.

got it, that makes sense, I feel that the code is neater and more cohesive with the change.
Thanks for the guidance and support in reviewing my PRs too!

PR #3945, opened 2025-08-20
fix [charts]: remove extra whitespace in charts/configmap

s-z-z:
## Description

remove extra whitespace in charts/configmap


thallgren (approved):
Thanks.

PR #3953, opened 2025-08-29
fix(Helm): Add Removed Value in config

Syndlex:
## Description

This was removed in github.com/telepresenceio/telepresence/commit/54cfaa5306c75f266b83dfaf81ed04705b1c3f4c

This makes Problems when using this Helm chart from Argo CD.
Because it tries to sync a empty string and k8s deletes the empty string.


thallgren:
Ping @Yuheng3107 . Can you please review this? I'm uncertain as to why the removal was made.

Yuheng3107:
LGTM

I think I shouldn’t have removed it and rather leave it defined explicitly as your change makes it

thallgren (approved):

thallgren:
Thanks!

Issue #3852, opened 2025-04-25
Add http-header option to telepresence Open Source


Versa78:
**Please describe your use case / problem.**
I work in a medium sized company with dussins of developers and many microservices. If we are to use telepresence to intercept these services in our shared kubernetes environments it is essential to us that we are able to do personal intercepts using different http headers. Without it we'd intercept too much traffic when we usually are interested in specific http calls for troubleshooting.

**Describe the solution you'd like**
I would like to be able to intercept only calls with specific http headers. As a bonus I would also like to be able to specify http request paths.

**Describe alternatives you've considered**
The commercial version

**Additional context**
We have been forced into the commercial product because of this single feature. Now the commercial option is changing and they are forcing an entirely different product upon us that we do not want or have any use for. 

I feel this is the last option that will allow us to use telepresence as we've decided not to continue using the commercial version as they are forcing us into a new product.

thallgren:
Hi @Versa78 . I understand your predicament. So far, we've been careful not introducing new features like "personal intercepts" into the OSS product that would be in direct competition with the commercial product. Ambassador's current approach of bundling their commercial Telepresence into Blackbird might change how we view this. Especially since Ambassador no longer contributes anything to the OSS project.

Meanwhile, I'd like to tell you what we have done already to make your life easier without personal intercepts.
1. It is possible to have any number of `ingests` on one workload at the same time as an intercept is active.
2. An improvement that might help is that multiple simultaneous intercepts are possible, provided that they use different ports.
3. The next minor release will contain a new `telepresence wiretap` command. Very similar to an intercept, but it mirrors the traffic instead of intercepting it, and thus, introduces no conflict if several developers wiretap the same workload at the same time.

Not quite http-header based intercepts, but you do get the ability for several developers to debug the same code simultaneously. And you don't need to worry about TLS certificates or header propagation.

How does that sound? Are these valuable improvements that will help your company?

Versa78:
Hi Thomas and thanks for the quick reply!

If I understand you correctly, wiretap allows us to mirror traffic without intercepting it, which could ease the pain in some situations. It would very much depend on what needs to be intercepted, in some situations where we need to intercept a specific api call to a workload that handles many calls every second it would become impractical trying to debug traffic locally as there would just be to many calls too single out the one you're interested in. But with less used workloads it could work better.

[...]

PR #3967, opened 2025-09-24
feat: HTTP Intercepts with HTTP header and path filtering

bgruszka:
This PR resolves the issue: https://github.com/telepresenceio/telepresence/issues/3852

## Description

This PR implements HTTP Intercepts, a new feature that enables fine-grained HTTP traffic filtering for Telepresence intercepts. Users can now intercept only specific HTTP requests based on headers
  and URL paths with explicit matching semantics, rather than intercepting all traffic to a service.

  Key Features

  - HTTP-aware interception with automatic mechanism detection
  - Header-based filtering with --http-header key=value (supports both = and : formats, AND logic for multiple headers)
  - Enhanced path-based filtering with explicit match types:
    - --http-path-equal - Exact path matching (O(1) performance)
    - --http-path-prefix - Path prefix matching (O(k) performance)
    - --http-path-regex - Regular expression matching (full regex support)
  - Backward compatibility with existing TCP intercepts
  - Available in both intercept and wiretap commands

  Architecture Changes

  New HTTP Interceptor: Complete HTTP-aware forwarding mechanism in pkg/forwarder/http.go
  Enhanced CLI: Added HTTP filtering flags with explicit match semantics for precise path matching
  Protocol Extensions: Extended gRPC protocol to support typed HTTP filtering parameters
  Mechanism Dispatch: Implemented DispatchByMechanism pattern for protocol-specific handling

  Implementation Details

  Core Components Added:
  - pkg/forwarder/http.go - HTTP interceptor with advanced filtering logic supporting all match types
  - pkg/matcher/request.go - Request matching infrastructure with :path-equal:, :path-prefix:, :path-regex: semantics

  CLI Enhancements:
  - pkg/client/cli/intercept/command.go - Added explicit HTTP path filtering flags and header validation
  - pkg/client/cli/intercept/state.go - Request creation with typed filtering parameters
  - Updated help documentation with clear match type descriptions for telepresence intercept and telepresence wiretap

  Protocol & Infrastructure:
  - rpc/manager/manager.proto - Extended InterceptSpec with HTTP filtering fields
  - cmd/traffic/cmd/agent/agent.go - Registered HTTP mechanism support
  - cmd/traffic/cmd/manager/state/intercept.go - Updated mechanism validation

  Usage Examples

  #### Enable HTTP-aware intercepts with header filtering
  telepresence intercept my-service --http-header "X-User-ID=dev123" --http-header "X-Environment=staging" --port 8080

  Path-based filtering with explicit match types
  #### Exact path matching
  telepresence intercept my-service --http-path-equal "/health" --port 8080

  #### Prefix matching
  telepresence intercept my-service --http-path-prefix "/api/v1/" --port 8080

  #### Regex matching
  telepresence intercept my-service --http-path-regex "^/api/v[0-9]+/" --port 8080

  Combined header and path filtering
  telepresence intercept my-service --http-header "X-Debug=true" --http-path-prefix "/api/" --port 8080

  Header format flexibility (curl -H compatible)
  telepresence intercept my-service --http-header "X-User-ID=dev123" --http-header "Authorization: Bearer token" --port 8080

  Traditional intercepts (unchanged behavior)
  telepresence intercept my-service --port 8080

  Backward Compatibility

  - All existing intercept and wiretap commands continue to work unchanged
  - Default mechanism remains TCP for non-HTTP intercepts
  - No breaking changes to existing APIs or protocols
  - Automatic HTTP mechanism detection when HTTP flags are used

## Checklist

### DIFF
--- a/cmd/traffic/cmd/manager/state/intercept.go
+++ b/cmd/traffic/cmd/manager/state/intercept.go
@@ -477,7 +477,7 @@ func (s *State) ensureAgent(parentCtx context.Context, wl k8sapi.Workload, exten
 }
 
 func (s *State) isExtended(spec *rpc.InterceptSpec) bool {
-	return spec.Mechanism != "tcp"
+	return spec.Mechanism != "tcp" && spec.Mechanism != "http"
 }
 
 func (s *State) ValidateAgentImage(agentImage string, extended bool) (err error) {
--- a/pkg/client/cli/intercept/command.go
+++ b/pkg/client/cli/intercept/command.go
@@ -66,6 +66,43 @@ func (c *Command) UsesHTTPMechanism() bool {
 	return len(c.HTTPHeaderFilters) > 0 || len(c.HTTPPathFilters) > 0
 }
 
+// parseHTTPHeader parses an HTTP header string that can use either "=" or ":" as separator.
+// Supports both formats:
+//   - "X-User-ID=dev123" (equals format)
+//   - "X-User-ID: dev123" (colon format, compatible with curl -H)
+//
+// Returns the key and value, or an error if the format is invalid.
+// When both separators are present, colon takes precedence (standard HTTP format).
+func parseHTTPHeader(header string) (string, string, error) {
+	// Try colon separator first (standard HTTP format, curl -H compatible)
+	if strings.Contains(header, ":") {
+		parts := strings.SplitN(header, ":", 2)
+		if len(parts) == 2 {
+			key := strings.TrimSpace(parts[0])
+			value := strings.TrimSpace(parts[1])
+			if key == "" {
+				return "", "", fmt.Errorf("invalid header format '%s': key cannot be empty", header)
+			}
+			return key, value, nil
+		}
+	}
+
+	if strings.Contains(header, "=") {
+		parts := strings.SplitN(header, "=", 2)
+		if len(parts) == 2 {
+			key := strings.TrimSpace(parts[0])
+			value := strings.TrimSpace(parts[1])
+			if key == "" {
+				return "", "", fmt.Errorf("invalid header format '%s': key cannot be empty", header)
+			}
+			return key, value, nil
+		}
+	}
+
+	// Neither separator found
+	return "", "", fmt.Errorf("invalid header format '%s': must be key=value or key: value", header)
+}
+
 func (c *Command) AddInterceptFlags(cmd *cobra.Command) {
 	what := "intercept"
 	how := "intercepted"
@@ -118,7 +155,8 @@ func (c *Command) AddInterceptFlags(cmd *cobra.Command) {
 	// HTTP Intercepts flags
 	flagSet.StringSliceVar(&c.HTTPHeaderFilters, "http-header", nil,
 		`HTTP header filters for HTTP Intercepts. Only requests with matching headers will be intercepted. `+
-			`Format: --http-header "X-User-ID=dev123" --http-header "X-Environment=staging". Multiple headers use AND logic.`)
+			`Supports both formats: --http-header "X-User-ID=dev123" or --http-header "X-User-ID: dev123" (curl -H compatible). `+
+			`Multiple headers use AND logic.`)
 
 	flagSet.StringSliceVar(&c.HTTPPathFilters, "http-path", nil,
 		`HTTP path filter patterns for HTTP Intercepts. Only requests matching these paths will be intercepted. `+
@@ -167,16 +205,11 @@ func (c *Command) Validate(cmd *cobra.Command, positional []string) error {
 	c.Cmdline = positional[1:]
 	c.FormattedOutput = output.WantsFormatted(cmd)
 
-	// HTTP Intercepts: validate header format and auto-detect HTTP mechanism
+	// HTTP Intercepts: validate header format
 	if c.UsesHTTPMechanism() {
-		// Validate header format (key=value)
 		for _, header := range c.HTTPHeaderFilters {
-			if !strings.Contains(header, "=") {
-				return fmt.Errorf("invalid header format '%s': must be key=value", header)
-			}
-			parts := strings.SplitN(header, "=", 2)
-			if parts[0] == "" {
-				return fmt.Errorf("invalid header format '%s': key cannot be empty", header)
+			if _, _, err := parseHTTPHeader(header); err != nil {
+				return err
 			}
 		}
 
--- a/pkg/client/cli/intercept/command.go
+++ b/pkg/client/cli/intercept/command.go
@@ -56,9 +56,14 @@ type Command struct {
 	NoDefaultPort   bool
 
 	// HTTP Intercepts fields
-	HeaderFilters []string // --header key=value pairs for HTTP header filtering
-	PathFilters   []string // --path patterns for HTTP path filtering
-	HTTPMechanism bool     // --http flag to enable HTTP-aware interception
+	HTTPHeaderFilters []string // --http-header key=value pairs for HTTP header filtering
+	HTTPPathFilters   []string // --http-path patterns for HTTP path filtering
+}
+
+// UsesHTTPMechanism returns true if any HTTP-specific flags were provided,
+// indicating that HTTP-aware interception should be used.
+func (c *Command) UsesHTTPMechanism() bool {
+	return len(c.HTTPHeaderFilters) > 0 || len(c.HTTPPathFilters) > 0
 }
 
 func (c *Command) AddInterceptFlags(cmd *cobra.Command) {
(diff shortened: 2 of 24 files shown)


bgruszka:
There are a lot of fmt errors like that. I’m not sure if I should handle them, since they also appear in files I haven’t changed?
image

thallgren:
> There are a lot of fmt errors like that. I’m not sure if I should handle them, since they also appear in files I haven’t changed?

These errors happen when golangci-lint release a new version with new nice features in it :-). I'm addressing them in #3968 where the version is bumped.

A note about that PR.
It changes the forwarder.Interceptor slightly so that it uses a `netip.AddrPort` target. That will create some conflicts with your PR. Can I ask you to please rebase it once I've merged it and deal with those? I want those changes in place for some other work that I have in progress.

bgruszka:
Of course, I can rebase my PR after you merge yours and resolve any conflicts that result from the changes to forwarder.Interceptor. 🙂

cmd/traffic/cmd/manager/state/intercept.go
```diff
@@ -477,7 +477,7 @@ func (s *State) ensureAgent(parentCtx context.Context, wl k8sapi.Workload, exten
 }
 
 func (s *State) isExtended(spec *rpc.InterceptSpec) bool {
-	return spec.Mechanism != "tcp"
+	return spec.Mechanism != "tcp" && spec.Mechanism != "http"
 }
```
thallgren, line 481: We should probably remove the "extended" concept at some point (not in scope for this PR), because with the pro-version being cancelled, there's no need to extend anything.

pkg/client/cli/intercept/command.go
```diff
@@ -54,6 +54,11 @@ type Command struct {
 	FormattedOutput bool
 	DetailedOutput  bool
 	NoDefaultPort   bool
+
+	// HTTP Intercepts fields
+	HeaderFilters []string // --header key=value pairs for HTTP header filtering
+	PathFilters   []string // --path patterns for HTTP path filtering
+	HTTPMechanism bool     // --http flag to enable HTTP-aware interception
```
thallgren: The old way of doing this, was to use the `mechanism` and `mechanismArgs` so that the relevant flags were forwarded to the traffic-agent and parsed there. The motivation behind that approach was that different traffic-agents would be capable of handling different mechanisms. It's only ever been used for passing http flags to the agent. In my mind, the `--mechanism` flag should be removed now when we introduce http to the OSS agent. I have some additional thoughts:

- The `--http` flag seems to have the same effect as the old `--mechanism http`, and that flag was rarely used because it could always be inferred from other flags. Unless I'm missing something, I think the same is true here. We don't really need the `--http` flag.
- If we extend the objective for this PR to also make it easy for existing users of the proprietary version, then we should consider using flags like `http-header`, `http-path-equal`, `http-path-prefix`, and `http-path-regexp`. This would also make it very easy to check if http is used (just visit the flagset and check for a "http-" prefix among the names).

bgruszka:
@thallgren I've rebased my branch and resolved all the conflicts. Can you run integration tests and take a look at PR? :)

pkg/client/cli/intercept/command.go
```diff
@@ -105,6 +110,20 @@ func (c *Command) AddInterceptFlags(cmd *cobra.Command) {
 		flagSet.Lookup("replace").Deprecated = "Use the replace command."
 	}
 
+	// HTTP Intercepts flags
+	flagSet.StringSliceVar(&c.HeaderFilters, "header", nil,
```
thallgren: Maybe also permit colon as a separator? That way, the flag value can also be used for the curl `-H` flag.

pkg/client/cli/intercept/command.go
```diff
@@ -105,6 +110,20 @@ func (c *Command) AddInterceptFlags(cmd *cobra.Command) {
...
+		`HTTP header filters for HTTP Intercepts. Only requests with matching headers will be intercepted. `+
+			`Format: --header "X-User-ID=dev123" --header "X-Environment=staging". Multiple headers use AND logic. `+
+			`Requires --http flag.`)
+
+	flagSet.StringSliceVar(&c.PathFilters, "path", nil,
```
thallgren: Would be nice if this was extended to equal, prefix, and regexp matches, just like the pro-version has.

thallgren (commented):
This is great work. Well done!

I have some questions regarding flag names, but otherwise I think this is good to go. 
So very nice when tests are provided. Thanks!

bgruszka:
I totally forgot about list cmd and info displayed :D I've just added it.

cmd/traffic/cmd/manager/state/intercept.go
bgruszka: 👍

pkg/client/cli/intercept/command.go
bgruszka: I've introduced http-header, http-path-equal, http-path-prefix, and http-path-regexp :)

pkg/client/cli/intercept/command.go
bgruszka: Added a colon to support curl -H :)

pkg/client/cli/intercept/command.go
bgruszka: added :)

bgruszka:
@thallgren can you check PR again and run integration tests? :D

pkg/client/cli/intercept/command.go
```diff
@@ -54,6 +54,58 @@ type Command struct {
...
+		parts := strings.SplitN(header, ":", 2)
+		if len(parts) == 2 {
+			key := strings.TrimSpace(parts[0])
+			value := strings.TrimSpace(parts[1])
+			if key == "" {
+				return "", "", fmt.Errorf("invalid header format '%s': key cannot be empty", header)
+			}
+			return key, value, nil
+		}
+	}
```
thallgren, line 86: Nitpick. This part could be broken out to a function instead of being repeated with "=" and ":". Also, the strings.Contains is probably redundant given the SplitN call.

pkg/forwarder/udp.go
```diff
@@ -92,6 +92,15 @@ func (f *udp) forward(ctx context.Context, conn *net.UDPConn, intercept *manager
 	return f.forwardConn(ctx, conn)
 }
 
+// DispatchByMechanism implements the Interceptor hook for UDP. Currently, HTTP-aware
+// mechanisms are not supported on UDP; we log and proceed with normal UDP forwarding.
+func (f *udp) DispatchByMechanism(ctx context.Context, _ net.Conn, intercept *manager.InterceptInfo) (bool, error) {
+	if intercept != nil && intercept.Spec != nil && intercept.Spec.HttpMechanism {
+		dlog.Warn(ctx, "HTTP mechanism requested on UDP listener; not supported. Proceeding with plain UDP forwarding.")
```
thallgren: I think this should be an error. The user will cause a unintended disruption for everyone using the service.

rpc/manager/manager.proto
```diff
@@ -186,6 +186,18 @@ message InterceptSpec {
...
+  // these headers (key=value) will be intercepted. Multiple headers use AND logic.
+  map<string,string> header_filters = 27;
+
+  // Path filter patterns for HTTP Intercepts. Supports glob patterns like "/api/v1/*".
+  // If specified, only requests matching these paths will be intercepted.
+  repeated string path_filters = 28;
+
+  // Enable HTTP-aware interception mechanism for HTTP Intercepts.
+  // When true, the mechanism will parse HTTP requests to apply header/path filters.
+  bool http_mechanism = 29;
```
thallgren: This seems redundant. Why would we ever use the http mechanism unless there are header_filters and path_filters?

pkg/forwarder/tcp.go
```diff
@@ -249,3 +254,60 @@ func (f *tcp) rerouteConn(ctx context.Context, conn net.Conn, clientSession tunn
...
+// DispatchByMechanism implements mechanism-specific per-connection dispatch for TCP.
+// It currently routes HTTP-aware intercepts when requested.
+func (f *tcp) DispatchByMechanism(ctx context.Context, clientConn net.Conn, intercept *manager.InterceptInfo) (bool, error) {
+	var spec *manager.InterceptSpec
+	if intercept != nil {
+		spec = intercept.Spec
+	}
+
+	switch {
+	case spec != nil && spec.HttpMechanism:
```
thallgren: Suggest that you add a `UsesHTTPMechanism(spec *manager.InterceptSpec) bool` similar to the one on Command in the client.

thallgren (changes requested):
Just some nits, but you'll need to also sign your commits to get the DCO check to pass.

On my wishlist (but fine if we add later):
- Some rudimentary conflict detection mechanism that will prevent users from using the same headers, paths, or subset thereof.
- A test that verifies that you can run multiple simultaneous intercepts on the same port using different filters.

pkg/forwarder/tcp.go
bgruszka: Right now is like this:

```go
httpMechanism := spec != nil && (len(spec.HeaderFilters) > 0 || len(spec.PathFilters) > 0)

	switch {
	case httpMechanism:
```

rpc/manager/manager.proto
bgruszka: removed :)

pkg/forwarder/udp.go
bgruszka: I actually did it like:

```go
// DispatchByMechanism implements the Interceptor hook for UDP. Currently, HTTP-aware
// mechanisms are not supported on UDP; all UDP traffic is forwarded unconditionally.
func (f *udp) DispatchByMechanism(ctx context.Context, _ net.Conn, intercept *manager.InterceptInfo) (bool, error) {
	return false, nil
}
```

Because for now UDP doesn't support any mechanism.

And I added validation to `command.Validate`:
```go
if c.UsesHTTPMechanism() {
	for _, header := range c.HTTPHeaderFilters {
		if _, _, err := parseHTTPHeader(header); err != nil {
			return err
		}
	}

	// Validate HTTP mechanisms aren't used with UDP ports
	for _, portSpec := range c.Ports {
		if pp, err := types.ParsePortAndProto(portSpec); err == nil && pp.Proto == types.ProtoUDP {
			return errcat.User.Newf("HTTP filters cannot be used with UDP port %s", portSpec)
		}
	}

	// Auto-detect and set mechanism to "http"
	c.Mechanism = "http"
}
```

pkg/client/cli/intercept/command.go
bgruszka: changed. I've introduced new function:

```go
// tryParseHeaderWithSeparator attempts to parse a header with the given separator.
// Returns the key, value, and true if successful; empty strings and false otherwise.
func tryParseHeaderWithSeparator(header, separator string) (string, string, bool) {
	parts := strings.SplitN(header, separator, 2)
	if len(parts) != 2 {
		return "", "", false
	}

	key := strings.TrimSpace(parts[0])
	if key == "" {
		return "", "", false
	}

	value := strings.TrimSpace(parts[1])
	return key, value, true
}
```

bgruszka:
> Just some nits, but you'll need to also sign your commits to get the DCO check to pass.
> 
> On my wishlist (but fine if we add later):
> 
> * Some rudimentary conflict detection mechanism that will prevent users from using the same headers, paths, or subset thereof.
> * A test that verifies that you can run multiple simultaneous intercepts on the same port using different filters.

@thallgren I made changes you asked for (in comments) and I've also singed my commits :)

Regarding conflict detection, would be great to have it :) but I hope we can add it as a next step? :)

thallgren (approved):
Perfect. Thanks!

Issue #3969, opened 2025-09-26
Personal intercept conflict


thallgren:
When I try to run two personal intercepts on the same workload simultaneously, I run into an error similar to:

```
✘ connector.CreateIntercept: Failed to establish intercept: intercept in error state AGENT_ERROR: Conflicts with the currently-served intercept                     9223372036.9s 
   "b85f2618-dac1-42e7-93d7-12a46659ec77:echo-one"
```

Steps to reproduce using `integration_test/testdata/k8s/echo-easy.yaml`

```bash
telepresence intercept --port 8080:80 --http-header 'x-user=adam' --mount=false --workload echo-easy echo-one
telepresence intercept --port 8081:80 --http-header 'x-user=bertil' --mount=false --workload echo-easy echo-two
```

@bgruszka do you want to take this one?

bgruszka:
I can take it :)

bgruszka:
I think we can close it :)

thallgren:
Closed by #3971

PR #3971, opened 2025-09-29
feat(intercept): implement interceptSpecsConflict function and add tests

bgruszka:
## Description

Issue #3969:
  - Personal intercepts with different HTTP headers now coexist
  - x-user=adam vs x-user=bertil = NO conflict

Performance Optimized:
  - Fast path: TCP conflicts detected immediately
  - Complex HTTP logic only runs when needed

Architecture Respects Design:
  - Different ports → Different InterceptState → No conflicts possible
  - Same port → Same InterceptState → Smart conflict detection applied

Backward Compatibility Preserved:
  - TCP intercepts still conflict correctly
  - All existing tests pass

## Checklist

### DIFF
--- a/cmd/traffic/cmd/agent/fwdstate.go
+++ b/cmd/traffic/cmd/agent/fwdstate.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"net/textproto"
 	"path/filepath"
 	"slices"
 	"strings"
@@ -111,15 +112,16 @@ func (pm *ProviderMux) CreateClientStream(ctx context.Context, tag tunnel.Tag, s
 	return pm.AgentProvider.CreateClientStream(ctx, tag, sessionID, id, roundTripLatency, dialTimeout)
 }
 
-// normalizeHeaderFilters returns a new map with all header keys normalized to lowercase.
-// HTTP headers are case-insensitive per RFC 7230.
+// normalizeHeaderFilters returns a new map with all header keys normalized to canonical MIME format.
+// HTTP headers are case-insensitive per RFC 7230, so we use the same canonicalization as net/http.
+// Examples: "x-user" -> "X-User", "content-type" -> "Content-Type"
 func normalizeHeaderFilters(headers map[string]string) map[string]string {
 	if len(headers) == 0 {
 		return headers
 	}
 	normalized := make(map[string]string, len(headers))
 	for key, value := range headers {
-		normalized[strings.ToLower(key)] = value
+		normalized[textproto.CanonicalMIMEHeaderKey(key)] = value
 	}
 	return normalized
 }
@@ -257,14 +259,14 @@ func isHeaderSubset(subset, superset map[string]string) bool {
 // priority tier than intercepts with only paths.
 //
 // Conflict Rules:
-// 1. Global intercepts (no headers, no paths) conflict with everything
-// 2. Headers vs Paths: One spec with headers, another with only paths → NO CONFLICT
-//    (different priority tiers - headers are checked first, then paths)
-// 3. Both have headers: Conflict if headers form a subset AND paths overlap
-//    - Within each intercept, filters use AND logic (must match ALL headers AND ALL paths)
-//    - Example: {x-user:adam} vs {x-user:adam, x-session:xyz} → CONFLICT (first is subset)
-//    - Example: {x-user:adam}+/api/* vs {x-user:adam}+/admin/* → NO CONFLICT (different paths)
-// 4. Both have only paths (no headers): Conflict if paths overlap
+//  1. Global intercepts (no headers, no paths) conflict with everything
+//  2. Headers vs Paths: One spec with headers, another with only paths → NO CONFLICT
+//     (different priority tiers - headers are checked first, then paths)
+//  3. Both have headers: Conflict if headers form a subset AND paths overlap
+//     - Within each intercept, filters use AND logic (must match ALL headers AND ALL paths)
+//     - Example: {x-user:adam} vs {x-user:adam, x-session:xyz} → CONFLICT (first is subset)
+//     - Example: {x-user:adam}+/api/* vs {x-user:adam}+/admin/* → NO CONFLICT (different paths)
+//  4. Both have only paths (no headers): Conflict if paths overlap
 func interceptSpecsConflict(spec1, spec2 *manager.InterceptSpec) bool {
 	hasHeaders1 := len(spec1.HeaderFilters) > 0
 	hasHeaders2 := len(spec2.HeaderFilters) > 0
@@ -282,11 +284,11 @@ func interceptSpecsConflict(spec1, spec2 *manager.InterceptSpec) bool {
 
 	// Rule 2: Headers take precedence - different priority tiers don't conflict
 	// If one spec has headers and the other has only paths, they operate at different tiers
-	if hasHeaders1 && !hasHeaders2 && !isGlobal2 {
+	if hasHeaders1 && !hasHeaders2 {
 		// spec1 has headers (high priority), spec2 has only paths (low priority)
 		return false
 	}
-	if hasHeaders2 && !hasHeaders1 && !isGlobal1 {
+	if hasHeaders2 && !hasHeaders1 {
 		// spec2 has headers (high priority), spec1 has only paths (low priority)
 		return false
 	}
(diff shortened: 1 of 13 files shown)


cmd/traffic/cmd/agent/fwdstate.go
```diff
@@ -94,6 +94,159 @@ func (pm *ProviderMux) CreateClientStream(ctx context.Context, tag tunnel.Tag, s
 	return pm.AgentProvider.CreateClientStream(ctx, tag, sessionID, id, roundTripLatency, dialTimeout)
 }
 
+// interceptSpecsConflict determines if two intercept specs would conflict with each other.
+// Two specs conflict if they would route the same traffic to different destinations.
+func interceptSpecsConflict(spec1, spec2 *manager.InterceptSpec) bool {
+	// Fast path: For TCP intercepts (no filters), they always conflict on the same port
+	if len(spec1.HeaderFilters) == 0 && len(spec1.PathFilters) == 0 &&
+		len(spec2.HeaderFilters) == 0 && len(spec2.PathFilters) == 0 {
```
thallgren: I would claim that any other spec is in conflict with `spec1` here, since `spec1` is global.

cmd/traffic/cmd/agent/fwdstate.go
```diff
@@ -94,6 +94,159 @@ func (pm *ProviderMux) CreateClientStream(ctx context.Context, tag tunnel.Tag, s
...
+		return true
+	}
+
+	// For HTTP intercepts, check header filter conflicts
+	if len(spec1.HeaderFilters) > 0 && len(spec2.HeaderFilters) > 0 {
+		// Check if any header key has the same value in both specs
+		for key1, value1 := range spec1.HeaderFilters {
+			if value2, exists := spec2.HeaderFilters[key1]; exists && value1 == value2 {
+				// Same header key with same value = traffic would match both intercepts
```
thallgren: This is can be false when there are more than one header match. An intercept for `A:a, B:b` will not be in conflict with `A:a, B:c` because B is different and we're using AND logic). It doesn't conflict with `A:a, C:c` either.

You'll need to check for subsets. E.g. A has has the exact same headers as B, or a A has a subset of the headers in B, but no other headers.

thallgren: In addition, you may have two intercepts with the exact same set of headers that matches different paths. They are not in conflict because were using AND logic between the headers and the paths.

cmd/traffic/cmd/agent/fwdstate_test.go
```diff
@@ -104,3 +104,134 @@ func TestGenerateMechanismDescription(t *testing.T) {
...
+				},
+			},
+			spec2: &manager.InterceptSpec{
+				Mechanism:   "http",
+				PathFilters: []string{"/admin/*"},
+			},
+			conflicts: false,
+		},
+		{
+			name: "Multiple headers with one overlap - should conflict",
```
thallgren: I don't see why those would conflict. All headers must be matched.

thallgren (changes requested):
I thing you're on the right track when enabling that multiple intercepts can be active on one port, but the conflict detection needs some more work.

cmd/traffic/cmd/agent/fwdstate.go
```diff
@@ -160,56 +313,19 @@ func (fs *fwdState) HandlePort(ctx context.Context, cepts []*manager.InterceptIn
...
+	candidateIntercepts := slices.Clone(waiting)
+
+	for i, ii := range candidateIntercepts {
+		if ii.Spec.Wiretap {
+			reviews = append(reviews, fs.processWiretapIntercept(ii))
+			continue
 		}
+
+		// Check for conflicts and process regular intercept
+		review := fs.processRegularIntercept(ctx, ii, i, active, candidateIntercepts, &activeIntercept)
```
thallgren, line 541: Thinking more about this. The concept of an `activeIntercept` is actually invalid now, because there can be several non-conflicting active intercepts on the same port/proto combination. Shouldn't this be used only for the one and only "global intercept" if there is such an intercept?

bgruszka: Yes, I've also discoverd that.  Active intercept is only valid for tcp (not http). I will think how to change it.

bgruszka:
@thallgren I think... headers should take precedence over paths, right?

dev1 sets header x-user: adam
dev2 sets path /admin/*

It's still ok, if we will first check header and then paths, right?

all the request will work for dev1 even for /admin/* paths

all the request will work for dev2 for /admin/* paths

So my question is - should we use prioritization/precedence model, not an AND logic model for conflict detection?

So conflict only happens when:
  1. Two specs have the same headers (or one is a subset) → conflict
  2. Two specs BOTH have no headers and have overlapping paths → conflict
  3. If one has headers and another has only paths → NO conflict (headers take precedence)

The routing prioritizes:
  1. Most specific header matches
  2. Fallback to path matching if no headers match
  3. Resolve conflicts by comparing header and path specificity

  This means different specs can coexist without conflict if they target distinct routing tiers - headers provide the primary filtering mechanism, with paths as a secondary routing strategy.

thallgren:
> I think... headers should take precedence over paths, right?
> 
Yes I like that. Let's enforce that rule and document it. It removes a lot of ambiguities.

> The routing prioritizes:
> 
> 1. Most specific header matches

Not sure what you mean here. If you have two intercepts where one is less specific, then that's a conflict right?

Oh  wait. You mean that if you have three headers or more, and one intercept matches one header and the other matches two headers, then the second one handles the request? That's sounds OK to me.

> 2. Fallback to path matching if no headers match

If a header match is requested, and the request doesn't match those headers, then there's no fallback in my mind. Path matching is only made when there are no headers to match or when the headers match.

> 3. Resolve conflicts by comparing header and path specificity

Yes, the priority rule resolves that.

cmd/traffic/cmd/agent/fwdstate.go
```diff
@@ -94,48 +111,217 @@ func (pm *ProviderMux) CreateClientStream(ctx context.Context, tag tunnel.Tag, s
...
+	isGlobal2 := !hasHeaders2 && !hasPaths2
+
+	// Rule 1: Global intercepts conflict with anything
+	if isGlobal1 || isGlobal2 {
+		return true
+	}
+
+	// Rule 2: Headers take precedence - different priority tiers don't conflict
+	// If one spec has headers and the other has only paths, they operate at different tiers
+	if hasHeaders1 && !hasHeaders2 && !isGlobal2 {
```
thallgren: `isGlobal1` and `isGlobal2` are both guaranteed to be `false` for the remainder of this function.

cmd/traffic/cmd/agent/fwdstate.go
```diff
@@ -94,48 +111,217 @@ func (pm *ProviderMux) CreateClientStream(ctx context.Context, tag tunnel.Tag, s
...
-		return true
+// normalizeHeaderFilters returns a new map with all header keys normalized to lowercase.
+// HTTP headers are case-insensitive per RFC 7230.
+func normalizeHeaderFilters(headers map[string]string) map[string]string {
+	if len(headers) == 0 {
+		return headers
+	}
+	normalized := make(map[string]string, len(headers))
+	for key, value := range headers {
+		normalized[strings.ToLower(key)] = value
```
thallgren: An alternative to using `strings.ToLower` here, which I think produces more readable headers once they are displayed, is to use the `textproto.CanonicalMIMEHeaderKey`.

thallgren (approved):
This is great. I have a suggestion that I think improves how headers are displayed and some nits that you can ignore (but I suspect the linter might complain).

cmd/traffic/cmd/agent/fwdstate.go
bgruszka: changed

cmd/traffic/cmd/agent/fwdstate.go
bgruszka: changed :)

cmd/traffic/cmd/agent/fwdstate_test.go
bgruszka: changed

cmd/traffic/cmd/agent/fwdstate.go
bgruszka: Introduced new logic :)

cmd/traffic/cmd/agent/fwdstate.go
bgruszka: removed unnecessary condition

cmd/traffic/cmd/agent/fwdstate.go
bgruszka: changed to `textproto.CanonicalMIMEHeaderKey`

PR #3979, opened 2025-10-08
feat: add mcp server to cli

njayp:
## Description

I wrote [ophis](https://github.com/njayp/ophis), which converts cobra.Command trees to mcp servers. With this small PR, Telepresnce has a mcp server that can do basic commands, though you have to have the daemons running first:

"telepresence connect",
"telepresence intercept",
"telepresence ingest",
"telepresence leave",
"telepresence list",
"telepresence quit",
"telepresence status",

Please try it out and let me know if you think this would be a good addition to telepresence. If so, I can start on docs, work on shrinking the server context footprint, and refine some flag descriptions and error messages to help the AI (and humans) out a bit.

## QuickStart

- Build Telepresence
- Start the root daemon
- Enable the mcp server with 'telepresence mcp claude enable' or 'telepresence mcp vscode enable'
- Use ai to create intercepts

## Checklist


thallgren (changes requested):
Very cool! Can you please run a `make generate` so that the proper CLI-docs gets included?
Also, please sign your commits ;-)

thallgren:
An entry in the CHANGELOG that describes this new feature would be great so that it gets advertised properly when we release.

njayp:
I decided to remove kubeflags from the telepresence connect mcp tool for two reasons:
- they are special-case, and probably shouldn't be controlled inaccurately by an AI
- less flags = less context footprint

What do you think of the flags I chose to keep/remove?

pkg/client/cli/cmd/mcp.go
```diff
@@ -0,0 +1,60 @@
...
+				LocalFlagSelector:     ophis.NoFlags,
+				InheritedFlagSelector: ophis.NoFlags,
+			},
+			{
+				CmdSelector: ophis.AllowCmds(
+
+					"telepresence intercept",
+					"telepresence ingest",
+					"telepresence leave",
+					"telepresence list",
```
bgruszka, line 50: should we add `telepresence wiretap` too?

pkg/client/cli/cmd/mcp_test.go
```diff
@@ -0,0 +1,22 @@
...
+	cmd := Telepresence(ctx, nil)
+	tools := test.GetTools(t, cmd)
+	test.ToolNames(t, tools,
+		"telepresence_quit",
+		"telepresence_status",
+		"telepresence_connect",
+		"telepresence_intercept",
+		"telepresence_ingest",
+		"telepresence_leave",
+		"telepresence_list",
```
bgruszka, line 20: should we add telepresence wiretap too?

pkg/client/cli/cmd/mcp.go
thallgren: We absolutely need `telepresence replace`. The `telepresence intercept --replace` option is deprecated in favor of that command.

thallgren (changes requested):
Looking good now, but please add the replace command as well.

Also, I forgot to mention, but the way things work right now you'll need to add the CLI to `docs/doc-links.yml` manually. Look at how `compose` and its subcommands are added.

thallgren:
> What do you think of the flags I chose to keep/remove?

I think that's fine. I'm not too happy about those flags in any command actually. Those settings typically belong in the kubeconfig IMO.

pkg/client/cli/cmd/mcp.go
njayp: I will add both, thanks

thallgren (approved):

PR #3980, opened 2025-10-10
Allow regex support for http header

nmane047:
Regular expression support for http header

Issue: if one user one uses --http-header x-intercept-id: some-name::.* , then if another tries using --http-header with the same key 'x-intercept-id' , it  wouldn't allow. 

Fix: This change reduces false positives, assuming most different regexes won't match the same input.
Returns true by default, meaning assumes possible match/conflict unless proven otherwise.

This is very conservative, especially about regexes.

### DIFF
--- a/cmd/traffic/cmd/agent/fwdstate.go
+++ b/cmd/traffic/cmd/agent/fwdstate.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"regexp"
 	"slices"
 	"strings"
 	"time"
@@ -101,7 +102,12 @@ func pathsCanMatch(filter1, filter2 string) bool {
 	return valuesCanMatch(matcher.PathValue(filter1), matcher.PathValue(filter2))
 }
 
-// valuesCanMatch determines if two matcher.Values can match the same string value.
+// valuesCanMatch determines if two matcher.Value objects could potentially match the same input.
+// It is used to detect conflicts between intercept path filters.
+// Supported operations: Equal, Prefix, Regex.
+//
+// The logic ensures that overlapping or equivalent patterns (e.g., /api/* vs /api/v1/*)
+// are treated as conflicts, while unrelated ones (e.g., /api/v1/* vs /api/v2/*) are not.
 func valuesCanMatch(v1, v2 matcher.Value) bool {
 	op1 := v1.Op()
 	op2 := v2.Op()
@@ -114,7 +120,7 @@ func valuesCanMatch(v1, v2 matcher.Value) bool {
 		return pattern1 == pattern2
 
 	case op1 == matcher.ValueOpPrefix && op2 == matcher.ValueOpPrefix:
-		// Both prefixes - conflict if one is prefix of the other
+		// Both are prefixes - conflict if one is prefix of the other
 		return strings.HasPrefix(pattern1, pattern2) || strings.HasPrefix(pattern2, pattern1)
 
 	case op1 == matcher.ValueOpPrefix && op2 == matcher.ValueOpEqual:
@@ -124,11 +130,98 @@ func valuesCanMatch(v1, v2 matcher.Value) bool {
 	case op1 == matcher.ValueOpEqual && op2 == matcher.ValueOpPrefix:
 		// Exact value matches prefix if it starts with the prefix
 		return strings.HasPrefix(pattern1, pattern2)
+
+	case op1 == matcher.ValueOpRegex && op2 == matcher.ValueOpRegex:
+		// --- Regex vs Regex ---
+		if pattern1 == pattern2 {
+			return true
+		}
+
+		anchored1 := strings.HasPrefix(pattern1, "^")
+		anchored2 := strings.HasPrefix(pattern2, "^")
+
+		prefix1 := regexLiteralPrefix(pattern1)
+		prefix2 := regexLiteralPrefix(pattern2)
+
+		// Both anchored → only overlap if one is prefix of the other
+		if anchored1 && anchored2 {
+			return strings.HasPrefix(prefix1, prefix2) || strings.HasPrefix(prefix2, prefix1)
+		}
+		// Fallback conservative
+		return true
+
+	case op1 == matcher.ValueOpRegex && (op2 == matcher.ValueOpEqual || op2 == matcher.ValueOpPrefix):
+		return regexCanMatchValue(pattern1, op2, pattern2)
+
+	case op2 == matcher.ValueOpRegex && (op1 == matcher.ValueOpEqual || op1 == matcher.ValueOpPrefix):
+		return regexCanMatchValue(pattern2, op1, pattern1)
 	}
 
-	// Conservative: assume regexes can overlap
-	// Proper regex intersection is computationally expensive
+	// Fallback: conservatively assume potential match
 	return true
+}
+
+// regexCanMatchValue checks if a regex could match an equal or prefix value.
+// Conservative: returns true if regex is invalid or overlap can't be ruled out.
+func regexCanMatchValue(regexPattern string, op matcher.ValueOp, value string) bool {
+	if value == "" {
+		return true
+	}
+
+	anchored := strings.HasPrefix(regexPattern, "^")
+
+	switch op {
+	case matcher.ValueOpEqual:
+		if anchored {
+			re, err := regexp.Compile(regexPattern)
+			if err != nil {
+				return true // invalid regex → assume conflict
+			}
+			return re.MatchString(value)
+		}
+		// Unanchored regex → can match anywhere in the string
+		return true
+
+	case matcher.ValueOpPrefix:
+		if anchored {
+			// Anchored regex → behaves like HasPrefix using literal prefix
+			prefix := regexLiteralPrefix(regexPattern)
+			if prefix == "" {
+				return true
+			}
+			return strings.HasPrefix(value, prefix) || strings.HasPrefix(prefix, value)
+		}
+
+		// Unanchored regex → can match anywhere in the string
+		return true
+	}
+	// fallback
+	return true
+}
+
+// regexLiteralPrefix safely extracts a literal prefix from a regex pattern.
+// Stops at first regex metacharacter. Conservative for complex regexes.
+func regexLiteralPrefix(pattern string) string {
+	if pattern == "" {
+		return ""
+	}
+
+	// Remove leading '^' (safe: TrimPrefix is a no-op if not present)
+	pattern = strings.TrimPrefix(pattern, "^")
+
+	// Remove trailing '.*' (safe: TrimSuffix is a no-op if not present)
+	pattern = strings.TrimSuffix(pattern, ".*")
+
+	var prefix strings.Builder
+	for _, r := range pattern {
+		// Stop at any regex metacharacter
+		if strings.ContainsRune("[](){}?+*|$.\\^", r) {
+			break
+		}
+		prefix.WriteRune(r)
+	}
+
+	return prefix.String()
 }
 
 // explainConflict generates a human-readable explanation of why two intercept specs conflict.
--- a/cmd/traffic/cmd/agent/fwdstate_test.go
+++ b/cmd/traffic/cmd/agent/fwdstate_test.go
@@ -491,7 +491,7 @@ func TestInterceptSpecsConflict(t *testing.T) {
 			conflicts: false,
 		},
 		{
-			name: "Path regex - :path-regex: always conflicts conservatively",
+			name: "Path regex - :path-regex: should not conflict (distinct paths)",
 			spec1: &manager.InterceptSpec{
 				Mechanism:   "http",
 				PathFilters: []string{":path-regex:^/api/.*$"},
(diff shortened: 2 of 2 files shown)


cmd/traffic/cmd/agent/fwdstate.go
```diff
@@ -124,11 +127,15 @@ func valuesCanMatch(v1, v2 matcher.Value) bool {
...
+	// If different, assume they don't overlap (optimistic heuristic)
+	case op1 == matcher.ValueOpRegex && op2 == matcher.ValueOpRegex:
+		return pattern1 == pattern2
 	}
 
-	// Conservative: assume regexes can overlap
-	// Proper regex intersection is computationally expensive
-	return true
+	// Default fallback — assume they don’t match to avoid false positives
+	return false
```
thallgren: This is too relaxed. It means that no regexps are in conflict unless they are exactly equal.

I could accept a solution that checked for anchored regexps using the same prefix or the same suffix. CoPilot actually suggested that in a review [here](https://github.com/telepresenceio/telepresence/pull/3972#discussion_r2394330182).

thallgren (changes requested):
This check is too relaxed.

cmd/traffic/cmd/agent/fwdstate.go
nmane047: All right , will work on this thanks @thallgren

cmd/traffic/cmd/agent/fwdstate.go
```diff
@@ -124,10 +127,46 @@ func valuesCanMatch(v1, v2 matcher.Value) bool {
...
 		// Exact value matches prefix if it starts with the prefix
 		return strings.HasPrefix(pattern1, pattern2)
+
+	case op1 == matcher.ValueOpRegex && op2 == matcher.ValueOpRegex:
+		// Regex vs Regex
+		if pattern1 == pattern2 {
+			return true
+		}
+		prefix1 := strings.TrimSuffix(strings.TrimPrefix(pattern1, "^"), ".*")
+		prefix2 := strings.TrimSuffix(strings.TrimPrefix(pattern2, "^"), ".*")
```
thallgren: This looks a odd to me. There's no check that the pattern actually has the prefix that is trimmed off. I think the test must be conditioned with:
```
		if strings.HasPrefix(pattern1, "^") && strings.HasPrefix(pattern2, "^") {
		  // trim and compare here
		}
```

I guess trimming (or not) a `".*"` is safe because it's a no-op at the end of a regexp. The regexp has no end-anchor anyway.

cmd/traffic/cmd/agent/fwdstate.go
```diff
@@ -124,10 +127,46 @@ func valuesCanMatch(v1, v2 matcher.Value) bool {
...
+		// Regex vs Exact or Prefix
+		switch op2 {
+		case matcher.ValueOpEqual:
+			re1, err := regexp.Compile(pattern1)
+			if err != nil {
+				return true
+			}
+			return re1.MatchString(pattern2)
+		case matcher.ValueOpPrefix:
+			prefix := strings.TrimSuffix(strings.TrimPrefix(pattern1, "^"), ".*")
```
thallgren: Same here. Must be conditioned with a check for the existence of the prefix

cmd/traffic/cmd/agent/fwdstate_test.go
```diff
@@ -500,8 +500,264 @@ func TestInterceptSpecsConflict(t *testing.T) {
 				Mechanism:   "http",
 				PathFilters: []string{":path-prefix:/admin/"},
 			},
+			conflicts: false,
+		},
+		{
+			name: "Regex vs Regex with same prefix - should conflict",
+			spec1: &manager.InterceptSpec{
+				Mechanism:   "http",
+				PathFilters: []string{":path-regex:^/api/.*"},
```
thallgren, line 509: Maybe add a similar check for `":path-regex:/api/.*"` (not anchored, and hence a possible conflict to, say "/v2/api/.*")

cmd/traffic/cmd/agent/fwdstate.go
```diff
@@ -124,10 +127,46 @@ func valuesCanMatch(v1, v2 matcher.Value) bool {
...
+			re2, err := regexp.Compile(pattern2)
+			if err != nil {
+				return true
+			}
+			return re2.MatchString(pattern1)
+		case matcher.ValueOpPrefix:
+			prefix := strings.TrimSuffix(strings.TrimPrefix(pattern2, "^"), ".*")
+			return strings.HasPrefix(prefix, pattern1) || strings.HasPrefix(pattern1, prefix)
+		}
 	}
```
thallgren, line 158: Maybe use a function that compares a regexp with an equal|prefix instead of repeating this block?

thallgren (changes requested):
Looks promising, but needs some more work.

cmd/traffic/cmd/agent/fwdstate.go
nmane047: hi @thallgren ,  I am new to Go programming and Github, please let me know if any other changes are required.

Please review the PR

This PR extends the existing valuesCanMatch logic to safely handle cases involving regular expressions, while retaining existing behavior for Equal and Prefix match types.

The new logic introduces lightweight regex analysis to detect likely overlaps based on literal prefixes, avoiding expensive or undecidable regex computations.

Details

No changes to existing Equal and Prefix handling — these remain exact and prefix-based comparisons.

New logic for regex cases:

Added regexLiteralPrefix() to extract a literal prefix from a regex pattern (stopping at the first metacharacter).

Added regexCanMatchValue() to check limited overlap between regex and prefix/equal values.

For regex vs regex, overlap is determined conservatively if one regex’s literal prefix starts with the other’s.

Fallback: For invalid or complex regex patterns where overlap can’t be determined safely, the function assumes a possible match (to avoid false negatives).

Rationale

Determining whether two arbitrary regexes overlap exactly is computationally expensive and, in general, undecidable without constructing and intersecting finite automata.
This implementation takes a pragmatic and conservative approach — detecting simple overlaps efficiently while avoiding exponential computation.


Future Improvements

Enhance literal extraction to handle escaped metacharacters.

Consider optional DFA-based overlap detection under a performance-flagged mode if higher accuracy is required.

cmd/traffic/cmd/agent/fwdstate_test.go
nmane047: Add the below test case and few more 

{
			name: "Regex vs Regex with same prefix - non-anchor- should conflict",
			spec1: &manager.InterceptSpec{
				Mechanism:   "http",
				PathFilters: []string{":path-regex:/api/.*"},
			},
			spec2: &manager.InterceptSpec{
				Mechanism:   "http",
				PathFilters: []string{":path-regex:/api/v2/.*"},
			},
			conflicts: true,
		},

cmd/traffic/cmd/agent/fwdstate.go
```diff
@@ -124,13 +130,82 @@ func valuesCanMatch(v1, v2 matcher.Value) bool {
...
 
+// regexLiteralPrefix safely extracts a literal prefix from a regex pattern.
+// Stops at first regex metacharacter. Conservative for complex regexes.
+func regexLiteralPrefix(pattern string) string {
+	if pattern == "" {
+		return ""
+	}
+
+	// Remove leading '^' (safe: TrimPrefix is a no-op if not present)
+	pattern = strings.TrimPrefix(pattern, "^")
```
thallgren, line 210: Unless the regexp is anchored at the beginning (starts with '^'), it cannot be used to match prefixes.

Example:
The regexp "/api" will match "anythinghere/api". The regexp "^/api" will not.

cmd/traffic/cmd/agent/fwdstate_test.go
thallgren: You'll also need something like:
```go
{
  name: "Regex vs different Regex - non-anchored - should conflict",
  spec1: &manager.InterceptSpec{
    Mechanism: "http",
    PathFilters: []string{":path-regex:/api"},
  },
  spec2: &manager.InterceptSpec{
    Mechanism: "http",
    PathFilters: []string{":path-regex:/foo"},
  },
conflicts: true,
},
```
The are in conflict because they're unanchored and both match the string "/foo/api"

```go
{
  name: "Regex vs different Prefix - non-anchored - should conflict",
  spec1: &manager.InterceptSpec{
    Mechanism: "http",
    PathFilters: []string{":path-regex:/api"},
  },
  spec2: &manager.InterceptSpec{
    Mechanism: "http",
    PathFilters: []string{":path-prefix:/foo/api"},
  },
conflicts: true,
},
```

Again, both the prefix and the regexp match the string "/foo/api"

thallgren (changes requested):
Better, but you still strip the anchor without checking if its there.

cmd/traffic/cmd/agent/fwdstate.go
nmane047: Added the below test cases, don't see the above issue, please let me know if I missed something

{
			name: "Regex vs different Regex - non-anchored - should not conflict",
			spec1: &manager.InterceptSpec{
				Mechanism:   "http",
				PathFilters: []string{":path-regex:/api"},
			},
			spec2: &manager.InterceptSpec{
				Mechanism:   "http",
				PathFilters: []string{":path-regex:/foo"},
			},
			conflicts: false,
		},
		{
			name: "Regex vs different Prefix - non-anchored - should not conflict",
			spec1: &manager.InterceptSpec{
				Mechanism:   "http",
				PathFilters: []string{":path-regex:/api/"},
			},
			spec2: &manager.InterceptSpec{
				Mechanism:   "http",
				PathFilters: []string{":path-prefix:/foo/api/"},
			},
			conflicts: false,
		},
		{
			name: "Regex vs overlapping Regex - non-anchored - should not conflict",
			spec1: &manager.InterceptSpec{
				Mechanism:   "http",
				PathFilters: []string{":path-regex:/api/"},
			},
			spec2: &manager.InterceptSpec{
				Mechanism:   "http",
				PathFilters: []string{":path-regex:/foo/api/"},
			},
			conflicts: false,
		},

nmane047: Here’s the same full explanation 

---

Test case

{
name: "Regex vs different Prefix - non-anchored - should not conflict",
spec1: &manager.InterceptSpec{
Mechanism:   "http",
PathFilters: []string{":path-regex:/api/"},
},
spec2: &manager.InterceptSpec{
Mechanism:   "http",
PathFilters: []string{":path-regex:/foo/api/"},
},
conflicts: false,
}

How valuesCanMatch handles it

When both operands are regexes (ValueOpRegex):

case op1 == matcher.ValueOpRegex && op2 == matcher.ValueOpRegex:
if pattern1 == pattern2 {
return true
}

```
prefix1 := regexLiteralPrefix(pattern1)
prefix2 := regexLiteralPrefix(pattern2)

return strings.HasPrefix(prefix1, prefix2) || strings.HasPrefix(prefix2, prefix1)
```

Here:
pattern1 = "/api/"
pattern2 = "/foo/api/"

Step 1: regexLiteralPrefix("/api/")

* Remove leading ^ → no change
* Remove trailing .* → no change
* Loop until metacharacter → none found
  Result: prefix1 = "/api/"

Step 2: regexLiteralPrefix("/foo/api/")

* Remove leading ^ → no change
* Remove trailing .* → no change
* Loop until metacharacter → none found
  Result: prefix2 = "/foo/api/"

Step 3: Prefix comparison
strings.HasPrefix("/api/", "/foo/api/") → false
strings.HasPrefix("/foo/api/", "/api/") → false
Both false → function returns false (no conflict)

Final outcome
valuesCanMatch returns false
Matches the test expectation: no conflict

Why this is correct

* Both patterns are non-anchored regexes (no leading ^), meaning they can appear anywhere in a path.
* Their literal prefixes differ (/api/ vs /foo/api/), so they describe different match regions.
* There’s no guaranteed overlap, only incidental. Therefore, not a conflict.

If Anchors Were Present
If they were anchored:
pattern1 = "^/api/"
pattern2 = "^/foo/api/"
regexLiteralPrefix("^/api/") → "/api/"
regexLiteralPrefix("^/foo/api/") → "/foo/api/"
Still no conflict.

If they were:
pattern1 = "^/api/"
pattern2 = "^/api/v1"
Then prefixes overlap:
strings.HasPrefix("/api/v1", "/api/") == true
So that would conflict.

Summary

Case                         | Anchored? | Conflict? | Reason
/api/ vs /foo/api/            | No        | No        | different literal prefixes
^/api/ vs ^/foo/api/          | Yes       | No        | still different
^/api/ vs ^/api/v1            | Yes       | Yes       | one prefix extends the other

Conclusion:
current implementation already correctly handles both anchored and non-anchored regex patterns for conflict detection.

thallgren: This:
```go
{
  name: "Regex vs different Prefix - non-anchored - should not conflict",
  spec1: &manager.InterceptSpec{
    Mechanism: "http",
    PathFilters: []string{":path-regex:/api/"},
  },
  spec2: &manager.InterceptSpec{
    Mechanism: "http",
    PathFilters: []string{":path-prefix:/foo/api/"},
  },
  conflicts: false,
},
```
is incorrect. This is indeed a conflict. The regexp "/api/" and the prefix "/foo/api/" will both match a path like "/foo/api/hello". Again, the regexp is not anchored, so in its current form it's similar to `string.Contains(s, "/api/")`. It is not similar to `string.HasPrefix(s, "/api/")`.

thallgren: Not sure what you mean with "only incidental" conflict. In my view it's very binary. There's either a conflict or there isn't. Whether it's incidental doesn't matter.

nmane047: Thanks @thallgren for the detailed explanation, also appreciate your patience , made changes so that they conflict now

{
			name: "Regex vs different Prefix - non-anchored - should conflict",
			spec1: &manager.InterceptSpec{
				Mechanism:   "http",
				PathFilters: []string{":path-regex:/api/"},
			},
			spec2: &manager.InterceptSpec{
				Mechanism:   "http",
				PathFilters: []string{":path-prefix:/foo/api/"},
			},
			conflicts: true,
		},
		{
			name: "Regex vs overlapping Regex - non-anchored - should conflict",
			spec1: &manager.InterceptSpec{
				Mechanism:   "http",
				PathFilters: []string{":path-regex:/api/"},
			},
			spec2: &manager.InterceptSpec{
				Mechanism:   "http",
				PathFilters: []string{":path-regex:/foo/api/"},
			},
			conflicts: true,
		},
		
Question: Should the below scenario also conflict? it does currently

{
			name: "Header regex - patterns with common suffix should conflict",
			spec1: &manager.InterceptSpec{
				Mechanism: "http",
				HeaderFilters: map[string]string{
					"x-user": "^adam::.*",
				},
			},
			spec2: &manager.InterceptSpec{
				Mechanism: "http",
				HeaderFilters: map[string]string{
					"x-user": "bertil::adam::.*",
				},
			},
			conflicts: true,
		},

cmd/traffic/cmd/agent/fwdstate.go
```diff
@@ -124,13 +130,114 @@ func valuesCanMatch(v1, v2 matcher.Value) bool {
...
+
+		// Both anchored → only overlap if one is prefix of the other
+		if anchored1 && anchored2 {
+			return strings.HasPrefix(prefix1, prefix2) || strings.HasPrefix(prefix2, prefix1)
+		}
+
+		// Both unanchored → overlap if their literal parts contain each other
+		if !anchored1 && !anchored2 {
+			return strings.Contains(prefix1, prefix2) || strings.Contains(prefix2, prefix1)
+		}
```
thallgren: Sorry, but no. An unanchored regexp will always conflict with everything but an exact match. They might match completely different parts of the string. E.g. `/b` and `/c` does not contain each other both they both match the path `/a/b/c/d`

Since maching unanchored against an exact match is done below, I suggest that you just return true as soon as you find at least one unanchored regexp.

thallgren (changes requested):
Sorry, but you still don't take into account that an unanchored regexp can match anywhere in the string.

cmd/traffic/cmd/agent/fwdstate.go
nmane047: Thomas, I have updated to return true as soon as least one unanchored regexp is found

cmd/traffic/cmd/agent/fwdstate.go
```diff
@@ -124,13 +130,106 @@ func valuesCanMatch(v1, v2 matcher.Value) bool {
...
+		// Both anchored → only overlap if one is prefix of the other
+		if anchored1 && anchored2 {
+			return strings.HasPrefix(prefix1, prefix2) || strings.HasPrefix(prefix2, prefix1)
+		}
+
+		// Both unanchored → overlap if their literal parts contain each other
+		if !anchored1 || !anchored2 {
+			return true
+		}
+
```
thallgren: Seems those lines can be removed. You return `true` anyway just below.

thallgren (changes requested):
Just a tiny nit, then you're good to go :-)

cmd/traffic/cmd/agent/fwdstate.go
nmane047: Yes removed them, Thanks.

thallgren:
Thanks for your contribution!

cmd/traffic/cmd/agent/fwdstate.go
nmane047: On a side note, Thomas when we do plan to release the recent changes ( along with the http header with path filtering ) :)

thallgren: When #3981 is merged.

nmane047: Thanks, Thomas! Super excited — can’t wait for the release! 😄

PR #3981, opened 2025-10-13
TLS/mTLS Intercept Support 

thallgren:
This PR adds TLS/mTLS support to Telepresence intercepts and wiretaps. The support wasn't needed before because all intercepts were made on the Transport Layer using TCP/UDP only and we could safely disregard everything related to encryption, but with the introduction of HTTP-filters, we're now at the Application Layer, above the presentation layer where the encryption occurs.

This PR is extensive and therefore divided into several distinct commits that follow a certain flow of changes. It's **highly recommended** that you look at each commit separately when reviewing.


cmd/traffic/cmd/agent/tls/portconfig.go
```diff
@@ -0,0 +1,204 @@
...
+	if p == nil {
+		return "", ""
+	}
+	if p.downstreamSecretPath != "" {
+		dsPath = filepath.Join("/tel_app_mounts", cn, p.downstreamSecretPath)
+	}
+	if p.upstreamSecretPath != "" {
+		usPath = filepath.Join("/tel_app_mounts", cn, p.upstreamSecretPath)
+	}
+	return dsPath, usPath
```
bgruszka: Should we do `filepath.Clean()` first and validate no `..` in?

```
cleaned := filepath.Clean(p.downstreamSecretPath)
```

```
cleaned := filepath.Clean(p.downstreamSecretPath)
```

```
if strings.Contains(cleaned, "..") {
```

cmd/traffic/cmd/agent/tls/portconfig.go
```diff
@@ -0,0 +1,204 @@
...
+		if err != nil {
+			return err
+		}
+	}
+	p.certMutex.Lock()
+	p.downstreamCertificate = dsCert
+	p.upstreamCertificate = usCert
+	p.certMutex.Unlock()
+	return nil
+}
```
bgruszka, line 319: should we validate `dsCert` and `usCert` certificates?

Using for example this:

```go
func validateCertificate(cert *tls.Certificate, direction string) error {
      if cert == nil {
              return nil
      }

      if len(cert.Certificate) == 0 {
              return fmt.Errorf("%s certificate has no certificate data", direction)
      }

      // Parse the certificate to check expiration
      x509Cert, err := x509.ParseCertificate(cert.Certificate[0])
      if err != nil {
              return fmt.Errorf("failed to parse %s certificate: %w", direction, err)
      }

      now := time.Now()

      // Check if certificate is expired
      if now.After(x509Cert.NotAfter) {
              return fmt.Errorf("%s certificate expired on %v (current time: %v)",
                      direction, x509Cert.NotAfter.Format(time.RFC3339), now.Format(time.RFC3339))
      }

      // Check if certificate is not yet valid
      if now.Before(x509Cert.NotBefore) {
              return fmt.Errorf("%s certificate not valid until %v (current time: %v)",
                      direction, x509Cert.NotBefore.Format(time.RFC3339), now.Format(time.RFC3339))
      }

      // Warn if certificate expires soon (within 7 days)
      if now.Add(7 * 24 * time.Hour).After(x509Cert.NotAfter) {
              dlog.Warnf(context.Background(), "%s certificate will expire on %v",
                      direction, x509Cert.NotAfter.Format(time.RFC3339))
      }

      return nil
}
```

cmd/traffic/cmd/agent/tls/manager.go
```diff
@@ -0,0 +1,404 @@
...
+	allReady := sync.WaitGroup{}
+	allReady.Add(len(m.portConfigs))
+	for p, cn := range m.portConfigs {
+		g.Go(fmt.Sprintf("watch-tls/%d", p), func(ctx context.Context) error {
+			return cn.watchPaths(ctx, &allReady)
+		})
+	}
+	allReady.Wait()
+	close(certsReady)
+}
```
bgruszka, line 198: Should we add timeout to prevent indefinite hang if certificate loading fails?

Maybe something like this:

```go
done := make(chan struct{})
go func() {
        allReady.Wait()
        close(done)
}()

select {
case <-done:
        dlog.Debugf(context.Background(), "All TLS certificates loaded successfully")
        close(certsReady)
case <-time.After(30 * time.Second):
        dlog.Errorf(context.Background(), "Timeout waiting for TLS certificates to load. Agent will continue without TLS support.")
        close(certsReady)
}
```

pkg/tunnel/stream_conn.go
```diff
@@ -0,0 +1,215 @@
...
+	writeProbe    *CounterProbe
+
+	// The lastIncoming message and the offset into it are protected by readLock.
+	readLock     sync.Mutex
+	offset       int
+	lastIncoming Message
+
+	// Using atomic.LoadInt64/atomic.StoreInt64 to avoid locking
+	readDeadline  int64
+	writeDeadline int64
```
bgruszka, line 45: I am not sure if it will be safe on 32-bit ARM.

Maybe use `sync.Mutex` ?

cmd/traffic/cmd/agent/fwd/http.go
```diff
@@ -0,0 +1,281 @@
...
+	if !useISV && cert == nil {
+		return trn
+	}
+	var certs []tls.Certificate
+	if cert != nil {
+		certs = []tls.Certificate{*cert}
+	}
+	trn.TLSClientConfig = &tls.Config{Certificates: certs, InsecureSkipVerify: useISV}
+	return trn
+}
```
bgruszka, line 203: Should we add log warning (audit logging)  when `InsecureSkipVerify` is enabled ?

cmd/traffic/cmd/agent/tls/manager.go
```diff
@@ -0,0 +1,404 @@
...
+	connInfo, ok := m.connInfos[port]
+	if ok && connInfo.TLS != ValueUnknown && connInfo.HTTP2 != ValueUnknown {
+		return connInfo.TLS == ValueSupported
+	}
+	dlog.Debugf(ctx, "Probing port %d for TLS and HTTP/2 support", port)
+	addr := netip.AddrPortFrom(m.podIP, port).String()
+	bc := backoff.NewExponentialBackOff()
+	bc.MaxElapsedTime = 2 * time.Second
+	bc.MaxInterval = 300 * time.Millisecond
+	bc.InitialInterval = 100 * time.Millisecond
```
bgruszka: Should it be configurable via environment variable/variables?

cmd/traffic/cmd/agent/tls/portconfig.go
```diff
@@ -0,0 +1,204 @@
...
+	if err != nil {
+		return nil, nil, fmt.Errorf("secret %q failed to load: %w", certPath, err)
+	}
+	privatePath := filepath.Join(path, privateKey)
+	keyData, err = os.ReadFile(privatePath)
+	if err != nil {
+		return nil, nil, fmt.Errorf("secret %q loaded successfully but %q failed with: %w", certPath, privatePath, err)
+	}
+	return certData, keyData, nil
+}
```
bgruszka: Should we have file size limits for certificates? For example 10MB?

bgruszka (commented):
Great work on this PR! 😄  The architecture and code structure are really good. However, I've identified some issues that need attention in my opinion:

1. Certificate loading timeout missing
2. Path traversal vulnerability (Certificate paths not validated, allows `../../` attacks)
3. Certificate validation missing
4. 32-bit race condition (Atomic int64 operations unsafe on 32-bit architectures)
5. `InsecureSkipVerify` should be logged when enabled
6. Hardcoded timeouts - TLS probe timeouts should be configurable

 I've left detailed comments. Overall, this is solid work 😄

cmd/traffic/cmd/agent/tls/portconfig.go
thallgren: That's a very good idea. As things stand here it would be possible to reach a cert in the wrong container. Good catch.

cmd/traffic/cmd/agent/tls/manager.go
thallgren: I don't think that's needed. The loading does not wait for network. It only examines files and it will immediately close on failure.

bgruszka: 👍

pkg/tunnel/stream_conn.go
thallgren: I don't think we really care about that. We've never provided 32-bit images. They might still have a future when it comes to small embedded systems, but in a Kubernetes cluster? Nah. Also, we're already using atomic operations on 64-bit quantities in a lot of other places.

bgruszka: 👍

cmd/traffic/cmd/agent/fwd/http.go
thallgren: There is a log warning already actually.  See line 231.

bgruszka: Oh... I missed it 😄

cmd/traffic/cmd/agent/tls/manager.go
thallgren: Yeah. A slow starting app-container might need longer poll-times here.

So far we've used annotations, not environment variables. The traffic-agent container isn't present in the workload so the only way to add environment variables is by having the traffic-agent injector do it when it creates the container. The traffic-agent injector is controlled by the Helm chart but the Helm chart config applies to all pods. But we might consider having two ways to configure this. One Helm-chart based that is global and then one annotation-based that is for a specific container port in a workload.

What do you think about this?

cmd/traffic/cmd/agent/tls/portconfig.go
thallgren: We're not managing the secrets, we're just reading them. If users install huge certificates, then that's their problem IMO. We don't need to impose any limits here.

cmd/traffic/cmd/agent/tls/manager.go
bgruszka: Good point about maintaining consistency (with the annotations) 😄 Helm-chart based and annotation-based approach sounds great 😄 

So we will have something like this, right?

```yaml
agent:
    tls:
        probeTimeout: "2s"  # Global default for all pods
```

Per-port (most specific) ?
```
telepresence.io/tls-probe-timeout.8080: "5s"
telepresence.io/tls-probe-timeout.8443: "3s"
```

Or per-workload (applies to all ports) ?
```
telepresence.io/tls-probe-timeout: "5s"
```

cmd/traffic/cmd/agent/tls/portconfig.go
bgruszka: 👍

cmd/traffic/cmd/agent/tls/manager.go
thallgren: Thanks. Yes, that sounds very reasonable, but thinking more about this, I'm actually inclined to keep things as they are. Here's why:

Probing is a last resort. It's only done when the protocol cannot be safely established by looking at the service `name`, `appProtocol`, or `number`. So when the user is faced with problems due to this timeout, there's already a simple solution for it. Simply add `appProtocol: https` or similar to the service. If we add helm chart configs and annotations, the user will have more choices, which is perhaps positive, but on the other hand, getting rid of the probing is always preferred. 

In short, we can decide to add this, but I don't think it will be very useful. A section in the docs explaining the importance of the `appProtocol` service field is probably more helpful.

cmd/traffic/cmd/agent/tls/manager.go
```diff
@@ -0,0 +1,404 @@
...
+				dlog.Debugf(ctx, "Port %d does not support HTTP/2: %v", port, err)
+				connInfo.HTTP2 = ValueNotSupported
+			}
+		}
+		m.connInfos[port] = connInfo
+		return nil
+	}, backoff.WithContext(bc, ctx))
+	if err != nil {
+		dlog.Warnf(ctx, "unable to dial port %d: %v", port, err)
+	}
```
bgruszka: Maybe improve the error message when probing times out?

For example:

```go
if err != nil {
    dlog.Warnf(ctx, "TLS probe timeout for port %d after %v. "+
        "To avoid probing and improve startup time, add 'appProtocol: https' "+
        "(or 'appProtocol: h2c' for HTTP/2 cleartext) to your Service definition. "+
        "See: https://kubernetes.io/docs/concepts/services-networking/service/#application-protocol "+
        "Error: %v",
        port, bc.MaxElapsedTime, err)
}
```

cmd/traffic/cmd/agent/tls/manager.go
bgruszka: That's a really good point, and I completely agree with your reasoning. Making the probe configurable would be optimizing for the wrong use case.

Maybe we should add documentation instead explaining:
- When/why TLS probing happens
- How to avoid it with appProtocol: https / appProtocol: https / h2c / etc.
- What to do if probing times out

I've also proposed improved error message (this guides users toward the correct solution rather than giving them another knob to turn) when probing fails in another comment 😄

cmd/traffic/cmd/agent/tls/manager.go
bgruszka: or even better (not only timeout):

```go
if err != nil {
      dlog.Warnf(ctx, "Failed to probe port %d for TLS support: %v. "+
          "To avoid probing and improve startup time, add 'appProtocol: https' "+
          "(or 'appProtocol: h2c' for HTTP/2 cleartext) to your Service definition. "+
          "See: https://kubernetes.io/docs/concepts/services-networking/service/#application-protocol",
          port, err)
  }
```

cmd/traffic/cmd/agent/tls/manager.go
thallgren: I decided to add a setting for the timeout after all. I also added info to the docs and better warnings. But most importantly, I modified the ListenerSwitch so that it starts the HTTP-accept loop when the first accept is made that is intended to be HTTP-filtered. This gives the app more time to get ready because the traffic-agent will need to start and communicate with the traffic-manager, send intercept reviews etc. before that happens.

So all in all, I think the probability for failure is greatly reduced, but if it still fails, the user will have a chance to remedy the problem.

cmd/traffic/cmd/agent/tls/portconfig.go
```diff
@@ -0,0 +1,451 @@
...
+	// Send the required settings frame (empty).
+	_, err = conn.Write([]byte{0, 0, 0, byte(http2.FrameSettings), 0, 0, 0, 0, 0})
+	if err != nil {
+		dlog.Debugf(ctx, "failed to send empty settings frame on port %d: %v", port, err)
+		return false
+	}
+
+	// Read the initial response (expect SETTINGS frame).
+	buf := make([]byte, 9) // Frame header is 9 bytes.
+	n, err := conn.Read(buf)
```
bgruszka, line 229: no timeout - can block forever?

cmd/traffic/cmd/agent/tls/portconfig.go
```diff
@@ -0,0 +1,451 @@
...
+	// Check if it's a valid, but empty SETTINGS frame with no flags.
+	if fr.Type != http2.FrameSettings || fr.Flags != 0 {
+		dlog.Debugf(ctx, "expected SETTINGS frame and empty flags, got %v, %b", fr.Type, fr.Flags)
+		return false
+	}
+
+	// The frame payload should be empty for initial SETTINGS.
+	if fr.Length > 0 {
+		buf = make([]byte, fr.Length)
+		n, err = conn.Read(buf)
```
bgruszka, line 251: no timeout - can block forever?

bgruszka (commented):
Just thinking about timeouts... What do you think about adding this in `probeHTTP2ClearText`?

```go
// Set a deadline for all I/O operations on this connection
// Use the same timeout as the dialer for consistency
deadline := time.Now().Add(200 * time.Millisecond)
if err := conn.SetDeadline(deadline); err != nil {
        dlog.Debugf(ctx, "failed to set connection deadline on port %d: %v", p.port, err)
        return false
}
```

cmd/traffic/cmd/agent/tls/portconfig.go
thallgren: Good catch. I'll add the proposed `SetDeadline` call.

bgruszka: 👍

cmd/traffic/cmd/agent/tls/portconfig.go
bgruszka: 👍

PR #3986, opened 2025-10-18
feat: add port information to workload list output

bgruszka:
# Add Port Information to Workload List Output

## Summary
This PR adds port information to the `telepresence list` command output, allowing developers to see which ports are available on interceptable workloads without needing to use `kubectl`. This is particularly useful for VS Code extensions and other tooling that need to display port information to users.

## Problem Statement
Previously, when running `telepresence list`, users could see workload names, namespaces, and intercept status, but had no visibility into which ports were exposed by each workload. To discover ports, users had to run separate `kubectl` commands, which:
- Required additional permissions
- Added friction to the workflow
- Made it difficult for IDE extensions to provide port information

## Solution
The feature now includes port information in the list output by:
1. Using the existing sidecar generation logic to discover service-to-container port mappings
2. Extracting port information from the generated sidecar configuration
3. Propagating port data through the RPC layer to the client
4. Displaying ports in multiple formats

### Protocol Definitions
- **rpc/manager/manager.proto**: Added `WorkloadPortInfo` message with fields:
  - `container_port_name` - Symbolic name of the container port (mandatory field)
  - `container_port` - Numeric container port number (mandatory, always present)
  - `protocol` - Protocol used by the port (TCP or UDP)
  - `service_port_name` - Symbolic name of the service port (may be empty for headless services or replacements)
  - `service_port` - Numeric service port number (may be 0 for headless services or replacements)
- **rpc/connector/connector.proto**: Updated to use `manager.WorkloadPortInfo` directly (no duplicate definition)

### Traffic Manager
- **cmd/traffic/cmd/manager/state/workload_info_watcher.go**:
  - Added `extractPortsFromSidecar()` - Extracts **both container and service port** information from sidecar configuration
  - Added `extractPortsForWorkload()` - Generates sidecar and extracts ports using service discovery
    - **Performance optimization**: Reuses existing sidecar config when traffic-agent is already installed, avoiding unnecessary regeneration
  - Modified `rpcWorkload()` - Includes port information in WorkloadInfo
  - Uses `types.PortAndProto` for protocol-aware deduplication (handles TCP/UDP on same port)
  - Populates all five WorkloadPortInfo fields: container port name/number, service port name/number, and protocol
  - Fixed import conflicts: aliased k8s types as `k8stypes` to avoid conflict with `types` package

### Client
- **pkg/client/userd/trafficmgr/session.go**:
  - Simplified to use `manager.WorkloadPortInfo` directly (no type conversion needed)
  - Propagates port information through RPC layer

### Tests
- **cmd/traffic/cmd/manager/state/workload_info_watcher_test.go**: Created with 13 comprehensive unit tests
  - Nil/empty sidecar handling
  - Single and multiple ports
  - Multiple containers
  - Port deduplication (same port/protocol)
  - TCP/UDP protocol support
  - Port name handling
  - Protocol string conversion
  - Complex scenarios with mixed configurations
  - Service port handling - Tests with both container and service port information
  - Headless service handling - Tests for workloads without service ports (container ports only)

- **integration_test/list_watch_test.go**: Added 4 integration tests
  - `Test_ListWithPorts()` - Verifies ports appear in JSON output with required fields
  - `Test_ListWithPortsAndNamedPorts()` - Tests named port handling
  - `Test_ListWithServicePorts()` - Verifies both container and service port fields are populated for normal services
  - `Test_ListWithHeadlessServicePorts()` - Verifies container ports are present for headless services

### Documentation
- **CHANGELOG.yml**: Added v2.26.0 feature entry describing the new port information feature

## Breaking Changes
None. This is a purely additive feature.

## Migration Guide
No migration needed. The feature is automatically available when using `telepresence list`.

## Checklist


cmd/traffic/cmd/manager/state/workload_info_watcher.go
```diff
@@ -171,7 +174,60 @@ func rpcWorkloadState(s workload.State) (state rpc.WorkloadInfo_State) {
...
+
+	// Extract ports from all containers in the sidecar
+	for _, container := range sc.Containers {
+		for _, intercept := range container.Intercepts {
+			// Create a unique key using port number and protocol
+			// This allows the same port with different protocols (TCP/UDP)
+			portKey := types.PortAndProto{
+				Port:  intercept.ServicePort,
+				Proto: intercept.Protocol,
+			}
```
thallgren, line 194: This will only work when there actually is a service port. Headless workloads won't have this. I suggest that you use the container port and proto as key instead. It's guaranteed to be present.

cmd/traffic/cmd/manager/state/workload_info_watcher.go
```diff
@@ -171,7 +174,60 @@ func rpcWorkloadState(s workload.State) (state rpc.WorkloadInfo_State) {
...
+
+			if !seenPorts[portKey] {
+				seenPorts[portKey] = true
+				protocol := intercept.Protocol.String()
+				ports = append(ports, &rpc.WorkloadPortInfo{
+					Name:     intercept.ServicePortName,
+					Port:     int32(intercept.ServicePort),
```
thallgren: While this is OK for most intercepts, it doesn't fully cover all use-cases. A headless deployment will only have container ports (and they will then be found by the intercept). A `replace` doesn't really care about service ports. It will always target container ports (all of them by default).

I suggest that you add information about the mandatory container port.

cmd/traffic/cmd/manager/state/workload_info_watcher.go
```diff
@@ -171,7 +174,60 @@ func rpcWorkloadState(s workload.State) (state rpc.WorkloadInfo_State) {
...
+					Protocol: protocol,
+				})
+			}
+		}
+	}
+
+	return ports
+}
+
+func (wf *workloadInfoWatcher) extractPortsForWorkload(ctx context.Context, wl k8sapi.Workload) []*rpc.WorkloadPortInfo {
```
thallgren: This function works, but it wastes unnecessary resources when a traffic-agent is already present. I suggest adding a check for that. Something like:

```go
  	m := mutator.GetMap(ctx)
  	sc := m.Get(wl.GetName(), wl.GetNamespace())
  	if sc == nil {
```

thallgren (changes requested):
Looks like a useful addition.

cmd/traffic/cmd/manager/state/workload_info_watcher.go
bgruszka: 👍 I added a check

cmd/traffic/cmd/manager/state/workload_info_watcher.go
bgruszka: I changed it to use container port and proto as key

cmd/traffic/cmd/manager/state/workload_info_watcher.go
bgruszka: In the list info we will have container port name, container port, service port name and service port

cmd/traffic/cmd/manager/state/workload_info_watcher.go
```diff
@@ -171,7 +174,68 @@ func rpcWorkloadState(s workload.State) (state rpc.WorkloadInfo_State) {
...
+					ServicePort:       int32(intercept.ServicePort),
+				})
+			}
+		}
+	}
+
+	return ports
+}
+
+func (wf *workloadInfoWatcher) extractPortsForWorkload(ctx context.Context, wl k8sapi.Workload) []*rpc.WorkloadPortInfo {
```
thallgren: Having the `*workloadInfoWatcher` as the receiver here seems redundant. It's not used.

rpc/connector/connector.proto
```diff
@@ -336,6 +336,24 @@ message WatchWorkloadsRequest {
...
+  // Protocol used by the port (TCP or UDP)
+  string protocol = 3;
+
+  // Symbolic name of the service port (may be empty for headless services or replacements)
+  string service_port_name = 4;
+
+  // Numeric service port number (may be 0 for headless services or replacements)
+  int32 service_port = 5;
+}
+
```
thallgren: There's no need to repeat this declaration. Just import the one from manager like we do with several other structs.

thallgren (changes requested):
LGTM. Just some minor nits.

cmd/traffic/cmd/manager/state/workload_info_watcher.go
bgruszka: I've converted it to a standalone function

rpc/connector/connector.proto
bgruszka: I've removed the duplicate definition and use the manager's version

thallgren (approved):

PR #3995, opened 2025-11-03
Customizable pats for config and logging

thallgren:
This pull request introduces several new features, improvements, and refactors related to Telepresence daemon management, workload watching, and build configuration. The most significant changes are the addition of customizable daemon config- and log file paths and improved naming for daemon processes. These improvements lay will make it easier to run the daemons as proper services.

**Daemon Management and Logging:**

* Added support for customizable daemon log file paths via the `--logfile` command-line flag, allowing logs to be redirected to custom locations or to stdout/stderr. Two new log-level configuration entries for `cli` and `kubeAuthDaemon` were introduced, expanding log-level controls.
* Improved daemon process names for clarity: `connector-foreground` is now `userd`, `daemon-foreground` is now `rootd`, and `kubeauth-foreground` is now `kubeauthd`. Documentation and usage examples were updated accordingly. 

**Workload Watcher Refactor:**

* Refactored workload watcher logic by moving `pkg/workload/watcher.go` to `cmd/traffic/cmd/manager/state/watcher.go` because only the traffic-manager uses this watcher now.

**gRPC Error Refactor.**
The connect, intercept, and some other gRPC calls made to the user daemon now propagate errors to the caller so that all calls either return a proper result and nil, or no result at all and an error. Earlier, errors were often propagated in the result itself, leading to overly complex logic.

**Build Configuration:**

* Improved build configuration flexibility by making `EXTERNAL_FUSEFTP`, `LINKED_FUSEFTP`, and `CGO_ENABLED` variables overridable, and updated the `GOLANGCI_VERSION` to `v2.6.0`. Also, adjusted environment setup for cross-compilation on ARM64 Linux.

**Bugfixes and Protocol Updates:**

* Fixed a bug in the network endpoint creation logic by correcting the capitalization of `AddrIpv4` and `AddrIpv6` fields.
* Added a restriction to prevent using the `--docker` flag when `intercept.useFtp=true` is enabled, with a clear error message explaining FTP protocol limitations in this scenario.## Description


pkg/client/logging/initcontext.go
```diff
@@ -45,17 +48,20 @@ func InitContext(ctx context.Context, name string, strategy RotationStrategy, ca
 				maxFiles = uint16(mx)
 			}
 		}
-		rf, err := OpenRotatingFile(ctx, filepath.Join(filelocation.AppUserLogDir(ctx), name+".log"), "20060102T150405", true, 0o600, strategy, maxFiles)
+		rf, err := OpenRotatingFile(ctx, logFile, "20060102T150405", true, 0o600, strategy, maxFiles)
```
bgruszka, line 58: The new `--logfile` flag accepts arbitrary file paths without any validation or sanitization. This creates a path traversal vulnerability that could allow users to write logs to arbitrary filesystem locations.

Maybe we can use something like this:

```go
func ValidateLogFilePath(logFile string) (string, error) {
    // Special cases for stdout/stderr don't need validation
    if logFile == "stdout" || logFile == "stderr" || logFile == "-" {
        return logFile, nil
    }

    // Clean the path to resolve any .. or . components
    cleanPath := filepath.Clean(logFile)

    // Convert to absolute path
    absPath, err := filepath.Abs(cleanPath)
    if err != nil {
        return "", fmt.Errorf("invalid log file path: %w", err)
    }

    // Verify parent directory exists or can be created
    dir := filepath.Dir(absPath)
    if err := os.MkdirAll(dir, 0o700); err != nil {
        return "", fmt.Errorf("cannot create log directory %q: %w", dir, err)
    }

    // Optional: Verify the path is within expected directories
    // (e.g., within user's home directory or /var/log/telepresence)

    return absPath, nil
}
```

And then:

```go
// Validate path before using it
        validatedPath, err := ValidateLogFilePath(logFile)
        if err != nil {
            return ctx, fmt.Errorf("invalid log file path: %w", err)
        }

        rf, err := OpenRotatingFile(ctx, validatedPath, "20060102T150405", true, 0o600, strategy, maxFiles)
```

bgruszka (approved):
Great work on this PR :) This is a really solid refactoring that significantly improves daemon management and logging capabilities :)

##### One Suggestion: Path Validation for `--logfile`

I noticed the new `--logfile` flag accepts user-provided paths directly without validation. While the daemons run with user permissions, it would be safer to add some path sanitization to prevent potential issues.

pkg/client/logging/initcontext.go
thallgren: The path is validated in the CLI (lines 25 - 49 in cli/connect/daemon.go), but I'll add this too because it's necessary when the daemon is started as a service.

PR #3996, opened 2025-11-06
Add startupProbe support and refactor manager startup logic

thallgren:
This PR introduces a `startupProbe` configuration for the Traffic Manager to ensure better service readiness detection and reorganizes the manager boot sequence by delaying the HTTP server start until all configurations are initialized.

Some general cleanup was also made. See commit comments for details.


cmd/traffic/cmd/manager/cluster/nodewatcher.go
```diff
@@ -249,8 +219,62 @@ nextN:
 	return added, oldSubnets
 }
 
-func nodeSubnets(ctx context.Context, node *corev1.Node) []netip.Prefix {
+// compareNodeConditions compares two NodeCondition objects and returns an integer indicating their
+// relative order based no transition time (last is highest), type (Ready is higher), and status (true is higher).
```
bgruszka: should be "on transition time" ?

cmd/traffic/cmd/manager/cluster/nodewatcher.go
```diff
@@ -249,8 +219,62 @@ nextN:
...
+	var lastCond *core.NodeCondition
+	conds := node.Status.Conditions
+	for i := range conds {
+		cond := &conds[i]
+		if compareNodeConditions(cond, lastCond) > 0 {
+			lastCond = cond
+		}
+	}
+	return lastCond != nil && lastCond.Type == core.NodeReady && lastCond.Status == core.ConditionTrue
+}
```
bgruszka, line 252: The function finds the condition with the most recent LastTransitionTime (of any type), then checks if that specific condition is NodeReady=True.

Kubernetes nodes have multiple independent conditions (Ready, DiskPressure, MemoryPressure, etc.) and a node can be NodeReady=True and have other conditions like DiskPressure=True and the most recent condition change might NOT be the NodeReady condition.

For example:

Ready Node with more recent DiskPressure
Conditions:
  - NodeReady=true at T1 (5 minutes ago)
  - DiskPressure=true at T2 (2 minutes ago, more recent)

Expected: true (node is ready despite disk pressure)
Actual:   false
RESULT:  fail

1. Multiple independent conditions exist simultaneously:
    - NodeReady=True + DiskPressure=True is valid
    - NodeReady=True + MemoryPressure=True is valid
    - These conditions are independent
2. NodeReady is the definitive condition:
  - Kubernetes uses the NodeReady condition specifically to determine if a node accepts pods
  - Other conditions (DiskPressure, MemoryPressure) are warnings, not readiness indicators

Am I right or I missed something?

cmd/traffic/cmd/manager/cluster/nodewatcher.go
```diff
@@ -60,92 +84,53 @@ func newNodeWatcher(ctx context.Context, lister licorev1.NodeLister, informer ca
...
+	var lastSent subnet.Set
+	sendUpdate := func() {
+		w.lock.Lock()
+		doSend := !w.subnets.Equals(lastSent)
+		if doSend {
+			lastSent = w.subnets.Clone()
+		}
+		w.lock.Unlock()
+		if doSend {
+			dlog.Debugf(ctx, "nodeWatcher calling updateSubnets with %v", w.subnets)
```
bgruszka: in line 94 is release lock and we do a read `w.subnets` here. So we read it without lock. Is it ok?

Should we use lastSent here instead?

```go
dlog.Debugf(ctx, "nodeWatcher calling updateSubnets with %v", lastSent)
```

bgruszka (commented):
Just few comments / questions :)

thallgren:
Thanks for reviewing and (as always) finding issues. I think I've addressed them, so please have a look.

bgruszka (approved):
Looks good to me :)

PR #4027, opened 2026-01-08
Allow cluster admin revoke other intercepts

### LINKED ISSUE #4009: Able to forcefully kill an intercept from other
When we have a big team, someone may mistakenly intercept service and then go away, causing an impact on all other users.
Based on [this comment](https://github.com/telepresenceio/telepresence/issues/3151#issuecomment-1537064487), we can reset the state by restarting the traffic manager, but it will cause trouble on other legitimate intercepts.

Provide an api for manager to list and kill an intercept by id.
We can protect it by basic auth via environment variable, so only cluster admin can access to the api.

```
OSS Client         : v2.25.1
OSS Root Daemon    : v2.25.1
OSS User Daemon    : v2.25.1
OSS Traffic Manager: v2.25.1
Traffic Agent      : ghcr.io/telepresenceio/tel2:2.25.1
Kubernetes: 1.33
```

**Additional context**
None

yuyuvn:
## Description

Fix https://github.com/telepresenceio/telepresence/issues/4009

## Checklist

### DIFF
--- a/cmd/traffic/cmd/manager/service.go
+++ b/cmd/traffic/cmd/manager/service.go
@@ -889,9 +889,9 @@ func (s *service) RemoveIntercept(ctx context.Context, riReq *rpc.RemoveIntercep
 // by the AGENT_K8S_ADMIN_GROUPS environment variable (default: system:masters).
 func (s *service) RevokeIntercept(ctx context.Context, riReq *rpc.RevokeInterceptRequest) (*empty.Empty, error) {
 	// Verify the authentication token
-	tokenResult, error := k8sapi.VerifyToken(ctx, riReq.Token)
-	if error != nil {
-		dlog.Warnf(ctx, "Authentication failed for RevokeIntercept: %v", error)
+	tokenResult, err := k8sapi.VerifyToken(ctx, riReq.Token)
+	if err != nil {
+		dlog.Warnf(ctx, "Authentication failed for RevokeIntercept: %v", err)
 		return nil, status.Errorf(codes.PermissionDenied, "authentication failed")
 	}
 
--- a/cmd/traffic/cmd/manager/service.go
+++ b/cmd/traffic/cmd/manager/service.go
@@ -888,19 +888,19 @@ func (s *service) RemoveIntercept(ctx context.Context, riReq *rpc.RemoveIntercep
 // Requires authentication via token and membership in system:telepresence-admin group.
 func (s *service) RevokeIntercept(ctx context.Context, riReq *rpc.RevokeInterceptRequest) (*empty.Empty, error) {
 	// Verify the authentication token
-	tokenResult := k8sapi.VerifyToken(ctx, riReq.Token)
-	if tokenResult.Error != nil || !tokenResult.Authenticated {
-		dlog.Warnf(ctx, "Authentication failed for RevokeIntercept: %v", tokenResult.Error)
+	tokenResult, error := k8sapi.VerifyToken(ctx, riReq.Token)
+	if error != nil {
+		dlog.Warnf(ctx, "Authentication failed for RevokeIntercept: %v", error)
 		return nil, status.Errorf(codes.PermissionDenied, "authentication failed")
 	}
 
 	// Check if user belongs to system:telepresence-admin group
-	if !(k8sapi.IsMemberOfGroup(tokenResult.Groups, "telepresence:admin") || k8sapi.IsMemberOfGroup(tokenResult.Groups, "system:masters")) {
+	if !(tokenResult.IsMemberOfGroup("telepresence:admin") || tokenResult.IsMemberOfGroup("system:masters")) {
 		dlog.Warnf(ctx, "User %s is not a member of telepresence:admin or system:masters group", tokenResult.Username)
 		return nil, status.Errorf(codes.PermissionDenied, "user must be a member of telepresence:admin or system:masters group")
 	}
 
-	dlog.Infof(ctx, "User %s authorized to revoke intercepts", tokenResult.Username)
+	dlog.Infof(ctx, "Authorized to revoke intercepts")
 
 	interceptID := riReq.InterceptId
 	dlog.Debugf(ctx, "Revoking intercept ID %s", interceptID)
--- a/cmd/traffic/cmd/manager/service.go
+++ b/cmd/traffic/cmd/manager/service.go
@@ -883,6 +883,47 @@ func (s *service) RemoveIntercept(ctx context.Context, riReq *rpc.RemoveIntercep
 	return &empty.Empty{}, nil
 }
 
+// RevokeIntercept allows the manager to revoke any client's intercept by intercept ID.
+// This is an administrative operation that can revoke intercepts for any client.
+// Requires authentication via token and membership in system:telepresence-admin group.
+func (s *service) RevokeIntercept(ctx context.Context, riReq *rpc.RevokeInterceptRequest) (*empty.Empty, error) {
+	// Verify the authentication token
+	tokenResult := k8sapi.VerifyToken(ctx, riReq.Token)
+	if tokenResult.Error != nil || !tokenResult.Authenticated {
+		dlog.Warnf(ctx, "Authentication failed for RevokeIntercept: %v", tokenResult.Error)
+		return nil, status.Errorf(codes.PermissionDenied, "authentication failed")
+	}
+
+	// Check if user belongs to system:telepresence-admin group
+	if !(k8sapi.IsMemberOfGroup(tokenResult.Groups, "telepresence:admin") || k8sapi.IsMemberOfGroup(tokenResult.Groups, "system:masters")) {
+		dlog.Warnf(ctx, "User %s is not a member of telepresence:admin or system:masters group", tokenResult.Username)
+		return nil, status.Errorf(codes.PermissionDenied, "user must be a member of telepresence:admin or system:masters group")
+	}
+
+	dlog.Infof(ctx, "User %s authorized to revoke intercepts", tokenResult.Username)
+
+	interceptID := riReq.InterceptId
+	dlog.Debugf(ctx, "Revoking intercept ID %s", interceptID)
+
+	// Get the intercept to verify it exists and to update metrics
+	intercept, ok := s.state.GetIntercept(interceptID)
+	if !ok {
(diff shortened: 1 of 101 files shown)


thallgren:
I don't see anything here preventing a client from revoking an intercept. You'll need to provide a test that proves that only an admin account can revoke an intercept.

yuyuvn:
sorry, after I commit DCO, the change related to token review was disapeared...

I added it and add unit test, too

pkg/k8sapi/tokenreview.go
```diff
@@ -0,0 +1,72 @@
...
+type VerifyTokenResult struct {
+	Authenticated bool
+	Username      string
+	Groups        []string
+	Error         error
+}
+
+// VerifyToken verifies a bearer token using Kubernetes TokenReview API
+// and returns information about the authenticated user including their groups.
+func VerifyToken(ctx context.Context, token string) *VerifyTokenResult {
```
thallgren: I'm not a fan of returning an error as part of the result. I think it's more in line with other API's if you remove the `Error` field from the `VerifyTokenResult` and instead return `(VerifyTokenResult, error)`.

Maybe also consider renaming the function and the struct. The function doesn't just verify a token, it actually obtains user information from that token. So perhaps something like:

`func UserInfoForToken(ctx context.Context, token string) (UserInfo, error)` ?

pkg/k8sapi/tokenreview.go
```diff
@@ -0,0 +1,72 @@
...
+	return &VerifyTokenResult{
+		Authenticated: true,
+		Username:      result.Status.User.Username,
+		Groups:        result.Status.User.Groups,
+		Error:         nil,
+	}
+}
+
+// IsMemberOfGroup checks if a user belongs to a specific group.
+func IsMemberOfGroup(groups []string, targetGroup string) bool {
```
thallgren: Maybe this could be `func (u *UserInfo) IsMemberOfGroup(group string) bool` ?

pkg/k8sapi/tokenreview.go
```diff
@@ -0,0 +1,72 @@
...
+	for _, group := range groups {
+		if group == targetGroup {
+			return true
+		}
+	}
+	return false
```
thallgren:
```suggestion
	return slices.Contains(groups, targetGroup)
```

cmd/traffic/cmd/manager/service.go
```diff
@@ -882,6 +882,47 @@ func (s *service) RemoveIntercept(ctx context.Context, riReq *rpc.RemoveIntercep
...
+		return nil, status.Errorf(codes.PermissionDenied, "authentication failed")
+	}
+
+	// Check if user belongs to system:telepresence-admin group
+	if !(k8sapi.IsMemberOfGroup(tokenResult.Groups, "telepresence:admin") || k8sapi.IsMemberOfGroup(tokenResult.Groups, "system:masters")) {
+		dlog.Warnf(ctx, "User %s is not a member of telepresence:admin or system:masters group", tokenResult.Username)
+		return nil, status.Errorf(codes.PermissionDenied, "user must be a member of telepresence:admin or system:masters group")
+	}
+
+	dlog.Infof(ctx, "User %s authorized to revoke intercepts", tokenResult.Username)
```
thallgren: Please don't include the user name in the log. It might be considered sensitive information and prevent logs from being shared.

thallgren (commented):
This code looks great overall. I made some improvement suggestions.

I'm missing one major part here though. The client side calling this new API endpoint. The PR won't be much use to an end user without that. So please also provide the new `telepresence revoke` command (or similar), where the user passes a token (preferably from stdin).

It would also be very valuable if you could add a section to the docs explaining how to add a user with the correct group affinity to the cluster and obtain the necessary token.

pkg/k8sapi/tokenreview.go
yuyuvn: resolved in https://github.com/telepresenceio/telepresence/pull/4027/commits/80274bd49c278aceda152cc42f47438b70d32e54

pkg/k8sapi/tokenreview.go
yuyuvn: resolved in https://github.com/telepresenceio/telepresence/pull/4027/commits/80274bd49c278aceda152cc42f47438b70d32e54

pkg/k8sapi/tokenreview.go
yuyuvn: resolved in https://github.com/telepresenceio/telepresence/pull/4027/commits/80274bd49c278aceda152cc42f47438b70d32e54

cmd/traffic/cmd/manager/service.go
yuyuvn: resolved in https://github.com/telepresenceio/telepresence/pull/4027/commits/80274bd49c278aceda152cc42f47438b70d32e54

yuyuvn:
I refactored to use `AGENT_K8S_ADMIN_GROUPS` env to control allowed admin groups because it's hard to test in kind cluster for harded code group (with a dynamic group, it can also allow service account token to be used)

Can quickly test by creating a service account in ambassador namespace and add this env to manager deployment
```
          - name: AGENT_K8S_ADMIN_GROUPS
            value: "system:masters system:serviceaccounts:ambassador"
```
with token created from
```
kubectl create token test-admin -n ambassador --duration=1h
```

it also works with OIDC token from `aws eks get-token`

yuyuvn:
Fixed lint 🙏

pkg/client/cli/cmd/revoke.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/telepresenceio/telepresence/v2/pkg/errcat"
+	"github.com/telepresenceio/telepresence/v2/pkg/grpc"
+)
+
+func revokeCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:  "revoke <token> <intercept_id>",
+		Args: cobra.ExactArgs(2),
+
+		Short: "Revoke an intercept by intercept ID. The intercept ID must be in the format <session_id>:<intercept_name>",
```
thallgren, line 23: I see two problems with exposing and using the intercept ID like this.
1. This ID is supposed to be opaque. The fact that it's composed by the client's session-id and a name isn't necessarily something that we want to expose as API and then stick with it.
2. How is a user supposed to find another users session_id? AFAIK, the only place it's visible today is if you run the traffic-manager with log level debug or trace and then scan the logs.

thallgren (commented):
A general question. Can we control this somehow using RBAC? Telepresence is often used in situations where it's limited to a specific set of namespaces. It would be great if the permission to revoke an intercept was associated with a role/rolebinding or clusterrole/clusterrolebinding. Which one that is in use depends on scope determined by the namespace selector. This way, any user could be permitted to do this by only associating this binding with an existing user account.

yuyuvn:
> A general question. Can we control this somehow using RBAC? Telepresence is often used in situations where it's limited to a specific set of namespaces.

I can try if SubjectAccessReview  can work without define custom resource for intercept.
it would be more k8s native if intercept state is stored in k8s storage with crd for intercept, but I feel like it would be a big refactor. that also resolve problem how to get list other's intercept

pkg/client/cli/cmd/revoke.go
yuyuvn: I can fetch it from sidecar logs (that's how I plan to use in my environment)

thallgren: I think that depends on what type of conflict you're encountering. If the traffic-manager detects that you have a `--http-header` based conflict, then you'll get an error similar to:
```
"intercept in error state AGENT_ERROR: Conflicts with the currently-served intercept "ae29c3db-0e14-4517-a538-7953e19fd155:echo-easy": header filters form a subset
```

but if the intercept is global, the user will instead see:
```
container port 8080 is already intercepted by root@3029256df965, intercept echo-easy
```

I think this is unfortunate. It would be far better, and also consistent with your approach, if both errors displayed the intercept ID. It would also remove the need to look into logs to find the ID. Do you think it would be possible for you to fix so that these errors both report the intercept ID?

thallgren:
When I looked into this more deeply, I realized that passing a certificate validated in this way won’t work in all situations. Perhaps authorization isn’t the right approach here at all.

## The problem

An intercept is similar to booking a seat on a plane without having paid for it yet. You can hold that reservation for a while, but unless you complete the purchase, it is eventually canceled and the seat becomes available to someone else.

The issue we are facing is that the time before an intercept is automatically canceled can be very long. Telepresence allows you to close the lid on your computer and continue working the next day with your intercept still intact. The time-to-live is controlled by `grpc.connectionTTL` and defaults to 24 hours. During this period, a client is considered “alive,” meaning any intercepts it has started remain active and block others from creating conflicting intercepts.

## A possible solution

What if we introduced a second time-to-live setting that controls how long a currently unreachable client can block others from taking over an intercept? Let’s call it `grpc.activeTTL`.

This value would be significantly shorter, perhaps defaulting to around 10 minutes. If a client hasn’t pinged the system within that time, any intercepts it holds would still exist, but Telepresence would allow other clients to terminate them. This would remove the need for special authorization and would also prevent someone who is actively working on an intercept from being unexpectedly interrupted by another client terminating it.

Thoughts?

yuyuvn:
> This value would be significantly shorter, perhaps defaulting to around 10 minutes. If a client hasn’t pinged the system within that time, any intercepts it holds would still exist, but Telepresence would allow other clients to terminate them. This would remove the need for special authorization and would also prevent someone who is actively working on an intercept from being unexpectedly interrupted by another client terminating it.

I see, this apporach look good to me. It would allow user resolve problem by themself instead depend on admin, which is a good point.

In that case, I would like to have a way to detect who is actively intercepting (hostname or IP). (just in case some one is testing  intercept then forget about it and do other works, blocking other member)

thallgren:
A solution involving a revoke command is needed anyway, because a user might have an ongoing intercept that makes the workstation look active, even though the users isn't using it. A local intercept handler that periodically sends alive polls another service would have that effect.

I created PR #4033 that includes your commit with the `telepresence revoke` command. But instead of using gRPC to call the traffic-manager, it uses a new "admin command" feature which essentially adds the command to a configmap that the traffic-manager listens to. Hence, a user must have RBAC set up to get and update this configmap, and that's how Telepresence will differentiate an "admin" user from a "regular" user.

thallgren:
Superseded by #4033

PR #4058, opened 2026-02-14
Fix h2c prior knowledge by setting HTTP1=false on transport

### LINKED ISSUE #4056: Header-based intercept breaks non-matching gRPC traffic — reverse proxy uses HTTP/1.1 instead of h2c
Description                                                                                                            
                                                                                                                         
When using a header-based intercept on a gRPC service, all non-matching requests silently hang and timeout. Matching requests work correctly (with --plaintext). Global intercepts (no header matching) work fine.                          
                                                                                                                         
Environment

  - Telepresence CLI: v2.26.1
  - Traffic Manager: v2.26.1
  - Kubernetes: EKS
  - Service mesh: Istio ambient mode
  - Go version used by telepresence: 1.24+
  - Service: gRPC on port 9090, appProtocol: grpc, no TLS (h2c)

Steps to Reproduce

  1. Have a gRPC service running with appProtocol: grpc and no TLS
  2. Create a header-based intercept:
  telepresence intercept my-grpc-svc --port 9090 --http-header 'x-test=123' --plaintext
  3. Send a request with the matching header → works, routed to local
  4. Send a request without the matching header → hangs forever, never reaches the app

Expected Behavior

Non-matching requests should be forwarded to the app via the reverse proxy, using h2c (HTTP/2 cleartext) since the app is a gRPC server.

Actual Behavior

Non-matching requests hang silently. The reverse proxy connects to the app using HTTP/1.1 instead of h2c. The gRPC server responds with an HTTP/2 SETTINGS frame, which the HTTP/1.1 transport can't parse:

```
  {"error":"net/http: HTTP/1.x transport connection broken: malformed HTTP response
  \"\\x00\\x00\\x06\\x04\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x05\\x00\\x00@\\x00\""}
```
The \x00\x00\x06\x04... bytes are the HTTP/2 SETTINGS frame. The agent returns a 502 internally but the client just
  sees a timeout.

How to Reproduce the Error Directly

With the intercept active, send a non-matching request directly to the agent's proxy port from inside the pod:
```
  kubectl debug -n <ns> <pod> -it --image=curlimages/curl --target=traffic-agent -- \
    curl --http2-prior-knowledge -s \
    -H "content-type: application/grpc" \
    -X POST http://localhost:9900/<grpc-method>
```
This returns the 502 with the HTTP/1.x transport connection broken error above.

For comparison, the app itself accepts h2c just fine:
```
  kubectl debug -n <ns> <pod> -it --image=curlimages/curl --target=traffic-agent -- \
    curl --http2-prior-knowledge -v -s -o /dev/null http://localhost:9090/<grpc-method>
  -- HTTP/2 connection established, app responds normally
````
And rejects HTTP/1.1:

  kubectl debug -n <ns> <pod> -it --image=curlimages/curl --target=traffic-agent -- \
    curl -v -s -o /dev/null http://localhost:9090/<grpc-method>
  -- "Received HTTP/0.9 when not allowed" — connection closed

Root Cause

  The bug is in https://github.com/telepresenceio/telepresence/blob/v2.26.1/cmd/traffic/cmd/agent/fwd/http.go.

  1. protocols() always enables HTTP/1.1 alongside h2c
```
  func (f *tcp) protocols(ctx context.Context, plainText bool) *http.Protocols {
      pr := new(http.Protocols)
      pr.SetHTTP1(true)              // ← Always enabled
      if tm.UseHTTP2(ctx, tp) {
          if !plainText && tm.UseTLS(ctx, tp) {
              pr.SetHTTP2(true)
          } else {
              pr.SetUnencryptedHTTP2(true) // ← Also enabled for gRPC/h2c targets
          }
      }
      return pr
  }
```
  2. Go 1.24's http.Transport requires HTTP1=false for h2c

  From https://github.com/golang/go/blob/go1.24.0/src/net/http/transport.go:
```
  unencryptedHTTP2 := pconn.tlsState == nil &&
      t.Protocols != nil &&
      t.Protocols.UnencryptedHTTP2() &&
      !t.Protocols.HTTP1()           // ← HTTP1 must be FALSE for h2c to activate
```
  Since HTTP1 is always true, the h2c path is never taken. The transport falls back to HTTP/1.1.

  3. The fix exists for the intercept path but not the default handler

  configureUpstreamTransport (matching requests → local) correctly disables HTTP/1.1:
```
  if plaintext {
      if requestProto == 2 && trn.Protocols.UnencryptedHTTP2() {
          trn.Protocols.SetHTTP1(false)  // ← Forces h2c!
      }
      return trn
  }
```
  But acceptHTTPLoop (non-matching requests → app) never does this:
```
  defaultHandler.Transport = f.configureTransport(ctx, false) // HTTP1 stays true
```
Suggested Fix:

In acceptHTTPLoop, disable HTTP/1.1 on the default handler transport when the target is an h2c server:

```
  trn := f.configureTransport(ctx, false)
  if !f.targetUsesTLS(ctx) && trn.Protocols != nil && trn.Protocols.UnencryptedHTTP2() {
      trn.Protocols.SetHTTP1(false)
  }
  defaultHandler.Transport = trn
```
Or use a per-request RoundTripper that checks req.ProtoMajor before choosing the transport, matching the logic in configureUpstreamTransport.

Agent Logs

The agent correctly detects no TLS and enables h2c — but the transport ignores it:

```
  Probing port 9090 for TLS and HTTP/2 support
  Port 9090 does not support TLS: tls: first record does not look like a TLS handshake
  Starting HTTP intercept forwarder on [::]:9900
```

thallgren:
## Summary
- Since Go 1.24, having both HTTP1 and UnencryptedHTTP2 enabled on an HTTP transport causes Go to default to HTTP/1.1 instead of using h2c prior knowledge
- Move the `SetHTTP1(false)` logic into `configureTransport()` so all callers (default handler and intercepted traffic) get the fix automatically
- Remove the now-redundant conditional block from `configureUpstreamTransport()`

Fixes #4056

## Test plan
- [x] `go build ./cmd/traffic/...` compiles successfully
- [x] `go test ./cmd/traffic/cmd/agent/fwd/...` unit tests pass
- [ ] Verify h2c intercepts work end-to-end against a gRPC service using cleartext HTTP/2

### DIFF
--- a/CHANGELOG.yml
+++ b/CHANGELOG.yml
@@ -77,6 +77,14 @@ items:
           loglevel failed with "root daemon is embedded". The user daemon now correctly delegates to the in-process
           root daemon session instead of returning an error.
         docs: https://github.com/telepresenceio/telepresence/issues/4049
+      - type: bugfix
+        title: Fix h2c (HTTP/2 cleartext) prior knowledge not working on transport
+        body: >-
+          Since Go 1.24, having both HTTP1 and UnencryptedHTTP2 enabled on an HTTP transport causes Go to
+          default to HTTP/1.1 instead of using h2c prior knowledge. The transport now explicitly disables
+          HTTP1 when UnencryptedHTTP2 is enabled, fixing h2c communication for both the default forwarding
+          handler and intercepted traffic.
+        docs: https://github.com/telepresenceio/telepresence/issues/4056
   - version: 2.26.1
     date: 2026-01-26
     notes:
--- a/build-aux/main.mk
+++ b/build-aux/main.mk
@@ -31,7 +31,7 @@ bindir ?= $(or $(shell go env GOBIN),$(shell go env GOPATH|cut -d: -f1)/bin)
 # https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/syntax.md.
 export DOCKER_BUILDKIT := 1
 
-GOLANGCI_VERSION:=v2.8.0
+GOLANGCI_VERSION:=v2.9.0
 
 .PHONY: FORCE
 FORCE:
--- a/cmd/traffic/cmd/agent/fwd/http.go
+++ b/cmd/traffic/cmd/agent/fwd/http.go
@@ -40,6 +40,9 @@ func (f *tcp) protocols(ctx context.Context, plainText bool) *http.Protocols {
 func (f *tcp) configureTransport(ctx context.Context, plainText bool) *http.Transport {
 	trn := http.DefaultTransport.(*http.Transport).Clone()
 	trn.Protocols = f.protocols(ctx, plainText)
+	if trn.Protocols.UnencryptedHTTP2() {
+		trn.Protocols.SetHTTP1(false)
+	}
 	return trn
 }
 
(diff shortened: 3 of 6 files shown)


thallgren:
Merged via release/v2.26

PR #4061, opened 2026-02-17
Fix race condition in intercept handling when InterceptInfo is updated concurrently

bgruszka:
This PR fixes a race condition in the traffic-agent that can cause requests to hang or be forwarded to the wrong port when multiple pods are running with HPA (Horizontal Pod Autoscaler).

The issue occurs when:
1. `Forward()` obtains an `interceptController` while holding the mutex
2. `Forward()` releases the mutex
3. `reconcile()` updates the `ic.InterceptInfo` pointer (with new `TargetPort`)
4. `Forward()` reads the now-stale `ic.InterceptInfo` → uses wrong `TargetPort`

This manifests as requests "hanging" indefinitely because the traffic-agent tries to forward traffic to an old port (e.g., 443) instead of the updated port (e.g., 3000).

#### `cmd/traffic/cmd/agent/fwd/tcp.go`
- Added `interceptSnapshot` struct to capture controller + InterceptInfo atomically
- Modified `Forward()` to capture `InterceptInfo` while holding the mutex
- Refactored to use consistent snapshot pattern for both intercepts and wiretaps
- Added `interceptConnWithInfo()` function that accepts pre-captured `InterceptInfo`
- Removed unused `interceptConn()` function
- Ensures consistent `TargetPort` value throughout connection handling

#### `cmd/traffic/cmd/agent/fwd/http.go`
- Uses shared `interceptSnapshot` struct for consistency
- Modified `handleHTTPRequest()` to create snapshots under mutex protection
- All subsequent `InterceptInfo` accesses use the captured snapshots

#### `cmd/traffic/cmd/agent/fwd/interceptcontroller.go`
- Added documentation warning about the race condition at pointer assignment site
- Guides future maintainers to use mutex-protected captures

### Improvements

1. **Standardized Pattern** - Both TCP and HTTP now use the same `interceptSnapshot` approach for consistency
2. **Removed Dead Code** - Eliminated unused `interceptConn()` function
3. **Comprehensive Testing** - Added two race condition tests:
   - `TestInterceptInfoRaceCondition` - Tests TCP intercept race scenarios
   - `TestHTTPInterceptInfoRaceCondition` - Tests HTTP intercept race scenarios
4. **Documentation** - Added race condition warnings
5. **Race Detector Verified** - All tests pass with `-race` flag

## Root Cause Analysis

The `interceptController` struct embeds `*manager.InterceptInfo` as a pointer. The `reconcile()` function in `interceptcontroller.go` updates this pointer when new intercept information arrives:

```go
if ok {
    ic.InterceptInfo = ii  // Pointer update without synchronization
}
```

However, `Forward()` and `handleHTTPRequest()` were accessing `ic.InterceptInfo` after releasing the mutex, creating a data race where the pointer could be updated mid-operation.

## Checklist

### DIFF
--- a/cmd/traffic/cmd/agent/fwd/http.go
+++ b/cmd/traffic/cmd/agent/fwd/http.go
@@ -113,13 +113,6 @@ func (f *tcp) acceptHTTPLoop(ctx context.Context, listener net.Listener) {
 	}
 }
 
-// interceptSnapshot holds an interceptController with its InterceptInfo captured at a point in time.
-// This prevents race conditions where InterceptInfo pointer could be updated while using it.
-type interceptSnapshot struct {
-	ic   *interceptController
-	info *manager.InterceptInfo
-}
-
 func (f *tcp) handleHTTPRequest(writer http.ResponseWriter, req *http.Request, defaultHandler http.Handler) {
 	// Copy wiretaps and intercepts to avoid holding a lock during request processing.
 	// Also capture InterceptInfo pointers while holding the mutex to avoid race condition
--- a/cmd/traffic/cmd/agent/fwd/http.go
+++ b/cmd/traffic/cmd/agent/fwd/http.go
@@ -113,11 +113,30 @@ func (f *tcp) acceptHTTPLoop(ctx context.Context, listener net.Listener) {
 	}
 }
 
+// interceptSnapshot holds an interceptController with its InterceptInfo captured at a point in time.
+// This prevents race conditions where InterceptInfo pointer could be updated while using it.
+type interceptSnapshot struct {
+	ic   *interceptController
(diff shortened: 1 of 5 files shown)


thallgren (changes requested):
I see the potential race condition, but I don't like the solution. The problem IMO, is that the `interceptController` is mutable. It really shouldn't be. I believe this race could be fixed in interceptcontroller.go, line 49, by simply changing:

```go
		if ok {
			ic.InterceptInfo = ii
		} else {
```

into

```go
		if !ok {
```

That way, the pointer will never get updated and the need for the extra indirection using an `interceptSnapshot` is gone.

Note! I don't think this is the cause of the original problem that your reported in Slack.

PR #4063, opened 2026-02-23
fix: work with custom go versions, like `go1.26.0-X:nodwarf5`

cwrau:
## Description

Only use the first column of the GOVERSION file to determine the Go version to use.
This allows to use custom go versions like `go1.26.0-X:nodwarf5` and `go1.25.1 X:nodwarf5`

## Checklist


thallgren (changes requested):
The substitution yields an empty version string and breaks the build.

cwrau:
> The substitution yields an empty version string and breaks the build.

Ah, missed the case where there is no suffix 😅

thallgren (approved):

PR #4089, opened 2026-03-29
Fix Docker Compose default network subnet conflicts with Kubernetes cluster CIDRs

jssosa10:
## Summary

- When using `telepresence compose`, Docker Compose auto-creates a `<project>_default` network whose CIDR could overlap with Kubernetes cluster CIDRs (e.g., `172.20.0.0/16` on EKS), causing cluster traffic to be misrouted to the local bridge
- Telepresence now configures the default network in the rendered compose file with an explicit non-conflicting RFC 1918 subnet, reusing the existing `FindNonConflictingSubnet` logic from the teleroute package
- Also includes `AlsoProxy` subnets in the compose connection's cluster CIDRs to match the teleroute path behavior

## Changes

- `pkg/client/docker/teleroute/network.go` — Export `FindNonConflictingSubnet` for reuse
- `pkg/client/cli/docker/compose/connection.go` — Include `AlsoProxy` in connection subnets
- `pkg/client/cli/docker/compose/transform.go` — Add `configureDefaultNetwork` method, wire into `applyEngagements`
- `integration_test/compose_test.go` — Add `Test_ComposeDefaultNetworkNoSubnetConflict`
- `CHANGELOG.yml` + docs — Add changelog entry under 2.27.3

## Test plan

- [x] Run `Test_ComposeDefaultNetworkNoSubnetConflict` — uses 3 named connections to consume Docker network address space, pushing allocation toward the conflicting range
- [x] Run full `Compose` test suite — all 8 tests pass, no regressions
- [x] Run `Test_DockerDaemon_networkNoSubnetConflict` — confirms exported function still works
- [x] Run teleroute unit tests — all pass

### Kind cluster setup for integration tests

The integration test requires a kind cluster with `serviceSubnet: 172.20.0.0/16` so Docker's sequential IPAM allocation (`172.17`, `172.18`, `172.19`, `172.20`...) would collide with the cluster's service CIDR if the fix is not in place.

```bash
cat <<KINDEOF > /tmp/kind-config.yaml
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
networking:
  serviceSubnet: "172.20.0.0/16"
KINDEOF
kind create cluster --name tp-test --config /tmp/kind-config.yaml
```

On arm64 hosts, a local arm64 echo-server image must be built and loaded into kind since `ghcr.io/telepresenceio/echo-server:0.3.1` is amd64-only.

### DIFF
--- a/CHANGELOG.yml
+++ b/CHANGELOG.yml
@@ -36,6 +36,13 @@ items:
           Telepresence now pre-computes a non-conflicting RFC 1918 subnet by checking both existing
           Docker networks and the cluster's routing CIDRs before creating the teleroute network.
         docs: https://github.com/telepresenceio/telepresence/issues/4005
+      - type: bugfix
+        title: Fix Docker Compose default network subnet conflicts with Kubernetes cluster CIDRs
+        body: >-
+          When using <code>telepresence compose</code>, Docker Compose auto-creates a default network
+          whose subnet could overlap with Kubernetes cluster CIDRs, causing cluster traffic to be
+          misrouted. Telepresence now configures the default network with an explicit non-conflicting
+          subnet in the rendered compose file.
   - version: 2.27.2
     date: 2026-03-09
     notes:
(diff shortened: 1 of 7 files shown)


thallgren (approved):

thallgren:
Thanks again @jssosa10 . Great work!

PR #4097, opened 2026-04-09
Clear agent snapMap on reconnect to prevent stale intercepts

### LINKED ISSUE #4095: Intercepts silently break after traffic-manager pod restart no recovery without agent reinstall
## What happens

When the traffic-manager pod restarts (rollout, eviction, node drain, whatever the cause), all existing intercepts just stop working. Traffic doesn't reach my local machine, and it doesn't reach the pod either. The weird part is that `telepresence status` still shows "Connected" and `telepresence list` still shows workloads normally. Only `telepresence list --intercepts` reveals that intercepts are gone.

The only way I've found to fix it is `telepresence uninstall --agent <workload>` followed by re-intercepting. That restarts the pod with a fresh agent and it works immediately. This is what pointed me toward the reconnection logic being broken.

This has been hitting my team repeatedly. Two devs intercepting on the same cluster, and every time the manager restarts we both lose everything with zero warning.

## Environment

- Client: v2.27.2 (macOS arm64)
- Traffic Manager: v2.27.1
- Traffic Agent: ghcr.io/telepresenceio/tel2:2.27.1
- Kubernetes: EKS v1.34.4
- Manager and workloads in separate namespaces
- Istio ambient mode on the workload namespace (with waypoint proxy). Manager namespace is NOT in the mesh.
- traffic-manager service is headless (`clusterIP: None`)
- Workloads are Argo Rollouts + regular Deployments

### 1. Agent detects the break but never recovers

One agent logged the failure explicitly:

```
20:48:42.1638 INFO  Connected to Manager 2.27.1
21:47:52.1704 WARN  remain: remain: rpc error: code = Unavailable desc = connection error:
  desc = "error reading server preface: read tcp <agent-ip>:48062-><old-manager-ip>:8081:
  read: connection reset by peer"
```

The agent saw the connection to the old manager pod die 15 seconds after the restart. After that: nothing. No reconnection logged, no more errors, complete silence for over an hour.

Another agent had active intercepts going and just went silent:

```
21:20:30.5502 INFO  Connected to Manager 2.27.1
21:20:30.5680 INFO  handleIntercept: Allowing non-conflicting intercept "<session-a>:<workload>"
21:22:09.3695 INFO  handleIntercept: Allowing non-conflicting intercept "<session-b>:<workload>"
[manager restarted at 21:47 — zero entries after 21:22]
```

### 2. New manager shows no agent activity

The new manager pod logged only its startup config. Over an hour later: nothing else. No agents registering, no sessions, no intercept watches.

I checked the source and `ArriveAsAgent` logs at DEBUG, so I can't be 100% sure nothing got through. But the netstat evidence below suggests RPCs aren't landing.

### 3. Connector actually recovered fine (correcting my earlier assumption)

I initially thought the connector was stuck because the log was silent for hours after the manager restart. But checking the full overnight log, the connector was working fine — `Remain` was just succeeding silently (no error = no log entry). It reconnected multiple times overnight on its own:

```
01:13:00 ERROR error calling Remain: the traffic manager gRPC API timed out.
01:34:10 ERROR error calling Remain: the traffic manager gRPC API timed out.
01:52:03 INFO  Connected to OSS Traffic Manager v2.27.1   ← reconnectManager worked
02:20:32 INFO  Connected to OSS Traffic Manager v2.27.1   ← again
02:55:21 INFO  Connected to OSS Traffic Manager v2.27.1   ← again
03:31:26 INFO  Connected to OSS Traffic Manager v2.27.1
04:08:36 INFO  Connected to OSS Traffic Manager v2.27.1
04:28:30 ERROR error calling Remain: ... executable aws failed with exit code 1  ← SSO expired (normal, unrelated)
```

So the connector's `reconnectManager` + `clear(snapMap)` mechanism works. The problem is agent-side only.

### 4. Agent TCP connections look wrong

Running `netstat` inside the agent containers after the restart shows two TCP connections to the new manager IP:

```
# From agent:
tcp  0  0  <agent-ip>:56390  <new-manager-ip>:8081  ESTABLISHED   <-- captured by ztunnel
tcp  0  0  <agent-ip>:56398  <new-manager-ip>:8081  ESTABLISHED   <-- direct

# From manager:
tcp  0  0  <new-manager-ip>:8081  <agent-ip>:56398  ESTABLISHED   <-- only sees ONE
```

The manager only sees the direct connection. The other one goes through the Istio ztunnel:

```
tcp  0  0  127.0.0.1:15001  <agent-ip>:56390  ESTABLISHED
```

Same pattern on multiple agents. Since the traffic-manager service is headless, DNS resolves to the pod IP directly. After the restart, gRPC re-resolved to the new IP, and one of the new connections got captured by the ztunnel while the other went through directly. I can't tell which one gRPC is actually using for RPCs.

### 5. Fresh agents work fine

An agent pod that started 35 minutes after the restart connected without issues:

```
22:22:23.9822 INFO  Traffic Agent v2.27.1
22:22:24.8818 INFO  Connected to Manager 2.27.1
```

The new manager is healthy. The problem is only with agents that were running before the restart.

### 6. Tunnel failures from stale intercept state

From a previous occurrence of the same bug, the agent logged what happens when traffic comes in during the broken state:

```
21:02:06.8640 ERROR unable to create tunnel to client <session-id>... for id TCP <src> -> 127.0.0.1:9090
21:02:06.8642 ERROR failed to create client stream: context canceled
```

The agent intercepts the traffic (preventing it from reaching the app), tries to tunnel it to a client session that doesn't exist on the new manager, fails, and drops the connection.

### 7. `status` and `list` hide the problem

`telepresence status` says "Connected" and `telepresence list` shows workloads correctly while the whole thing is broken. I checked the source:
- `status()` returns cached fields like `managerName`/`managerVersion`, no live RPC (`session.go:806`)
- `WorkloadInfoSnapshot()` uses cached agent data plus direct k8s API watchers, doesn't talk to the manager (`session.go:608`)

## What I found in the source code

I went through the `release/v2` branch to understand why reconnection doesn't work. Found a few things:

### Agent doesn't clear `snapMap` on reconnect (the connector side was fixed, agent was missed)

Commit `4a17cf15` ("Use gRPC delta methods for efficient map updates") introduced `WatchInterceptsDelta` with delta-based streaming. That commit changed the following files:

```
pkg/client/userd/trafficmgr/agents.go      ← connector: added clear(snapMap) in repair
pkg/client/userd/trafficmgr/intercept.go   ← connector: added clear(snapMap) in repair
pkg/client/userd/trafficmgr/session.go     ← connector: reconnectManager
cmd/traffic/cmd/manager/service.go         ← manager: WatchInterceptsDelta endpoint
cmd/traffic/cmd/manager/state/state.go     ← manager: delta subscribe
cmd/traffic/cmd/manager/state/intercept.go ← manager: intercept delta
pkg/grpc/watcher/watcher.go               ← shared: WatchWithRetry
pkg/maps/utils.go                          ← shared: DeltaUpdate
pkg/cache/map.go                           ← shared: Subscribe with delta
```

Notice what's not in that list: `cmd/traffic/cmd/agent/client.go`. The agent-side `interceptWatchLoop` also uses `WatchInterceptsDelta` with a `snapMap`, but the `clear(snapMap)` was only added to the connector repair functions. The agent's repair function (`reconnectAgent`) was left without it.

In `cmd/traffic/cmd/agent/client.go`, `interceptWatchLoop` keeps a `snapMap` outside the `WatchWithRetry` call:

```go
snapMap := make(map[string]*rpc.InterceptInfo)
err := watcher.WatchWithRetry(ctx, "WatchInterceptsDelta", retryInterval,
    func(ctx context.Context) (grpc.ServerStreamingClient[rpc.InterceptInfoDelta], error) {
        return manager.WatchInterceptsDelta(ctx, session)
    },
    func(delta *rpc.InterceptInfoDelta) error {
        maps.DeltaUpdate(snapMap, delta.Upserts, delta.Removals)
        snapshots <- maps.Values(snapMap)
        return nil
    }, reconnectAgent)
```

If `WatchWithRetry` reconnects to the new manager, the manager sends an empty initial delta (no intercepts to report). `DeltaUpdate` with empty upserts and nil removals does nothing (`pkg/maps/utils.go` just iterates empty collections). Old intercepts stay in `snapMap`.

The connector already handles this. Both `agents.go` and `intercept.go` clear the map in their repair functions:

```go
// connector agents.go
func() error {
    clear(snapMap)
    return nil
}

// connector intercept.go
func() error {
    clear(snapMap)
    return s.reconnectManager()
}
```

The agent doesn't:

```go
// agent client.go
reconnectAgent := func() error {
    _, err := manager.ReconnectAgent(ctx, &rpc.ReconnectAgentRequest{Session: session, Agent: info})
    return err
}
```

### `WatchWithRetry` retries forever with no bound

In `pkg/grpc/watcher/watcher.go`:

```go
backoff.WithContext(backoff.NewConstantBackOff(retryInterval), ctx)
```

No `MaxElapsedTime`, no `MaxRetries`. If the inner retry keeps failing, it never returns. Since `TalkToManager` runs four goroutines through `errgroup.Wait()` (all of them either stuck in `WatchWithRetry` or swallowing errors), it blocks forever too. The outer `TalkToManagerLoop` never gets a chance to do a full clean reconnect with a new `grpc.NewClient` + `ArriveAsAgent`.

The defer that would clean up stale state also never fires:

```go
defer func() {
    state.HandleIntercepts(ctx, nil)  // would clear stale intercepts
    manager.Depart(ctx, session)
}()
return wg.Wait()  // stuck here
```

### Agent `remain()` swallows errors

Agent side (`client.go`):
```go
if _, err := manager.Remain(ctx, &rpc.RemainRequest{Session: session}); err != nil {
    clog.Warnf(ctx, "remain: %v", err)
    // never returns the error
}
```

When the new manager returns `NotFound` for the old session, the agent just warns and continues. No reconnection is triggered. The connector side has the same pattern (`session.go:412` always returns nil), but in practice the connector recovers through its watcher streams calling `reconnectManager`, so it's not an issue there.

### gRPC keepalive set to 24 hours (connector side, may affect agents too)

In `pkg/client/k8s/connect.go`:

```go
grpc.WithKeepaliveParams(keepalive.ClientParameters{
    Time:    24 * time.Hour,
    Timeout: 20 * time.Second,
})
```

This is the connector's keepalive. In practice, the connector recovers fine through its watcher retry mechanism despite this. But the agent uses `grpc.NewClient` with no keepalive at all (default), which means the agent's gRPC transport detection after a manager pod restart depends entirely on DNS re-resolution and subchannel management, which may be slow or broken when Istio ambient ztunnel is involved (see #4 in the logs section).

## What I don't know yet

1. Which TCP connection gRPC picks after the agent reconnect. I can see two (one through ztunnel, one direct), but I don't know which one carries RPCs. If it's the ztunnel one, that could be why the manager sees nothing from pre-existing agents.

2. Whether the agent's `WatchWithRetry` actually succeeds or stays stuck retrying. Retries happen at DEBUG level. At INFO I see nothing. Could be looping forever, or it could have reconnected with stale `snapMap` data. Both break things the same way.

## How to get better evidence

Reproduce with DEBUG logging:

```bash
telepresence connect
telepresence loglevel DEBUG -d 0
telepresence intercept <workload> --port <port>
# verify intercept works, then:
kubectl delete pod -n <manager-namespace> -l app=traffic-manager
# wait a couple minutes, then:
telepresence gather-logs
```

At DEBUG, the agent should show `WatchWithRetry` retries and `handleIntercept` snapshots (revealing if `snapMap` keeps stale entries). The manager should show `ArriveAsAgent`/`ReconnectAgent` calls. That would close the remaining gaps.

## Summary

After a manager restart, the connector recovers on its own (the `reconnectManager` + `clear(snapMap)` mechanism works). But agents don't recover. Intercepts die silently. `status` and `list` show cached data so there's no visible indication that anything broke.

The most concrete issue from the source code is: commit `4a17cf15` added `clear(snapMap)` to the connector's repair functions but missed the agent's `interceptWatchLoop`. That means if the agent's `WatchWithRetry` reconnects to a fresh manager, the empty initial delta leaves stale intercepts in `snapMap`, causing tunnel failures.

On top of that, `WatchWithRetry` has no retry limit (can block the outer `TalkToManagerLoop` from doing a clean reconnect), and the agent's `remain()` swallows errors instead of triggering recovery.

The Istio ambient ztunnel + headless service combination might also play a role in preventing the agent's gRPC from reconnecting at all, but I can't prove it from the logs I have.

Workaround: `telepresence uninstall --agent <workload>` then re-intercept after any manager restart.

## Related issues

- #2931 — "Traffic Manager keeps getting stuck and needs to be restarted" (closed, v2.6.6). Similar symptoms but different root cause (k8s API throttling). The ask there was the same: self-healing after failures.
- #2378 — "Intercept stop working after some time and won't work until traffic-agent removed" (closed, turned out to be a GitOps tool reverting the port rename). Same workaround though: remove agent, re-intercept.

obs: Maybe my investigation isn't correct, so I would love any feedbacks and insights = )

fabricio-neri:
## Description

Fixes #4095

When the traffic-manager pod restarts, the traffic-agent's `interceptWatchLoop` reconnects via `WatchWithRetry` but its local `snapMap` retains intercepts from the old manager. The new manager sends an empty initial delta (no intercepts), and `DeltaUpdate` with empty upserts/removals is a no-op — so stale intercepts persist, causing tunnel failures and silently dropped connections.

The connector side already clears `snapMap` in its repair functions (added in commit `4a17cf15`), but the agent side was missed. This PR adds `clear(snapMap)` to the agent's `reconnectAgent` repair function, matching the connector's behavior.

## Checklist

- [ ] Once my PR is ready to have integration tests ran, I posted the PR in #telepresence-oss channel on the CNCF Slack so that the "ok to test" label can be applied.


cmd/traffic/cmd/agent/client.go
```diff
@@ -159,15 +159,16 @@ func interceptWatchLoop(
 	snapshots chan<- []*rpc.InterceptInfo,
 	retryInterval time.Duration,
 ) error {
+	// Call WatchIntercepts and publish the snapshots on the channel
```
danielavelez12, line 162: Should we keep this comment right before line 172? Also, we could add a comment explaining the `clear(snapMap)`

danielavelez12 (approved):
Nice, seems like we also clear snapMap in other places before doing a Watch Delta

Makes sense!

thallgren:
Thank you for @fabricio-neri. Good catch.

PR #4098, opened 2026-04-10
Support multi-namespace intercepts

mindstorms6:
## Description

This PR introduces the ability to support multi namespace intercepts.

For our use case - we have services that live in multiple namespaces (due to isolation/security constraints) that we often need to co-develop. We'd like to be able to intercept those multi namespace services on one laptop/all at once to produce some true end to end validation. This adds the ability to do that. 

## Checklist

 - [x] My change is adequately tested. (testing details below)
       
Testing done:

Unit/build coverage:
- Added CLI request tests for omitted vs explicit engagement namespace.
- Added coverage that engagement-local `--namespace` is not copied into implicit connect flags.
- Added agent port-forward coverage for namespace-scoped `WaitForAgentIP` fallback behavior.
- Added a namespace integration test that connects to one namespace with `alpha,beta` mapped, then creates two simultaneous non-global HTTP intercepts in different namespaces using unique intercept names and `--workload echo`.

Commands run:
- `make protoc`
- `make docs-files` with `EXTERNAL_FUSEFTP=1`
- `make build` with `EXTERNAL_FUSEFTP=1`
- `go test ./pkg/client/cli/intercept ./pkg/client/cli/daemon ./pkg/client/agentpf ./pkg/client/rootd ./cmd/traffic/cmd/manager`
- `go test ./integration_test/... -run '^$'`
- `git diff --check`

Live cluster validation:
- Built a dev `tel2` image and temporarily upgraded the `traffic-manager` Helm release in `bdawg-3646` to that image.
- Created temporary namespaces `tp-multins-a` and `tp-multins-b`, each with an `echo` workload.
- Connected with `telepresence connect --namespace tp-multins-a --manager-namespace ambassador --mapped-namespaces tp-multins-a,tp-multins-b`.
- Created two simultaneous non-global HTTP intercepts:
  - `tp-a-local` targeting `echo` in `tp-multins-a` with `--http-header x-tp-ns=a`
  - `tp-b-local` targeting `echo` in `tp-multins-b` with `--http-header x-tp-ns=b`
- Verified header-matched traffic reached the correct local handler in each namespace, unmatched traffic still reached the in-cluster echo service, and `telepresence list --namespace` showed the correct intercept per namespace.
- Repeated the validation with branch-built host user/root daemons and confirmed client, user daemon, root daemon, traffic-manager, and traffic-agent all reported the dev version.
- Cleaned up temp namespaces/intercepts/secrets and rolled the Helm release back to `ghcr.io/telepresenceio/tel2:2.27.2`.


rpc/manager/manager.proto
```diff
@@ -776,6 +777,16 @@ service Manager {
...
+  // WatchAgentPodsInNamespaces notifies a client of the set of known Agents from
+  // the requested namespaces that the client can connect to when port-forwards
+  // are allowed.
+  rpc WatchAgentPodsInNamespaces(AgentsRequest) returns (stream AgentPodInfoSnapshot);
+
+  // WatchAgentPodsInNamespacesDelta notifies a client of changes to the set of
+  // known Agents from the requested namespaces that the client can connect to
+  // when port-forwards are allowed.
+  rpc WatchAgentPodsInNamespacesDelta(AgentsRequest) returns (stream AgentPodInfoDelta);
+
```
thallgren: Why would we need both? The reason both exists historically was a need to retain backward compatibility when introducing the delta. Moving forward, there should only be delta IMO.

CHANGELOG.yml
```diff
@@ -24,6 +24,20 @@
 #
 # For older changes, see CHANGELOG.OLD.md
 items:
+  - version: 2.27.5
```
thallgren: This is a new feature so it has to go into 2.28.0

cmd/traffic/cmd/manager/service.go
```diff
@@ -423,13 +452,36 @@ func (f *AgentStateFile) UnmarshalJSON(data []byte) error {
...
+		namespaces = []string{clientInfo.Namespace}
+	}
+	nss := make([]string, 0, len(namespaces))
+	for _, namespace := range namespaces {
+		if namespace == "" {
+			namespace = clientInfo.Namespace
+		}
+		if !s.State().ManagesNamespace(ctx, namespace) {
+			return nil, status.Errorf(codes.FailedPrecondition, "namespace %s is not managed by this traffic-manager", namespace)
+		}
```
thallgren: This looks like a duplicate of `managedTargetNamespace`

docs/reference/engagements/cli.md
```diff
@@ -6,13 +6,23 @@ title: Configure workload engagements using CLI
...
+The `intercept`, `wiretap`, and `replace` commands also accept `--namespace` when the workload
+to engage is in a different mapped namespace. `ingest` continues to use the connected namespace.
 
 ```shell
 telepresence connect --namespace myns
 telepresence replace/ingest/intercept/wiretap hello
 ```
 
+To intercept a workload in another mapped namespace without reconnecting, pass `--namespace`
+to the engagement command:
```
thallgren: Perhaps use the word "engage" here instead of "intercept".

pkg/client/agentpf/clients.go
```diff
@@ -362,7 +413,7 @@ func (s *clients) GetRandomAgent(ctx context.Context) (aa agent.AgentClient) {
 // The function returns nil when there are no agents for the given workload in the connected namespace.
 func (s *clients) GetWorkloadClient(workload string) (pvd tunnel.Provider) {
 	s.clients.Range(func(_ string, ac *client) bool {
-		if ac.info.WorkloadName == workload {
+		if ac.info.WorkloadName == workload && ac.info.Namespace == s.Namespace {
```
thallgren, line 417: The comment describing this function is no longer accurate given that the client can be in a different namespace.

pkg/client/agentpf/clients.go
```diff
@@ -406,7 +457,40 @@ func (s *clients) WatchAgentPods(rmc manager.ManagerClient) error {
...
+			clog.Debugf(ctx, "No delta support in traffic-manager, starting WatchAgentPodsInNamespaces instead")
+			return rmc.WatchAgentPodsInNamespaces(ctx, s.agentsRequest())
+		},
+		func(snapshot *manager.AgentPodInfoSnapshot) error {
+			return s.updateClients(snapshot.Agents)
+		}, nil)
+	if err == nil || status.Code(err) != codes.Unimplemented {
+		return err
+	}
+
```
thallgren: The non-delta function fills no purpose.

thallgren (changes requested):
This looks very good overall. Well done!

It's a major new feature, so it must land in 2.28.0.

Also, the documentation mentions that DNS will allow single label name access to the connected namespace. A clarification that workloads intercepted in other namespaces than the connected one doesn't change that - they still need to be accessed using `<name>.<namespace>` - would be welcome.

integration_test/namespaces_test.go
```diff
@@ -330,3 +332,78 @@ func (s *nsSuite) Test_NamespacesStatic() {
 	rq.Len(st.UserDaemon.MappedNamespaces, 2)
 	itest.TelepresenceDisconnectOk(ctx)
 }
+
+func (s *nsSuite) Test_MultiNamespaceHTTPIntercepts() {
+	if !(s.ManagerIsVersion(">2.24.x") && s.ClientIsVersion(">2.24.x")) {
+		s.T().Skip("HTTP intercepts require Telepresence 2.25.0 or later")
+	}
```
thallgren: The versions here are wrong.

thallgren (approved):
I believe the version check in the integration test is incorrect, but other than that, this is good to go.

Oh, and you need a new `make docs-files`.

mindstorms6:
Thank you for the quick turnaround @thallgren! I think I got all the remaining things in - let me know if there's any other changes you need from me. I think the last part is waiting for your approval to run the checks/workflows.

Thanks again!

thallgren:
The `Test_MultiNamespaceHTTPIntercepts` is constantly failing.

thallgren:
Thanks @mindstorms6 !

PR #4101, opened 2026-04-15
Preserve default gateway for never-proxy routes

mindstorms6:
Howdy again!

We hit an issue with telepresence specifically related to routing. 

When installing static routes for never-proxy subnets, Telepresence routed them to the default interface but discarded the default route's gateway and source address.

On some networks this creates an on-link route such as:

    10.224.1.135 dev enp1s0

instead of the usable route:

    10.224.1.135 via 10.0.2.1 dev enp1s0

Preserve the matching-family gateway and local source address from the selected default route when constructing never-proxy routes. This keeps the existing default-route selection behavior while avoiding unusable on-link routes for destinations that require the default gateway.

I _believe_ this is a safe change that should not break other uses of `never-proxy` - and it did unblock my specific vm test case. 

## Description

## Checklist

### DIFF
--- a/CHANGELOG.yml
+++ b/CHANGELOG.yml
@@ -38,6 +38,15 @@ items:
           different namespaces, while preserving the connected namespace as the default
           when no engagement namespace is specified.
         docs: reference/engagements/cli
+      - type: bugfix
+        title: Preserve the default gateway for never-proxy routes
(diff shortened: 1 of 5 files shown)


thallgren (approved):
Looks good to me. Thanks.

PR #4112, opened 2026-04-21
Avoid probing cleartext HTTP app protocols

mindstorms6:
## Description

Kubernetes Services can advertise cleartext HTTP/1 with `appProtocol`: `http` or `kubernetes.io/http`. The traffic agent previously treated those values as unknown, so it still probed TLS and HTTP/2 support through the inactive proxy port before forwarding. When that probe cannot connect, filtered intercepts can pay seconds of avoidable latency per request. In our environment- this latency was as high as 4s! With this patch - the latency was a much more tolerably ~150ms.

Classify common HTTP/1 appProtocol spellings as cleartext HTTP/1 and mark both TLS and HTTP/2 unsupported up front. This skips the probe path for known HTTP/1 services while leaving unknown app protocols unchanged.

## Checklist

### DIFF
--- a/CHANGELOG.yml
+++ b/CHANGELOG.yml
@@ -73,6 +73,14 @@ items:
           determined. This avoids relying on workload DNS search paths to resolve
           the manager service from another namespace, while preserving the
           existing short name fallback.
+      - type: bugfix
+        title: Avoid probing cleartext HTTP app protocols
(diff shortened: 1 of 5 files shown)


thallgren (approved):
This is great. Thanks!

PR #4116, opened 2026-04-29
Reconnect root daemon when it drops

mindstorms6:
## Description

We had cases where (for reasons yet to be determined, possibly remote cluster control plane instability or $network things) we ended up with losing a watch and ending up with an `EOF`. This left the user daemon in a weird state in some cases / broken / hard to fix.  

The user daemon already has the network config it needs to rebuild the root daemon side of the session, so use that instead of making users bounce the whole connection when rootd goes away.

This wires the activity watcher failure path into a reconnect loop, swaps in the new root daemon client once it is ready, and re-posts DNS domains after reconnecting. Calls that need rootd get a short failed-precondition error while the new client is being established.

Also add the changelog entry and regenerate the release notes.

## Checklist

### DIFF
--- a/CHANGELOG.yml
+++ b/CHANGELOG.yml
@@ -87,6 +87,14 @@ items:
           Ensures /usr/sbin is in the PATH used by the root daemon when it is run
           by systemd.  This ensures iptables can be found by the daemon.
       - type: bugfix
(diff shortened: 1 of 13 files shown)


mindstorms6:
Fixed the linter issues - sorry about that

PR #4120, opened 2026-05-02
Client Stability: Bounded Intercept Dials

mindstorms6:
## Description

Sometimes when using intercepts with web sites (especially things like vite + HMR) - the website would loads _thousands_ of resources at once (sigh, I know) and that could cause telepresence to die. 

After some back and forth - I think this is the right fix: Limit selected-intercept dial responders so bursty workloads cannot make the client daemon fan out without bound. Let me know if you want to change this number, for now it's hard coded. This seems to work for us, and now the super bursty traffic doesn't make telepresence die. 

Also make VIF open failures return a plain nil device and tolerate partially initialized Linux devices during cleanup. 

Tests cover the concurrency cap and nil-device error path.

## Checklist

### DIFF
--- a/CHANGELOG.yml
+++ b/CHANGELOG.yml
@@ -95,6 +95,14 @@ items:
           connect to the application through the pod IP reach the expected
           redirect and DNAT rules, avoiding missed routing when the pod IP does
           not use the loopback interface.
+      - type: bugfix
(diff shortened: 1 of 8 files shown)


thallgren (approved):
LGTM. Thanks1

PR #4132, opened 2026-05-29
Support multi-namespace ingest

fabricio-neri:
## Description

This PR introduces multi-namespace support for `telepresence ingest`, closing the gap left by PR #4098 (which added the same capability for `intercept`, `wiretap`, and `replace` in v2.28.0).

A user connected with `--mapped-namespaces a,b,c` can now run `telepresence ingest <workload> --namespace b` to ingest a workload in a non-connected mapped namespace, while preserving the connected namespace as the default when `--namespace` is omitted.

The change is contained: three additive proto fields, one manager-side handler update (reusing the existing `managedTargetNamespace` helper), userd validation that mirrors `intercept.CanIntercept` verbatim, and the same fallback / error-message pattern as the original multi-namespace work. The lower-level agent routing infrastructure added in #4098 (`AgentConfigRequest.namespace`, `WaitForAgentIPRequest.namespace`, `WatchAgentPodsInNamespacesDelta`) is reused as-is.

The `leave` command also gains `--namespace`, since ingests have no unique name and are identified by `workload + container + namespace` — without it you cannot disambiguate two ingests with the same workload across mapped namespaces.

## Checklist

- [x] I made sure to update `./CHANGELOG.yml` (release notes are generated from this file).
- [x] My change is adequately tested. (details below)
- [ ] Once my PR is ready, I posted it in #telepresence-oss on CNCF Slack so the "ok to test" label can be applied.

## Testing done

**Unit/build coverage**
- Extended `TestSession_findIngest` with two new cases: cross-namespace lookup with same workload+container resolves correctly; ambiguous lookup (no namespace) errors clearly.
- Updated the existing exact-key-miss case to provide a namespace, since the empty-namespace path now goes through scan-and-resolve.

**Integration**
- Added `Test_MultiNamespaceIngests` in `integration_test/namespaces_test.go`, mirroring `Test_MultiNamespaceHTTPIntercepts`. Connects with `--mapped-namespaces alpha,beta`, ingests `echo` in both namespaces, asserts the JSON output contains the correct namespace and that the two pod IPs differ.

**Commands run**
- `make protoc`
- `make docs-files`
- `GOEXPERIMENT=jsonv2 go build ./...`
- `GOEXPERIMENT=jsonv2 go vet ./...`
- `make lint-go` (0 issues)
- `make lint-rpc`
- `go test ./pkg/client/userd/... ./pkg/client/cli/... ./cmd/traffic/cmd/manager/...`

**Live cluster validation**
- TBD before merge — would mirror PR #4098's pattern: spin up two namespaces with `echo`, run two simultaneous ingests in different namespaces against the same connection, verify environments come from the correct namespace.

## Compatibility

- `EnsureAgentRequest.namespace` is additive (proto3); older traffic-managers silently ignore the field and look in the client's connected namespace. Cross-namespace ingest against a pre-v2.28.1 manager fails with "no such workload" — same compatibility model as #4098.
- `IngestInfo.namespace` is additive; the new column in `list`/`status` output is the only behavior change for existing consumers.
- The handler-cleanup ID format changed from `workload/container` to `workload/container/namespace`; the parser accepts both for backward compatibility within a session.


pkg/client/cli/cmd/leave.go
```diff
@@ -77,10 +78,11 @@ func leaveCmd() *cobra.Command {
 		},
 	}
 	cmd.Flags().StringVarP(&containerName, "container", "c", "", "Container name")
+	cmd.Flags().StringVarP(&namespace, "namespace", "n", "", "Namespace of the ingest. Required to disambiguate ingests with the same workload name across mapped namespaces")
```
thallgren, line 81: This is not consistent with how intercept works. An intercept must have a unique name. The ingestion didn't need a name because the container name made it unique anyway - you can't ingest a container twice. That changes now. Patching the generic `leave` command with an ingestion namespace doesn't seem right.

I think this calls bigger change. We need a `--name` flag on the ingest command. If not given, a default name is generated from the ingested container. If there's a desire to ingest the same workload/container in two different namespaces, then the `--name` flag must be used to make them unique. The name is then what's used when listing ingests and when leaving the ingest.

We should probably remove the `--container` flag here too and rely 100% on the name of the engagement.

pkg/client/userd/trafficmgr/ingest.go
```diff
@@ -14,15 +14,17 @@ import (
 	"github.com/telepresenceio/telepresence/rpc/v2/manager"
 	"github.com/telepresenceio/telepresence/v2/pkg/client"
 	"github.com/telepresenceio/telepresence/v2/pkg/client/remotefs"
+	"github.com/telepresenceio/telepresence/v2/pkg/errcat"
 )
 
 type ingestKey struct {
 	workload  string
 	container string
+	namespace string
```
thallgren, line 23: If the ingest is named, the ingestKey can be a string.

thallgren (changes requested):

pkg/client/cli/cmd/leave.go
thallgren: Thinking more about this. I'm willing to accept this change as-is. Adding the `--name` to the ingest command is a bigger refactor and deserves its own PR.

thallgren (approved):
You'll need to run a `make docs-files` and fix the merge conflicts, but I'll merge this PR once that's done and the integration tests pass.

pkg/client/cli/cmd/leave.go
fabricio-neri: Thanks for the explanation, @thallgren! I tried to mirror what was done for intercepts but it's true that the behavior is different. Just pushed the merge with release/v2 and the regenerated docs, so CI should be able to take another pass.
 
On my team we use ingest a lot to create local workers from Temporal, so multi-namespace support unblocks a real workflow for us. Let me know if there's anything else!

fabricio-neri:
The pipe is failing, I will check what i need to fix

thallgren:
It's failing because you need a `make generate`. The `make docs-files` that I recommended isn't enough.

fabricio-neri:
> It's failing because you need a `make generate`. The `make docs-files` that I recommended isn't enough.

Thanks, Thomas! I merged the latest release/v2 into the branch and ran make generate

thallgren:
Thanks @fabricio-neri for your work on this.

PR #4162, opened 2026-06-12
Short-circuit local traffic to intercepted destinations

### LINKED ISSUE #4125: RFC: Should we support "local intercepts"
We use telepresence in a multi service setup - where there can be a number of micro services that all talk to each other. Telepresence does a great job of helping the local running process use in cluster dns names like normal, as well as intercepts so the local process can participate in the remote cluster.

However, when running multiple things in this way, sometimes `service A` will need to talk to `service B`. We have instrumented our services such that header routing keys are propagated through the stack, so that any service can be intercepted with the same specific routing key and anywhere in the stack, what ever you are running, will get hit.

This means that (with out special app specific/client specific logic) - 2 local running processes that are intercepted, with the same routing key will end up having to round trip to the cluster - despite running locally on the same machine.

So, the proposal is: what if we introduced a port redirect/proxy concept - eg:

`telepresence local-intercept add service-a.default.svc.cluster.local:8000:8000`

This would instruct telepresence locally to take any request that matches the dns name `service-a.default.svc.cluster.local` outbound on port `8000` to redirect to `localhost:8000`

Sort of similar to some of the existing `--reroute*` calls we have today, but happens outside of the `connect` call.

I'm thinking `telepresence local-intercept {add,remove,list} $dnsName:$outboundPort:$localReroutePort`

This would all be client side, I don't think we'd need any in cluster changes. 

I can sort of do this today with `/etc/hosts` and some `socat` hackery - but `telepresence` feels cleaner.  

  - Output of `telepresence version` (just in case the feature exists in future versions): `2.27.4`
  - 1.32 - talos 

**Additional context**
Thanks for the consideration!

thallgren:
## Problem

When the handler of one intercept dials a cluster destination that is covered by another of the client's active intercepts (or its own), the connection round-trips through the cluster only to arrive back at a local intercept handler. With several interdependent services engaged on the same workstation, local-to-local latency becomes a function of the cluster.

## Solution

The root daemon connects such traffic directly to the local intercept handler:

- The user daemon derives, from each ACTIVE intercept in its snapshot, the workload identity, intercepted container port and protocol, the service ClusterIP:port, and the local handler address, and pushes them to the root daemon with the new `SetInterceptShortcuts` daemon RPC (replace-all semantics per push).
- The root daemon matches outbound connections against the declared service addresses, and against the pod IPs of intercepted workloads via the agent-pod snapshot it already maintains — covering every replica, including pods created after the intercept. A match pipes the connection locally instead of opening a tunnel stream.
- The service cluster IPs are resolved by the traffic-manager (using the shared service informer) when the intercept is prepared, and travel to the client in the new `service_ips` fields of `PreparedIntercept` and `InterceptSpec`. The manager also fills in the previously unset resolved `service_name`. Clients therefore need no permission to read services.

## Configuration

- `intercept.localShortcut` (default `true`) enables the shortcut, also settable cluster-wide via the chart's `client.intercept.localShortcut`.
- `intercept.localShortcutIsGlobal` (default `true`) resolves a dilemma: the shortcut is made when a connection is established, so it cannot evaluate HTTP header or path filters. Either the shortcut serves all local traffic to a filtered intercept's destination, or filtered intercepts always take the cluster round trip. The default applies the shortcut to all intercepts, on the assumption that filters exist to limit how an intercept impacts other users of the cluster, not the developer's own traffic. Setting it to `false` makes local traffic honor the filters exactly. The dilemma is documented in `docs/reference/config.md`.

## Version skew

An old root daemon answers the new RPC with `Unimplemented`, which the user daemon tolerates. An old manager leaves the new spec fields empty, degrading the shortcut to pod-IP matching only. An old user daemon never calls the RPC, and unknown spec fields are ignored by old consumers. No synchronized upgrade is required in any direction.

## Testing

Unit tests cover the root daemon's shortcut table (service-address and workload-pod matching, wholesale replacement, malformed-entry skipping), the eligibility filter, and the config defaults. A new `LocalShortcut` integration suite verifies the shortcut for service and pod addresses via the daemon log, the `localShortcut=false` round-trip behavior, and the `localShortcutIsGlobal=false` exemption of a header-filtered intercept. The integration test harness disables the shortcut by default so that the many tests whose purpose is to verify the cluster path (agent routing, header filters, TLS termination) keep their meaning; the LocalShortcut suite enables it explicitly.

Closes #4125


mindstorms6:
Wow this is incredible- thanks so much @thallgren !

PR #4171, opened 2026-06-22
Fix slow routing on windows

zoriya:
## Description

When running `telepresence connect`, we list all interfaces and use `net.InterfaceByIndex(idx)` on each of them. This function takes ~10ms so when you have LOTS of interfaces it takes a while (we have ~2000 interfaces in our company due to cloudflare's warp that bind every interfaces available).

This PR reduces connect time from ~45s to ~6s.

## Checklist


pkg/routing/routing_windows.go
```diff
@@ -69,26 +87,44 @@ retryInconsistent:
...
-								return r, nil
-							}
-						}
-					}
+			if !owners[int(row.InterfaceIndex)] {
+				continue
+			}
+			r, err := rowAsRoute(&row, localIP, ifaces)
+			if err != nil {
+				if err == errInconsistentRT {
```
thallgren: The `err == errInconsistentRT` will never happen. With this change, owner rows are by construction present in the `ifaces` snapshot, so `rowAsRoute` can no longer return `errInconsistentRT`.

Please flatten `getRouteForIP` to a single pass - drop the `retryInconsistent:` label, the outer loop,, and the `if err == errInconsistentRT` check.

CHANGELOG.yml
```diff
@@ -24,6 +24,15 @@
 #
 # For older changes, see CHANGELOG.OLD.md
 items:
+  - version: 2.29.1
+    date: 2026-06-20
```
thallgren:
```suggestion
    date: (TBD)
```

thallgren (changes requested):
Some nits, but this looks good in general. Thanks for contributing this!

Please run `make docs-files`. It's required after a change to CHANGELOG.yml. Also, please sign your commit.

pkg/routing/routing_windows.go
zoriya: The err == errInconsistentRT check can happen if the interface changes (created) while we're iterating. I kept the retry logic for the same reason but i can delete it.

zoriya: done

thallgren: It can't happen. You do `if !owners[int(row.InterfaceIndex)] { continue }` just before you call `rowAsRoute`. So the `row` is in the `owners` table and that table only contain entries from the snapshotted `ifaces`. There's no way `rowAsRoute` can have a cache-miss in the `ifaces` passed to it.

It's all good. The snapshotting removes the inconsistency window the loop used to cover.

thallgren:
You commit is still missing a sign-off.

zoriya:
what do you mean? my commit is signed
image

thallgren:
It's signed, but it lacks a "Signed-off-by:" in the comment. Sorry if I was unclear.

zoriya:
i didn't know about this `Signed-off-by:` convention, done!

PR #4189, opened 2026-07-09
DO NOT MERGE: validate CI integration retry orchestration

thallgren:
Draft-only validation of the check-integration-ci retry flow (see the three commits below the flake). Carries an intentional flake that fails on attempt 1 and passes on retry; a green `build_and_test` proves on the real runners that the suite runs to completion, the retry is scoped to the failing test, and the step summary labels it flaky. Will be closed unmerged; the real PR follows from thallgren/ci-integration-retries.


thallgren:
Validation succeeded: attempt 1 ran the full suite to completion with the flake failing, attempt 2 re-ran only the scoped test in ~20s, and the summary labeled it flaky (see the job log). Closing unmerged; the real PR follows.

PR #4195, opened 2026-07-11
Prove the node-agent cluster default and its per-attachment override

thallgren:
## Description

Adds the NodeAgentClusterDefault integration suite, closing an e2e coverage gap in the node-agent mode-selection precedence: the Helm chart's `client.nodeAgent.enabled=true` cluster default and its explicit `--node-agent=false` per-attachment override were only proven at the unit seam (`pkg/client/cli/flags/nodeagent_test.go`); nothing exercised the full path from the Helm value through the cluster-provided client config to the workstation's mode selection.

The suite installs the traffic-manager with `nodeAgent.enabled=true,client.nodeAgent.enabled=true` and asserts both directions:

- a flagless `telepresence intercept` is served by a node-agent — a Job appears, the workload's pod is left untouched, and the Job is reaped on detach;
- `telepresence intercept --node-agent=false` overrides the default — a traffic-agent sidecar is injected and no Job is created.

The two cases use separate workloads, since an injected sidecar outlives its intercept and makes the workload inadmissible for later node-agent attachments.

Verified locally against a kind cluster (both tests pass in ~11s after setup).

### DIFF
--- a/integration_test/node_agent_cluster_default_test.go
+++ b/integration_test/node_agent_cluster_default_test.go
@@ -1,0 +1,131 @@
+package integration_test
+
+import (
+	"strconv"
+	"time"
+
+	"github.com/telepresenceio/telepresence/v2/integration_test/itest"
+)
+
+// nodeAgentClusterDefaultSuite exercises the cluster-provided client default
+// for node-agent mode (Helm value client.nodeAgent.enabled=true): a flagless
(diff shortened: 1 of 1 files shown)


thallgren:
Moving this commit onto the v2.30.0 release branch (PR #4194) instead.

PR #4196, opened 2026-07-11
Refresh the compose CLI model for Docker Compose v5.3

thallgren:
Docker Compose renamed `--no-TTY` to `--no-tty` on the run command. This regenerates `dc-cli.json` against Compose v5.3.0 and the compose CLI reference doc with it. Until committed, every local build regenerates the file and leaves the working tree dirty for anyone on a current Compose.


thallgren:
Superseded by #4197, which carries the same compose CLI refresh as part of the dependency update.

PR #4201, opened 2026-07-12
Report agent type and node-agent creation in usage reports

thallgren:
Usage reporting previously said nothing about which agent serves an attachment. Now:

- Every successful manager-side attachment emits a `manager.attach` report with `agent.type` (`sidecar` or `node`): from `AddIntercept` for the intercept-family modes (with the `mechanism` included), and from `EnsureAgent` for ingests.
- The manager emits `manager.nodeagent.create` whenever it actually creates a node-agent Job, with a `trigger` entry distinguishing first-time creation (`ensure`) from stale-Job replacement (`replace`). Idempotent reuse and failed creates stay silent.

Both values are bounded, tool-intrinsic strings per the pkg/usg ground rules; no workload, namespace, or user-derived data is added. No context plumbing was needed: gRPC handler contexts fall back to the manager's root context, which carries the producer, and the pod-watch reconciler runs on the state's background context.

Unit tests assert the report on the AddIntercept node-agent path and on both Job-creation triggers, including that a replacement emits exactly one report.

### DIFF
--- a/cmd/traffic/cmd/manager/state/intercept.go
+++ b/cmd/traffic/cmd/manager/state/intercept.go
@@ -38,6 +38,7 @@ import (
 	"github.com/telepresenceio/telepresence/v2/pkg/maps"
 	"github.com/telepresenceio/telepresence/v2/pkg/tunnel"
 	"github.com/telepresenceio/telepresence/v2/pkg/types"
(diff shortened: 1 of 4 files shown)


thallgren:
Superseded: this lands on the v2.30.0 release branch ahead of the rc.1 prepare commit instead, so the release candidate exercises it.

PR #4216, opened 2026-07-21
Bugfixes for a 2.30.2 patch release

thallgren:
Collects the bugfixes for a 2.30.2 patch release. The branch originates at v2.30.1 (`f8f416c7a`, "Tidy the 2.30.1 changelog") so the same commits can be tagged for the patch release without dragging in any 2.31 development.

- **Fix a "symblic" typo in the `--vnat` and `--proxy-via` flag help** — also regenerates the CLI reference pages, which still carried the typo.
- **Let every progress Stop end its own Start cycle** — the TTY progress renderer guarded a per-cycle channel with a `sync.Once`, so only the first stop ever worked; stale render goroutines kept re-rendering spinner blocks over whatever the terminal showed next.
- **Keep a stale render goroutine from stopping the next run's ticker** — follow-up to the above: the goroutine selected on a mutable field and deferred a stop of the shared ticker, letting a goroutine from a previous run silence a run that had already begun.
- **Match warning events to owned Pods and ReplicaSets only** — the event watch's name-prefix match accepted any object whose name starts with the workload's name and a dash, attributing a sibling object's warnings to the workload and aborting install waits with someone else's error.
- **Add the 2.30.2 release notes.**

The progress and eventwatch fixes were extracted from #4214, where they were unrelated to the setup feature; #4214 carries the same patches to release/v2, while this PR brings them to the 2.30 line. The same four fixes were also contributed to the OpenAI fork in breland-openai/telepresence#3.


thallgren:
Superseded by #4217

Issue #4222, opened 2026-07-26
Closing a port-forward stream pair corrupts the shared per-pod stream connection


thallgren:
## Description

The client's port-forward machinery (`pkg/client/portforward/streamconn.go`) maintains one shared stream connection per pod (`podDialer`, keyed by pod UID). Each `dial` creates an error/data stream pair on that connection, and `portConn.Close` closes its own stream pair while leaving the connection up for other users (full teardown only happens when the refcount reaches zero).

In practice, closing a stream pair leaves the shared connection unusable for any stream created afterwards:

- a subsequent `CreateStream` on the same connection times out, and
- frames written on a stream pair created after the close never reach the pod (writes appear to succeed; the peer never sees them).

Verified empirically against a kind cluster with both the WebSocket-tunneled and (stream-created-after-close case) SPDY transports: handshake-exchange on port A, close it, then dial port B on the same connection — port B hangs; a second exchange on port A also hangs.

## Why nobody has hit it

Every current consumer opens exactly one long-lived stream pair per pod connection (the manager gRPC channel, and one gRPC channel per agent pod), and closes it only at teardown, where the refcount reaches zero and the whole connection is torn down with it. The corrupted state is unobservable in that pattern.

The one latent trigger is a gRPC transport redial over a still-healthy tunnel: `grpc.ClientConn` re-invokes the dialer, `dialContext` finds the existing `podDialer`, the new transport gets a second stream pair on the same connection, and the old transport's close then corrupts it under the new one. This requires the h2 transport to die while the underlying tunnel stays healthy (h2 protocol error, keepalive timeout with 24h keepalive time), which is rare, and the failure self-heals after one connect deadline (the broken connection is eventually torn down and evicted, and the next dial builds a fresh one). Applies equally to manager and agent channels.

## How it was found

The x509 client-authentication work (2.31.1) initially dialed the manager's auth-only TLS port as a second, short-lived stream pair on the manager channel's connection — the first caller ever to close a stream pair while another stayed in use. The handshake succeeded and its cleanup deterministically hung the gRPC channel (`ArriveAsClient` HEADERS written into a dead connection, 60s deadline). That path now uses `portforward.OneShotDialer`, a dedicated connection per exchange, which removes the only deterministic trigger.

## Suggested fixes (either suffices)

[...]

PR #4223, opened 2026-07-26
Add x509 client-certificate authentication of clients

thallgren:
A kubeconfig whose credentials are a client certificate yields no bearer token, so an `enforcing` traffic-manager rejected such clients (common on kubeadm, bare-metal, and kind clusters).

The traffic-manager now runs an auth-only TLS listener when authentication is enforced. A cert-only client performs a one-shot TLS handshake against it, presenting its kubeconfig client certificate; the manager verifies the certificate against the cluster client CA and returns a short-lived bearer token for the regular gRPC channel. The client thereby gets an authenticated identity, so session binding and the SubjectAccessReview-based intercept authorization now work for certificate clients too.

Active by default under `security.authentication.mode: enforcing`; opt out with `security.authentication.x509.enabled=false`. The chart then also creates a RoleBinding in `kube-system` to the stock `extension-apiserver-authentication-reader` Role so the manager can read the client CA. See `docs/reference/authentication.md` for details.

The handshake uses a dedicated port-forward connection; sharing the gRPC channel's connection surfaced #4222.

The `docs/plans/x509-client-auth` folder accompanies the review and will be dropped before merge.


thallgren:
I performed a detailed review of this PR and wanted to capture the findings here.

Overall I think this is a well-designed approach. The separation of the x509 exchange onto its own connection is clean, the identity mapping matches Kubernetes' client-certificate conventions, and the implementation is easy to follow.

## 1. Minted tokens are tied to a single manager process (blocking)

`MintedTokens` is an in-memory map. The x509 handshake is performed over one dedicated port-forward connection, while the subsequent gRPC connection may reconnect to a different traffic-manager pod (or the original pod may have restarted). A token minted by pod A will be unknown to pod B, fall through to TokenReview, and fail.

This introduces a hidden dependency on a single manager replica and also affects reconnects after manager restarts.

Possible approaches:
- make the token self-verifying (for example manager-signed),
- force re-authentication whenever the manager pod changes, or
- explicitly enforce/document the single-manager assumption.

At minimum the client should discard an unrecognized minted token and immediately perform a fresh x509 exchange.

## 2. Expired tokens are only removed on lookup

Entries are only evicted when that exact token is looked up after expiry. A long-running manager can therefore accumulate expired entries indefinitely. A periodic cleanup or opportunistic pruning during `Mint()` would avoid unbounded growth.

## 3. `activate()` preserves cached tokens

If the client reconnects to another manager pod, `activate()` updates the dial target but leaves any cached token intact. Clearing the cached token whenever activation changes would avoid reusing a token minted by a previous manager.

## 4. Listener shutdown

`Serve()` appears to continue looping if the listener is closed independently of context cancellation. Treating `net.ErrClosed` as a terminal condition would make shutdown behavior more robust.

The first point is the only one I'd consider blocking. The remaining points are relatively minor.

thallgren:
## Security review

Reviewed at head `41faf2389435b0413b30acfadab7fa71d02efade`.

**Verdict: changes recommended before merge.** I found no broken cryptography or direct private-key disclosure, but there are two significant security-boundary issues and several denial-of-service and credential-lifecycle weaknesses.

### 1. High in clusters using certificate UIDs: distinct identities can collapse into one

The new verifier derives a principal using only:

```go
Username: leaf.Subject.CommonName
Groups:   leaf.Subject.Organization
```

It leaves `Principal.UID` empty.

That no longer matches current kube-apiserver x509 semantics. Kubernetes' `CommonNameUserConversion` also extracts the x509 UID extension and includes a certificate credential ID. UID parsing has been enabled by default since Kubernetes 1.33, and Telepresence currently builds against Kubernetes 1.36.2 libraries.

This becomes security-relevant because `Principal.SameAs` treats an absent UID as a wildcard:

```go
if p.UID == "" || o.UID == "" {
    return true
}
```

Consequently, two certificates with the same CN but different Kubernetes UIDs are considered the same principal for session ownership. A second identity could operate an existing session if it obtained its session ID.

The missing UID is also propagated into `SubjectAccessReview`, potentially making authorization results differ from those obtained by the certificate holder at the API server.

There is a related discrepancy: kube-apiserver adds `system:authenticated` to every successfully authenticated non-anonymous identity, according to its authenticated-group rules, but the new x509 path does not. With standard RBAC this will mostly create false denials for bindings to `system:authenticated`; with custom webhook authorizers, omitting an identity attribute can change the decision in either direction.

**Recommended fix:**

- Convert the successful verified chain using Kubernetes' x509 user-conversion semantics.
- Preserve the UID in `Principal`.
- Add `system:authenticated` using the same rules as kube-apiserver.
- Consider extending `Principal` and the SAR construction to preserve relevant `Extra` fields, particularly the x509 credential ID, if exact authorization parity is the goal.
- Add a test with two certificates sharing a CN but carrying different UID extensions.

This is the strongest security finding because the PR's purpose is to establish a Kubernetes identity, and its derived identity is currently weaker than the one established by kube-apiserver.

### 2. High availability risk in flat pod networks: the listener and token store are unbounded

The listener binds to `0.0.0.0`, accepts connections indefinitely, and starts an unrestricted goroutine for each connection.

Because TLS uses `RequireAnyClientCert`, an attacker can generate an arbitrary self-signed certificate, complete an expensive TLS 1.3 handshake, reach manual certificate verification, and produce a warning log entry. No valid cluster identity is required for that part of the attack. A compromised ordinary workload that can reach the manager Pod IP could therefore combine:

- file-descriptor and goroutine pressure;
- TLS and x509 CPU consumption;
- ten-second slow-connection occupancy;
- warning-log flooding;
- periodic API-server reads through failure-triggered CA reloads.

A valid but low-privilege certificate holder can do more: each successful handshake mints a new 256-bit token and adds it to an unbounded map. There is no per-principal reuse, connection rate limit, or map capacity.

Expired entries are removed only if that exact expired token is later presented. Normal clients stop using their tokens before expiry, so old entries will usually never be looked up again. The map therefore grows even under normal long-running use, and an authorized client can intentionally grow it much faster.

Periodic expiry cleanup alone would not fully address the attack: a client could still fill the heap with one hour's worth of tokens. The store needs a hard bound.

**Recommended fix:**

- Put a global limit on simultaneous TLS handshakes.
- Use a bounded TTL cache with an explicit maximum entry count.
- Prefer one reusable active token per certificate fingerprint or principal instead of minting an unlimited number.
- Add rate limiting after identity verification, ideally keyed by certificate fingerprint.
- Rate-limit rejection logs or log routine invalid certificates at debug level while incrementing metrics.
- Add a short timeout and singleflight/serialization around CA reloads.
- Document a NetworkPolicy that restricts direct access to the listener, or bind more narrowly if Kubernetes port-forwarding permits it.

I would treat this as blocking where manager Pod IPs are reachable from untrusted workloads, which remains common in clusters without default-deny network policies.

### 3. Medium: the bearer token can outlive the certificate and its trust chain

`MintedTokens.Mint` always uses a fixed one-hour expiry. It does not consider the leaf certificate's `NotAfter`, any intermediate certificate expiry, or the current CA-bundle generation.

The verifier already receives the verified chains from `leaf.Verify`, but discards them.

This permits scenarios such as:

- A certificate that expires in 30 seconds authenticates and receives a token usable for another hour.
- A CA is removed from the cluster trust bundle, but tokens previously minted under it remain usable.
- Because the CA pool refreshes every ten minutes, an old certificate can mint a token shortly before refresh and remain accepted for approximately another hour afterward.
- If subsequent ConfigMap reads fail or contain invalid data, the last successfully loaded trust pool remains active indefinitely.

Some grace during CA rotation may be desirable, but the current behavior is implicit and extends trust beyond the source credential's validity.

**Recommended fix:**

- Set token expiry to the minimum of:
  - the configured token TTL;
  - the leaf certificate's `NotAfter`;
  - the earliest `NotAfter` in the selected verified chain.
- Give each successfully loaded CA pool a generation number or fingerprint.
- Associate minted tokens with that generation and invalidate them when the trusted bundle changes, or explicitly document a deliberate grace policy.
- Serialize CA reloads so an older concurrent GET cannot overwrite a newer trust pool.
- Consider a substantially shorter token TTL; the exchange is already designed to be repeated.

### 4. Medium: a hostile or misrouted auth endpoint can hang or exhaust the client

The client reads the auth response using an unrestricted:

```go
body, err := io.ReadAll(conn)
```

There is no response-size limit. A server can stream arbitrary data until the client exhausts memory.

The connection receives a deadline only when the calling context already has one. More importantly, the port-forward `net.Conn` implementation silently returns success from `SetDeadline` when its underlying HTTP stream is not itself a `net.Conn`; in that case the requested deadline has no effect.

After the TLS handshake, `io.ReadAll` is not context-aware. A server that simply never closes can therefore block the token source indefinitely. `Token()` also holds its mutex across the whole network operation, so every concurrent RPC waiting for credentials is then blocked behind the same connection.

The legitimate manager will not behave this way, but this matters because server-certificate verification is deliberately disabled and endpoint authenticity rests entirely on correct port-forward routing.

**Recommended fix:**

- Apply an independent, short timeout to the complete exchange.
- Arrange for context cancellation to actively close the connection rather than relying on `SetDeadline`.
- Decode one JSON object through an `io.LimitReader`; 8–16 KiB is already much larger than this response needs.
- Reject trailing data after the single object.
- Avoid holding the main state mutex during network I/O; use a separate singleflight mechanism for refresh.
- Clamp the server-supplied expiry to a locally defined maximum.

### 5. Medium-low: the selected Pod UID is recorded but not enforced

The client resolves the manager Service to a specific Pod and records its UID. The x509 address includes that UID:

```text
pod/<name>.<namespace>:<port>#<uid>
```

However, `OneShotDialer` parses the UID and then discards it. It opens the port-forward using only Pod name and namespace.

This weakens the stated trust argument that skipping server verification is safe because the API server routes the connection to the already selected manager Pod. Between the unauthenticated `Version` call and the x509 exchange, a deleted-and-recreated Pod with the same name would be accepted without checking its UID.

This does **not** expose the private key—the TLS handshake proves possession without transmitting it—but it means the code is not enforcing the endpoint identity it claims to rely on.

Additionally, `activate()` does not invalidate a cached token when the target changes, while the gRPC Service resolver can select another Pod after reconnecting.

**Recommended fix:**

- Verify the current Pod UID before opening the second port-forward, or make `OneShotDialer` enforce `PodID`.
- Clear the cached token whenever Pod name, UID, port, or manager process changes.
- A stronger design would have the initial `Version` response advertise an ephemeral public-key or certificate fingerprint and pin the x509 handshake to it.

### Additional hardening

The runtime code starts and advertises the listener whenever authentication is anything except `disabled`, whereas the Helm chart intends x509 authentication to exist only in `enforcing` mode. The chart prevents the normal mismatch, but a manually supplied `AUTH_X509_PORT` enables unnecessary attack surface in `permissive` mode. The code should enforce `ModeEnforcing` too.

The plan calls the minted credential "session-scoped," but the implementation is actually a replayable, principal-scoped bearer token. It is not linked to a Telepresence session and remains valid after a client session departs. Either bind it to a session/one-time bootstrap exchange or change the documentation and threat model accordingly.

### What looks sound

The cryptographic primitives are appropriate:

- TLS 1.3 is required.
- The server certificate and key are generated using ECDSA P-256 and `crypto/rand`.
- Certificate verification uses Go's standard chain verifier with the client-auth EKU.
- Tokens contain 256 bits of randomness generated by `crypto/rand`.
- Token contents are not written to logs.

The Helm RBAC addition also appears appropriately narrow: it binds only the traffic-manager ServiceAccount to Kubernetes' stock `extension-apiserver-authentication-reader` Role. I found no chart-level privilege-escalation issue.

### Security tests I would add before merge

1. Two valid certificates with the same CN and different UID extensions remain distinct session owners.
2. The x509 principal includes `system:authenticated` according to kube-apiserver's rules.
3. A token cannot outlive the verified certificate chain.
4. A CA-bundle change invalidates or predictably limits previously minted credentials.
5. The token store remains bounded under repeated valid handshakes.
6. Excess simultaneous TLS connections are rejected or queued within a fixed limit.
7. A hanging or oversized auth response cannot block or exhaust the client.
8. A Pod UID mismatch prevents the one-shot exchange.
9. Reconnection to another manager process clears and refreshes the minted token.

I would block the merge on findings **1 and 2**, and strongly recommend fixing **3 and 4** in the same patch because this is new authentication code being introduced in a patch release. Finding 5 may be acceptable as a follow-up only if the singleton-manager and manager-namespace trust assumptions are stated explicitly.

This was a source-and-test review; I did not execute the test suite locally.

thallgren:
## Re-review after security fixes

Reviewed at head `65ad72beef42f4b02b94f3e207b0f48a4a89e6c0`.

**Verdict: changes still recommended before merge.** Most of the original findings have been addressed well, but I found two remaining blockers, one important timeout gap, and three smaller identity-parity issues.

The visible build, lint, and unit-test jobs pass on Ubuntu, macOS, and Windows; the license and image-scan workflows also pass.

### 1. P1: a CA-bundle change can still mint a token from the old CA after invalidation

The new CA-change invalidation has a race:

1. A handshake loads the old `CertPool` and successfully verifies a certificate.
2. The reload goroutine stores the new pool.
3. `OnChange` calls `MintedTokens.InvalidateAll()`.
4. The already-verified handshake resumes and calls `Mint`, adding a new token for a certificate accepted only by the old CA.

Verification and minting are separate from the CA-pool swap and token invalidation. The reload stores the new pool, unlocks, and only then invokes the invalidation callback. `InvalidateAll` only replaces the current maps; it has no generation or epoch that a concurrent `Mint` must match.

This matters particularly when an old or compromised CA is removed. An attacker with a certificate under that CA can continuously run handshakes around the reload and potentially obtain a token valid for up to another hour.

**Suggested fix:** make the CA pool an atomic snapshot containing both the pool and a generation number. Verification returns that generation, and `Mint` refuses to insert the token unless it still matches the token store's current CA generation. Alternatively, put an RW lock around the complete verify-and-mint sequence and have reload-plus-invalidation take the write lock.

A deterministic test can pause a handshake after old-pool verification, reload the bundle and invalidate, then allow the handshake to continue and verify that no old-generation token is accepted.

### 2. P1: invalidated or process-local tokens still have no refresh path

The manager now deliberately invalidates all minted tokens when the CA bundle changes. Tokens can also disappear because:

- the 1,024-entry store evicts one;
- the manager container restarts;
- a rollout replaces the manager Pod;
- the gRPC resolver moves the connection to another manager process.

The client nevertheless keeps returning its cached token until its locally calculated expiry. There is no callback from an `Unauthenticated` response into `x509TokenSource`, and `GetRequestMetadata` only obtains the next token—it cannot see why the previous call was rejected.

`activate()` clears the cache, but it is called once after the initial `Version` exchange. It is not called when the gRPC resolver changes Pods. The resolver can independently update the gRPC channel to a newly resolved Pod, while the x509 source retains the original Pod name and UID.

This is relevant during a normal rolling update: the Deployment permits a surge Pod while the old Pod is still serving.

The documentation currently says that after restart or CA-bundle change, the client transparently repeats the handshake when its token is no longer accepted. That behavior is not currently implemented.

**Suggested fix:**

- Let the resolver notify the x509 source whenever the selected manager Pod changes, updating the Pod address and invalidating the cached token.
- Add a client interceptor that invalidates the x509 token after an `Unauthenticated` response. The next RPC can then perform a new handshake. Retrying the current RPC automatically should be limited to calls known to be safe to retry.
- Alternatively, make the short-lived tokens verifiable across manager processes using a shared or persisted signing key and CA generation.

Tests should cover CA-bundle change, token-store eviction, same-Pod container restart, and a rolling update to a new Pod.

### 3. P2: the ten-second timeout still does not bound the entire exchange

#### Server side

`handleConn` sets a network deadline, but it passes the manager's long-lived context into `verify`. If certificate verification fails, `ReloadOnFailure` can perform a Kubernetes API `GET` using that context.

The reload holds `ClientCAPool.mu` while making the API call. There is no per-connection deadline on that request.

If the API server stalls:

- one invalid handshake can block indefinitely in the ConfigMap `GET`;
- subsequent rejected handshakes can block waiting for the same mutex;
- all 16 handshake slots can eventually be occupied indefinitely.

The TCP deadline does not interrupt a Kubernetes API request.

Use a per-exchange context:

```go
exchangeCtx, cancel := context.WithTimeout(ctx, x509HandshakeTimeout)
defer cancel()
```

That context should be used for both `HandshakeContext` and CA reload. A separate, shorter reload timeout may be preferable.

#### Client side

The client timeout is created inside `handshake`, after waiting for `exchangeMu`. Therefore waiting for another exchange is not covered by the caller's context or `x509ExchangeTimeout`.

More importantly, the real one-shot port-forward path uses the outer cluster context when it creates the stream connection. Only the preliminary Pod-UID `GET` uses the timed `grpcCtx`.

The SPDY path then invokes `Dial` without a context. Kubernetes' SPDY dialer creates a plain `http.NewRequest`, and the Telepresence-created `http.Client` has no timeout.

Consequently, a stalled API server, proxy, or port-forward upgrade can still hang indefinitely before the watchdog goroutine is installed.

The port-forward upgrade needs either a context-bearing request or an `http.Client.Timeout` derived from the remaining exchange deadline. Waiting behind the in-flight exchange should also be context-aware, for example through a result channel or a singleflight-like mechanism rather than an ordinary mutex.

### 4. P2: `system:anonymous` is still mapped differently from kube-apiserver

The new conversion correctly uses Kubernetes' `CommonNameUserConversion`, but the subsequent group addition is not quite identical to kube-apiserver.

The PR adds `system:authenticated` whenever neither authentication group is already present. Kube-apiserver first checks whether the username is `system:anonymous`. When it is, the user is returned unchanged and `system:authenticated` is not added.

Therefore, a valid client certificate with:

```text
CN=system:anonymous
```

and no organization is assigned `system:authenticated` by the manager but not by kube-apiserver. A RoleBinding granting `pods/portforward` to `system:authenticated` could consequently produce a different intercept-authorization result.

The fix is small:

```go
if info.GetName() != user.Anonymous &&
    !slices.Contains(groups, user.AllAuthenticated) &&
    !slices.Contains(groups, user.AllUnauthenticated) {
    groups = append(groups, user.AllAuthenticated)
}
```

Rejecting a certificate whose CN is `system:anonymous` would also be a safe, stricter policy, but would no longer exactly reproduce kube-apiserver.

### 5. P2: a missing UID is still treated as a wildcard identity

The original same-CN/different-UID issue is fixed when both certificates contain UIDs. However, `SameAs` still returns true when only one principal has a UID:

```go
if p.UID == "" || o.UID == "" {
    return true
}
```

The test suite explicitly codifies this behavior.

This means:

- certificate A: `CN=alice`, `UID=uid-a`
- certificate B: `CN=alice`, no UID

are treated as the same session owner. If UID is intended to distinguish Kubernetes identities, an identity that lacks the UID should not act as a wildcard for a UID-bound session.

I would use:

```go
if p.Username != o.Username {
    return false
}
return p.UID == o.UID
```

That treats two UID-less identities with the same username as equal, while treating empty-versus-nonempty as different. If the wildcard behavior is needed for compatibility, it should be explicitly documented as a weakening of session identity rather than described as exact identity matching.

### 6. P3: x509 UID semantics follow the manager's library version, not necessarily the cluster

`CommonNameUserConversion` parses the x509 UID only when the process-global `AllowParsingUserUIDFromCertAuth` feature gate is enabled.

The traffic-manager is built against Kubernetes 0.36.2 libraries. In those libraries, UID parsing is enabled by default beginning with Kubernetes 1.33.

That feature gate belongs to the **traffic-manager process**, not the remote kube-apiserver. On:

- a pre-1.33 cluster, or
- a newer API server where the feature gate is explicitly disabled,

the traffic-manager may extract and submit an x509 UID that the API server itself would not have assigned to that certificate.

Standard RBAC generally keys user grants by username, so this will often have no effect. It can matter to webhook authorizers that inspect UID or extra attributes, and it makes "exactly as the API server would" stronger than the implementation can guarantee.

This is inherent in locally reproducing kube-apiserver authentication. At minimum, document that conversion follows the Kubernetes library semantics compiled into the manager. A cluster-version-aware switch plus an operator override would reduce the mismatch, though it still cannot discover an explicitly overridden API-server feature gate.

## Previously reported issues that are now addressed

The updated PR substantially improves the implementation:

- Kubernetes' x509 conversion is used, and UID, credential ID, groups, and `Extra` are forwarded into `SubjectAccessReview`.
- Tokens are reused per certificate fingerprint, capped at 1,024 entries, pruned on mint, and clamped to certificate-chain expiry.
- Concurrent TLS handshakes are capped, rejection logging is rate-limited, and independent listener closure now terminates the accept loop.
- Client responses are size-limited, trailing content is rejected, and a watchdog closes established connections when the exchange context expires.
- The one-shot dial verifies that the Pod UID still matches, and both listener creation and capability advertisement are now restricted to enforcing mode.

The two blockers are therefore narrower than in the original review: **make CA invalidation atomic with minting, and provide an actual client refresh/rebinding path after a token becomes invalid.**

thallgren:
## Re-review

Reviewed at current head `19b73dfb4cba4143c3a073f6fe633f04480f1fe6`.

**Verdict: changes recommended before merge.** The previous findings have been addressed substantially, but I found three remaining blocking correctness problems and several smaller concurrency issues.

### 1. P1: x509 authentication never recovers if the initial CA load fails

The service initializes `MintedTokens.caGeneration` from the pool's initial snapshot. If the initial ConfigMap read fails, that generation is `0`.

A later successful reload does this:

```go
changed := p.bundleHashSet && hash != p.bundleHash
first := !p.bundleHashSet

if first || changed {
    p.generation++
}
...
p.snapshot.Store(...)
...
if changed && onChange != nil {
    onChange(generation)
}
```

On the first successful load after a startup failure:

- `first` is true;
- the pool becomes generation 1;
- `changed` is false;
- `onChange` is therefore not called.

The token store remains at generation 0, while verification returns generation 1. `Mint` consequently rejects every certificate with `ErrCAGenerationMismatch`. Reverification produces generation 1 again, so the retry cannot recover either.

This can happen after a transient API-server failure, temporarily missing ConfigMap, delayed RBAC propagation, or an operator fixing manually managed RBAC without restarting the manager. The listener appears healthy but can never mint a token until the manager restarts or the CA content changes a second time.

**Suggested fix:** notify the token store on every generation transition, including the first successful load after generation 0. Because the callback is installed after the constructor's initial attempt, changing the final condition to account for `first && onChange != nil` would handle delayed recovery.

Add a test that:

1. constructs the pool while the ConfigMap is unavailable;
2. installs the callback and initializes the token store at generation 0;
3. makes the ConfigMap available;
4. reloads;
5. verifies that minting with generation 1 succeeds.

### 2. P1: the x509 exchange is still not bound to the process serving the gRPC connection

`ConnectToManager` initially resolves one specific pod, `pap`, and uses it as the known address for the gRPC connection. But the x509 source does not retain that selected pod. It installs a resolver that independently resolves the Service again on every handshake.

Meanwhile, the gRPC resolver keeps its existing `lastPA` and waits for gRPC to request a new resolution. It does not proactively move to whichever pod a subsequent independent Service lookup would select.

This violates the required invariant:

> The pod minting a process-local token must be the pod receiving the RPC carrying that token.

It can fail during an ordinary rolling update because the Deployment deliberately permits an old and new manager pod to coexist using `maxSurge: 1` and `maxUnavailable: 0`.

For example:

1. The gRPC channel is connected to manager A.
2. Manager B becomes ready during a rollout.
3. A token refresh independently resolves B.
4. B mints the token.
5. The RPC still goes to A.
6. A cannot find B's token and returns `Unauthenticated`.

The unary interceptor invalidates the token, but it deliberately does not retry the failed RPC. If this happens on the initial `ArriveAsClient`, `connectMgr` returns the error and the connection attempt fails; there is no subsequent RPC within that connection attempt to trigger the promised recovery.

The same problem can occur when:

- the token expires during a rolling-update overlap;
- the token is evicted from the bounded store;
- the client CA changes during a rollout;
- the configured x509 port changes between old and new manager pods.

Re-resolving "the current Service pod" is not equivalent to identifying the pod serving the existing gRPC subconnection.

**Suggested fix:** expose the actual pod selected by the gRPC resolver/dialer to the x509 source, and perform the separate auth connection against that same pod UID. Another robust solution is making tokens verifiable across manager processes using an installation-scoped signing key.

At minimum, the initial handshake should use `pap`, but full correctness also requires notifying the x509 source whenever the gRPC resolver switches pods.

### 3. P1: the resolver discards the exchange deadline

`x509TokenSource.Token` now correctly creates a ten-second context before waiting for the exchange gate, and passes that context into `handshake`.

However, the resolver installed by `ConnectToManager` ignores that context:

```go
resolvePod := func(context.Context) (*portforward.PodAddress, error) {
    return portforward.ResolveSvcToPod(kc, "traffic-manager", namespace, "8081")
}
```

It uses the long-lived cluster context `kc` instead. `ResolveSvcToPod` relies on its supplied context for API calls and its pod-wait timeout; without a deadline, it falls back to one minute for the watch, while the initial Service and Pod API requests can still wait on the underlying transport.

Consequences:

- The advertised ten-second bound does not cover manager-pod resolution.
- A stalled API request can leave the first `Token` call blocked indefinitely.
- That call retains the one-slot exchange gate.
- Later callers time out waiting for the gate, but none can initiate a replacement handshake.
- The source remains wedged until the whole manager connection is replaced.

The port-forward upgrade has also been given its own fixed ten-second timeout rather than the remaining duration of the exchange context. Because `OneShotDialer` uses its outer cluster context for `newPodDialer`, the upgrade can continue after the original exchange context expires, making the total operation additive rather than bounded to ten seconds.

**Suggested fix:** preserve the cluster values from `kc` while applying the deadline and cancellation from the context passed to the resolver. Also derive the port-forward upgrade timeout from the remaining exchange deadline rather than starting a new ten-second budget.

Add tests with:

- a resolver that blocks until its context is cancelled;
- a port-forward upgrade that outlives the remaining exchange budget;
- a second caller waiting behind the first.

### 4. P2: streaming `Unauthenticated` responses usually bypass the invalidation interceptor

The stream interceptor invalidates the cache only when `streamer(...)` itself returns `Unauthenticated`. It returns the resulting `grpc.ClientStream` unchanged.

For an established stream, the final server status is generally returned later by `Header` or `RecvMsg`. In particular, a server-streaming authentication failure will normally surface when the client receives from the stream, not necessarily while the stream object is being created.

The current test mocks a `streamer` that directly returns an error, so it does not exercise the normal case in which stream creation succeeds and the manager's authentication rejection appears on the first receive.

A watch that reconnects after token invalidation can therefore repeatedly present the same stale token until:

- a unary RPC happens to receive `Unauthenticated`;
- the cached token reaches its local expiry;
- or the entire manager connection is recreated.

**Suggested fix:** either wrap the returned `grpc.ClientStream` and inspect errors from `Header`, `SendMsg`, and especially `RecvMsg`, or append `grpc.OnFinish` to the stream call options so the final RPC status can invalidate the cache.

A real in-process gRPC test should have the server reject a server-streaming RPC in its auth interceptor and verify that the x509 token cache is cleared when `RecvMsg` returns `Unauthenticated`.

### 5. P2: CA generation publication and token invalidation are still not atomic or monotonic

The new generation check is a good improvement, but the sequence remains:

1. publish the new CA snapshot;
2. release `ClientCAPool.mu`;
3. invoke the token-store callback.

That leaves two races.

#### Old-generation minting window

After the new snapshot is published but before `InvalidateAll(newGeneration)` runs, the token store still considers the old generation current. A handshake that had already verified against the old snapshot can therefore still pass the generation check and mint a token. The callback will eventually invalidate it, but the mint is not atomically excluded as the code comments claim.

#### Out-of-order callbacks can regress the generation

Because the callback runs after releasing `ClientCAPool.mu`, another reload can publish a later generation and invoke its callback first. A possible order is:

```text
publish generation 2
publish generation 3
InvalidateAll(3)
InvalidateAll(2)
```

`InvalidateAll` assigns the supplied generation unconditionally, so the token store can regress to generation 2 while the current snapshot is generation 3. All generation-3 mint attempts then fail indefinitely.

The existing race test calls `ReloadOnFailure` synchronously and only attempts minting after the callback has completed, so it does not exercise either publication gap or callback reordering.

**Suggested fix:** make the CA snapshot transition and token-generation transition one ordered operation. At minimum:

- make `InvalidateAll` monotonic—ignore generations older than the current one;
- update the token-store generation before exposing the new CA snapshot;
- or use shared synchronization around verification/minting and CA rotation.

A token entry can also carry its CA generation and have `Lookup` reject it against an atomically published current generation, which avoids depending entirely on map clearing.

### 6. P3: the defensive reload timeout can leak one goroutine per attempt

`reload` starts the ConfigMap `Get` in a goroutine and stops waiting after five seconds. That bounds the mutex hold time even if the client ignores context cancellation, but the uncooperative `Get` goroutine itself remains blocked.

Rejected certificates can trigger another reload every 30 seconds, so the exact failure mode the code is designed to survive can accumulate blocked goroutines indefinitely.

The current test unblocks the fake reactor during cleanup, so it confirms that `reload` returns but does not test persistent goroutine accumulation.

The production client-go HTTP transport should respect context cancellation, so a direct context-bounded `Get` is probably sufficient. If protection against a genuinely uncooperative implementation is desired, retain at most one outstanding detached request rather than starting another after every cooldown.

## What is now correctly addressed

The previous review's identity concerns are fixed:

- Kubernetes' x509 conversion is used.
- UID and credential ID are preserved.
- `system:anonymous` is handled consistently with kube-apiserver.
- `Extra` is forwarded into `SubjectAccessReview`.
- Session identity now compares an empty UID exactly rather than treating it as a wildcard.

The token and listener hardening is also materially better:

- one active token per certificate fingerprint;
- bounded token-store size;
- expiry clamped to the certificate chain;
- concurrent-handshake limit;
- response-size limit;
- connection watchdog;
- listener shutdown handling;
- Pod UID verification;
- enforcing-mode-only activation.

The visible GitHub workflows are green: image scanning, license verification, and build/unit testing all passed; build, lint, and tests succeeded on Ubuntu, macOS, and Windows.

I would block merge on findings **1–3**, and fix **4** before relying on transparent token recovery. Finding **5** should also be resolved in authentication-critical code rather than left as a theoretical generation race.

This was a source review plus inspection of GitHub's reported CI results; I did not run the test suite locally.

PR #4237, opened 2026-08-05
Reconcile agent configs after Service selector updates

breland-openai:
## Description

When a Service selector changes, the watcher currently only revisits configs already tied to that Service UID. That leaves newly selected workloads stale, and a selectorless Service can accidentally look like it matches every workload.

This makes updates consider both existing claimants and newly selected workloads, while keeping selectorless Services from matching by labels. I added a focused regression test for switching a selector from one workload to another and then removing it.

Tested with `GOEXPERIMENT=jsonv2 go test ./cmd/traffic/cmd/manager/mutator -run TestUpdatedServiceIncludesNewlySelectedWorkload -count=1`.

## Checklist

### DIFF
--- a/cmd/traffic/cmd/manager/mutator/service_watcher.go
+++ b/cmd/traffic/cmd/manager/mutator/service_watcher.go
@@ -24,7 +24,7 @@ type affectedConfig struct {
 	sc  *agentconfig.Sidecar
 }
 
-func (c *configWatcher) configsAffectedBySvc(ctx context.Context, svc *core.Service, trustUID bool) []affectedConfig {
+func (c *configWatcher) configsAffectedBySvc(ctx context.Context, svc *core.Service, includeSelectorMatches bool) []affectedConfig {
 	references := func(ac *agentconfig.Sidecar) (k8sapi.Workload, error, bool) {
 		for _, cn := range ac.Containers {
 			for _, ic := range cn.Intercepts {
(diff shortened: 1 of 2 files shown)


cmd/traffic/cmd/manager/mutator/service_watcher.go
```diff
@@ -102,7 +109,7 @@ func (c *configWatcher) watchServices(ctx context.Context, ix cache.SharedIndexI
 			},
 			UpdateFunc: func(oldObj, newObj any) {
 				if newSvc, ok := newObj.(*core.Service); ok {
-					c.updateSvc(ctx, newSvc, true)
+					c.updateSvc(ctx, newSvc, false)
```
thallgren: I didn't fully understand this change at first. How can an UID change on an existing service? However, that's not what this boolean signifies. It's more like `matchUIDOnly` (i.e. "do not include selector matches", so passing `false` means "include selector matches").

Perhaps we should use opposite polarity for this parameter and call it `includeSelectorMatches`? We'd get:
```
AddFunc:    includeSelectorMatches = true
UpdateFunc: includeSelectorMatches = true
DeleteFunc: includeSelectorMatches = false
```

thallgren (approved):
LGTM. Fix the suggestion if you want, but it's OK to merge without.

PR #4238, opened 2026-08-05
Only evict pods owned by the workload being reconciled

breland-openai:
## Description

When two workloads share a selector, `workloadPods` currently returns every matching pod. During agent config reconciliation, that can let one workload evict pods belonging to another.

This resolves each matching pod back to its owning workload and keeps only pods owned by the workload being reconciled. It also follows Rollout pods through their owning ReplicaSet when ReplicaSets are not enabled as top-level workloads, so those pods still get reconciled.

I added regression coverage for both cases.

Tested with `GOEXPERIMENT=jsonv2 go test ./cmd/traffic/cmd/manager/mutator -count=1`.

## Checklist


thallgren (approved):
LGTM aside from the lint error.

PR #4239, opened 2026-08-05
Let warning watches stop when delivery is canceled

breland-openai:
## Description

`WatchWarnings` can block while delivering an event to a slow consumer. If the context is canceled at that point, the goroutine cannot unwind and the underlying Kubernetes watch stays alive.

This makes event delivery cancellation-aware, so the watch stops promptly even when nobody is receiving from the returned channel. I added a regression test that cancels while a matching warning is blocked in delivery.

Tested with `GOEXPERIMENT=jsonv2 go test ./pkg/eventwatch`.

## Checklist


thallgren (approved):
LGTM!

PR #4242, opened 2026-08-05
Avoid racing on cached Sidecar configs during injection

breland-openai:
## Description

`MarshalTight` temporarily clears creation-only fields on the shared Sidecar config before serializing it. Concurrent admission requests can observe that temporary state and build an injected container with an empty image.

This strips those fields from a local value copy instead, so serializing a tight config never mutates the cached Sidecar. I added both a direct mutation check and a concurrent container-build regression test.

Tested with `GOEXPERIMENT=jsonv2 go test ./pkg/agentconfig` and `GOEXPERIMENT=jsonv2 go test -race ./pkg/agentconfig -run TestMarshalTight -count=1`.

## Checklist


breland-openai:
I know we are sort of moving away from the side car - but this was a real issue/bug we hit.

thallgren:
I don't think we're moving away from the sidecar. It serves two prominent use-cases. One is the `telepresence replace` command. You can't really do that unless the container is replaced. The other is about what OS-capabilities that are needed on the node. A sidecar doesn't need any (when symbolic port names are used), but the node-agent is fairly intrusive. Cluster admins can be very restrictive with this.

thallgren (approved):

PR #4243, opened 2026-08-05
Avoid cache sends racing with subscription shutdown

breland-openai:
## Description

A cache update can race with subscription cleanup: one goroutine closes the subscriber channel while another is about to send a delta, which can panic with a send on a closed channel.

This guards close and send with the same subscription lock and makes late sends no-ops after closure. I added a focused regression that exercises sending after the subscription has closed, plus a repeated race run.

Tested with `GOEXPERIMENT=jsonv2 go test ./pkg/cache` and `GOEXPERIMENT=jsonv2 go test -race ./pkg/cache -run TestAllDeltaSendAfterSubscriptionClose -count=100`.

## Checklist

### DIFF
--- a/pkg/cache/map.go
+++ b/pkg/cache/map.go
@@ -39,11 +39,22 @@ func (delta *Delta[K, V]) Merge(other Delta[K, V]) {
 }
 
 type subscription[K comparable, V any] struct {
+	sync.Mutex
 	channel     chan Delta[K, V]
 	include     func(K, V) bool
-	initialized atomic.Bool
+	initialized bool
 	mark        atomic.Bool
 	doneCh      <-chan struct{}
+	closed      bool
+}
+
+func (sb *subscription[K, V]) close() {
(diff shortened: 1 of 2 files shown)


pkg/cache/map.go
```diff
@@ -308,6 +319,11 @@ func (ad *allDelta[K, V]) filteredDelta(initialized bool, include func(K, V) boo
 }
 
 func (ad *allDelta[K, V]) send(sb *subscription[K, V]) {
+	sb.Lock()
+	defer sb.Unlock()
+	if sb.closed {
+		return
+	}
 	initialized := sb.initialized.Swap(true)
```
thallgren: With a mutex present, there's no reason for `sb.initialized` to be an `atomic.Bool`. Please change it to a `bool`. Keep `mark` as `atomic.Bool`.

thallgren (approved):

PR #4244, opened 2026-08-05
Recover promptly when manager keepalives fail

breland-openai:
## Description

The session keepalive path can leave the client treating a half-dead manager connection as healthy and retrying indefinitely.

This repairs the manager connection instead of ending the client session. An expired session reconnects immediately, repeated transient failures reconnect after a short threshold, and a successful keepalive resets that threshold. The manager transport also uses a regular keepalive interval so a stalled connection is noticed sooner.

I added focused gRPC coverage for expired sessions, repeated transient failures, reset-after-success, and reconnect failures.

Tested with `GOEXPERIMENT=jsonv2 go test ./pkg/client/userd/trafficmgr ./pkg/client/k8s -count=1`.

## Checklist


thallgren (changes requested):
The premise is real: a half-dead manager connection is a permanent zombie today. The conn's gRPC keepalive is `Time: 24h` (`pkg/client/k8s/connect.go:166`) so the transport never detects death, the watcher streams block in `Recv()` without erroring — which means their `WatchWithRetry` repair hook (`reconnectManager()`, the actual recovery path) never fires — and the only call that notices, `Remain`, logs and swallows. Worth fixing.

But propagating the error doesn't recover — it disconnects. Codex's P1 has this right, and it goes a bit further than stated there:

- **One transient failure kills the session.** A single slow Remain (manager pod rescheduling, a port-forward hiccup, brief overload) now exits `remainLoop`. Today the next ping a minute later simply succeeds.
- **Nothing restarts it.** `session.Run()` returning just `clearSession`s (`daemon/grpc.go:148`) — and in `rootSessionInProc` mode the daemon quits. Worse, `remainLoop`'s deferred `Depart` runs on the way out: if the blip healed by then, Depart succeeds and deletes the session and all its intercepts server-side. The user reconnects and re-creates everything by hand.
- The changelog's "once again propagated" isn't accurate: even v2.19 only escalated `NotFound` (as `ErrSessionExpired` → reconnect) and swallowed everything else. That NotFound escalation is the valuable thing that got lost since.

Suggested shape:

1. Fix the root cause: a real keepalive on the manager conn (`connect.go:166`, e.g. Time 30-60s). The transport then detects half-death, every stream errors, and the existing `reconnectManager()` repair (`session.go:442`) heals the session with intercepts intact. This one line likely fixes your whole scenario by itself.
2. In `remain()`: `NotFound` → `reconnectManager()` immediately; 2-3 consecutive `DeadlineExceeded`/`Unavailable` → also try `reconnectManager()`; only return an error (ending the session) when the reconnect itself fails.

The bufconn test verifies the propagation but not the consequence; with the above it should target the reconnect behavior instead.

thallgren:
One addendum to my review — a concrete scenario that makes the requested redesign easy to validate: **close the laptop lid, go home, open it hours later, and resume with live intercepts.**

This works today, by design, on both halves:

- The manager keeps a silent client's session and intercepts for `CLIENT_CONNECTION_TTL` (default 24h), so hours of silence are budgeted for.
- On wake, the dead sockets *error* (interface change), the watcher streams fail out of `Recv()`, and `WatchWithRetry`'s repair runs `reconnectManager()` — fresh port-forward, `ReconnectClient` with `getCurrentInterceptInfos()`, which even revives a session the manager already GC'd. Meanwhile `remain()`'s failures during the bumpy re-establishment window (wifi handshake, VPN, OIDC token refresh on the port-forward dial) are swallowed — that swallow is load-bearing: it keeps the session alive during exactly the minutes when failures are guaranteed.

With this PR, the first post-wake `Remain` races network re-establishment, and the session usually loses: the ticker fires while the VPN is still handshaking, `remainLoop` exits, and the session is torn down before `reconnectManager()` gets its chance. The deferred `Depart` then either fails (network still down → orphaned intercept pinned to the workload until TTL) or succeeds (network just came up → deletes the session and intercepts that were about to be rescued). A deterministic "always survives" becomes a wake-up coin toss.

Note the sleep/wake case is the *opposite* of the zombie this PR targets: here the transport does error, so the existing repair path already handles it — the PR only subtracts tolerance. Whatever shape the fix takes (keepalive + reconnect-from-remain per my review), "survive lid-close, resume with live intercepts" is the regression test it must pass — including sizing the gRPC keepalive so an hours-long suspend doesn't itself count as death on wake; keepalive failure should funnel into reconnect, not session exit.

thallgren (approved):

thallgren:
Pushed two commits to this branch: a merge of the current `release/v2`, and a fix for the namespace watcher raised in the review comments. `StartNamespacesFromManager` now takes a manager-client provider, the same way the pod watcher does, so a retry after a manager repair opens its stream on the replacement connection instead of the closed one. Full regression suite passed locally on top of this together with #4246.

PR #4246, opened 2026-08-05
Make intercept-only lists independent of workload watchers

breland-openai:
## Description

`telepresence list --intercepts` only needs the current intercept specs, but it currently waits on the workload watcher snapshot first. If that snapshot is empty or slow, an active intercept can be missing from an intercept-only list.

This builds the intercept-only snapshot directly from current intercepts, grouping them by workload identity from the spec and keeping the existing full-workload path unchanged. When watcher data is available, the row still keeps its services, UID, replica counts, and agent version; without it, the intercept still appears.

The regression tests cover an empty watcher snapshot, cached workload enrichment, grouping, namespace filtering, and omitted intercept types.

Tested with `GOEXPERIMENT=jsonv2 go test ./pkg/client/userd/trafficmgr -count=1`.

## Checklist


thallgren (changes requested):
I like this approach, but the spec-only rows drop fields that are readily available:

- Access URLs vanish.
- JSON output loses `uid`, `replica counts`, `agent_version`, and `services`
- `workload_resource_type` flips from "DEPLOYMENT" to "Deployment"

Since `AddIntercept` pre-warms the watcher, `s.workloads[ns]` is populated in every normal flow. I'd suggest enriching rows from that cache when present and normalizing the kind string.

thallgren (approved):

thallgren:
Pushed two commits to this branch: a merge of the current `release/v2`, and a changelog entry for the 2.32.0 release notes. Full regression suite passed locally on top of this together with #4244.

Issue #4183, opened 2026-07-06
When Intercept, muti pod only one pod can forward traffic


id-id-id:
The problem is like https://github.com/telepresenceio/telepresence/issues/4085
I use intercept for a Rollout service，which has 2 pods. Only one pod can receiver traffic.

Command I used is  telepresence intercept my-service --port 8080:http
telepresence version  is 2.27.4
Log of the pod which can not forward traffic as follow:

2026-07-06 09:38:53.0621 DEBUG Accept got connection from 10.53.144.210:43912
2026-07-06 09:38:53.0622 DEBUG Creating tunnel to client 146c1b9e-2e7f-4eeb-963f-fb67c563e586 for id TCP 10.53.144.210:43912 -> 127.0.0.1:8080

telepresence version output is
OSS Client : v2.27.4
OSS Root Daemon : v2.27.4
OSS User Daemon : v2.27.4
OSS Traffic Manager: v2.27.4
Traffic Agent : xxxxxx:telepresence-tel2-2.27.4

All 2 pods log SetIntercepting 1 intercept But only 1 pod can send traffic to my local service.

The log for success pod
receiver a intercept ：
2026-07-06 11:35:17.3197 DEBUG handleIntercept: fwdState.HandlePort 8080 called with 1 intercepts
2026-07-06 11:35:17.3197 DEBUG SetWiretapping 0 wiretaps
2026-07-06 11:35:17.3197 DEBUG Controller for intercept database-service created
2026-07-06 11:35:17.3198 DEBUG SetIntercepting 1 intercepts
2026-07-06 11:35:17.3592 DEBUG tunneling/WatchDial-6: started call
2026-07-06 11:35:17.3592 DEBUG tunneling/WatchDial-6: WatchDial called from client 146c1b9e-2e7f-4eeb-963f-fb67c563e586

when receiver request:
2026-07-06 11:42:41.9846 DEBUG Accept got connection from 10.53.176.187:55902
2026-07-06 11:42:41.9846 DEBUG Creating tunnel to client 146c1b9e-2e7f-4eeb-963f-fb67c563e586 for id TCP 10.53.176.187:55902 -> 127.0.0.1:8080
2026-07-06 11:42:42.0234 DEBUG tunneling/Tunnel-4: started call

The log for fail pod
receiver a intercept
2026-07-06 11:35:17.3202 DEBUG handleIntercept: fwdState.HandlePort 8080 called with 1 intercepts
[...]

PR #4251, opened 2026-08-05
Support Service intercepts across multiple workloads

breland-openai:
## Description

A Service can select pods owned by more than one workload, but an intercept currently follows only the explicitly requested workload. Traffic that lands on another selected workload is then left out of the intercept.

For a concrete example we currently do something like: (this is usually an envoy or something similar doing weight based random request distribution)

                         | - > "main" deployment
                         |
kube svc --->  |
                         |
                         | - > "canary" deployment

We want to be able to intercept both - so that I can not worry about cases where the request might land on either.

This lets agents advertise the Service UID and port targets they can serve, discovers every selected workload, waits for one review from each participant, and publishes the participating workload identities on the intercept. Older agents keep the historical single-workload matching path during a rolling upgrade. Shared-Service expansion is limited to HTTP intercepts without --replace, where one logical intercept can safely span the selected workloads.

It also discovers selector-matching workloads before their pods are ready, keeps the participant set aligned when the selector changes, and will not activate shared coverage while any live replica is missing the selected Service target. That keeps a scaled-zero canary or a rolling upgrade from quietly bypassing the local handler.

Restored intercepts rebuild their participants before reconciliation, each agent uses its own matched container, and unsupported expansion cases fall back to the requested workload instead of failing the intercept.

Tested with:

    GOEXPERIMENT=jsonv2 go test ./cmd/traffic/cmd/manager/... -count=1
    GOEXPERIMENT=jsonv2 go test ./cmd/traffic/cmd/agent ./pkg/agentconfig -count=1

## Checklist

 - [x] I made sure to update ./CHANGELOG.yml (our release notes are generated from this file).
 - [ ] I updated CONTRIBUTING.md with any special dev tricks I had to use to work on this code efficiently.

### DIFF
--- a/CHANGELOG.yml
+++ b/CHANGELOG.yml
@@ -44,6 +44,20 @@ items:
           pod creation cannot corrupt the cached config or block a workload
           rollout.
       - type: feature
+        title: HTTP intercepts can follow every workload selected by one Service
+        body: >-
+          A Kubernetes Service can route requests to more than one workload,
+          such as stable and canary Deployments, while an intercept previously
+          followed only the workload named by the user. HTTP intercepts
+          without <code>--replace</code> now let each selected workload
+          participate in one logical intercept, so filtered traffic reaches
+          the same local handler regardless of which selected workload
+          receives it. If the selected workloads or their traffic-agents
+          cannot support shared interception, Telepresence warns and keeps
+          the historical single-workload behavior instead of failing the
+          intercept.
+        docs: reference/attachments/cli
+      - type: feature
         title: Guided traffic-manager setup with telepresence setup
         body: >-
           The new <code>telepresence setup</code> command analyzes the
--- a/cmd/traffic/cmd/agent/agent.go
+++ b/cmd/traffic/cmd/agent/agent.go
@@ -260,6 +260,50 @@ func TalkToManagerLoop(ctx context.Context, s State, info *rpc.AgentInfo) {
 	}
 }
 
+func advertisedInterceptTargets(ac *agentconfig.Sidecar) []*rpc.AgentInfo_InterceptTarget {
+	interceptTargets := make([]*rpc.AgentInfo_InterceptTarget, 0)
+	seenTargets := make(map[string]struct{})
+	for _, cn := range ac.Containers {
+		for _, ic := range cn.Intercepts {
+			if ic.ServiceUID == "" {
+				continue
+			}
+			key := fmt.Sprintf("%s/%d/%s/%s/%d",
+				ic.ServiceUID, ic.ServicePort, ic.Protocol, cn.Name, ic.ContainerPort)
+			if _, ok := seenTargets[key]; ok {
+				continue
+			}
+			seenTargets[key] = struct{}{}
+			interceptTargets = append(interceptTargets, &rpc.AgentInfo_InterceptTarget{
+				ServiceUid:      string(ic.ServiceUID),
+				ServiceName:     ic.ServiceName,
+				ServicePortName: ic.ServicePortName,
+				ServicePort:     int32(ic.ServicePort),
+				Protocol:        ic.Protocol.String(),
+				ContainerName:   cn.Name,
+				ContainerPort:   int32(ic.ContainerPort),
+			})
+		}
+	}
+	sort.Slice(interceptTargets, func(i, j int) bool {
+		a, b := interceptTargets[i], interceptTargets[j]
+		if a.ServiceUid != b.ServiceUid {
+			return a.ServiceUid < b.ServiceUid
+		}
+		if a.ServicePort != b.ServicePort {
+			return a.ServicePort < b.ServicePort
+		}
+		if a.Protocol != b.Protocol {
+			return a.Protocol < b.Protocol
+		}
+		if a.ContainerName != b.ContainerName {
+			return a.ContainerName < b.ContainerName
+		}
+		return a.ContainerPort < b.ContainerPort
+	})
+	return interceptTargets
+}
+
 func StartServices(g log.Group, config Config, srv State) (*rpc.AgentInfo, error) {
 	ac := config.AgentConfig()
 
@@ -340,6 +384,7 @@ func StartServices(g log.Group, config Config, srv State) (*rpc.AgentInfo, error
 			Mounts:      appMounts.ToRPC(),
 		}
 	}
+	interceptTargets := advertisedInterceptTargets(ac)
 
 	return &rpc.AgentInfo{
 		Name:      ac.AgentName,
@@ -367,7 +412,8 @@ func StartServices(g log.Group, config Config, srv State) (*rpc.AgentInfo, error
 				Version: version.Version,
 			},
 		},
-		Containers: containers,
+		Containers:       containers,
+		InterceptTargets: interceptTargets,
 	}, nil
 }
 
--- a/cmd/traffic/cmd/agent/fwdstate.go
+++ b/cmd/traffic/cmd/agent/fwdstate.go
@@ -86,12 +86,23 @@ func (pm *ProviderMux) CreateClientStream(ctx context.Context, tag tunnel.Tag, s
 	return pm.AgentProvider.CreateClientStream(ctx, tag, sessionID, id, roundTripLatency, dialTimeout)
 }
 
+// containerForIntercept returns the local container that should review and
+// serve an intercept. A service-scoped intercept can be shared by workloads
+// whose matching containers have different names, so the matched forwarder
+// target wins over the primary workload's container name in that case.
+func (fs *fwdState) containerForIntercept(spec *manager.InterceptSpec) string {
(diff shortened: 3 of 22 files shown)


thallgren (changes requested):
The gap is real (traffic escaping to unnamed workloads behind a shared Service) and the architecture is right — agent-advertised claims, per-workload reviews, published participants. Proto changes are wire-safe (fields 17/23 free, no new RPCs, both degradation directions covered). But there's one blocker and a few regressions that bite outside the feature:

- **Reverts #4183's fix.** `78d93aa3f` (v2.29.3) deleted `interceptNeedsDialWatcherOnAllPods` — "the mechanism/filter distinction had no basis" — and made every agent pod of an intercepted workload report intercepted. This PR reintroduces that function and narrows `IsInterceptedBy` back to recorded-PodIp-or-http/filters, so plain-`tcp` and `--replace` intercepts on multi-replica workloads again time out on non-recorded replicas. The rewritten `TestIsInterceptedBy` assertions (`require.False` for a sibling pod of the tcp intercept) invert the regression test that guarded this. The fix strengthens your feature: return true for any pod whose agent passes `AgentMatchesIntercept` — no PodIp clause, no mechanism function. `integration_test/multi_replica_intercept_test.go` should catch this.
- **Pod-churn flapping.** `consolidateAgentSessionIntercepts` drops the `agent.PodIp != intercept.PodIp` guard, so *any* sibling-pod death now flips an ACTIVE intercept to WAITING and clears PodIp/PodName. Keep the recorded-pod guard for non-service intercepts; clear only the participant review for service ones.
- **Restore ordering** (extends the Codex note): agents reconnect before `ReconnectClient` restores intercepts, so `participants` stays empty until new pods arrive, and the first churn hits the `len(participants) == 0 → NO_AGENT` branch. `RestoreIntercepts` needs to call `initializeParticipants` (nil reviews are fine while ACTIVE).
- **Container names** (confirming the other Codex note): a non-primary agent resolves the primary's `Spec.ContainerName` at `fwdstate.go:161` → `AGENT_ERROR`, and `applyServiceReview` applies that rejection to the whole intercept — breaking the mixed main/canary case this feature exists for. Agents should resolve the container from their own matched `InterceptTarget`.
- **Degrade, don't error.** `ensureServiceWorkloads` runs for every service-backed intercept and hard-fails setups that work today: tcp/`--replace` behind a multi-workload service, http with not-yet-upgraded agents, or a service selecting a pod with an unresolvable owner. These should warn and fall back to the single-workload path. Also worth a stated decision: agents are now auto-injected into workloads the user never named (with no cleanup on leave), and a participant scaling to zero takes the whole intercept to NO_AGENT.

Please split the (good) `RestoreAgent` idempotency fix and the `UpdateIntercept` no-op skip into their own PR — the latter changes ModifiedAt/event semantics for all callers. Needs a `CHANGELOG.yml` entry and docs.

thallgren:
Addendum to my review — a correctness finding against the current revision that makes the expansion a silent no-op in a real manager:

**`serviceSelectsPod` reads a field the shared informer strips.** Discovery lists pods through the factory's shared informer (`kf.Core().V1().Pods().Lister()`), but that informer's `SetTransform` nils `Status.Conditions` (`mutator/watcher.go:426`) — it deliberately keeps only container `State` for the eviction/podwatcher consumers. So in production, `serviceSelectsPod`'s scan for `PodReady` falls through and returns false for **every** pod (unless `PublishNotReadyAddresses`), `serviceWorkloads` never finds a secondary workload, and the multi-workload path never engages. The unit tests pass because they construct their own Pod objects and never go through the transform. The codebase documents this exact trap for Services: `quictunnel/discovery.go` can't use the shared Services informer "because that informer's SetTransform zeroes Status — exactly the field the ingress addresses live in."

Preferred fix: drop the readiness filter entirely — Codex's scaled-to-zero P1 already argues it's the wrong filter on independent grounds. Discovery should answer "which workloads can this Service route to" (a label/owner question), not "which pods are ready this instant." If readiness must stay, the transform can retain just the `PodReady` condition, but then it serves a third consumer and that coupling needs a comment.

Related, on the open question of watching pod-label changes (Codex's latest P2): pods informers already run in every managed namespace with two consumers (`mutator/watcher.go:408`, `cluster/podwatcher.go:93`), so the cache and watch streams are already paid for. One additional `AddEventHandler` that short-circuits on unchanged labels — a single handler feeding an intercept-scoped dispatcher, not one per intercept — closes that window cheaply, if you decide it's worth closing.

thallgren:
@breland-openai and @chatgpt-codex-connector, I'm concerned that this review/fix loop isn't going in the right direction. I asked my AI to evaluate the situation, and it confirmed my suspicion.

Here is my assessment of the review/fix loop on this PR.

**Verdict:** the loop is producing real findings but it is not converging, and it won't on its own — the termination condition is currently "the Codex bot runs out of edge cases," and it never will. And there has been drift — not sideways into unrelated problems, but downward: the PR has been pulled from its stated scope (static discovery of co-selected workloads at intercept creation, HTTP-only, no --replace) into an unscoped and much harder problem: live reconciliation of intercept state against arbitrary cluster mutation.

**What the record shows:** the loop has exactly one human round — the day-one changes-requested review with five substantive blockers plus scope fencing. That part worked: the requested split happened (#4258 for the RestoreAgent idempotency, #4259 for the UpdateIntercept no-op skip), and the worst blocker (the #4183 revert) is gone from the current diff — interceptNeedsDialWatcherOnAllPods no longer appears anywhere in it.

**Everything since is author versus bot:** 16 bot review rounds over three days, each push earning one or two new findings, 19 of 22 threads unresolved, and the newest P2 arriving four minutes after the latest squashed force-push. The finding trajectory tells the story better than the count. Day one: core lifecycle gaps (participants lost on manager restore, requeue on workload add). By day two and three: a Service selector shrinking mid-intercept, mixed legacy/new-agent rollouts, a pod manually relabeled into the Service while its owner's template doesn't match, the Service deleted and recreated with a new UID, AgentMaxIdleTime interplay, restart concurrent with selector change. Each finding is individually legitimate — the bot is not wrong — but they are monotonically more exotic while the arrival rate stays constant. That is the signature of an unbounded state space being enumerated one patch at a time.

**The structural cause:** the pivot point is visible in the thread history — the P1 "Reconcile workloads selected after intercept creation" (Aug 6, 09:47). Responding to it, the PR grew service_intercept_watch.go — a live watcher that doesn't exist on release/v2, isn't in the PR description, and is now the primary generator of new findings. Once the PR promises dynamic participant reconciliation, the correctness surface becomes the cross-product of selector changes x pod lifecycle x agent versions x manager restarts x intercept mechanisms, and the implementation handles that product as accumulated per-event patches. Every patch legitimately exposes the adjacent cell of the matrix. A tireless bot reviewer will walk that matrix forever.

**What I'd do:** this ends with a maintainer scope ruling, not with more fixes. Two coherent stopping points. Option one (recommended): freeze participants at creation. Rip the watcher back out; discovery happens once, at intercept creation, matching the PR description and the actual use case (a stable envoy-weighted main/canary pair). Selector drift while intercepted gets detected and fails or degrades the intercept with a clear message rather than being chased. Most of the 19 open threads become resolvable as out-of-scope-by-policy, the diff shrinks, and the feature ships. Option two: accept dynamic reconciliation as scope — but then demand the architecture for it: a single idempotent reconcile function that computes desired participants from (current selector, current agent inventory) and is requeued on any input event, replacing the accumulated patches. That collapses the edge-case list into one mechanism the bot can't nibble at — but it is a redesign of the manager-side half of the PR and deserves its own design round rather than emerging from review pressure.

**The untenable position is the current middle:** dynamic scope, patch-per-edge-case implementation. Whichever way the ruling goes, the bot posture should also be stated explicitly on the PR — findings against declared-out-of-scope scenarios get resolved as such, so the thread count stops functioning as a treadmill.

thallgren:
Hey @breland-openai. My thoughs at this point.

**#4251 is scoped to static discovery: participants are computed once, at intercept creation, and frozen.** Please rip the watcher back out: `service_intercept_watch.go`, the reconcile loop, and the fallback machinery.

The review loop went off course at the P1 "Reconcile workloads selected after intercept creation." That finding assumed the participant set must track the cluster, but nothing in this PR's description or use case (a stable main/canary pair of workloads behind one Service) requires that. Accepting it turned a bounded feature into open-ended reconciliation of intercept state against arbitrary cluster mutation. Every finding since has been an edge of that unbounded surface, not a defect in what the PR set out to do. To be clear: the reconcile rework itself is good engineering, and if dynamic reconciliation ever gets a real use case it should start from that design, in its own issue.

In this PR, selector drift is detected, not repaired. If the participant set stops matching reality, the intercept reports a clear degraded/failed state and the user re-creates it. I'll resolve the open threads that target removed code or drift-repair scenarios as out-of-scope, and further findings against those scenarios will be closed the same way.

This also matters for sequencing. #4266 restructures the manager's state layer around session authorization. A creation-time-only discovery runs inside the already-authorized `AddIntercept` path and rebases cleanly over that work, where an autonomous watcher mutating intercept state would not.

PR #4126, opened 2026-05-12
Stability improvements for intercepts

mindstorms6:
**Still a draft. I want to see if I can simplify this further. Reviews/feedback welcome. I can also split this if needed in to some (partly) smaller change sets.** 

Selected intercepts have a few sharp edges when a watch stream drops, a local endpoint is slow to answer, or traffic arrives in a burst. This makes that path steadier without making quiet sessions noisier.

Reconnect agent dial watchers with backoff, clear stale agent clients before retrying, reuse HTTP intercept transports, and clean up pending forwards more carefully so old work cannot step on newer state.

Also add dedicated timeout knobs for traffic-agent connects and intercepted endpoint dials, cap dial responder and rejecter work, and log the slow or flappy paths with enough context to diagnose real stalls.

Update the 2.28.0 release notes for the stability work.

## Description

A few sentences describing the overall goals of the pull request's commits.

## Checklist

 - [x] I made sure to update `./CHANGELOG.yml` (our release notes are generated from this file).
 - [x] I made sure to add any documentation changes required for my change.
 - [ ] My change is adequately tested.
 - [ ] I updated `CONTRIBUTING.md` with any special dev tricks I had to use to work on this code efficiently.
 - [ ] Once my PR is ready to have integration tests ran, I posted the PR in #telepresence-oss channel on the
       [CNCF Slack](https://slack.cncf.io/) so that the "ok to test" label can be applied.

PR #4253, opened 2026-08-05
Reuse HTTP transports for selected intercepts

breland-openai:
## Description

This is the transport-reuse half of #4126, split out so it can be reviewed independently of reconnect behavior.

The HTTP forwarder currently builds a fresh transport for every selected request. That repeats connection setup and makes connection-level tunnel accounting follow request lifetime instead of the reused connection itself.

This caches selected-intercept transports per protocol mode, keeps HTTP/1 and HTTP/2 behavior separate, closes idle transports when the intercept changes, and reports tunnel metrics when each reused connection actually closes. The current protocol routing behavior is preserved.

Tested with `GOEXPERIMENT=jsonv2 go test ./cmd/traffic/cmd/agent/fwd -count=1`.

## Checklist

### DIFF
--- a/cmd/traffic/cmd/agent/fwd/http.go
+++ b/cmd/traffic/cmd/agent/fwd/http.go
@@ -12,6 +12,7 @@ import (
 	"net/http/httputil"
 	"net/netip"
 	"net/url"
+	"sync"
 
 	"golang.org/x/net/http2"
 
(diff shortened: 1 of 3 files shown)


thallgren:
Pushed three commits to this branch: a merge of the current `release/v2`, a change that keeps only the raised per-host idle limit (the cloned default transport has no cap on total idle or concurrent connections per host, and `max(0, 512)` would have introduced one), and a changelog entry. Full regression suite passed locally on top of this together with #4261 and #4262.

thallgren (approved):

PR #4259, opened 2026-08-05
Avoid rewriting unchanged intercepts

breland-openai:
## Description

Some intercept updates run through a mutator that can leave the stored value semantically unchanged. Rewriting that value anyway refreshes `ModifiedAt` and adds an unnecessary compare-and-swap cycle.

This returns the current intercept when the mutation is a no-op, while real changes still get a fresh timestamp and stored value.

Tested with `GOEXPERIMENT=jsonv2 go test ./cmd/traffic/cmd/manager/state -count=1`.

## Checklist

### DIFF
--- a/cmd/traffic/cmd/manager/state/state.go
+++ b/cmd/traffic/cmd/manager/state/state.go
@@ -720,8 +720,9 @@ func (s *State) WatchWorkloads(ctx context.Context, ns string) (ch <-chan []Even
 }
 
 // UpdateIntercept applies a given mutator function to the stored intercept with interceptID;
-// storing and returning the result.  If the given intercept does not exist, then the mutator
-// function is not run, and nil is returned.
+// storing and returning the result. If the mutator leaves the intercept semantically unchanged,
+// the current value is returned without updating ModifiedAt. If the given intercept does not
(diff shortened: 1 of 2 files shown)


cmd/traffic/cmd/manager/state/state.go
thallgren, line 750: I think this whole function can be simplified. There's no caller that locks inside the `apply` so the CAS logic can be removed completely, leaving us with one single call to `interceptEqual`.
```go
	v, _ := s.intercepts.Compute(interceptID, func(cur *Intercept, loaded bool) (*Intercept, xsync.ComputeOp) {
		if !loaded {
			return nil, xsync.CancelOp
		}
		n := cur.Clone()
		apply(n)
		if interceptEqual(cur, n) {
			return cur, xsync.CancelOp
		}
		n.ModifiedAt = timestamppb.Now()
		return n, xsync.UpdateOp
	})
	return v
```

thallgren: Done in 41d4c3338: UpdateIntercept is now a single Compute on the intercept map, cancelling on a missing id or a semantically equal result, with the CAS loop removed.

PR #4260, opened 2026-08-05
Add local client redirect commands

breland-openai:
## Description

Sometimes I want a local client to send one remote host and port to a localhost listener without creating or changing an intercept.

This adds `telepresence local-client-redirect add|list|remove`, carries the redirect through the user and root daemons, and checks the mapping before normal tunnel routing. The active session owns the redirect table, so it is easy to inspect and remove.

Tested with `GOEXPERIMENT=jsonv2 go test ./pkg/client/rootd ./pkg/client/cli/connect ./pkg/client/cli/cmd ./pkg/client/userd/daemon -count=1`.

### DIFF
--- a/docs/reference/cli/telepresence.md
+++ b/docs/reference/cli/telepresence.md
@@ -56,6 +56,7 @@ result in a password prompt.
 | [list](telepresence_list) | List current intercepts |
 | [list-contexts](telepresence_list-contexts) | Show all contexts |
 | [list-namespaces](telepresence_list-namespaces) | Show all namespaces |
+| [local-client-redirect](telepresence_local-client-redirect) | Manage local client redirects |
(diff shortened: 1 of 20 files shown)


thallgren:
🚀 This is a pretty cool new feature, and I'd like for it to be announced as such 😃 

Here are some additions that I'd love to see in this PR:
- The manifest for the new `telepresence apply` command should cover this, perhaps with an associated local service that is started automatically.
- The documentation should explain the concept and why it's a great feature. It's your own local personal intercept!
- Please add a feature entry to the CHANGELOG.yml that talks about the advantage of being able to actually intercept the outgoing traffic from your handler and redirect it to something that perhaps doesn't run in the cluster (yet).
- Add an entry to the comparison docs, describing `localRedirects`. I don't think mirrord or GeFyra has this.

 I do question the name because "client" is somewhat redundant in the vocabulary. What do you think about using just "local-redirect"?

PR #4261, opened 2026-08-05
Keep workload watch delivery safe when streams cancel

breland-openai:
## Description

Workload watch subscribers can disappear while a batched update is being delivered. The previous delivery path could keep waiting on a subscriber whose stream had already been canceled.

Track each subscription's cancellation alongside its channel, skip canceled subscribers during dispatch, and avoid closing channels that an in-flight delivery may still hold. The watcher continues to preserve namespace filtering for active subscribers.

Tested with `GOTOOLCHAIN=local GOEXPERIMENT=jsonv2 /opt/homebrew/bin/go test ./cmd/traffic/cmd/manager/state -run '^TestWatcherDispatchSkipsCanceledSubscription$' -count=1` and `GOTOOLCHAIN=local GOEXPERIMENT=jsonv2 /opt/homebrew/bin/go test ./cmd/traffic/cmd/manager/state -count=1`.

## Checklist


cmd/traffic/cmd/manager/state/watcher.go
```diff
@@ -135,6 +111,34 @@ func NewWatcher(ctx context.Context, ns string, enabledWorkloadKinds k8sapi.Kind
...
+	w.Lock()
+	ss := make([]subscription, 0, len(w.subscriptions))
+	for _, sub := range w.subscriptions {
+		ss = append(ss, sub)
+	}
+	events := w.events
+	w.events = nil
+	w.Unlock()
+	for _, s := range ss {
+		filtered := make([]Event, 0, len(events))
```
thallgren, line 124: This creates a slice for the whole events batch once for every subscription, even when there's no match. In cluster-wide mode, this can cause a substantial memory spike.

The old version wasn't ideal either because it didn't do any preallocation at all.

I think it would be better to partition the `events` on namespaces prior to the loop and then write matching partitions to the subscriber.

cmd/traffic/cmd/manager/state/watcher.go
```diff
@@ -135,6 +111,34 @@ func NewWatcher(ctx context.Context, ns string, enabledWorkloadKinds k8sapi.Kind
 	return w, nil
 }
 
+func (w *watcher) dispatch(ctx context.Context) {
+	w.Lock()
+	ss := make([]subscription, 0, len(w.subscriptions))
+	for _, sub := range w.subscriptions {
+		ss = append(ss, sub)
+	}
```
thallgren, line 119: Nit for sure, but why change this into an append? Please retain this:
```suggestion
	ss := make([]subscription, len(w.subscriptions))
	i := 0
	for _, sub := range w.subscriptions {
		ss[i] = sub
		i++
	}
```

thallgren (changes requested):
Good catch and solution!

cmd/traffic/cmd/manager/state/watcher.go
```diff
@@ -135,6 +111,34 @@ func NewWatcher(ctx context.Context, ns string, enabledWorkloadKinds k8sapi.Kind
...
+		}
+		if len(filtered) == 0 {
+			continue
+		}
+		select {
+		case <-ctx.Done():
+			return
+		case <-s.done:
+		case s.ch <- filtered:
+		}
```
thallgren, line 138: Thinking more about this. If a subscriber is slow but not cancelled, it will stall the whole delivery to all clients. So in addition to the `s.done`, I think we need a timer here too.

Before the loop:
```
    timer := time.NewTimer(time.Duration(math.MaxInt64))
    defer timer.Stop()
```
and then:
```suggestion
		timer.Reset(sendTimeout)
		select {
		case <-ctx.Done():
			return
		case <-s.done:
		case s.ch <- filtered:
		case <-timer.C:
			clog.Warnf(ctx, "workload watch subscriber for %s stopped reading; dropping subscription", s.namespace)
			s.cancel()
		}
		timer.Stop()
```

cmd/traffic/cmd/manager/state/watcher.go
thallgren: Done in 425e659c6: events are partitioned by namespace once per batch and each subscriber receives its partition; no per-subscriber slice is built.

cmd/traffic/cmd/manager/state/watcher.go
thallgren: Done in 425e659c6: the subscription snapshot is preallocated and index-assigned again.

cmd/traffic/cmd/manager/state/watcher.go
thallgren: Done in 425e659c6: a 10-second send timer per subscriber; on expiry the subscription is cancelled with a warning and delivery continues to the rest. Covered by TestWatcherDispatchDropsStalledSubscriber.

thallgren:
Pushed three commits to this branch: a merge of the current `release/v2`, the changes requested in review (namespace partitioning once per batch, the preallocated subscription snapshot, and a 10-second send timer that drops a subscriber that stops reading), and a changelog entry. Full regression suite passed locally on top of this together with #4262 and #4253.

thallgren (approved):

PR #4262, opened 2026-08-05
Keep agent watch payloads compact until details are needed

breland-openai:
## Description

Agent watch streams can carry container environment maps for every pod even though clients only need identity and routing fields until an ingest is started. That makes reconnect snapshots larger than they need to be.

Let capable clients request compact watch payloads, mark omitted environments explicitly, and keep compact snapshots from being restored as authoritative state after a manager restart. Clients that do not advertise support keep receiving the existing full payloads, and full details are still fetched through EnsureAgent when an ingest needs them.

Tested with `GOTOOLCHAIN=local GOEXPERIMENT=jsonv2 /opt/homebrew/bin/go test ./cmd/traffic/cmd/manager -count=1` and `GOTOOLCHAIN=local GOEXPERIMENT=jsonv2 /opt/homebrew/bin/go test ./pkg/client/userd/trafficmgr -count=1`.

## Checklist

### DIFF
--- a/cmd/traffic/cmd/manager/service.go
+++ b/cmd/traffic/cmd/manager/service.go
@@ -305,6 +305,14 @@ func (s *service) ReconnectClient(ctx context.Context, info *rpc.ReconnectClient
 	now := time.Now()
 	st.RestoreClient(sessionID, client, auth.PrincipalFrom(ctx), now)
 	agents := slices.DeleteFunc(slices.Clone(info.Agents), func(agent *rpc.AgentInfo) bool {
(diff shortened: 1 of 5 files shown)


thallgren:
Pushed two commits to this branch: a merge of the current `release/v2` and a changelog entry. The proto change was reviewed for wire compatibility: both new fields are additive with unused numbers, and old clients or managers fall back to full payloads. Full regression suite passed locally on top of this together with #4261 and #4253.

thallgren (approved):

PR #4266, opened 2026-08-07
Minimize the client RBAC footprint

### LINKED ISSUE #4264: Minimize the client RBAC footprint
# Minimize the client RBAC footprint

## Problem

A telepresence client today needs RBAC grants that exist for purely mechanical reasons — discovering the traffic-manager, tunneling to it, watching namespaces, and fetching logs — not because the client must be the one doing the work. The manager already authenticates every caller and authorizes intercepts using the verified identity and RBAC, so it can do all of this as the client's deputy.

## Goal

A client that needs **no Kubernetes API access at all** when an admin publishes an external manager endpoint, and a single name-scoped `pods/portforward` rule otherwise. Concretely:

- Namespace watching and log retrieval become manager RPCs, gated on the caller's identity via Telepresence's own policy attributes (not `pods/log`, which the reduced client no longer holds). Log streaming always requires an authenticated principal, in every authentication mode — the manager must not exercise its own log access for a caller who has proven nothing.
- The manager pod name becomes deterministic, so the connect role shrinks to one name in a `resourceNames`-scoped `pods/portforward` rule. The need for discovery disappears. The conversion trades the current surge-based upgrade continuity for a brief no-ready-manager window per upgrade; clients reconnect and restore their intercepts automatically.
- An admin can publish the manager's control endpoint — a separate TLS gRPC listener exposing only client-facing methods — so the client dials it directly instead of bootstrapping over a port-forward. QUIC remains the tunnel data plane, bootstrapped after authentication; the TLS connection doubles as the tunnel fallback where UDP is blocked. Bearer authentication is unchanged; a cert-only kubeconfig authenticates via direct, optional mTLS on that listener.
- Publishing an endpoint requires `security.authentication.mode: enforcing`, validated at chart render and again at manager startup, and the authentication path is rate-limited before TokenReview (an unseen invalid token costs up to two reviews, and unique tokens bypass the negative cache).

Backward compatibility at all times is a requirement: old and new clients and managers interoperate in every combination, with capability detection over the existing handshake.

## Authorization: policy without mechanism

Today `pods/portforward` is simultaneously the policy ("may intercept here") and the mechanism, with the API server as enforcement point. Moving the mechanics server-side breaks that coincidence, with three consequences:

1. **Connect must be authorized, not just authenticated.** Today, reaching the manager's gRPC port at all requires `pods/portforward` on the manager pod — the connect Role *is* the connect control point. With an external endpoint that control point is gone, so session establishment gets its own review; without one, any valid cluster identity reaching the endpoint gets namespace lists, DNS, and routing. The review must also run before the first side effect — today's single check sits in `CreateIntercept`, after preparation has already provisioned agent configuration — and again when a fresh manager restores a client's sessions and intercepts, which the current reconnect path accepts without any review. Restored state is also renormalized — the payload is treated as desired intercept specs, never as authoritative manager state.
2. **What the review asks.** Reviewing Telepresence's own attributes in a `telepresence.io` group: `create` on `connections` to connect, and an attachment review in the target namespace — `attachments` is the umbrella for intercept (verb `create`) and ingest (verb `get`), with the workload as resource name, so a Role reads `resources: ["attachments"], resourceNames: ["payments"]`. The review carries no workload-kind subresource: which kinds exist at all is a global chart decision (`workloads.*.enabled`), no plausible policy admits a user to Deployments but not StatefulSets, and RBAC has no `attachments/*` form, so per-kind subresources would force every hand-written Role to enumerate them; a Deployment and a StatefulSet sharing a name are both covered by one name-scoped grant — an accepted imprecision. This works without a CRD (the RBAC authorizer matches rule strings without consulting discovery; verified against a live cluster) and buys what `pods/portforward` can't express: Telepresence access separable from port-forward access in both directions, name-scoping to stable workload names instead of random pod names, and legible intent. One bound: wildcard identities such as `cluster-admin` pass any review, unregistered resources included — RBAC is additive and allow-only, so the review separates Telepresence from *narrowly scoped* roles but cannot deny a wildcard holder.
3. The manager grants pod-level reach to callers who may hold no `pods/portforward`. In essence, the admin that installs the manager delegates enforcement of the `telepresence.io` group to Telepresence.

Proposed knob: `security.authorization.requiredGrant: portforward | telepresence | any` — the grant a client must hold to be authorized — initially defaulting to `any`, a non-breaking superset of today's behavior that lets `telepresence.io`-only Roles work immediately. Publishing an external endpoint already withholds the chart's `pods/portforward` grant (external clients never port-forward); setting `requiredGrant: telepresence` then makes the external endpoint the only path in, for every identity short of a wildcard role.

## Decisions

- The namespace list is served unfiltered — no per-caller filtering. This is what the DNS and dialer expose anyway once fully qualified names are used, so filtering would add review volume (one authorization check per namespace × client) without hiding anything.
- A TCP/TLS gRPC fallback listener for the external endpoint is in scope, for environments where UDP (and hence QUIC) is blocked.
- The gate defaults to `any` in every authentication mode from the start (non-breaking), and while it resolves to `any` or `portforward` the manager logs a per-session warning whenever a caller authorized only via the legacy `pods/portforward` grant — the admin's inventory of hand-rolled Roles still needing migration. The enforcing-mode default then flips to `telepresence`, no earlier than two minor releases after the external endpoint ships and preferentially at the next major version; that flip breaks only hand-rolled portforward-only Roles, since chart-rendered Roles migrate with the chart. Non-enforcing modes stay on `any`.
- No periodic re-review: authorization runs at session and intercept creation only, so revoking a grant does not tear down running sessions. When a revocation is time critical, the admin restarts the manager: clients re-establish their sessions and intercepts automatically, and the restore path re-runs the reviews — a check this work adds, since today's reconnect restores client-provided state without one — so the revoked caller is the one that does not come back.

thallgren:
Implementation of the client RBAC minimization proposed in #4264: an authorization foundation in the traffic-manager, manager-mediated namespace watching and log streaming, a deterministic manager pod name, and an external client-only control endpoint that removes the last mechanical RBAC requirements. The resulting client footprint is documented in `docs/reference/rbac.md`.

Closes #4264.


breland-openai:
This is a very exciting change. I will try to review a bit more in depth. We currently suffer somewhat large `telepresence connect` times due to the client RBAC calls so very much looking forward to this.

breland-openai:
Are there any areas you want me to deeply look at?

thallgren:
> Are there any areas you want me to deeply look at?

If there are any areas in this PR that might create friction for you or in other ways overlap/conflict with the work you're doing or have planned. The #4251's service-intercept restore bookeeping will require some conflict resultion but other than that, I haven't found any such problems in the PRs that you've posted.

PR #4270, opened 2026-08-11
Return NODATA for single-stack DNS lookups

breland-openai:
## Why

An AAAA lookup for an IPv4-only hostname currently returns NXDOMAIN even though the hostname exists. A lookups for IPv6-only hostnames have the same problem. Clients using per-record DNS lookups can mistake a valid single-stack hostname for a missing one.

## What changed

When the requested address family has no records, check the other family. If that family exists, return a successful empty response. Genuine NXDOMAIN, temporary resolver failures, cancellation, and deadline behavior stay unchanged.

## Test plan

- `GOEXPERIMENT=jsonv2 go test ./pkg/dnsproxy -count=1`
- `GOEXPERIMENT=jsonv2 go test -race ./pkg/dnsproxy -run TestLookupIPAddressFamilies -count=3` on Linux
- `GOEXPERIMENT=jsonv2 go test ./cmd/traffic/cmd/manager -run Test -count=1`
- `GOEXPERIMENT=jsonv2 go vet ./pkg/dnsproxy`
- `make lint GOOS=linux`
- Confirmed both single-stack cases fail against unchanged `release/v2`.


pkg/dnsproxy/lookup.go
```diff
@@ -187,6 +187,20 @@ func lookupIP(ctx context.Context, network, qName, noSearchDomain string, r *net
 			IsNotFound: true,
 		}
 	}
+	var dnsErr *net.DNSError
```
thallgren: The `r.LookupIP` on line 178 could use `"ip"` instead of `network` here. The result will be a slice containing both ip4 and ip6 addresses that can then be correctly filtered using `network`. That'll make an extra call to `LookupIP` unnecessary.

thallgren (changes requested):

thallgren (approved):

